Issue #8, opened 2025-11-27
Prevent new AI-DLC workflow from overwriting incomplete existing workflow


ankitsri-aws:
## Problem
When a user starts a new AI-DLC workflow (using "Using AI-DLC, ...") while an incomplete workflow already exists, sometimes the system removes the old aidlc-state.md and starts from the beginning, losing all previous progress.

## Current Behavior
- aidlc-state.md gets overwritten without warning
- User loses all progress from previous workflow
- No option to resume or preserve existing work

## Expected Behavior
Before starting a new AI-DLC workflow, the system should:
1. Check if aidlc-state.md exists with incomplete stages
2. If found, block new workflow initiation
3. Present user with options:
   - **A) Resume existing workflow** - Continue from last completed stage
   - **B) Abandon and start fresh** - Archive old aidlc-docs/ folder with timestamp, then start new workflow
4. Only proceed after explicit user decision

## Proposed Solution
Add explicit workflow state check in at the Workspace Detection stage (before logging initial user request):

Files to Modify: aidlc-rules/aws-aidlc-rule-details/inception/workspace-detection.md
```
## Step 1: Check for Existing AI-DLC Project

Check if `aidlc-docs/aidlc-state.md` exists:
- **If exists with incomplete stages**:
  * STOP - do not start new workflow
  * Ask user: "Incomplete workflow found. A) Resume or B) Abandon and start fresh?"
  * If A: Load state and continue from last stage
  * If B: Rename `aidlc-docs/` to `aidlc-docs-backup-[timestamp]/` and continue
- **If exists and complete**: Resume from last phase (load context from previous phases)
- **If not exists**: Continue with new project assessment
```

raj-jain-aws:
Stale. Hasn't been reported elsewhere. Pls. reopen if you're still encountering this issue.

PR #9, opened 2025-11-27
Protection check to avoid overwriting of the audit file

SiddhJog:
Issue 8: https://github.com/awslabs/aidlc-workflows/issues/8

added a protection check before the audit file is updated at the beginning of the workflow


raj-jain-aws (changes requested):
@SiddhJog Thanks for this fix. 

- Can you add an example of what happens before this fix?

- One rule statement says ""...STOP and present protection message". How does the protection message look like? Do we provide guidance to user how to handle this situation?

raj-jain-aws:
@wmatosjr Agreed to the verbosity concern. I think that's a bigger concern we need to handle in general, shed the fat first and then ensure that whatever new behaviors we put on later are as lean an possible.

PR #37, opened 2026-01-09
Support for other platforms

harmjeff:
*Issue #: 34*

*Adding instructions for:*

- Claude
- Cline
- CoPilot
- Cursor

Only change to existing README is links to new READMEs and moving the rule details to the <.aidlc-rules-details> folder.  The move is to support platforms that read all contents of their instructions folders regardless of path and maintaining consistency.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #36, opened 2026-01-07
feat: add automated release pipeline

leandrodamascena:
Closes #17 

## Summary

This PR introduces Agent Skills format support and an automated release pipeline for the AI-DLC methodology.

## Problem

The AI-DLC methodology currently has no versioning or release tracking. Users have no way to know which version they're using, there's no changelog to track changes between versions, and manual distribution leads to inconsistencies over time. With the addition of new standards like Agent Skills, we need a clear way to distribute multiple formats without duplicating source files.

### Files in one single place

The `aidlc-rules/aws-aidlc-rule-details/` directory remains the source of truth. All other formats are generated from it automatically during release, eliminating duplication and drift.

### Copying to new formats

Adds support for the [Agent Skills specification](https://agentskills.io), making the methodology compatible with Claude Desktop, Cursor, Windsurf, and any editor implementing the standard. The skills format is generated automatically from source files during release.

### Automated release pipeline

Release now is automated when you push a version tag (`git tag v1.0.0 && git push origin v1.0.0`), the pipeline:

1. Generates the Agent Skills format with version metadata
2. Creates two distribution artifacts (skills zip and rules zip)
3. Publishes a GitHub Release with artifacts attached
4. Creates a PR to update CHANGELOG.md using git-cliff

## Commit convention

For proper changelog generation, use conventional commits: `feat:` for features (minor bump), `fix:` for bug fixes (patch bump), and others.

All GitHub Actions used are well-established and pinned to specific commit SHAs:

| Action | Purpose |
|--------|---------|
| `actions/checkout` | Official GitHub action |
| `softprops/action-gh-release` | 6k+ stars, widely adopted |
| `orhun/git-cliff-action` | Official action from git-cliff author |

This is working in local repo

image

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #41, opened 2026-01-17
feat: Add Kiro CLI Knowledge Base support

inariku:
## Summary

This PR adds support for Kiro CLI's Knowledge Base feature, enabling efficient on-demand loading of AI-DLC rule details.

## Problem

Currently, all rule detail files are loaded into context when using steering mode. For large rule sets, this consumes significant tokens even when only specific rules are needed.

## Solution

Add a custom agent configuration (`agents/ai-dlc.yaml`) that:
- Keeps `core-workflow.md` as always-loaded steering context
- Indexes `aws-aidlc-rule-details/` as a Knowledge Base for efficient search and retrieval

## Changes

1. **`agents/ai-dlc.yaml`** - New custom agent configuration with Knowledge Base resource
2. **`aidlc-rules/aws-aidlc-rules/core-workflow.md`** - Added Loading Method section supporting both Steering and Knowledge Base modes
3. **`README.md`** - Added installation instructions for Knowledge Base mode

## Compatibility

- **Backward compatible**: Existing steering mode continues to work unchanged
- **Additive**: Knowledge Base mode is an additional option for Kiro CLI users
- **Independent**: Does not conflict with PR #36 (release pipeline) or PR #37 (multi-platform support)

## Benefits

- **Token efficiency**: Rule details loaded on-demand via search instead of all at once
- **Scalability**: Better suited for growing rule sets
- **User choice**: Users can choose between ste- Keeps `core-workflow.md` as always-loaded steering 

By subm- Indexes `aws-aidlc-rule-details/` as you can use, modify, c
## Changes

1. **`agents/ai-dlc.yaml`** - New custom agent configuratio

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,39 @@ If you do not see the `aws-aidlc-rules` rules loaded, please check the directory
 
 ![AI-DLC Rules in Kiro CLI](./assets/images/kiro-cli-aidlc-rules-loaded.png?raw=true "AI-DLC Rules in Kiro CLI")
 
+### Kiro CLI (Knowledge Base Mode)
+
+For larger projects or when you want better token efficiency, you can use AI-DLC as a custom agent with Knowledge Base support. This mode indexes the rule details and loads them on-demand instead of including everything in context.
+
+Copy the custom agent configuration to your Kiro CLI agents directory:
+
+```bash
+# Global installation (available in all projects)
+mkdir -p ~/.kiro/agents
+cp ../aidlc-workflows/agents/ai-dlc.yaml ~/.kiro/agents/
+
+# Or project-specific installation
+mkdir -p .kiro/agents
+cp ../aidlc-workflows/agents/ai-dlc.yaml .kiro/agents/
+```
+
+Also copy the rule files to the expected location:
+
+```bash
+cp -R ../aidlc-workflows/aidlc-rules .kiro/
+```
+
+To use the AI-DLC agent:
+
+```bash
+kiro-cli --agent ai-dlc
+```
+
+Benefits of Knowledge Base mode:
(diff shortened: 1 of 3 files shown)


dreamorosi (commented):
Hi @inariku thank you for the PR.

I think this should be an opt-in add-on to the workflow rather than a mandatory item in the core workflow.

This change ties the rules too much to Kiro and I think we want customers who are using other LLM harnesses to be able to use the workflows.

leandrodamascena:
Hi @inariku, thank you for this PR and for your patience waiting on the extensions framework.

We're going to close this one because the problem it solves has already been addressed in the core workflow itself.

## How it works now

The core workflow already supports on-demand loading through the extensions mechanism:

1. At workflow start, the model scans `extensions/` but only loads lightweight `*.opt-in.md` files (a few lines each with a multiple-choice question)
2. During Requirements Analysis, the opt-in questions are presented to the user
3. Only when the user opts IN does the model load the full rules file (convention: `foo.opt-in.md` → `foo.md`)

This gives us the same token efficiency benefit you were after, without tying the core workflow to a Kiro-specific feature (Knowledge Base / custom agents YAML).

## Why we're closing

- The on-demand loading problem is solved in an agent-agnostic way
- Modifying `core-workflow.md` to reference Kiro's Knowledge Base mode would couple the core rules to one specific agent, which goes against the project's goal of being platform-neutral
- The `agents/ai-dlc.yaml` format is specific to Kiro CLI and would need to live in a Kiro-specific extension or separate repo rather than in the core

We appreciate the contribution and the thinking behind it. If Kiro's Knowledge Base feature evolves in a way that could be expressed as a standalone extension (without modifying the core workflow), we'd be happy to look at that in the future.

Thank you!

PR #52, opened 2026-01-23
Add Kiro Power Support and Installation Instructions

osataken:
*Description of changes:*
- Add POWER.md with metadata and complete documentation
- Add README.md with multi-platform installation instructions
- Include shell commands for Linux/macOS and Windows
- Document power structure and verification steps
- Add troubleshooting guide for common issues

core-workflow.md
- Add variable system for RULE_LOADER_TOOL and RULE_FILE_PATH
- Support Amazon Q Developer, Kiro CLI, and Kiro Power platforms
- Update all rule loading references to use platform variables
- Add platform detection logic with priority order
- Document file loading examples for each platform

Root README
- Link to Kiro Power installation guide in Quick Start section
- Position after Kiro CLI section for logical flow

### DIFF
--- a/aidlc-kiro-power/POWER.md
+++ b/aidlc-kiro-power/POWER.md
@@ -10,17 +10,27 @@ author: "AWS"
 
 ## Overview
 
-The AI-DLC (Adaptive Intelligent Development Lifecycle) is a comprehensive software development methodology that adapts to your project's needs. Unlike rigid frameworks, AI-DLC intelligently determines which stages to execute based on:
-
-- Project complexity and scope
-- Existing codebase state (greenfield vs brownfield)
-- Risk and impact assessment
-- User's stated intent and clarity
+The AI-DLC (AI-Driven Development Life Cycle) is a comprehensive software development methodology that adapts to your project's needs. Unlike rigid frameworks, AI-DLC intelligently determines which stages to execute based on project complexity, existing codebase state, risk assessment, and user intent.
 
 The methodology consists of three main phases:
 - **Inception Phase**: Planning, requirements gathering, and architectural decisions
 - **Construction Phase**: Detailed design, NFR implementation, and code generation
 - **Operations Phase**: Deployment and monitoring (placeholder for future expansion)
+
+## Getting Started
+
+**To understand the complete workflow**, load the core workflow file:
+```
+kiroPowers action: readSteering, powerName: ai-dlc-methodology, steeringFile: core-workflow.md
+```
+
+The core-workflow.md contains all workflow details including:
+- Platform-specific configuration
+- Adaptive workflow principles
+- Detailed stage definitions for all phases
+- References to all other steering files
+- Mandatory rules for content validation and audit logging
+- Progress tracking requirements
 
 ## Key Principles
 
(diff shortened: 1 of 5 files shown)


aidlc-kiro-power/POWER.md
raj-jain-aws, line 1: Can we build this in a way that we programmatically copy over or reuse the main README.md (or parts of it) in this directory instead of duplicating its contents here. That way any changes in the main README.md will be updated in Kiro `POWER.md` automatically and we won't have to maintain two separate files with duplicated content. We foresee major changes in the steering files structure to address the systemic over-bloating issues, so reusing original content will be highly beneficial.

aidlc-kiro-power/README.md
```diff
@@ -0,0 +1,189 @@
...
+**Step 1: Copy power files to your project**
+
+From your project directory (e.g., `my-project/`), run:
+
+**Linux/macOS (bash):**
+```bash
+mkdir -p .kiro/powers/ai-dlc-methodology/steering
+cp ../aidlc-workflows/aidlc-kiro-power/POWER.md .kiro/powers/ai-dlc-methodology/
+cp ../aidlc-workflows/aidlc-rules/aws-aidlc-rules/core-workflow.md .kiro/powers/ai-dlc-methodology/steering/
+cp ../aidlc-workflows/aidlc-rules/aws-aidlc-rule-details/common/*.md .kiro/powers/ai-dlc-methodology/steering/
```
raj-jain-aws, line 39: Is it possible to maintain the sub-directories and file arrangements in `common`, `inception` etc, as opposed to flattening that directory structure?

aidlc-kiro-power/README.md
```diff
@@ -0,0 +1,189 @@
...
+cp ../aidlc-workflows/aidlc-rules/aws-aidlc-rule-details/inception/*.md .kiro/powers/ai-dlc-methodology/steering/
+cp ../aidlc-workflows/aidlc-rules/aws-aidlc-rule-details/construction/*.md .kiro/powers/ai-dlc-methodology/steering/
+cp ../aidlc-workflows/aidlc-rules/aws-aidlc-rule-details/operations/*.md .kiro/powers/ai-dlc-methodology/steering/
+```
+
+**Windows (PowerShell or Command Prompt):**
+```powershell
```
raj-jain-aws, line 46: If we're able to maintain the original directory structure then we won't need so many commands and avoid the human error in missing a command. If we can't do that (because of Kiro power limitation) then suggest we wrap these multiple commands in a script so that the user has to issue a single command and there is no risk of human omission of a command.

aidlc-kiro-power/README.md
```diff
@@ -0,0 +1,189 @@
...
+2. Open Powers Panel (Command Palette → "Powers: Manage")
+3. Click "Install from Directory"
+4. Select `.kiro/powers/ai-dlc-methodology` from your project
+
+**Step 3: Verify**
+
+- Look for "AI-DLC Methodology" in the Powers Panel
+- Click to view details and available steering files
+
+### Option 2: Global Installation (Shell Script)
```
raj-jain-aws: If we wrapped this in a script then the script can ask the user question at the top - global versus project and then can copy material to right folders automatically.

aidlc-kiro-power/README.md
```diff
@@ -0,0 +1,189 @@
...
+4. Review and approve at quality gates
+5. Track progress in `aidlc-docs/aidlc-state.md`
+
+## Power Structure
+
+```
+ai-dlc-methodology/
+├── POWER.md                          # Main power documentation
+└── steering/                         # Detailed workflow rules (flattened structure)
+    ├── core-workflow.md              # Main workflow orchestration
```
raj-jain-aws, line 139: Suggest avoid enumerating the files list here as this will constantly be changing and this information will become obsolete quickly.

aidlc-rules/aws-aidlc-rules/core-workflow.md
```diff
@@ -1,6 +1,40 @@
 # PRIORITY: This workflow OVERRIDES all other built-in workflows
 # When user requests software development, ALWAYS follow this workflow FIRST
 
+## Platform-Specific Variables
+
+**CRITICAL**: Set these variables based on your platform at workflow start:
+
+### For Amazon Q Developer:
+```
+RULE_LOADER_TOOL = "readFile"
```
raj-jain-aws: Are these environment variables? If so, should there be `export` keywords mentioned for Linux/MacOS and equivalent for Windows. Could these be wrapped in the script idea mentioned above?

raj-jain-aws (commented):
Hi @osataken - Nice that the actual rule files are copied and not duplicated. Wondering if we can avoid other content duplications. Also, today many customers ask for customizing AI-DLC. We tell them to drop their rules files in along side our rule files. How would that be done for Kiro Power?

aidlc-kiro-power/POWER.md
osataken: Yes, we can combine the install instruction into single README. I originally created a separated file to minimize the changes on the core-workflow artifacts as Kiro Power is quite specific to Kiro only.

aidlc-kiro-power/README.md
osataken: With the current limitations of Kiro Power, we need to flatten the folder for now.
This will be for the first time installation only (from local file). Once the AIDLC Kiro Power got approved and goes to official Kiro Market place, installation will be much easier. 

Kiro will use internal command "**readSteering**" to read the related rule/steering files. Currently, this command doesn't support nested files/folder structure.

There is also another Kiro internal command "**readFile**" but this command will work only the files located in current workspace. When we install Kiro power, the files will be copied to global installation folder (only the files under steering folder). Using "readFile" command will cause permission error.

aidlc-kiro-power/README.md
osataken: Copied comments from previous one 

> With the current limitations of Kiro Power, we need to flatten the folder for now.
This will be for the first time installation only (from local file). Once the AIDLC Kiro Power got approved and goes to official Kiro Market place, installation will be much easier.

> Kiro will use internal command "readSteering" to read the related rule/steering files. Currently, this command doesn't support nested files/folder structure.

> There is also another Kiro internal command "readFile" but this command will work only the files located in current workspace. When we install Kiro power, the files will be copied to global installation folder (only the files under steering folder). Using "readFile" command will cause permission error.

aidlc-kiro-power/README.md
osataken: Actually, both installation methods will goes to global installation. 

The first approach is to copy/flatten the files/folder structure, then install via IDE. 
The second approach will just run the script to copy the files to Kiro global installation folder. 

I believe that we could remove the second approach to avoid confusion and simplify the installation process.

aidlc-kiro-power/README.md
osataken: Will try to use files/folder structure definitions on core-workflow instead

aidlc-rules/aws-aidlc-rules/core-workflow.md
osataken: No, it is just an internal variable for Kiro/Amazon Q Dev/other AI agents to use as reference. 
This is similar to strategy pattern where we determine the running AI agents to determine the root folder or adjust some agents specific stuffs.

I'm still thinking if we could separate these instructions to a separated file outside core-workflow.md

leandrodamascena:
Hi @osataken, thank you for this PR and for your patience. We appreciate the effort you put into making AI-DLC work with Kiro Power.

We're going to close this one for a few reasons:

## 1. The core workflow should remain agent-agnostic

This PR adds ~40 lines of platform detection logic (if Amazon Q → do X, if Kiro Workspace → do Y, if Kiro Power → do Z) directly into `core-workflow.md`. The core workflow needs to stay neutral — it shouldn't contain conditional logic tied to specific agents or their file-loading mechanisms.

## 2. The file mapping approach doesn't scale

The `kiro-power-file-mapping.md` requires manual updates every time a rule file is added, renamed, or moved. This creates a maintenance burden and a source of bugs when things get out of sync.

## 3. The on-demand loading problem is already solved

Since this PR was opened, we introduced the extensions framework with `*.opt-in.md` files. The core workflow now instructs the model to load only lightweight opt-in files at startup and then load full rule files on-demand when the user opts in. This gives us token efficiency without platform-specific workarounds.

## 4. The PR is significantly behind main

The `core-workflow.md` has changed substantially since January. The base this PR was built on no longer reflects the current state of the project.

---

If Kiro Power evolves to support subdirectory structures (eliminating the need for the flat-file mapping), a much simpler integration could work without modifying the core workflow at all. We'd be happy to look at that in the future.

Thank you for contributing!

Issue #55, opened 2026-01-29
Audit timestamps are inaccurate - need explicit real-time capture with timezone awareness


jimini55:
## Problem

  AI assistants generate inaccurate timestamps in `audit.md`. Common patterns observed:

  1. **Placeholder times** - Default values like `T00:00:00Z` or incremental patterns (`00:00`, `00:01`,
  `00:02`...)
  2. **Wrong time values** - Date may be correct but hours/minutes don't match actual time

  ## Observed Behavior

  Tested with multiple AI tools (Claude, Cursor) - both exhibit unreliable timestamp capture:
  - Sometimes fabricate timestamps entirely
  - Sometimes get date correct but time (hours/minutes) wrong
  - Sometimes use obvious placeholder patterns

  ## Root Cause

  Current workflow specifies ISO 8601 format but doesn't enforce how AI must obtain accurate time. AI models are
  unreliable at capturing real-world time without explicit, strict instructions.

  ## Impact

  - Audit logs lose forensic and debugging value
  - Cannot accurately track workflow duration or event sequence

  ## Proposed Solution

  Add explicit instructions in `core-workflow.md` requiring AI to:
  1. Execute system command to get actual time before each audit entry
  2. Never estimate, guess, or use placeholder values
  3. Verify the captured time appears reasonable

  ## Affected Files

  - `aidlc-rules/aws-aidlc-rules/core-workflow.md`

raj-jain-aws:
Thanks @jimini55 for clarification. This is a good example of custom behavior for different tools. 

@wilmatos I think this will make sense to put in tool specific directories. Your thoughts?

jimini55:
Hi @raj-jain-aws - Yes. I tested with Haiku 4.5 in Kiro and works well.

<img width="639" height="86" alt="Image" src="https://github.com/user-attachments/assets/3aefcf64-efa0-4447-851f-11c2ecfe65a6" />

jimini55:
Hey @raj-jain-aws - I don't have a Windows machine, but I presume AI tools are smart enough to find the corresponding command if `date` fails. 

From my testing, unless I explicitly specified the `date` command, Cursor auto mode didn't pick up the correct timestamp.

raj-jain-aws:
Hi @jimini55 - I tested and found that the `date` command you've proposed doesn't work on Windows. Pls. see attached.

[...]

PR #56, opened 2026-01-29
fix: require actual system time for audit timestamps

jimini55:
AI tools often fabricate timestamps (e.g., 00:00, 00:01 patterns) or capture incorrect time. This adds explicit instruction to execute a system command and use local timezone with offset for human readability.

*Issue #, if available:* #55

*Description of changes:*

  ## Summary
 AI tools fabricate timestamps in audit.md instead of capturing actual time. This adds explicit instruction to execute a system command for real timestamps.

  ## Problem Observed
  - Placeholder patterns: `00:00:00`, `00:01:00`, `00:02:00`...
  - Incorrect time even when date tool was used (minutes didn't match)
  - **Issue observed on**:  Claude (via Claude Code), Cursor
  - **Reproducible**: Start any AIDLC workflow and check audit.md timestamps

  ## Change
  - Requires ISO 8601 with local timezone offset for human readability
  - Adds CRITICAL instruction to execute system command - never fabricate

  ## Testing
  - **Platform tested**: Claude Code (Opus 4.5)
  - **Result**: AI correctly executed `date +"%Y-%m-%dT%H:%M:%S%z"` and captured actual local time `2026-01-29T14:19:02-0800`
  - **Verified**: Timestamp matches real clock time 

  
### DIFF
--- a/aidlc-rules/aws-aidlc-rules/core-workflow.md
+++ b/aidlc-rules/aws-aidlc-rules/core-workflow.md
@@ -452,11 +452,8 @@ The Operations stage will eventually include:
 - **CRITICAL**: ALWAYS append changes to EDIT audit.md file, NEVER use tools and commands that completely overwrite its contents
 - **CRITICAL**: Using file writing tools and commands that overwrite contents of the entire audit.md and cause duplication
 - When adding an entry to audit.md:
-  1. Run a system command to get the current timestamp:
-     - **macOS/Linux**: `date +"%Y-%m-%dT%H:%M:%S%z"`
-     - **Windows (PowerShell)**: `Get-Date -Format "yyyy-MM-ddTHH:mm:sszzz"`
-     - **Windows (CMD)**: `echo %date:~10,4%-%date:~4,2%-%date:~7,2%T%time:~0,2%:%time:~3,2%:%time:~6,2%%time:~9,2%`
-  2. Use the output exactly as the **Timestamp** value
+  1. Run a system command to get the current date and time in ISO 8601 format (YYYY-MM-DDTHH:MM:SS). Use whatever command is appropriate for the current OS and shell environment.
+  2. Use the command output exactly as the **Timestamp** value.
   Do not use placeholder (e.g., T00:00:00Z), estimated, or fabricated times.
 - Include stage context for each entry
 


aidlc-rules/aws-aidlc-rules/core-workflow.md
```diff
@@ -451,7 +451,8 @@ The Operations stage will eventually include:
 - **MANDATORY**: Record every user response with timestamp after receiving it
 - **CRITICAL**: ALWAYS append changes to EDIT audit.md file, NEVER use tools and commands that completely overwrite its contents
 - **CRITICAL**: Using file writing tools and commands that overwrite contents of the entire audit.md and cause duplication
-- Use ISO 8601 format for timestamps (YYYY-MM-DDTHH:MM:SSZ)
+- Use ISO 8601 format with local timezone offset (e.g., `2026-01-29T13:45:22-0800`)
+- **CRITICAL**: Execute a system command to get actual local time - NEVER fabricate or estimate timestamps
```
raj-jain-aws: We've excessive `CRITICAL` rules in the workflow at the moment. We believe this is causing the models to not pay attention or honor them. Can we avoid the `CRITICAL` keyword here and still get the intended outcome?

jimini55: Agreed on avoiding CRITICAL.

I tested different approaches - Claude worked fine without it, but Cursor often missed timestamps entirely or used wrong times. The fix was making the instruction more explicit and step-by-step:

  - When adding an entry to audit.md:
    1. Run `date +"%Y-%m-%dT%H:%M:%S%z"` to get the current timestamp
    2. Use the output exactly as the **Timestamp** value
    Do not use placeholder (e.g., T00:00:00Z), estimated, or fabricated times.

Tested with both Claude and Cursor - both now run the command and use actual output. Updated the PR.

aidlc-rules/aws-aidlc-rules/core-workflow.md
```diff
@@ -451,7 +451,12 @@ The Operations stage will eventually include:
 - **MANDATORY**: Record every user response with timestamp after receiving it
 - **CRITICAL**: ALWAYS append changes to EDIT audit.md file, NEVER use tools and commands that completely overwrite its contents
 - **CRITICAL**: Using file writing tools and commands that overwrite contents of the entire audit.md and cause duplication
-- Use ISO 8601 format for timestamps (YYYY-MM-DDTHH:MM:SSZ)
+- When adding an entry to audit.md:
+  1. Run a system command to get the current timestamp:
+     - **macOS/Linux**: `date +"%Y-%m-%dT%H:%M:%S%z"`
+     - **Windows (PowerShell)**: `Get-Date -Format "yyyy-MM-ddTHH:mm:sszzz"`
```
raj-jain-aws: This should help. One more request - we've run into situations where the Windows users don't have access to PowerShell (security enforcement in the org). Can we offer a Windows CMD variant also?

jimini55: @harmjeff - Could you help me on this?

harmjeff:
Findings for PR 56 that fixes the timestamp issue in audit

I have verified that it is working in Linux and Windows for:
- AmazonQ in VS Code
- Cline
- CoPilot (works as well as co-pilot ever does, which is badly)
- Claude
- Kiro (Ignored the requirement to get timestamps not just dates)
- Kiro-cli (Linux only)

Impact: Some clients ask permission to use system time functions (`date` and `Get-Date`)

aidlc-rules/aws-aidlc-rules/core-workflow.md
harmjeff: `echo %date:~10,4%-%date:~4,2%-%date:~7,2%T%time:~0,2%:%time:~3,2%:%time:~6,2%%time:~9,2%`

jimini55: Done

harmjeff:
So I've been reviewing.  Having 3 different instruction sets based on platform is very inefficient.

Would a python solution be acceptable?

`- When adding an entry to audit.md:
  1. Run this command to get the current timestamp (works on all platforms and shells):
     `python -c "from datetime import datetime; print(datetime.now().strftime('%Y-%m-%dT%H:%M:%S'))"`
  2. Use the output exactly as the **Timestamp** value
  Do not use placeholder (e.g., T00:00:00Z), estimated, or fabricated times.`

harmjeff:
Jimin is pushing the updates with the cleaned up rules for datestamps.

Here are the test result screen shots:
Claude Code VS Code Windows
Cline in VS Code Windows
Co-Pilot VS Code Windows
Cursor IDE Mac
Cursor IDE Windows
Kiro CLI Ubuntu
Kiro IDE on Mac
Kiro IDE Windows
Q in VS Code Windows

Jimin will add the remaining screen shots from the Mac.

harmjeff:
@raj-jain-aws ready for your approval

raj-jain-aws:
Nice! Thanks @jimini55 and @harmjeff for this. Long thread but concluded now.

PR #68, opened 2026-02-07
docs: add Windows CMD setup instructions and ZIP download note

### LINKED ISSUE #57: Add setup instructions for Windows CMD
The `mkdir -p`, and `cp -R` commands listed in README do not work on Windows CMD. Add setup commands for users with Windows CMD users. Also, some organizations block `git clone` access for security reasons. They're sent zipped version of the repo. The ZIP file structure is different and the setup commands need to be adjusted. Add instructions in README that we're now delivering ZIP as a build artifact that can be directly used.

inariku:
## Description

Add setup instructions for Windows CMD users and a note about ZIP download directory structure, as requested in #57.

### Changes

- Added Windows CMD equivalents (`mkdir`, `xcopy`) for both Kiro and Amazon Q Developer setup sections
- Added a note explaining that ZIP downloads from the Releases page may contain a top-level directory that users need to navigate into first

### Affected Files

- `README.md`

Closes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,15 @@ cp -R aidlc-rules/aws-aidlc-rules .kiro/steering/
 cp -R aidlc-rules/aws-aidlc-rule-details .kiro/
 ```
 
+On Windows (CMD):
+```cmd
+mkdir .kiro\steering
+xcopy aidlc-rules\aws-aidlc-rules .kiro\steering\aws-aidlc-rules\ /E /I
+xcopy aidlc-rules\aws-aidlc-rule-details .kiro\aws-aidlc-rule-details\ /E /I
+```
+
+> **Note**: If you downloaded a ZIP from the [Releases page](../../releases/latest), the extracted folder may contain a top-level directory (e.g., `aidlc-workflows-0.1.0/`). Navigate into it first so that `aidlc-rules/` is directly accessible.
+
 Your project should look like:
 ```
 <project-root>/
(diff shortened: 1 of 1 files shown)


inariku:
Hey @raj-jain-aws 👋 Just a quick reminder — could you use "Create a merge commit" when merging this one? Last time the squash merge didn't count me as a contributor. Thanks!

README.md
```diff
@@ -27,6 +27,15 @@ cp -R aidlc-rules/aws-aidlc-rules .kiro/steering/
...
 ```
 
+On Windows (CMD):
+```cmd
+mkdir .kiro\steering
+xcopy aidlc-rules\aws-aidlc-rules .kiro\steering\aws-aidlc-rules\ /E /I
+xcopy aidlc-rules\aws-aidlc-rule-details .kiro\aws-aidlc-rule-details\ /E /I
+```
+
+> **Note**: If you downloaded a ZIP from the [Releases page](../../releases/latest), the extracted folder may contain a top-level directory (e.g., `aidlc-workflows-0.1.0/`). Navigate into it first so that `aidlc-rules/` is directly accessible.
```
raj-jain-aws, line 37: This note should be moved up before the user issues the above above.

raj-jain-aws (approved):
@inariku Thanks for this. Tested these changes and they look good.

I've left a minor comment that can be addressed in the next PR. 

One issue that should also be addressed is that we're not clearly telling the user where to download the ZIP, in the project folder or elsewhere. Right now the instructions assume that it is in the project folder, then we should ask them to delete the original zip and its extracted folder after the copy commands to clean the clutter in project workspace. I think a safer approach might be to ask ZIP to be downloaded to an outside folder so user has a reference of the baseline rules. Let me know your thoughts and if this can be part of your next PR. Thanks!

PR #81, opened 2026-02-18
add workflow visualizer

roficas:
*Description of changes:*
I created this "visualizer" html - usefull to present to customers and analyze the steps, deliverables, questions, etc. 
HTML has a rich visualization and navigation (local anchors and dynamic layers). I used this to refresh customer's knowledge during the workshop and received great feedback. 

### DIFF
--- a/aidlc-workflow-visualizer.html
+++ b/aidlc-workflow-visualizer.html
@@ -1,0 +1,1486 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


harmjeff:
@roficas We like the idea but don't have anyone to support this longer term.

roficas:
we can put a Kiro hook to keep it current as we update the workflow. I actually updated and improved it 2 weeks ago. 
[aidlc-workflow-visualizer.html](https://github.com/user-attachments/files/28318902/aidlc-workflow-visualizer.html)

Issue #84, opened 2026-02-23
Kiro CLI: rule details path resolution broken after directory consolidation


rashimr:
After commit c6648db ("Cursor instructions and consistent directory handling for rule details"), Kiro CLI can no longer find the .aidlc-rule-details/ directory. The workflow fails
  immediately with:

  Tool validation failed:
  Failed to validate tool parameters: '<project>/.kiro/steering/aws-aidlc-rules/.aidlc-rule-details/common/welcome-message.md' does not exist

  Root Cause

  Kiro CLI resolves relative paths in core-workflow.md relative to the steering file location (.kiro/steering/aws-aidlc-rules/), not the project root.

  Before c6648db, core-workflow.md had platform-specific paths with a fallback order:

  - .aidlc-rule-details/ (for Cline)
  - .kiro/aws-aidlc-rule-details/ (for Kiro CLI)
  - .amazonq/aws-aidlc-rule-details/ (for Amazon Q)

  Check these paths in order and use the first one that exists.

  The Kiro-specific path .kiro/aws-aidlc-rule-details/ resolved correctly from the project root. The consolidation to a single .aidlc-rule-details/ path broke this because Kiro resolves it
   relative to the steering file, resulting in the lookup path: .kiro/steering/aws-aidlc-rules/.aidlc-rule-details/.

  Steps to Reproduce

  1. Clone/start a project
  2. Set up AI-DLC rules per the README:
```
  mkdir -p .kiro/steering
  cp -R /tmp/aidlc-workflows/aidlc-rules/aws-aidlc-rules .kiro/steering/
  mkdir -p .aidlc-rule-details
  cp -R /tmp/aidlc-workflows/aidlc-rules/aws-aidlc-rule-details/* .aidlc-rule-details/
```
  3. Start Kiro CLI and trigger the AI-DLC workflow
  4. Observe the path resolution error

  Workaround 

  Copy rule details into the steering directory so the relative path resolves:

  `cp -R .aidlc-rule-details .kiro/steering/aws-aidlc-rules/.aidlc-rule-details `
  
 Note: Kiro will search the directory and will find the file, but not a good user experience 

  Suggested Fix

  Either:
  - Restore the platform-specific fallback paths that existed before c6648db, or
  - Use absolute/project-root-relative paths in core-workflow.md that work regardless of where the steering file lives, or
  - Update the Kiro CLI README instructions to place .aidlc-rule-details/ inside the steering directory

  Environment

[...]

Issue #82, opened 2026-02-19
Incorrect GitHub Copilot setup instructions.


shchinnasamy:
Hello Team,

I suppose GitHub Copilot  setup instructions is incorrect. Could you confirm if
`.aidlc-rule-details` directory be created inside ` .copilot` directory? If so, I request to reflect corrections in [Option 1: .copilot Directory (Recommended)](https://github.com/awslabs/aidlc-workflows?tab=readme-ov-file#option-1-copilot-directory-recommended) snippet. 

Thank you!

raj-jain-aws:
@harmjeff has verified this. Pls. reopen if you're still encountering this issue.

PR #87, opened 2026-02-24
fix: correct GitHub Copilot instructions and Kiro CLI rule-details path resolution (#82, #84)

harmjeff:
# Summary

* Fix GitHub Copilot setup instructions to use the officially supported .github/copilot-instructions.md path instead of the non-functional .copilot/instructions.md and COPILOT.md approaches (#82)
* Fix Kiro CLI rule-details path resolution by updating README setup commands to place .aidlc-rule-details/ inside the steering directory, and restoring platform-specific fallback paths in core-workflow.md so the agent checks multiple locations (#84)
* Restore changes in README: moved Quick Start section, consolidate platform setup instructions with zip-based source paths, add Kiro IDE verification screenshots, and add an "Other Agents" section

### GitHub Copilot (#82)

The previous instructions directed users to place rules in .copilot/instructions.md or COPILOT.md, neither of which is recognized by GitHub Copilot. The correct approach per VS Code docs is .github/copilot-instructions.md, which is auto-detected and applied to all chat requests. Updated setup commands, verification steps, troubleshooting, and version control recommendations accordingly.

### Kiro CLI rule-details resolution (#84)
Kiro CLI resolves relative paths in steering files relative to the steering file location (.kiro/steering/aws-aidlc-rules/), not the workspace root. After commit c6648db consolidated rule-details to a single .aidlc-rule-details/ path, Kiro CLI could no longer find them because it resolved the path as .kiro/steering/aws-aidlc-rules/.aidlc-rule-details/.

Two changes fix this:
1. README: Updated Kiro CLI setup commands to place rule-details inside the steering directory where relative resolution finds them
1. core-workflow.md: Restored the original platform-specific fallback path order so the agent checks these locations and uses the first one that exists:
  * .aidlc-rule-details/ (Cursor, Cline, Claude Code, GitHub Copilot)
  * .kiro/aws-aidlc-rule-details/ (Kiro IDE and CLI)
  * .amazonq/aws-aidlc-rule-details/ (Amazon Q Developer)

## Test plan

[ ] Verify Kiro CLI can resolve rule-details files when set up per README instructions
[ ] Verify Kiro IDE can resolve rule-details files when set up per README instructions
[ ] Verify Amazon Q Developer can resolve rule-details files when set up per README instructions
[ ] Verify Cursor can resolve rule-details files via .aidlc-rule-details/
[ ] Verify GitHub Copilot auto-detects .github/copilot-instructions.md
[ ] Verify Cline and Claude Code still work with .aidlc-rule-details/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,213 +4,132 @@ AI-DLC is an intelligent software development workflow that adapts to your needs
 
 ## Table of Contents
 
-- [Tenets](#tenets)
-- [Prerequisites](#prerequisites)
-- [Get the AIDLC](#get-the-aidlc)
+- [Quick Start](#quick-start)
 - [Platform-Specific Setup](#platform-specific-setup)
 - [Usage](#usage)
 - [Three-Phase Adaptive Workflow](#three-phase-adaptive-workflow)
 - [Key Features](#key-features)
+- [Tenets](#tenets)
+- [Prerequisites](#prerequisites)
 - [Troubleshooting](#troubleshooting)
 - [Additional Resources](#additional-resources)
 
 ---
 
-## Tenets
+## Quick Start
 
-These are our core principles to guide our decision making.
-
-- **No duplication**. The source of truth lives in one place. If we add support for new tools or formats that require specific files, we generate them from the source rather than maintaining separate copies.
-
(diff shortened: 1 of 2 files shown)


raj-jain-aws (approved):
I've tested this PR in Kiro IDE, Kiro CLI, and Amazon Q and it is working. The workflow is reading the rules details folder files correctly in these three tools. The README instructions are now also aligned with the previous state of README.

Image

Image

Image

This PR addresses a breakage. It requires prompt action so it's being merged. If there are issues with other than the three tools mentioned above then those can be PRed separately.

PR #80, opened 2026-02-14
Add security extensions framework 

ai-ram-ramani:
## Overview
This PR introduces a comprehensive security extensions framework to the AI-DLC workflow, enabling enforcement of security best practices across all development phases.

### 1. Core Workflow Updates (`core-workflow.md`)
- Added mandatory Extensions Loading section with recursive scanning
- Each extension presents a qualifying question during requirements phase allowing users to enable/disable it per project
- Added blocking security finding behavior for non-compliant artifacts

### 2. Requirements Analysis Enhancement (`requirements-analysis.md`)
- Added Step 5.1: Extension Applicability Questions
- Collects user responses to extension qualifying questions
- Automatically records extension configuration in aidlc-state.md

### 3. Security Baseline Extension (`extensions/security/baseline/security-baseline.md`)
- Created 15 comprehensive OWASP Top 10 (2025) based security rules
- Covers encryption, logging, access control, authentication, input validation
- Includes verification criteria for each rule
- Implements blocking findings for non-compliant security issues
- Maps to OWASP categories: A01-A10

### 4. Extension Framework Structure
- Includes placeholder directories for compliance extensions (HIPAA, PCI-DSS, SOC2)
- Provides customer-specific extension directory for organization-specific rules
- Makes framework flexible and adaptable to diverse customer requirements

## Validation Results

Conducted comparative testing by building the same full-stack application twice:
- **First build**: AI-DLC without security extensions (frontend + backend on AWS)
- **Second build**: AI-DLC with security extensions enabled

**Security testing performed using:**
- AWS Security Agent for design review and penetration testing
- Amazon Inspector for code vulnerability scanning

**Results:**
- ✅ **70%+ reduction** in security findings when using security extensions
- ✅ OWASP Top 10 requirements automatically integrated into NFR requirements
- ✅ Security considerations properly reflected in application design artifacts
- ✅ Infrastructure design included appropriate security configurations
- ✅ Manual verification confirmed security controls in both code and AWS resource configurations

## Benefits
- Enforces security by default for production applications
- Allows flexibility for PoCs and prototypes via opt-in/opt-out mechanism
- Provides clear verification criteria at each workflow stage
- Aligns with industry-standard OWASP Top 10 framework
- Extensible for compliance and organization-specific requirements
- Proven to significantly reduce security vulnerabilities in generated applications

## Testing
Tested with full-stack application deployment on AWS with comprehensive security scanning.

PR #90, opened 2026-02-25
Address partial review comments on PR #80

raj-jain-aws:
*Address partial review comments on PR #80*

*Description of changes:* 

Addresses subset of review comments on https://github.com/awslabs/aidlc-workflows/pull/80

- .gitkeep files emptied
- Added Extensions mechanism description in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,57 +504,65 @@ Deployment and monitoring (future)
 
 ## Extensions
 
-AI-DLC supports an extension system that lets you layer additional rules on top of the core workflow. Extensions are markdown files placed in the `aws-aidlc-rule-details/extensions/` directory and are automatically loaded and enforced when enabled during the Requirements Analysis phase.
+AI-DLC supports an extension system that lets you layer additional rules on top of the core workflow. Extensions are markdown files organized under `aws-aidlc-rule-details/extensions/` and are automatically loaded and enforced when enabled during the Requirements Analysis phase.
 
 ### How Extensions Work
 
-The workflow currently comes pre-seeded with certain baseline security extensions.
-You may add your own extensions. For example, you can add security extensions for specific compliance standards within the `extensions/security/compliance` folder hierarchy.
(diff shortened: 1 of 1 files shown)


aidlc-rules/aws-aidlc-rule-details/extensions/security/compliance/hipaa/.gitkeep
```diff
@@ -1,5 +0,0 @@
-# HIPAA Compliance Extensions
```
ConnorKirk, line 1: My preference would be to remove the file entirely, in favour of documentation (with an example file structure) in the readme. Having partial folders around might be confusing.

raj-jain-aws: Thanks @ConnorKirk for the comment. I spoke to @ai-ram-ramani about this. He understands that these are empty directories at the moment but I believe there is something in works and there are some messaging objectives behind this.

ConnorKirk: That makes sense, but in the meantime, having these folders doesn't add much, and risks becoming clutter if there isn't follow up work (or it doesn't happen in the immediate future). It should be trivial to add these back when there is contents.

raj-jain-aws: Do not disagree @ConnorKirk . @ai-ram-ramani - Your thoughts?

ai-ram-ramani: @ConnorKirk @raj-jain-aws i thought about this a little more. I am ok with removing the directories for now and putting it into README.

raj-jain-aws:
Addressed review comments so far. Pls. review again.

PR #94, opened 2026-02-27
docs: add developer's guide for running CodeBuild locally

scottschreckengaust:
## Summary
- Adds `docs/DEVELOPERS_GUIDE.md` with instructions for running CodeBuild locally using the local agent
- Documents all `codebuild_build.sh` flags including `-b` for buildspec overrides, `-c` for AWS credentials, and `-e` for environment variables
- Adds `.codebuild/**` to `.gitignore` for local build artifacts

## Test plan
- [x] Verify markdown renders correctly on GitHub
- [x] Confirm `codebuild_build.sh` commands work locally with `-b` flag


scottschreckengaust:
An alternative is including instructions in this GitHub Wiki. This is viable, but will not be pulled down while developing and could be missed.

https://github.com/awslabs/aidlc-workflows/wiki

I would argue that the Wiki should be disabled.

raj-jain-aws:
> > `act` depends on `Go` runtime and requires `sudo`.
> 
> does not require `sudo`

Updated that in my comment. It was `Go` runtime installation that required `sudo`.

PR #96, opened 2026-03-01
fix: correct copy-paste error in error-handling.md

sungho-rpls:
## Summary

Fix misplaced content in `error-handling.md` where text from the "When to Suggest Starting Over" section appears to have been accidentally appended to the "Resumption Best Practices" section, likely due to a copy-paste error.

## Problem

Item 6 of "Resumption Best Practices" seems to have trailing content from another section:

```markdown
6. **Document recovery**: Log all recovery actions in audit.md State file is severely corrupted
- User requirements have changed significantly
- Architectural decision needs to be reversed

**Before Starting Over**:
1. Archive all existing work
...
```

These items appear to belong to the `Consider Fresh Start If` list and the `When to Suggest Starting Over` section, where similar content already exists.

### DIFF
--- a/aidlc-rules/aws-aidlc-rule-details/common/error-handling.md
+++ b/aidlc-rules/aws-aidlc-rule-details/common/error-handling.md
@@ -249,8 +249,17 @@
 **Consider Fresh Start If**:
 - Multiple phases have errors
 - State file is severely corrupted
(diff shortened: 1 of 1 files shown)


raj-jain-aws (approved):
Thanks for the fix. Looks good.

PR #135, opened 2026-03-20
docs: add Agent Skill setup option for Claude Code

w0yne:
## Summary

Add **Option 3: Agent Skill** under the Claude Code section of README, showing how to package AI-DLC as an [Agent Skill](https://agentskills.io).

## What Changed

**README.md only** — no new files, no content duplication.

The new option shows how to assemble a skill directory from the existing `aidlc-rules/` content:
- Creates `SKILL.md` with YAML frontmatter + `core-workflow.md` content (path refs updated via sed)
- Copies `aws-aidlc-rule-details/` as `references/`

## Why

- Enables `/aidlc` slash command in Claude Code
- Follows the open [Agent Skills standard](https://agentskills.io) — also works with Cursor, Gemini CLI, OpenCode, etc.
- No repo structure changes needed — just a documentation addition alongside the existing Option 1/Option 2

## Pre-built Skill

A ready-to-use pre-built version of this skill is available at [w0yne/aidlc-claude-code-skill](https://github.com/w0yne/aidlc-claude-code-skill), which contains the assembled `aidlc/` directory (SKILL.md + references/) that can be directly copied to `~/.claude/skills/` or `.claude/skills/`.

## Testing

Verified the skill works correctly with Claude Code in an empty project:
- `/aidlc` slash command triggers the full AI-DLC workflow
- Workspace Detection → Requirements Analysis stages execute as expected
- `aidlc-state.md`, `audit.md`, and structured question files are generated correctly
- Security Extension opt-in from `references/extensions/` is recognized

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,57 @@ xcopy "%USERPROFILE%\Downloads\aidlc-rules\aws-aidlc-rule-details" ".aidlc-rule-
     └── operations/
 ```
 
+#### Option 3: Agent Skill
+
+Package AI-DLC as an [Agent Skill](https://agentskills.io) for Claude Code. This enables the `/aidlc` slash command and automatic activation on development tasks. Also compatible with other Agent Skills-compliant tools (Cursor, Gemini CLI, OpenCode, etc.).
+
+**Unix/Linux/macOS:**
+```bash
+# Create the skill directory
+mkdir -p ~/.claude/skills/aidlc/references
+
+# Create SKILL.md with frontmatter + core workflow
+cat > ~/.claude/skills/aidlc/SKILL.md << 'EOF'
+---
+name: aidlc
+description: >-
+  AI-DLC (AI-Driven Development Life Cycle) adaptive workflow for software
+  development. Use when building new features, creating applications,
+  refactoring systems, or any software development task that benefits from
+  structured planning, requirements analysis, and phased execution with
+  quality gates.
+---
+
+EOF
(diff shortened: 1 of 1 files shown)


leandrodamascena:
Hi @w0yne, thanks for this. The Agent Skills standard is interesting and the PR is clean.

We're not sure about this approach though. A few concerns:

1. **Drift.** Once someone assembles a SKILL.md from `core-workflow.md` content, that copy is frozen in time. When we update the core files (and we do frequently), everyone who built a skill from them would be running an outdated version without knowing it.

2. **Regeneration burden.** Every time we change core files, users would need to re-run the sed/copy steps to regenerate their skill directory. There's no mechanism to notify them that their local skill is out of date.

3. **Implicit versioning problem.** With the current copy/symlink approach, symlinks point to the repo and get updates automatically. With a generated skill, there's no link back to the source — no way to tell which version of AI-DLC someone is actually running.

I appreciate the effort here, but I'm not sure this is the right path given the concerns above.

@raj-jain-aws @harmjeff @scottschreckengaust - thoughts?

raj-jain-aws:
Hi @w0yne, thanks for your contribution. The AI-DLC v1 implementation (the main branch)  is pretty much frozen right now. The AI-DLC v2 design is natively skills based. Claude code skills for v2 distribution will be added soon, and we'll forward to your contributions there.

Agreed @leandrodamascena on closing this for now.

PR #39, opened 2026-01-10
feat(cli): add comprehensive CLI tool with multi-platform AI assistan…

lsiten:
…t support

- Add complete TypeScript-based CLI implementation with Commander.js framework
- Add five core commands: init, setup, status, validate, and version
- Add service layer with AIDLC service and file management utilities
- Add comprehensive test suite with Jest configuration and initial test coverage
- Add ESLint configuration for TypeScript code quality enforcement
- Add package.json with all required dependencies and npm scripts
- Add build and publish scripts for distribution and CI/CD pipeline
- Add extensive documentation: CHANGELOG.md, EXAMPLES.md, QUICK-START.md, README-CLI.md
- Add platform integration documentation and installation success guide
- Add project summary and platform integration reference materials
- Add executable bin/aidlc entry point for global CLI access
- Update .gitignore to exclude dist and node_modules directories
- Update main README.md with CLI integration information
- Add TypeScript configuration with strict type checking
- Add package-lock.json for dependency version locking

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #166, opened 2026-04-01
feat: add interactive CLI installer for multi-agent rule setup

inariku:
# Summary

Adds an interactive CLI tool that installs AI-DLC rules to the correct paths for each supported agent, matching the README instructions exactly.

Closes #165

## Changes

- Added `cli/` directory with an npm package (`@inariku/aidlc-install`)
- Single file CLI (`src/index.mjs`) using `@clack/prompts` for interactive UI
- Supports all 6 documented agents: Kiro, Amazon Q, Cursor, Cline, Claude Code, GitHub Copilot
- Install paths match the README exactly for each agent
- Features: agent auto-detection, symlink/copy modes, lock file, uninstall, CI flags

## User experience

Before: Users manually run multiple `mkdir` and `cp` commands per agent, following README instructions.

After:
```bash
npx @inariku/aidlc-install
```
Interactive prompt auto-detects agents and installs rules to the correct paths.

## Checklist

* [x] Changes are documented

## Test Plan

1. Run `npx @inariku/aidlc-install` in a project directory
2. Select agents interactively
3. Verify files are installed to correct paths per README
4. Run `npx @inariku/aidlc-install list` to confirm tracking
5. Run `npx @inariku/aidlc-install remove` to clean up
6. Tested all 6 agents on macOS with both symlink and copy modes

## Acknowledgment


leandrodamascena:
Hi @inariku, thanks for the effort here. The CLI experience is well thought out and the code is clean.

Unfortunately we're going to close this. We've had similar PRs in the past (#39, #53, #308) that proposed installation scripts and tooling, and we closed all of them. We're not looking at tooling right now, it's something we're considering but nothing has been decided yet. 

From what I see in the code, there are also cross-platform concerns we'd need to address: symlinks on Windows require admin privileges or Developer Mode enabled, path separators and home directory resolution differ across OSes, etc. These are solvable but it's one more reason why we want to think this through carefully before committing to shipping tooling.

The install steps today are intentionally simple (copy or symlink markdown files) and for now we prefer keeping them as documentation in the README.

Thanks for the contribution. When we have a clearer picture of how we see tooling fitting in, we'll publish an RFC, no timeline for that yet though.

PR #178, opened 2026-04-08
fix: forks skip codebuild

scottschreckengaust:
# Summary

Forks shall not access secrets, so skip the "build".

## Changes

* Add logic when there are forks to skip running Code Build.
* Add comment on the purpose of the `env.ACT`

## User experience

## Checklist

* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented

## Test Plan

## Acknowledgment

### DIFF
--- a/.github/workflows/codebuild.yml
+++ b/.github/workflows/codebuild.yml
@@ -122,9 +122,13 @@ jobs:
           fi
 
   build:
+    # Fork PRs are skipped because they cannot access the repository secrets
+    # or OIDC credentials needed for AWS CodeBuild.
     if: >-
-      github.event_name != 'pull_request'
-      || contains(github.event.pull_request.labels.*.name, 'rules')
+      (github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/codebuild.yml
```diff
@@ -414,6 +420,7 @@ jobs:
           archive: false
 
       - name: Upload Evaluation Report
+        # env.ACT is set by the 'act' CLI tool for local testing
```
Kalindi-Dev, line 423: Do we need this comment everywhere? One should be okay?

MichaelWalker-git (approved):

scoropeza (approved):
LGTM, the duplication of the "act" comment on each line where env.ACT is mentioned could be redundant, yes. But its just a matter of preference.

.github/workflows/codebuild.yml
scoropeza: the duplication of the "act" comment on each line where env.ACT is mentioned could be redundant, yes. Its just a matter of preference.

Issue #72, opened 2026-02-09
Add unit context to audit.md logging format for multi-unit workflow traceability


junsjang:
## Problem

When working on projects with multiple units in the Construction phase, the audit.md file lacks unit-specific context, making it difficult to trace which interactions belong to which unit. This becomes particularly problematic when merging work from parallel unit development or analyzing the audit trail retrospectively.

## Observed Behavior

Current audit.md logging format as defined in `core-workflow.md`:

```
## [Stage Name or Interaction Type]
**Timestamp**: [ISO timestamp]
**User Input**: "[Complete raw user input - never summarized]"
**AI Response**: "[AI's response or action taken]"
**Context**: [Stage, action, or decision made]
```
The Context field is defined generically as "Stage, action, or decision made" without explicit guidance to include unit identification. When multiple units are processed:

- All unit interactions are logged to a single `audit.md` file
- No clear indication of which unit each log entry belongs to
- Difficult to filter or analyze interactions by unit
- Challenging to understand the sequence of events for a specific unit

## Root Cause

- Missing specification: No explicit requirement to include unit name in audit log entries
- Generic context definition: The Context field description doesn't mandate unit identification
- Documentation gap: Construction phase files (`functional-design.md`, `code-generation.md`, etc.) only specify "Log approval in audit.md with timestamp" without unit context requirements
- Single file design: While `audit.md` is intentionally a single file for complete audit trail, the format doesn't accommodate multi-unit scenarios

### Update Context field specification in core-workflow.md

Modify the existing "Prompts Logging Requirements" section in `core-workflow.md` to clarify Context field usage:

**Current text:**

```markdown
- Include stage context for each entry
```

**Updated text:**

```markdown
- Include stage context for each entry
- **For Construction phase per-unit stages**: Include unit name in Context field (e.g., "CONSTRUCTION - Code Generation - Unit: user-service")
```

### Add one example in core-workflow.md

Add a single example showing unit context usage in the existing "Audit Log Format" section:

```markdown
### Audit Log Format:
```markdown
## [Stage Name or Interaction Type]
**Timestamp**: [ISO timestamp]
**User Input**: "[Complete raw user input - never summarized]"
**AI Response**: "[AI's response or action taken]"
**Context**: [Stage, action, or decision made]

---
```

**Example with unit context:**

```markdown
## Code Generation
**Timestamp**: 2024-02-09T14:30:00Z
**User Input**: "Approve code generation plan"
[...]

PR #185, opened 2026-04-15
feat: add feature-folder isolation for multi-feature AI-DLC workflows

mpearcy13:
# Summary

Users that want to run AI-DLC multiple times in a repo are currently blocked from syncing the AI-DLC files to Git, as they create merge conflicts. To mitigate this, all generated documentation is now scoped under aidlc-docs/{feature-name}/ instead of a flat aidlc-docs/ directory, allowing retention of AI-DLC context per feature and minimizing merge conflicts.

## Changes

Core Workflow (`core-workflow.md`)

- Added "MANDATORY: Feature Name Capture and Path Resolution" section — prompts for feature name on new workflows, detects from existing state on resume, and offers migration for legacy flat
- Added Global Path Resolution Rule — all aidlc-docs/ paths implicitly resolve to aidlc-docs/{feature-name}/ throughout the workflow and rule detail files
- Updated Workspace Detection stage to include feature name capture as an explicit sub-step
- Updated directory structure diagram to reflect {feature-name}/ nesting level

Workspace Detection (`inception/workspace-detection.md`)

- Rewrote Step 1 to scan aidlc-docs/ for subdirectories containing aidlc-state.md (handles single feature, multiple features, and legacy flat structure)
- Added new Step 2 for feature name capture with sanitization rules (lowercase, alphanumeric + hyphens, max 50 chars)
- Updated all subsequent steps with aidlc-docs/{feature-name}/ paths and renumbered accordingly

Session Continuity (`common/session-continuity.md`)

- Added "Feature Detection on Resume" section with auto-select, multi-feature list, and legacy migration logic
- Updated Welcome Back prompt to show Resuming feature: {feature-name}
- Updated session continuity instructions to scan for feature folders before loading state

Error Handling (`common/error-handling.md`)

- Added "Legacy flat structure detected" error case with migration guidance
- Updated existing error recovery paths to reference aidlc-docs/{feature-name}/

Workflow Changes (`common/workflow-changes.md`)

- Updated pause/resume handling to reference feature-scoped paths and scan for feature folders on resume

Reverse Engineering (`inception/reverse-engineering.md`)

- Updated state tracking paths to aidlc-docs/{feature-name}/

Workflow Planning (`inception/workflow-planning.md`)

- Updated state tracking paths and added Feature Name field to state template

## User Experience

Users are prompted for a feature name when starting the AI-DLC workflow. This creates a partition under aidlc-docs/{feature-name}/, allowing users to keep AI-DLC documents in the repo and run multiple iterations in the same repo while maintaining historical context. On resume, existing features are auto-detected; legacy flat structures are offered a migration path.

## Checklist

* [x] Changes are documented

## Test Plan

1. Create a new AI-DLC project and initialize the AI-DLC workflow.
2. When prompted, enter a feature name. Verify the aidlc-docs/{feature-name}/ directory is created with audit files and phase folders.
3. Complete or pause the workflow, then resume. Verify the feature is auto-detected and state is restored correctly.
4. Start a second AI-DLC run with a different feature name. Verify both feature folders coexist under aidlc-docs/ without conflicts.
5. Test legacy migration: create a flat aidlc-docs/aidlc-state.md structure, start the workflow, and verify the migration prompt moves artifacts into a named feature folder.
6. Test feature name sanitization: enter names with uppercase, special characters, and >50 chars. Verify they are slugified correctly.

## Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ If your agent has no convention for rules files, place both folders at your proj
 4. Carefully review every plan that AI generates. Provide your oversight and validation
 5. Review the execution plan to see which stages will run
 6. Carefully review the artifacts and approve each stage to maintain control
-7. All the artifacts will be generated in the `aidlc-docs/` directory
+7. All the artifacts will be generated in the `aidlc-docs/{feature-name}/` directory
 
 ---
 
--- a/aidlc-rules/aws-aidlc-rule-details/common/error-handling.md
+++ b/aidlc-rules/aws-aidlc-rule-details/common/error-handling.md
@@ -43,7 +43,12 @@
 **Error**: Existing `aidlc-state.md` is corrupted
 - **Cause**: Manual editing, incomplete previous run
 - **Solution**: Ask user if they want to start fresh or attempt recovery
-- **Recovery**: Create backup, start new state file
+- **Recovery**: Create backup, start new state file under `aidlc-docs/{feature-name}/`
+
+**Error**: Legacy flat structure detected
+- **Cause**: Project created before feature-folder isolation was introduced
+- **Solution**: Prompt user for a feature name and migrate all artifacts from `aidlc-docs/` into `aidlc-docs/{feature-name}/`
+- **Recovery**: Move files, update state file with feature name, continue workflow
 
 **Error**: Cannot determine required stages
 - **Cause**: Insufficient information from user
--- a/aidlc-rules/aws-aidlc-rule-details/common/session-continuity.md
+++ b/aidlc-rules/aws-aidlc-rule-details/common/session-continuity.md
@@ -1,13 +1,45 @@
 # Session Continuity Templates
 
+## Feature Detection on Resume
+When a user returns, scan `aidlc-docs/` for existing feature folders (subdirectories containing `aidlc-state.md`):
+
+**Single feature found**: Auto-select and present the Welcome Back prompt below.
+
+**Multiple features found**: Present a feature selection prompt first:
+```markdown
+**Welcome back! I found multiple AI-DLC features in progress.**
+
+Available features:
+1. **{feature-name-1}** - [Current Stage from aidlc-state.md]
+2. **{feature-name-2}** - [Current Stage from aidlc-state.md]
+...
+
+A) Continue feature 1: {feature-name-1}
+B) Continue feature 2: {feature-name-2}
+...
+N) Start a new feature
+
+[Answer]: 
+```
+
+**Legacy flat structure found** (aidlc-state.md directly in aidlc-docs/ with no feature subfolder):
+```markdown
+**Welcome back! I found an existing AI-DLC project using the legacy flat structure.**
+
+To support concurrent feature development, I need to migrate your existing docs into a feature folder.
+
+What name should I use for this feature? (e.g., "User Authentication", "Payment Processing")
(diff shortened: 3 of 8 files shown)


dreamorosi:
Hi @mpearcy13 - thank you for opening this pull request.

In case I have missed it, could you please link the issue where you discussed this change with maintainers (this is mentioned in the contributing guidelines, [bullet point no3 in this list](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md#contributing-via-pull-requests))?

I'd like to get more context since I'm unsure we should be adding this at this point, while the feature is definitely something we should be doing - the team is working on a new iteration of the aidlc-workflow rules that already incorporates this idea but that we're not yet ready to share with the public.

augie09:
> Hi @mpearcy13 - thank you for opening this pull request.
> 
> In case I have missed it, could you please link the issue where you discussed this change with maintainers (this is mentioned in the contributing guidelines, [bullet point no3 in this list](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md#contributing-via-pull-requests))?
> 
> I'd like to get more context since I'm unsure we should be adding this at this point, while the feature is definitely something we should be doing - the team is working on a new iteration of the aidlc-workflow rules that already incorporates this idea but that we're not yet ready to share with the public.

Is there a rough timeframe when this will be shared?  Weeks or longer?
I put in a similar request today, then found this one.  https://github.com/awslabs/aidlc-workflows/issues/208

mpearcy13:
This is the issue I was working to resolve.
https://github.com/awslabs/aidlc-workflows/issues/72

dreamorosi:
@augie09 we're planning on sharing news on the new version soon - no exact date yet, but it's going to be (few) weeks and not months.

@mpearcy13 - understood, the issue was already assigned there was no further discussion with the maintainers. I'm inclined to leave the `do-not-merge` label and/or close it, but we're open to discuss whether this is still needed once we have the new version out.

leandrodamascena:
Hi @mpearcy13, thanks for this. The feature-folder approach makes sense for isolating multiple executions.

This PR addresses part of what #126 covers. The indicated issue is broader and includes both sequential re-runs and parallel execution in monorepos by different teams. The demand for this has been high so we want to prioritize it.

@raj-jain-aws @harmjeff @scottschreckengaust do you think we can leverage any of the work in this PR to move #126 forward? The feature-folder isolation approach here could be a good starting point for the namespacing piece.

leandrodamascena:
Hey there — thanks for this, and for the clear write-up. The problem you found is real: a flat `aidlc-docs/` makes it impossible to keep AI-DLC artifacts in the repo across multiple runs without merge conflicts.

v2 already solves this with **intents**. Each run gets its own record directory at `aidlc/spaces/<space>/intents/<YYMMDD>-<label>/`, so artifacts from different work never collide and each one keeps its history. There's also automatic migration from the old flat layout, and customers running several workflows in one repo are already on this model.

Since it's the same goal with different vocabulary (intent rather than feature), I'd rather not carry the v1 version forward — two partitioning schemes would mean solving the migration problem twice.

Would you take a look at v2 and tell us if intents cover your case? If something in your design handles a situation they don't, that's worth an issue against v2.

Thanks again, and sorry it landed on the branch being superseded.

PR #202, opened 2026-04-21
feat: detect and flag infrastructure failures in trend reports

Kalindi-Dev:
## Summary

- When Bedrock is unavailable during evaluation (throttling, service outages), runs produce 0 test passes and the trend report incorrectly signals a regression. This change adds infrastructure failure detection so the gate skips unreliable runs instead of firing false regressions.
- Adds `InfraFailureReason` enum and `InfraFailure` dataclass with conservative detection logic covering: Bedrock throttling, service unavailable, model errors, run failures/crashes, missing metrics, and server start failures.
- The regression gate now passes with an annotation when the latest run is an infra failure, and falls back to an older clean run when the comparison run is an infra failure.
- Both MD and HTML renderers show a prominent warning banner at the top of the report, and Section F is expanded with per-error-type columns (throttle, svc unavail, model error) and an infra failure flag.

## Test plan

- [x] All 174 existing + new unit tests pass (`uv run pytest packages/trend-reports/tests/ -v`)
- [x] Smoke tested with synthetic infra failure bundle — detection, gate skip, and report rendering all work correctly
- [ ] Verify trend report renders correctly in CodeBuild pipeline when Bedrock is healthy (no false positives)
- [ ] Verify trend report correctly flags infra failure when Bedrock is degraded

### DIFF
--- a/scripts/aidlc-evaluator/packages/trend-reports/src/trend_reports/__init__.py
+++ b/scripts/aidlc-evaluator/packages/trend-reports/src/trend_reports/__init__.py
@@ -19,6 +19,8 @@ from trend_reports.gate import check_regressions
 from trend_reports.models import (
     BaselineMetrics,
     GateResult,
+    InfraFailure,
+    InfraFailureReason,
     RunData,
     RunType,
     SemVer,
@@ -33,6 +35,8 @@ from trend_reports.render_yaml import render_trend_yaml
 __all__ = [
     "BaselineMetrics",
     "GateResult",
+    "InfraFailure",
+    "InfraFailureReason",
     "RunData",
     "RunType",
     "SemVer",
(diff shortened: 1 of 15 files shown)


Kalindi-Dev:
Have addressed Semgrep findings on the PR


   **`gitlab.bandit.B101` — 25 findings (false positives, no action)**
   These flag `assert` usage in test files (`test_*.py`). This is standard pytest
   practice — `assert` is the intended mechanism for test assertions. The rule doesn't
   distinguish test code from production code.

   **`is-function-without-parentheses` — 5 findings (fixed in c9b5b16)**
   Semgrep flagged `assert result.is_infra_failure is True` thinking `is_infra_failure`
    might be a method that needs `()`. It's actually a `bool` field on a dataclass.
   Fixed by using idiomatic form: `assert result.is_infra_failure` / `assert not
   result.is_infra_failure`."2>&1

Kalindi-Dev:
Fixed Semgrep finding: python.lang.correctness.common-mistakes.string-concat-in-list.string-concat-in-list

scottschreckengaust (approved):
LGTM

harmjeff (approved):
LGTM

PR #239, opened 2026-05-04
feat(extensions): B+C extensions framework with NIST 800-53 compliance extension

ClintEastman02:
## Summary

Implements the hybrid B+C extensions model proposed in #225:
- **Proposal B**: NIST 800-53 ships as official compliance extension alongside security-baseline and property-based-testing
- **Proposal C**: Empty `community-extensions-index.yaml` establishes the decentralized extension contract for future community contributions
- **Selection UX**: Consolidated menu replaces per-extension opt-in prompts — scales to N extensions with a single question

## What Changed

| File | Change |
|---|---|
| `inception/requirements-analysis.md` | Step 5.1: per-extension opt-in -> consolidated selection menu |
| `extensions/README.md` | Combined with NIST README, lists all available extensions |
| `extensions/compliance/nist-800-53/` | New official NIST 800-53 extension (opt-in + controls) |
| `community-extensions-index.yaml` | Empty index template at repo root for Proposal C |
| `core-workflow.md` | Added `.aidlc/` path for AI-assisted setup |

## NIST 800-53 Extension

~90 control mappings across 3 NIST families mapped to specific AWS mechanisms:

| Family | Controls | AWS Mechanisms |
|---|---|---|
| **AC** (Access Control) | AC-3, AC-4, AC-6, AC-12, AC-17, AC-22 + enhancements | SCPs, Declarative Policies, Config Rules, Security Hub |
| **AU** (Audit and Accountability) | AU-2, AU-3, AU-5(2), AU-6, AU-12 | CF Hooks, Config Rules, Security Hub |
| **CA** (Assessment and Monitoring) | CA-3, CA-7 | CF Hooks, Config Rules (GuardDuty) |

The model selectively applies controls relevant to each project — a network isolation project triggers AC-4/AC-17; a logging project triggers AU-2/AU-12.

## Extension Selection Menu

Instead of N sequential yes/no prompts, all opt-in extensions are presented in one table:

```
| # | Extension          | Category   | Description                       |
|---|--------------------|------------|-----------------------------------|
| 1 | Security Baseline  | security   | OWASP-mapped security rules       |
| 2 | NIST 800-53        | compliance | NIST 800-53 controls mapped to AWS|

Enter numbers to enable (comma-separated), "all", or "none".
```

## Alignment with #225

| #225 Consensus | This PR |
|---|---|
| Scott: "Proposal B + Proposal C" | Official extensions in `extensions/`, community via index |
[...]

Issue #225, opened 2026-04-30
RFC: Community Extensions Framework


scottschreckengaust:
# RFC: Community Extensions Framework

## Summary

Introduce a "Community Extensions" system that allows external contributors to author, publish, and share AI-DLC extensions (e.g., HIPPA, PCI DSS, SOX, FedRAMP, Accessibility/WCAG, Disaster Recovery, ISO 27001) that are **not** part of the core out-of-the-box distribution but can be discovered, vetted, and installed by users.

## Motivation

The current extension system (`extensions/security/baseline/` and `extensions/testing/property-based/`) ships two built-in extensions in the release zip. As AI-DLC adoption grows, organizations will need domain-specific compliance and quality extensions that:

- Are too specialized for the core distribution (HIPPA applies only to healthcare, PCI only to payment systems)
- Require domain expertise the core team may not have
- Need independent versioning and release cadence
- Benefit from community contribution and iteration

**Example extensions that could be community-contributed:**
- HIPPA (healthcare PHI handling)
- PCI DSS v4.0 (payment card security)
- SOX Level 1 / Level 2 (IT general controls)
- FedRAMP Moderate/High (US federal cloud)
- WCAG 2.2 AA Accessibility
- Disaster Recovery / Business Continuity
- ISO 27001 (information security management)
- GDPR (EU data protection)
- SOC 2 Type II controls

## Design Constraints

- Extensions are Markdown (no code execution) — low security risk, high quality risk
- Any solution must work across all supported platforms (Kiro, Q, Cursor, Cline, Claude Code, Copilot)
- Methodology first — no install tooling *required* (but optional tooling is acceptable)
- Must provide trust signals (maturity, review status) since compliance extensions carry legal weight
- Must not bloat the core release zip
- **Multi-extension repositories**: A single repository may host multiple extensions (e.g., an org publishes HIPPA, SOX, and PCI under one repo). The index must point to a specific path within a repo, not assume one extension per repo.
- **Private extensions**: Enterprises must be able to use private extensions from internal repos that are never published to the public index.

---

## Proposal A: Separate Registry Repository ("Extension Marketplace")

**Inspiration**: Terraform Registry, Homebrew core, GitHub Actions Marketplace

### Architecture

```text
github.com/awslabs/aidlc-community-extensions/
├── registry.yaml                    # Central index of all extensions
├── CONTRIBUTING.md                  # Submission & review process
├── EXTENSION_TEMPLATE/              # Starter template
[...]

PR #240, opened 2026-05-05
feat: Make security baseline extension default-enabled (opt-out instead of opt-in)

ai-ram-ramani:
## Summary

Changes the security baseline extension from opt-in to default-enabled (opt-out). Security rules are now loaded and enforced from workflow start. Users can explicitly disable them during Requirements Analysis if needed (e.g., for throwaway PoCs).

## Motivation

Security should be on by default — requiring users to opt in means most projects run without security rules unless someone actively remembers to enable them. This change flips the default so the path of least resistance is the secure path.

### `security-baseline.opt-in.md`
- Added `**Default**: Enabled` metadata flag
- Changed prompt from opt-in multiple choice (A=Yes, B=No) to opt-out informational (type disable to turn off, otherwise leave blank)
- Added Loading Behavior section documenting immediate load at workflow start

### `core-workflow.md` (Extensions Loading section)
- Added step 3: check each `*.opt-in.md` for `**Default**: Enabled`
- If present: load full rules file immediately (default-enabled behavior)
- If not present: existing deferred opt-in behavior unchanged
- Added Default-Enabled Extensions subsection documenting the opt-out flow
- No-answer behavior: default-enabled extensions stay enabled

### `requirements-analysis.md` (Step 5.1)
- Renamed from Extension Opt-In Prompts to Extension Opt-In / Opt-Out Prompts
- Now handles both flows: opt-out for default-enabled, opt-in for standard
- State tracking table includes Default column (On/Off)
- Documented no-answer behavior for both extension types

## Backward Compatibility

- Property-based testing extension is unchanged (still opt-in)
- Any existing extension without `**Default**: Enabled` in its opt-in file behaves exactly as before
- The `**Default**: Enabled` flag is additive — only extensions that declare it get the new behavior

## Testing

Validated end-to-end in Kiro IDE:

**Test 1: Default behavior (keep enabled)**
1. Workflow starts → security rules loaded immediately ✅
2. Requirements Analysis presents opt-out question ✅
3. User leaves answer blank → security stays enabled ✅
4. Requirements document includes security rule applicability mapping ✅

**Test 2: Explicit opt-out (disable)**
1. Workflow starts → security rules loaded immediately ✅
2. Requirements Analysis presents opt-out question ✅
3. User types "disable" → extension marked as disabled in aidlc-state.md ✅
4. Subsequent stages do not enforce security rules ✅
5. Audit log records the opt-out decision ✅

---

### DIFF
--- a/aidlc-rules/aws-aidlc-rule-details/extensions/security/baseline/security-baseline.opt-in.md
+++ b/aidlc-rules/aws-aidlc-rule-details/extensions/security/baseline/security-baseline.opt-in.md
@@ -1,18 +1,25 @@
-# Security Baseline — Opt-In
+# Security Baseline — Default Enabled (Opt-Out)
 
 **Extension**: Security Baseline
+**Default**: Enabled — rules are enforced unless the user explicitly opts out.
 
-## Opt-In Prompt
+## Opt-Out Prompt
 
 The following question is automatically included in the Requirements Analysis clarifying questions when this extension is loaded:
 
 ```markdown
 ## Question: Security Extensions
-Should security extension rules be enforced for this project?
+Security baseline rules (OWASP-based) are **enabled by default** for this project. No action is needed to keep them active.
 
-A) Yes — enforce all SECURITY rules as blocking constraints (recommended for production-grade applications)
-B) No — skip all SECURITY rules (suitable for PoCs, prototypes, and experimental projects)
-X) Other (please describe after [Answer]: tag below)
+If you want to disable security rules (e.g., for a throwaway PoC or experimental prototype), type "disable" below. Otherwise, leave blank or type "keep".
 
 [Answer]: 
 ```
+
+## Loading Behavior
+
+Unlike standard opt-in extensions, this extension's full rules file (`security-baseline.md`) is loaded immediately at workflow start — before the user answers the opt-out question. This ensures security rules are enforced from the earliest stages.
+
+If the user types "disable" during Requirements Analysis, the extension is marked as disabled in `aidlc-docs/aidlc-state.md` and enforcement stops from that point forward.
+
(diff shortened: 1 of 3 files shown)


scottschreckengaust:
The contributor statement is missing. Extensions being on by default (opt-out) vs. optionally enabled (opt-in) is interesting. The RFC for extensions and PR to choose which extensions are enabled in one question should be reviewed:

* #225 -> #239

aidlc-rules/aws-aidlc-rule-details/extensions/security/baseline/security-baseline.opt-in.md
```diff
@@ -1,18 +1,25 @@
...
-B) No — skip all SECURITY rules (suitable for PoCs, prototypes, and experimental projects)
-X) Other (please describe after [Answer]: tag below)
+If you want to disable security rules (e.g., for a throwaway PoC or experimental prototype), type "disable" below. Otherwise, leave blank or type "keep".
 
 [Answer]: 
 ```
+
+## Loading Behavior
+
+Unlike standard opt-in extensions, this extension's full rules file (`security-baseline.md`) is loaded immediately at workflow start — before the user answers the opt-out question. This ensures security rules are enforced from the earliest stages.
```
raj-jain-aws, line 21: Why should it be loaded "immediately at workflow start"? Shouldn't the loading behavior still be the same, just that opt-in language becomes opt-out language?

ai-ram-ramani: sorry i don't understand. We want it to be security by default, so always loaded during workspace detection but if opted out then security rules are not considered. It did work for me, but I don't think it undid the context add.

raj-jain-aws:
Discussed with @ai-ram-ramani that this will be handled by stronger guidance language in the security opt-in question

Issue #299, opened 2026-05-22
[Feature]: Propagate build-time discoveries (rate limits, API quirks, pivots) across units, independent of declared dependencies


Mameta29:
## Description

Construction runs as a per-unit loop. Today, cross-unit context is shared **only** by following the *statically declared* dependency matrix (`unit-of-work-dependency.md`, produced back in Units Generation). That matrix captures **design-time** dependencies.

**Runtime discoveries** made *while building* a unit — undocumented rate limits, API quirks, environment constraints, or pivots that weren't foreseeable at design time — never appear in that matrix. A unit that touches the same external resource **without a declared dependency** loads none of the prior unit's findings, so its agent repeats the same investigation and the same pivot independently.

This is a different layer from the static dependency graph: the graph models *who-depends-on-whom by design*, while this is about *what-was-learned-at-build-time that crosses unit boundaries regardless of declared dependencies*. The two are orthogonal — this proposal adds a layer on top of the dependency matrix, it does not change it.

> **Origin & attribution.** This was raised by @mayakost while reviewing #276. We agreed there that it is orthogonal to that PR (a session-resume path fix) and deserves its own issue. The example below is @mayakost's, carried over verbatim; filing on their behalf with credit as co-reporter.

## Use Case

A static dependency graph cannot represent build-time discoveries by design, so without a dedicated channel the same external-resource lesson gets re-learned (and re-paid-for in tokens, time, and risk of an *inconsistent* second pivot) by every unit that independently touches that resource.

**Concrete example (from @mayakost):**

> Unit 1 is a Notification module. While building it, the agent tries to use the platform's email API v2, discovers it's rate-limited to 10 req/s (undocumented), and pivots to batching with API v1. Unit 2 is a User Onboarding module that also sends emails. Since it has no connection to Unit 1, that context is never loaded because it only loads its own unit, so the agent independently tries API v2, hits the same wall, and has to repeat the same discovery and pivot.

Beyond the wasted re-discovery, there's a correctness risk: nothing guarantees Unit 2 pivots the *same* way Unit 1 did, so two units can end up integrating with the same external resource through divergent, mutually inconsistent approaches.

## Version

v1 (current)

## Area

Construction phase rules

### Expected vs. actual behavior

- **Actual**: each unit loads its own artifacts plus the artifacts of its *declared* dependency units. Cross-cutting lessons learned at build time do not propagate to later units.
[...]

Issue #269, opened 2026-05-15
Suggestion: Archive cycle-specific docs after Build and Test approval to prevent stale context on new requirements


MMVesap:
## Summary

When a completed AI-DLC project receives new requirements in a subsequent session, the AI loads all previous artifacts including cycle-specific documents (requirements,
user stories, workflow plans, per-unit functional design). These documents describe *intent* from the completed cycle — not reality — and can conflict with new
requirements, causing the AI to over-scope or anchor incorrectly on old decisions.

## Problem

On session resumption, `session-continuity.md` instructs the AI to load all previous stage artifacts unconditionally. For a project on its second (or later) cycle this
means:

- Old `requirements.md` is loaded alongside new requirements → conflation risk
- Old user stories may describe personas or flows that no longer apply
- Old workflow plans describe a scope that is already complete
- Per-unit functional design docs describe the *original* design intent, not necessarily what the code actually implements

The code itself is the authoritative record of what was built. Old planning documents are one step removed and decay in accuracy over time.

## Proposed Solution

Introduce a **cycle archive** step that executes automatically when the user approves Build and Test results.

### Archive step (Build and Test — on approval)

Move cycle-specific documents to `aidlc-docs/archive/cycle-N/` (where `N` increments per cycle):

| Source | Destination |
|---|---|
| `inception/requirements/` | `archive/cycle-N/requirements/` |
| `inception/user-stories/` | `archive/cycle-N/user-stories/` |
| `inception/plans/` | `archive/cycle-N/inception-plans/` |
| `construction/plans/` | `archive/cycle-N/construction-plans/` |
| Per-unit `functional-design/` dirs | `archive/cycle-N/units/{unit}/functional-design/` |

### Evergreen documents (never archived — stay in place)

- `inception/application-design/` — architecture decisions apply to all future cycles
- `inception/reverse-engineering/` — still authoritative for brownfield projects
- Per-unit `code/` summaries — useful reference for future changes
- `construction/build-and-test/` — test baseline history
- `aidlc-state.md`, `audit.md` — never archived

### Session continuity change

Add a rule to `session-continuity.md`:

> **NEVER read files under `aidlc-docs/archive/`** — archived documents describe intent from a completed cycle and will conflict with new requirements. They are
preserved for audit purposes only.

Split artifact loading into *evergreen* (always load) and *cycle-specific* (load only if not yet archived).

### Ground truth for new cycles

[...]

PR #300, opened 2026-05-22
feat: propagate build-time discoveries across units (#299)

Mameta29:
> **Companion to #299.** Opened so the proposed mechanism is concrete and reviewable. This touches the shared `common/` file, so per CONTRIBUTING ("align on scope before you invest time writing code") I'd still like the approach settled in #299 — happy to revise the mechanism (or close this) based on where that discussion lands. Putting it up for review rather than leaving it as a draft so it's easy to comment on inline.

See #299 for the problem statement and discussion.

## What this changes

Construction runs a per-unit loop. Today, cross-unit context is shared only by following the *statically declared* dependency matrix (`unit-of-work-dependency.md`). Build-time discoveries — undocumented rate limits, API quirks, pivots that weren't foreseeable at design time — don't appear in that matrix, so a later unit touching the same external resource re-discovers (and re-pays for) the same lesson, and may even pivot inconsistently.

This adds an **orthogonal build-time layer** on top of the dependency matrix — the matrix itself is unchanged.

| File | Change |
|------|--------|
| `common/session-continuity.md` | New **Cross-Unit Discoveries** loading rule: read `aidlc-docs/construction/cross-unit-discoveries.md` during Construction stages (also covers resumed sessions, the same section #276 just fixed). |
| `construction/code-generation.md` | **Step 1** reads the log and reuses prior findings (referencing the session-continuity rule, not duplicating it). **Step 11** appends an entry when a build-time discovery crosses a unit boundary. |

The log is append-only, one entry = *finding · scope of impact · action taken · rationale · originating unit*. The file is created on first discovery and only when relevant, so projects with no cross-cutting findings produce nothing.

## Example (from @mayakost, #299)

Unit 1 (Notification) discovers email API v2 is rate-limited to 10 req/s (undocumented) and pivots to batching with API v1. With this change, that pivot is recorded once; Unit 2 (User Onboarding), which also sends email but has no declared dependency on Unit 1, reads the log at Step 1 and reuses the v1 batching approach instead of re-hitting the same wall.

## Design notes / open questions for review

- **Single source of truth**: the loading rule lives in `common/session-continuity.md`; `code-generation.md` references it rather than restating it. Recording is a distinct responsibility, so it's defined where it happens (Step 11).
- **Where should the log live / who owns the schema?** Open to feedback on the file location, the entry shape, or whether loading belongs in the per-unit loop vs. the smart-context rule.
- **RFC?** If maintainers consider this design-level rather than a small enhancement, happy to move the discussion to the RFC template.

## Testing

These are rule (documentation) changes, not code. I verified:
- The new path `aidlc-docs/construction/cross-unit-discoveries.md` is referenced consistently across both files (no typos / no divergent paths).
- The cross-reference from `code-generation.md` Step 1 to the `common/session-continuity.md` rule resolves to the section added here.
- No directory names under `aidlc-rules/` were renamed or moved (public-contract paths intact).
- Walked the email-API scenario above against the edited rules to confirm the load → reuse → record loop is coherent for an agent driving the per-unit Construction loop.

Platform: reviewed against the Claude Code rule-loading flow (rule-details path resolution under `aidlc-rules/aws-aidlc-rule-details/`).

### DIFF
--- a/aidlc-rules/aws-aidlc-rule-details/common/session-continuity.md
+++ b/aidlc-rules/aws-aidlc-rule-details/common/session-continuity.md
@@ -38,11 +38,15 @@ B) Review a previous stage ([Show available stages])
      `unit-of-work-dependency.md`). The exact files in each subdirectory are enumerated by the
      corresponding construction stage rules.
    - **Code Stages**: Read all code files, plans, AND all previous artifacts
+   - **Cross-Unit Discoveries (Construction resume)**: Also read
+     `aidlc-docs/construction/cross-unit-discoveries.md` if it exists, so a resumed unit reuses prior
+     units' build-time findings rather than re-discovering them. The log's schema and its recording rule
+     are defined in `construction/code-generation.md`.
 4. **Smart Context Loading by Stage**:
    - **Early Stages (Workspace Detection, Reverse Engineering)**: Load workspace analysis
    - **Requirements/Stories**: Load reverse engineering + requirements artifacts
    - **Design Stages**: Load requirements + stories + architecture + design artifacts
-   - **Code Stages**: Load ALL artifacts + existing code files
+   - **Code Stages**: Load ALL artifacts + existing code files + the cross-unit discoveries log (if present)
 5. **Adapt options** based on architectural choice and current phase
 6. **Show specific next steps** rather than generic descriptions
 7. **Log the continuity prompt** in audit.md with timestamp
--- a/aidlc-rules/aws-aidlc-rule-details/construction/code-generation.md
+++ b/aidlc-rules/aws-aidlc-rule-details/construction/code-generation.md
@@ -19,6 +19,8 @@ This stage generates code for each unit of work through two integrated parts:
 
 ## Step 1: Analyze Unit Context
 - [ ] Read unit design artifacts from Unit Design Generation
+- [ ] If `aidlc-docs/construction/cross-unit-discoveries.md` exists, read it and reuse any finding whose
+      *Affected Resource* this unit also uses (see "Cross-Unit Discoveries Log" in Critical Rules)
 - [ ] Read unit story map to understand assigned stories
 - [ ] Identify unit dependencies and interfaces
 - [ ] Validate unit is ready for code generation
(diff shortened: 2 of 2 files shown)


Mameta29:
@mayakost since this came directly out of your review on #276, I'd really value your read on it when you have a moment. The diff is small (+15/-1) — the substance is the design choice, not the lines.

Two things I'd especially like your take on:
- Does the recording trigger (Step 11 in `code-generation.md`: "a build-time discovery that crosses a unit boundary") match the failure mode you had in mind, or is it too narrow/too broad?
- Your email-API example is the canonical case here — does the load → reuse → record loop actually close it the way you'd expect, or is there a gap?

No rush, and no pressure to merge — I'd rather get the approach right in #299 first. Happy to rework the mechanism (or fold it into the RFC track) based on what you think.

Mameta29:
Pushed a revision (self-review pass before you dig in, @mayakost — didn't want to waste your time on rough edges):

- **Entry schema now leads with an `Affected Resource` column** (service name / API + version / env variable). This is the key fix: without it a later unit can't reliably tell whether a recorded finding applies to it. Your email-API example now resolves cleanly — Unit 2 matches on the email API identifier at Step 1.
- **Recording trigger is now fact-based, not judgement-based**: "a finding about an external resource this unit does not own" rather than "a finding that affects work beyond this unit's boundary." The old phrasing asked the agent to predict which future units are affected, which it can't reliably do; keying off the external-resource fact is reproducible.
- **Single source of truth**: the schema, lifecycle, and creation header now live in one place (Critical Rules → "Cross-Unit Discoveries Log"); Step 1 and Step 11 reference it. `session-continuity.md` keeps only a one-line resume-coverage note and points back to `code-generation.md`, instead of redefining the rule.
- **Lifecycle**: the log is marked cycle-scoped (archive/reset with other cycle-specific docs), which keeps it from colliding with #269.
- **Housekeeping**: commit scope corrected to `feat:` since it touches `common/` too; markdownlint clean.

The substance of the proposal is unchanged — still an orthogonal build-time layer that leaves `unit-of-work-dependency.md` untouched. Open questions for #299 are the same (where the log lives, whether this is RFC-level). Diff is +38/-2.

Mameta29:
One more polish pass (own self-review, still ahead of your review @mayakost):

- **Recording trigger no longer hinges on "own"**: replaced "an external resource this unit does not own" with an enumerated, objective test — "a shared resource: a third-party API, a service this unit did not provision, or an env var/secret it only consumes." The earlier "own" phrasing was ambiguous for shared resources (e.g. a DynamoDB table one unit defines and another only writes to); the enumeration removes that judgement call.
- **Step 1 trimmed to one line**: it now just says "reuse any finding whose *Affected Resource* this unit also uses (see Critical Rules)." The matching example and rationale moved into the single Critical Rules definition, so the per-step instructions stay scannable.
- **Restored a dedicated `session-continuity.md` checklist item**: the cross-unit-discoveries entry is its own bullet again (under Code Stages), so it isn't easy to skim past on resume — while the rule itself still lives only in `code-generation.md`.
- Added a one-line definition of "cycle" for the cycle-scoped note. markdownlint clean; Steps stay 1–16.

No change to the design — orthogonal build-time layer, dependency matrix untouched. Diff is +20/-20.

PR #311, opened 2026-05-28
ci(security): exclude noisy semgrep rules that generate false positives

### LINKED ISSUE #206: Tackle Code Scanner Findings
@leandrodamascena noticed there are ~1000 code scanning findings. Most are noise rather than actual issues. We should clear them so actual issues can be detected.

[Findings](https://github.com/awslabs/aidlc-workflows/security/code-scanning)

leandrodamascena:
closes #206 

# Summary

Exclude 6 semgrep rules from the security scanner workflow that produce ~1,554 false positives (89% of all open findings)

As reported in #206, we have ~1,748 open code scanning findings. The vast majority are noise:

  | Rule | Count | Why it's noise |
  |------|------:|----------------|
  | `gitlab.bandit.B101` / `B101` | 960 | Flags `assert` in test files — correct usage in tests |
  | `ai.generic.detect-generic-ai-anthprop` | 549 | Flags mentions of "Anthropic/Claude" — irrelevant for an AI project |
  | `ai.generic.detect-generic-ai-oai` | 3 | Flags mentions of "OpenAI" — same reason |
  | `gitlab.bandit.B404` / `B404` | 42 | Informational "consider implications of subprocess import" — not actionable |

  The root cause is `--config=r/all` which loads every rule in the Semgrep registry, including experimental and low-quality community rules. This change excludes the 6 noisiest rules while keeping the `r/all` config for broader coverage on actual
  security findings

After this change, ~194 findings remain for proper triage.

## Changes

> replace with a description of the changes

## User experience

## Checklist

* [x] Changes are documented

## Test Plan

## Acknowledgment


.github/workflows/security-scanners.yml
```diff
@@ -95,6 +95,12 @@ jobs:
           set +e
           semgrep scan --oss-only --verbose --metrics=off --config=r/all \
             --max-log-list-entries=0 \
+            --exclude-rule ai.generic.detect-generic-ai-anthprop.detect-generic-ai-anthprop \
+            --exclude-rule ai.generic.detect-generic-ai-oai.detect-generic-ai-oai \
+            --exclude-rule gitlab.bandit.B101 \
+            --exclude-rule B101 \
+            --exclude-rule gitlab.bandit.B404 \
+            --exclude-rule B404 \
             --sarif-output semgrep-report_sarif.json $BASELINE_ARGS
```
scottschreckengaust, line 104: Outside of tests `assert` has no place, and notifications on subprocesses warrant reviews.

scottschreckengaust (commented):
The AI items makes sense.

.github/workflows/security-scanners.yml
leandrodamascena: The 4 assert outside of tests are type narrowing guards in `copilot.py` .None of these are actual security issues and strictly used to make the type check happy.

I think excluding the rule entirely makes sense here. But I'm open to hear from you.

dreamorosi (approved):
Thank you for the PR - I think it's important that we refine the scanners and remove all the noise so that we can focus on the signal that matters instead.

leandrodamascena:
Hey @scottschreckengaust can I have another review here? Thanks

harmjeff (approved):
I see the discussion.  I think we are ok with this logic to reduce unwanted noise

PR #350, opened 2026-06-10
fix: add steering to enforce orchestration for all code-change prompts

SiddhJog:
In brownfield repos, Kiro's default routing intercepts small code-change requests ("make buttons yellow", "add a grid to dashboard") and invokes its built-in context-gatherer instead of the aidlc orchestrator. The orchestration skill never gets activated.

**Fix**: Add an always-included steering file that tells Kiro to route ALL code-change requests through `aidlc-orchestration `first, regardless of workspace state.

- Added `aidlc-workflows.md` with `inclusion: always`
- Broadened `orchestration` skill description to cover small changes explicitly

### DIFF
--- a/kiro/build/kiro-ide/build.js
+++ b/kiro/build/kiro-ide/build.js
@@ -8,6 +8,7 @@
  *   src/stages/              → dist/kiro-ide/.kiro/stages/        (copy)
  *   src/conventions/         → dist/kiro-ide/.kiro/conventions/   (copy)
  *   src/tools/               → dist/kiro-ide/.kiro/tools/         (copy)
+ *   src/steering/            → dist/kiro-ide/.kiro/steering/     (copy)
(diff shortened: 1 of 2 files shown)


raj-jain-aws:
@SiddhJog - Does this happen even when the user starts their intent with "Using AIDLC, ..."? I think its okay if the user sometimes doesn't want to engage AI-DLC for their own reasons. The "Using AIDLC ...", acts as that dynamic choice.

SiddhJog:
@raj-jain-aws When the user starts with "aidlc" the orchestration skill is engaged. See your point. Holding off. lets discuss.

raj-jain-aws:
@SiddhJog I'm assuming this can now be closed given the latest `v2` refresh? If so, pls. close.

leandrodamascena:
Hi @SiddhJog! I'm closing this PR because the problem has been fixed in the v2 branch.

PR #358, opened 2026-06-15
feat: add MCP support for Kiro (same servers than claude code)

jeromevdl:
# Summary

Add the same MCP servers than Claude Code (AWS + Context7)

## Changes

- harness/kiro/settings/mcp.json (new) — 5-server registry (context7 + 4 AWS), mirrors Claude's .mcp.json.
- harness/kiro/manifest.ts — ships settings/mcp.json as a harnessFile.
- harness/kiro/agents/aidlc-architect-agent.json — includeMcpJson: true + 5 @server tools; dropped $schema.
- harness/kiro/agents/aidlc-developer-agent.json — same.
- harness/kiro/agents/aidlc.json — dropped $schema only (no MCP).
- harness/kiro/onboarding.fills.ts — MCP-servers line updated.
- docs/guide/harnesses/kiro-cli.md — comparison-table row updated.

\+ regenerated the dist
\+ version/changelog

## User experience

- mcp servers are added in .kiro/settings/mcp.json (kiro syntax).
- Use `includeMcpJson` in the 2 agents (not in main one).

## Checklist

* [x] Changes are documented

## Test Plan

- MCP are only enabled for Architect and Dev agents, not the main one (to be discussed)

## Acknowledgment


apackeer (commented):
Nice work — this lands MCP on Kiro cleanly, and the wiring is right: it edits the `harness/kiro/` source (not just `dist/`), adds the manifest row, and the regenerated tree stays in sync (`bun scripts/package.ts --check` passes all three harnesses on this branch). I checked the mechanics against the current Kiro docs and the Amazon Q agent schema, and they hold up. Three things before merge, one of them just a question.

**1. The `$schema` removal is the right call, but the stated reason is wrong.** The changelog and commit say it was "broken (wrong product, unreachable URL)." I checked both: the URL returns HTTP 200, so it's reachable, and it isn't the wrong product either — `amazon-q-developer-cli` is Kiro CLI's direct predecessor. Dropping it is still correct, just for a different reason. That old schema sets `"additionalProperties": false` and defines `useLegacyMcpJson`, not `includeMcpJson`, so keeping it while you add `includeMcpJson` would make each agent JSON fail validation against its own declared schema. And Kiro's current agent-config reference doesn't list `$schema` as a field anyway. Could you reword the changelog bullet to match? It ships to users, so I'd rather the reasoning be right.

**2. Nothing tests the new Kiro registry or the `@server` grants.** `tests/unit/t110-mcp-server-grants.test.ts` covers the Claude registry (`dist/claude/.mcp.json`) and the Claude agents, including a dangling-grant guard that fails if an agent names an undeclared server. Nothing yet reads the new `dist/kiro/.kiro/settings/mcp.json` or checks the `@context7` / `@aws-*` entries against it — so a typo'd `@aws-srvless`, or a server you grant but forget to declare, would sail through. A Kiro counterpart to t110 would close that. I'm happy to add it as a follow-up commit here if you want.

**3. (Note, not a blocker.) The two registries are byte-identical.** `harness/kiro/settings/mcp.json` is a byte-for-byte copy of `harness/claude/.mcp.json`, so the server list now lives in two hand-authored files and can drift the next time someone adds one. That's the existing pattern, though — Claude already authors its registry as a harnessFile rather than in `core/` — so it's not something you introduced. Feels like the right candidate for a later "author it once in `core/`, project per-harness" refactor. Defer it.

**One question:** you put the `@server` tools in `tools` but not `allowedTools`, so MCP calls will prompt. I'm assuming that's deliberate — it matches the rest of the allowlist, where `fs_write` and `execute_bash` aren't auto-approved either — but worth a quick confirm.

And on your open question (MCP on architect and developer, not the conductor): right split. Claude hands subagents the session's MCP tools automatically; Kiro makes you opt in per agent, and the conductor only delegates, so it doesn't need them.

jeromevdl:
@apackeer I addressed your comments:
- new set of tests for kiro mcp (t111)
- change the changelog with real reason (not the good format for kiro)
- agree for the mcp.json, we can probably clean all these in a later PR
- To your question: yes, I didn't want to allow directly, users will have the choice.

Issue #368, opened 2026-06-16
[Bug]: for_each: unit-of-work — engine emits one directive but conductor doesn't iterate over units


tejasavora:
### Description

Stages with `for_each: unit-of-work` (code-generation, functional-design, nfr-requirements, nfr-design, infrastructure-design) are supposed to run once per unit defined in the delivery plan. The engine emits a SINGLE `run-stage` directive and expects the conductor to iterate internally. But nothing in the stage body or conductor instructions makes this explicit — the conductor treats it as a single invocation, generates code for one unit, approves, and moves on.

Result: a 9-unit enterprise project gets code generated for 1 unit only.

### Root Cause

`for_each` is declared in YAML frontmatter but:
1. The engine doesn't emit 9 separate directives (one per unit)
2. The stage body doesn't say "loop over all units"
3. The conductor has no mechanism to know how many units exist without reading units-of-work.md
4. The approve guard checks for src/ changes (passes after 1 unit's code exists)

### Steps to Reproduce

1. Run enterprise scope with 9 units in delivery plan
2. Reach code-generation stage
3. Agent generates code for Unit 1, commits, approves
4. Engine advances to build-and-test — Units 2-9 have zero code

### Expected Behavior

Code-generation produces implementation for ALL units before the stage can be approved.

### Actual Behavior

Agent generates code for 1 unit (2-5 files), approves, engine advances. 8 units have zero implementation.

### Fix Implemented (in fork)

Added explicit "MANDATORY: Per-Unit Iteration" section to code-generation.md body instructing the conductor to read units-of-work.md and spawn a subagent for EACH unit. Cannot approve until all units have code.

### Recommended Upstream Fix

Either:
1. The engine should emit N separate `run-stage` directives (one per unit) for `for_each` stages — the conductor handles one unit per directive, engine tracks which units are done
2. OR the stage protocol should have a standard "per-unit iteration" pattern that the conductor is trained to follow without relying on per-stage prose instructions
3. OR the approve guard should count: if `for_each: unit-of-work` and units-of-work.md lists 9 units, check that 9 unit directories exist under `aidlc-docs/construction/*/code-generation/`

Option 1 is cleanest — it moves iteration responsibility to the engine (deterministic) instead of the conductor (LLM, unreliable).

### AI-DLC Version

v2 (v0.6.4)

### AI-DLC Phase

Construction (code-generation, functional-design, nfr-requirements, nfr-design, infrastructure-design — all per-unit stages)

[...]

PR #390, opened 2026-06-18
[Bug] Validate reviewer / reviewer_max_iterations stage fields

### LINKED ISSUE #389: [Bug]: reviewer / reviewer_max_iterations stage fields are not validated — bad values and non-existent reviewer agents pass silently
### Description

The AI-DLC stage `reviewer:` / `reviewer_max_iterations` frontmatter fields are carried through schema → graph → directive by recent reviewer work, but nothing **validates** them. A malformed cap (`reviewer_max_iterations: "two"`), a non-numeric / zero / negative cap, or a `reviewer:` naming an agent that has no `.claude/agents/*.md` file all pass validation today and surface only as a runtime failure (or a silently-disabled review loop). `lead_agent` / `support_agents` **are** roster-checked (stage-schema Rule 9); `reviewer` is not.

### Root Cause

- `validateStageFrontmatter` (`core/tools/aidlc-stage-schema.ts`) has no per-field check for the two reviewer fields, and Rule 9 never includes `reviewer` in its agent-roster cross-check.
- `validateDirective` (`core/tools/aidlc-directive.ts`) lists the reviewer fields in `RUN_STAGE_FIELDS` but never type-checks them.
- The graph compile (`core/tools/aidlc-graph.ts`) coerces the cap with a bare `Number(...)`, which is NaN-prone for non-numeric input.
- A cap present with no `reviewer` is silently dropped at compile (the compiler guards the whole reviewer block on `reviewer` being present), so the schema accepts what the compiler discards.

### Steps to Reproduce

Set `reviewer: aidlc-nonexistent-agent` (or `reviewer_max_iterations: "two"`) on a stage, then compile the graph / run doctor. No error is raised; the bad value ships to `stage-graph.json`.

### Expected Behavior

Schema + directive validation should reject:
- a non-string `reviewer`;
- a `reviewer` slug with no matching agent file;
- a `reviewer_max_iterations` that isn't a positive integer (>= 1);
- a `reviewer_max_iterations` present with no `reviewer`.

Tests pin all of it.

### Proposed Fix

- Add per-field type checks + a `reviewer` roster cross-check in the stage-schema validator.
- Mirror the type checks in the directive validator.
- Make the parser return the cap as a number, with a safe compile-time guard so NaN never reaches `stage-graph.json`.
- Add focused unit tests.

### Out of Scope

Cross-harness reviewer wiring (Codex/Kiro trustedAgents / SKILL steps), the stage-protocol §12a loop determinism, reviewer write-permissions (`fs_write` grants), the `for_each` × reviewer iteration semantics (already tracked as #368), and docs. These are separate, later efforts.

### AI-DLC Version

v2 (based on the merged reviewer baseline; ships as 2.0.2 with this change).

### AI-DLC Phase

Framework / engine validation.

apackeer:
## Summary

The stage `reviewer:` / `reviewer_max_iterations` frontmatter fields were carried through schema → graph → directive by the 2.0.0 reviewer mechanism but were **never validated**. A malformed cap, a non-positive-integer cap, a cap declared with no reviewer, or a `reviewer:` naming a non-existent agent all passed validation and surfaced only as a runtime failure or a silently-disabled review loop.

This PR makes those failures **loud and deterministic** with the smallest changes that achieve it. Reviewer **runtime behaviour is unchanged** — only malformed/unknown config now errors at authoring/compile time.

Closes #389

## What changed (V1–V5)

- **V1 — parser returns `reviewer_max_iterations` as a number** (`core/tools/aidlc-lib.ts`). `parseStageFrontmatter` coerces an integer-literal cap to a real number (a non-integer literal is left a string for V2 to reject); `emitStageFrontmatter` gains a numeric emit branch so the cap round-trips as an unquoted number. `reviewer` stays a string.
- **V2 — stage-schema per-field validation + coupling** (`core/tools/aidlc-stage-schema.ts`). `validateStageFrontmatter` checks `reviewer` (optional string, via `checkString` exactly like `lead_agent`) and `reviewer_max_iterations` (optional positive integer, via a new `checkPositiveInteger` helper), and rejects a cap declared with no reviewer. Neither field is required.
- **V3 — reviewer roster cross-check in Rule 9** (`core/tools/aidlc-stage-schema.ts`). `reviewer` is cross-checked against `ctx.agents` like `lead_agent` (exempting the reserved orchestrator slug); a reviewer with no matching `.claude/agents/*.md` is rejected. This closes the silent-failure path.
- **V4 — directive validator + safe cap coercion** (`core/tools/aidlc-directive.ts`, `core/tools/aidlc-graph.ts`). The directive validator type-checks `reviewer` (string) and `reviewer_max_iterations` (positive integer, via a new `checkOptionalPositiveInteger` helper); the graph cap coercion is guarded so a bad value can never ship NaN to `stage-graph.json`. `aidlc-orchestrate.ts` is unchanged — V1+V4 guarantee a valid number upstream, so its `?? 2` can't emit NaN.
- **V5 — tests** (`tests/unit/t62.test.ts`, `t113.test.ts`, `t64.test.ts`). Reject/accept cases for every new rule, the directive checks, and the parse→number→round-trip contract.

## Authoring / CI impact

A stage whose `reviewer:` names an unknown agent, or whose `reviewer_max_iterations` is non-numeric / zero / negative / non-integer / present-without-a-reviewer, now **FAILS** `validateStageFrontmatter` and the graph compile where it previously passed. Fix the value or remove the field. No new commands or flags.

## Version

Bumps `2.0.1 → 2.0.2` (`aidlc-version.ts`, README badge, `## [2.0.2]` CHANGELOG entry) and regenerates `dist/` for all four harnesses. Re-copy your `dist/<harness>/` to pick up the regenerated tools.

## Out of scope

Cross-harness reviewer wiring (Codex/Kiro `trustedAgents` / SKILL steps), the stage-protocol §12a loop, reviewer `fs_write` grants, the `for_each` × reviewer iteration semantics (#368), and docs. These are separate, later efforts.

## Testing

- `bun run typecheck` — 0 errors
- `bun scripts/package.ts --check` — all four harness trees in sync (no hand-edited dist)
- Default-tier suite (`bash tests/run-tests.sh --debug -P 8`) — 182 files / 3253 assertions; the only failure is `t72-stage-reverse-engineering`, a pre-existing live-SDK timing flake unrelated to this change
- Target tests `t62` / `t113` / `t64` / `t65` / `t68` — all green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.0.2] - 2026-06-18
+
+Validates the stage `reviewer:` / `reviewer_max_iterations` frontmatter fields, which were carried through schema → graph → directive by the 2.0.0 reviewer mechanism but never checked. A malformed cap, a non-positive-integer cap, a cap declared with no reviewer, or a `reviewer:` naming an agent with no `.claude/agents/*.md` file all passed validation before and surfaced only as a runtime failure or a silently-disabled review loop; they now fail validation/compile loudly and deterministically. This is a behaviour change at authoring time only — the reviewer's runtime behaviour is unchanged. Re-copy your `dist/<harness>/` to pick up the regenerated tools. Refs #389.
+
+* **Authoring impact (action required if you hand-edit stages):** a stage whose `reviewer:` names an unknown agent, or whose `reviewer_max_iterations` is non-numeric / zero / negative / non-integer / present-without-a-reviewer, now FAILS `validateStageFrontmatter` and the graph compile where it previously passed. Fix the value or remove the field.
+* **`reviewer_max_iterations` is now parsed as a number.** `parseStageFrontmatter` coerces an integer-literal cap to a real number (and `emitStageFrontmatter` round-trips it as an unquoted number), so the type is correct end-to-end instead of relying on a `Number()` coercion at compile.
+* **`reviewer` is roster-checked like `lead_agent`.** A reviewer slug with no matching agent file is rejected (stage-schema Rule 9), closing the silent-failure path.
+* **Directive validation** type-checks `reviewer` (string) and `reviewer_max_iterations` (positive integer), and the graph compiler guards the cap so a bad value can never ship NaN to `stage-graph.json`.
+* No new commands or flags; the only breaking change is that previously-accepted malformed reviewer config now errors.
+
 ## [2.0.1] - 2026-06-18
 
 Greens the default-tier test suite (smoke + unit + integration) after the 2.0.0 reviewer-mechanism merge, with no behaviour change to the reviewer feature itself. Three independent, mechanical defects rode in with the merge: the frontmatter emitter dropped the two `reviewer` fields on a round-trip, the version trio was inconsistent, and three agent-count assertions still expected the pre-merge roster of 11. This release reconciles all three so CI is green on a clean 2.0 baseline. No re-copy of `dist/<harness>/` is required for behaviour — the changes are test-correctness, metadata, and the regenerated dist copies of `aidlc-version.ts`. Refs #387.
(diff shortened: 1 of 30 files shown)


spraja08 (approved):
Tested in an e2e poc scope. GTG.

PR #401, opened 2026-06-18
feat: per-stage traceability enforcement with deterministic JSON sensor

jeromevdl:
## Summary

Adds element-level traceability enforcement across the AIDLC workflow — ensuring nothing from the initial intent (requirements, stories, acceptance criteria) is silently dropped between stages.

### Stable ID conventions (enforced in stage prose)
- `FR{n}`, `NFR{n}` — requirements-analysis
- `US{n}.{m}` — user-stories
- `AC{story}.{seq}` (3-segment) — user-stories
- `BR{group}.{seq}` — functional-design
- `NFRx.y` — nfr-requirements decomposition
- `U{n}` / `u{n}-{desc}` — units-generation naming

### Per-stage traceability.json
Each stage produces a structured JSON file mapping upstream elements to downstream coverage:
```json
{
  "stage": "functional-design",
  "unit": "u1-auth",
  "upstream_ids": ["AC1.1.1", "AC1.1.2"],
  "coverage": [
    { "id": "AC1.1.1", "status": "OK", "target": "BR01" },
    { "id": "AC1.1.2", "status": "GAP" }
  ],
  "reverse": [
    { "id": "BR03", "status": "ORPHAN" }
  ]
}
```

### Deterministic traceability sensor
`aidlc-sensor-traceability.ts` fires at each gate with three checks:
1. **Minimum coverage** — empty file fails
2. **Status scan** — GAP/ORPHAN entries are findings; Deferred/N/A without target are findings
3. **Upstream ID cross-check** — extracts IDs from upstream files via regex, compares against declared `upstream_ids`. Two independent sources must agree.

Per-unit stages scope via `unit-of-work-story-map.md` to avoid false positives across units.

### Phase boundaries
- **delivery-planning** (Inception exit): aggregates upstream traceability files, checks for gaps
- **build-and-test** (Construction exit): final coverage gate — verifies all FRs/ACs/NFRs from inception have implementing code across all units

## Other changes
- Moved Construction→Operation boundary from ci-pipeline (conditional) to build-and-test (always runs)
- Fixed typo in infrastructure-design.md
- Regenerated stale t66 designer-export fixture (agents 11→13, pre-existing)
- Updated docs and knowledge base

## Known limitations
- **Ideation→Inception boundary untouched** — no numbered IDs at that phase (exploratory by design)
- **Unit-name join assumes `u{n}` convention** — advisory warning emitted on naming mismatch
- **Free-text elements (component names) not diffable** — marker scan only for units-generation
- **build-and-test final gate is prose-instructed** — not sensor-enforced (cross-unit aggregation)

## Pre-existing test failures (not introduced by this PR)
- t65 round-trip: YAML emitter issue on v2 branch
- t68 version badge: v2 branch version mismatch (0.7.12 vs 2.0.0 in README)


apackeer (changes requested):
Thank you for pushing element-level traceability forward. This remains aligned with Goal 6, and the existing upstream-coverage sensor does not replace it. Stable IDs and structured JSON are useful foundations.

The first pass found several issues that need to be addressed before this can move to second review:

1. `traceability.json` sits outside each stage's `produces:` contract, so omitting it bypasses the write-triggered sensor, artifact guard, and reviewer artifact scope.
2. The sensor casts parsed JSON without validating its schema. Invalid shapes can crash and be recorded by the dispatcher as a passing script-error, while missing or unknown statuses can pass silently.
3. `OK` entries do not require or verify a downstream target, and the optional reverse map is trusted rather than derived, so the result does not yet prove coverage or detect omitted orphans.
4. Unit scoping assumes `U<n>` / `u<n>-*`, misses multi-unit mappings and Windows paths, and can turn an empty expected set into a passing result.
5. The Construction boundary was moved to Build and Test even when CI Pipeline still executes afterward, and this conflicts with the shared governance protocol.

The targeted sensor suite passes all 14 submitted tests, but it does not cover these integration and false-pass cases. The branch is now 80 commits behind and conflicting, so please rebase and redesign the traceability artifact around current `produces`/`consumes`, reviewer, unit-graph, and phase-boundary contracts. Please also remove the unrelated Python sensor plan and add the required release metadata and upgrade guidance.

Once those changes are in place, this will be ready for the second-pass review.

apackeer:
Heads-up @jeromevdl: rather than ask for a full rebase (the branch was 107
commits behind v2 with 63 conflicting files), we rebuilt your traceability
design in-house on current v2 and pushed it to this PR, with your
co-authorship on the commit. The design is still yours: stable IDs, the
per-stage traceability.json, the deterministic JSON sensor, and the
phase-level aggregation all carried through; the rebuild wires them into the
contracts that landed on v2 since June.

What changed relative to your original, per the two review rounds:

1. traceability.json is now a declared `produces:`/`outputs:` artifact of all
   eight emitting stages, and a new `artifactFilename()` seam resolves the
   JSON exception across directive resolution, the artifact guard, review
   fingerprints, and per-unit wave contracts, so omitting the file is
   detectable (review ask 1).
2. The sensor validates the JSON shape at runtime, enforces the closed status
   set (unknown statuses are findings, GAP in `reverse` counts), and never
   crashes into a script-error pass (ask 2).
3. `OK` entries require verified targets where deterministic: US ids against
   stories.md, BR ids against business-rules.md, existing workspace-relative
   files for code-generation; business-rule orphans are derived from
   business-rules.md rather than trusted from `reverse` (ask 3).
4. Unit scoping is driven by the authored unit DAG and unit table, normalizes
   Windows paths, and fails closed on an empty scoped set (ask 4).
5. The Construction phase boundary stays in ci-pipeline per the governance
   protocol; build-and-test gains an always-on stage-level cross-unit
   coverage gate that feeds it (ask 5). The ci-pipeline-skipped case remains
   a documented limitation.
6. Release metadata added (2.5.71, CHANGELOG, badge), the unrelated Python
   sensor plan is dropped, the sensor test moved off the taken t154 slot to
   t281, and the docs now enumerate six sensors (v2 gained claim-sources
   since your branch).

Evidence: full smoke+unit+integration tier green (316 files), packaging
parity across all seven harness trees, typecheck and lint clean. The PR
description above predates the rebuild, so its "pre-existing test failures"
section no longer applies.

apackeer (approved):
Approving at 8f8d91e4: all five asks from my 2026-08-03 review are addressed in the rebuilt head (artifact contract, runtime schema validation, verified targets and derived orphans, fail-closed unit scoping, boundary kept at the governed transition), release metadata and docs are in, and the branch is current with v2. Verification evidence is in the comment above.

PR #412, opened 2026-06-19
feat(kiro): ship the MCP registry disabled-by-default with per-persona grants (2.5.71)

jeromevdl:
# Summary

Kiro CLI now ships the same five MCP servers as Claude Code (`context7` + the four AWS servers), declared in `.kiro/settings/mcp.json` and disabled by default. Maintainer rebuild of @jeromevdl's original branch onto current `v2` (the original predated the 15-config Kiro agent layout); recreates #358.

**Enable a server:** flip `"disabled": false` on that entry in `.kiro/settings/mcp.json`.

## Changes

- `harness/kiro/settings/mcp.json` (new): five-server registry mirroring Claude's `.mcp.json`, every server shipped with `"disabled": true` so no fetched code runs until a user opts in per server. The `@latest` launchers stay aligned with the Claude registry; under opt-in they execute only after an explicit enable.
- **context7 ships keyless on Kiro.** Live verification on kiro-cli 2.12.1 showed Kiro sends MCP HTTP header values verbatim (`$VAR`, `${VAR}`, and `${env:VAR}` all arrive unexpanded), so a configured header is either a broken placeholder or a committed secret. The registry therefore declares no headers, and a test pins that policy.
- All 14 delegated persona configs set `includeMcpJson: true` and grant the five `@<server>` tools. The conductor (`aidlc.json`) is excluded and keeps its locked-down surface. Grants are never added to `allowedTools`, so MCP tool calls keep prompting.
- Persona configs drop the upstream `$schema`: that schema sets `additionalProperties: false` and does not define `includeMcpJson`, so keeping it would make each file fail validation against its own declared schema. The unchanged conductor retains its schema.
- Docs: Kiro CLI harness comparison row, Kiro onboarding (`AGENTS.md` fill), and the composer CodeKB paragraph in `docs/guide/05-scopes-and-depth.md` (the MCP mechanism now exists on Kiro CLI; Kiro IDE remains workspace-scan fallback).
- Tests: `tests/unit/t281-kiro-mcp-registry.test.ts` (registry integrity, Claude parity minus `disabled`, dynamic agent-roster grant matrix, no-headers policy, `allowedTools` guard) and the gated live e2e `tests/e2e/t-acp-kiro-mcp-headers.serial.test.ts` (workspace MCP discovery plus the verbatim-header limitation pin; runs under `AIDLC_KIRO_ACP_LIVE=1`, skips cleanly otherwise).
- Version 2.5.71, CHANGELOG entry, README badge; all dists regenerated (`bun scripts/package.ts --check` green).
- Kiro IDE is intentionally untouched; that surface is being reworked separately (#653).

## Verification

- Full default tier (smoke + unit + integration): 316 files, 7348 assertions, PASS.
- Live on authenticated kiro-cli 2.12.1: workspace `.kiro/settings/mcp.json` discovery and `@server` grants work end to end (a local MCP fixture received `tools/list`); `"disabled": true` servers never connect; header values arrive verbatim, which is what drove the keyless context7 decision.

## Checklist

* [x] Changes are documented

## Acknowledgment


harness/kiro/agents/aidlc-architect-agent.json
```diff
@@ -1,13 +1,18 @@
 {
-  "$schema": "https://raw.githubusercontent.com/aws/amazon-q-developer-cli/refs/heads/main/schemas/agent-v1.json",
   "name": "aidlc-architect-agent",
   "description": "AI-DLC Architect Agent \u2014 delegation target for the reverse-engineering (2.1) synthesis step. Use for delegated architecture-analysis tasks.",
   "prompt": "file://aidlc-architect-agent.md",
+  "includeMcpJson": true,
```
apackeer, line 5: The architect/developer-only grant policy was designed for the older three-config layout. Current `v2` ships 14 persona configs, and inline stages execute in the conductor rather than a delegate. Please define an explicit role-to-server policy for the current roster, including deliberate conductor and composer decisions, before wiring individual grants.

harness/kiro/settings/mcp.json
```diff
@@ -0,0 +1,32 @@
...
+      "type": "http",
+      "url": "https://mcp.context7.com/mcp",
+      "headers": {
+        "CONTEXT7_API_KEY": "${CONTEXT7_API_KEY}"
+      }
+    },
+    "aws-mcp": {
+      "command": "uvx",
+      "args": [
+        "mcp-proxy-for-aws@latest",
```
apackeer, line 11: Please settle the default and supply-chain policy before shipping this registry. Configured Kiro MCP servers are enabled by default, and these AWS entries execute mutable `@latest` packages that may use local AWS credentials. The implementation should reflect an explicit choice such as disabled-by-default entries and/or exact package versions, with the rationale documented.

tests/unit/t111-kiro-mcp-registry.test.ts
```diff
@@ -0,0 +1,194 @@
...
+const AGENTS_DIR = join(KIRO_ROOT, "agents");
+
+// The harnesses that ship the MCP registry — both must carry the identical shape.
+const REGISTRY_ROOTS = [
+  { harness: "kiro", path: join(REPO_ROOT, "dist", "kiro", ".kiro", "settings", "mcp.json") },
+  { harness: "kiro-ide", path: join(REPO_ROOT, "dist", "kiro-ide", ".kiro", "settings", "mcp.json") },
+] as const;
+
+// The three Kiro agent configs.
+const AGENT_FILES = ["aidlc.json", "aidlc-architect-agent.json", "aidlc-developer-agent.json"] as const;
```
apackeer: This fixed three-file list pins the old agent layout and cannot validate the current 14-persona access policy. Please discover the current agent configs and assert the agreed role/server matrix. A credential-free local MCP fixture or equivalent host discovery/access check is also needed; registry-shape checks alone do not prove Kiro exposes the tools as intended.

harness/kiro-ide/manifest.ts
```diff
@@ -56,6 +56,7 @@ const manifest: HarnessManifest = {
     // + the --check drift guard. (The roll-forward latch lines are inert on Kiro IDE,
     // which has no userPromptSubmit/preToolUse seam, but are kept for parity.)
     { src: "dot-gitignore", dst: ".gitignore", projectRoot: true },
+    { src: "settings/mcp.json", dst: "settings/mcp.json" },
```
apackeer: Please remove Kiro IDE from this PR and keep the scope to Kiro CLI v2. PR #653 is replacing the IDE’s CLI-style JSON agent surface with native Markdown agents, so landing a parallel MCP contract here would couple this change to an unsettled integration.

apackeer (changes requested):
Thanks, Jerome. This remains a useful direction, and the predecessor’s Kiro CLI mechanics were reviewed favorably. I’m requesting changes because current `v2` has outgrown the access model in this branch and the default/runtime policy is not settled yet.

Please scope this PR to Kiro CLI v2. Kiro IDE’s native agent/config surface is being reworked in #653, and Kiro CLI v3 is tracked separately in #582. Before implementation, we need an explicit decision on (1) whether the five servers ship enabled, disabled, or as documented configuration only, and (2) which current execution contexts receive which servers. Kiro now has 14 persona configs, inline stages run in the conductor, and the composer can use CodeKB when exposed.

After those decisions, the current-v2 integration should update the MCP/composer documentation, broaden the policy and host-access tests, and include the required version, changelog, and README badge updates. There is no need to chase the moving base while the design is unsettled; maintainers can handle the current-v2 integration afterward. I did not run the branch tests because the branch is conflicting and predates the current Kiro contracts.

apackeer (approved):
Approving the rebuilt head 30d3359e, which resolves all asks from my August 4 review: scoped to Kiro CLI v2 only (IDE untouched, reworked in #653), disabled-by-default enablement with the supply-chain rationale documented, the full 14-persona/conductor-none access matrix pinned by a dynamic-roster test, docs and composer/CodeKB guidance updated, and the version/changelog/badge contract met at 2.5.71. Runtime behavior is live-verified on kiro-cli 2.12.1: workspace registry discovery and @server grants work end to end, disabled servers never connect, and header values are sent verbatim, which is why context7 ships keyless and the registry declares no headers.

Issue #421, opened 2026-06-22
Front-load phases (ideation/inception) force frequent manual compaction — adopt a ralph-loop so subagents do the work and the main agent only tracks state


fanhongy:
## Problem

The early phases of v2 — **initialization, ideation, inception** — accumulate a large amount of context in the **main agent** session. Token *spend* itself is fine; the pain is that all this work runs **inline in the main agent context** (per `stage-graph.json`, 30/32 stages are `mode: "inline"`; only `reverse-engineering` and `code-generation` use `mode: "subagent"`).

Because everything lives in the main session, the context window fills up fast and I have to **compact many, many times** during a single run. Worse, the forwarding loop (`SKILL.md` → `aidlc-orchestrate next/report` → repeat, enforced by the `aidlc-stop.ts` Stop hook) **keeps the workflow moving and won't naturally yield for a compaction** — I have to explicitly interrupt/pause the orchestrator mid-stage to compact, then resume.

## Proposed direction — ralph-loop execution model

Invert the execution model so the **main agent stays thin**:

- The main agent runs only the **forwarding loop + state tracking** — read `aidlc-state.md`, dispatch, observe outcome, report.
- Each stage's actual work runs in a **dedicated subagent** (the persona/lead agent for that stage), which does the heavy lifting and eventually touches the state file.
- The main agent's context then grows roughly with the *number of stages*, not with the *full transcript of every stage's reasoning* — so compaction becomes rare and the loop no longer has to be manually paused.

This generalizes the pattern already used for `reverse-engineering` and `code-generation` (both `mode: "subagent"`) to the front-loaded inline phases that hurt most.

## Relevant current mechanics

- Forwarding loop + conductor: `harness/claude/skills/aidlc/SKILL.md`
- Orchestration engine + directive kinds (`dispatch-subagent` is defined but **never emitted** today): `core/tools/aidlc-orchestrate.ts`, `core/tools/aidlc-directive.ts`
- Stage modes: `core/.../data/stage-graph.json` (`mode: inline|subagent`)
- State + audit (already atomic via `withAuditLock()`, audit-first): `core/tools/aidlc-state.ts`
- Flow control: `aidlc-stop.ts` Stop hook; existing `SESSION_COMPACTED` / `acknowledge-compaction` recovery path

## Open questions / scope

1. Should *all* inline stages move to subagent dispatch, or just the heavy ideation/inception ones?
[...]

PR #422, opened 2026-06-22
feat: delegate inline-stage artifact drafting to throwaway subagents

fanhongy:
## Summary

Inline ideation/inception stages run their full body in the main conductor
context, so heavy artifact-drafting reasoning (large `consumes`, full `produces`
drafts) accumulates in the main session and forces the user to compact
repeatedly — and the forwarding loop won't yield for a compaction without a
manual interrupt.

This adds a **protocol-driven drafting hand-off** (ralph-loop style): the
conductor stays thin — it collects answers and runs the approval gate inline —
but delegates **only the artifact-drafting sub-step** to a single throwaway
`Task(directive.lead_agent)` whose context is discarded after it returns the §11
summary. The heavy drafting reasoning never lands in the main transcript.

Stages stay `mode: inline` — **no engine, schema, or directive-type change.**
This is deliberately distinct from whole-stage `mode: subagent`, which an
interactive stage cannot use (a subagent cannot drive the question UI or the
approval gate).

## Approach

Chosen over wiring the unused `dispatch-subagent` directive because the split we
need is *intra-stage*: it fits inside the one existing `run-stage` directive, so
it needs no engine/schema/directive churn and keeps the blast radius to prose +
docs + one test. The draft step slots into the existing atomic stage ritual
(`questions → artifact → reviewer → completion → learnings → gate`) — it
specifies *how* the artifact step runs, not a new step.

### Hook-safety
- The conductor never spawns the draft `Task` while any `[Answer]:` tag is still
  blank (the §3 Step 4 completeness check gates it).
- The subagent treats **any** `*-questions.md` in the stage dir as read-only
  (the Stop hook matches the `-questions.md` suffix, not just the canonical
  name), so it can never introduce/clear a blank tag that `aidlc-stop.ts`'s
  Tier-2 human-wait carve-out would misread.

## Changes
- `core/aidlc-common/protocols/stage-protocol.md` — new "Drafting delegation for inline stages" section (three-part execution, ordering, hook-safety, fallback)
- `harness/claude/skills/aidlc/SKILL.md` — draft sub-step in the `gate: true` arm
- `core/aidlc-common/conductor.md` — persona craft for the delegation
- `docs/reference/03,04,15` — document the hand-off, distinct from `mode: subagent`
- `tests/integration/t154-drafting-delegation-contract.test.ts` — locks the contract across protocol + SKILL
- `dist/` — regenerated for all four harnesses (drift-clean)

## Scope notes
- This lands the **shared mechanism**; it applies by shape to every gated
  artifact-generating inline stage. Verify on the heaviest stages
  (`requirements-analysis`, `application-design`) before relying on it broadly.
- **Versioning intentionally left to maintainers** — this PR does not touch
  `aidlc-version.ts`, `CHANGELOG.md`, or the README badge. Happy to add the
  bump/changelog entry if you'd prefer it in-PR.

## Testing
- `bun scripts/package.ts --check` — drift-clean across all four harnesses
- `t154` (new), `t32`, `t37`, `t38`, `t68`, `t119` — all pass
- Full default tier: 181/183 files pass; the 2 failures (`t140`, `t19`) are a
  pre-existing missing `@anthropic-ai/claude-agent-sdk` dependency in the local
  environment, unrelated to this change.

Refs #421


apackeer:
Thanks, Stan. The context-pressure problem here is real, and issue #421 should remain open. Since this PR was created, though, the `v2` execution model has changed substantially: mode is now an explicit four-way communication topology, Construction has engine-driven per-unit iteration, every dispatched worker receives deterministic steering, and the project ships five harnesses.

This commit cannot be safely brought forward as a normal rebase. The draft dispatch is currently placed after `run-stage` has already executed the body and written the artifacts; its worker does not receive the stage body, steering bundle, or support-agent perspectives; and only the Claude skill is wired while shared core prose uses Claude-specific `Task` terminology. The per-unit gate lifecycle also means the hook misses the expensive unit drafting it is intended to offload.

We already need a broader stage redesign, and that work will introduce the explicit setup for whether agents execute inline or as subagents. That is the right place to solve this context-pressure problem consistently across topologies and harnesses, with behavioral tests for routing, rule delivery, per-unit work, and the actual context reduction.

This is therefore superseded by the planned redesign rather than rejected as a product idea. I propose closing this PR to clear the stale implementation while retaining #421 as an input to that redesign. The core idea and this contribution should be credited in the replacement.

fanhongy:
thansk for the response! Love to see the issue been taken by the team!

Issue #431, opened 2026-06-24
[Feature]: Blocking sensor severity — make a sensor failure able to halt a gate


harmjeff:
## Summary

Today **every sensor is advisory** — a `SENSOR_FAILED` is recorded for telemetry
but **cannot halt a stage, fail a gate, or block the workflow**. This is the one
capability an extension (or core) genuinely cannot add on its own: real
enforcement (e.g. "fail the build if branch coverage < target" or "fail if a
requirement has no test") requires a **core change** — a `blocking` sensor
severity plus a conductor/hook that halts on it.

This came up directly while building the `test-pro` testing extension (#430): its
`coverage-threshold` and `requirement-coverage` sensors are correct and run, but
can only **report** — the framework has no way to make them gate. The bundle
ships them advisory and relies on agent prose to meet targets.

## Current state (verified)

- **The schema hard-rejects any severity but `advisory`.**
  `core/tools/aidlc-sensor-schema.ts` (`requireExactValue(obj, "default_severity",
  "advisory", ...)`) — `advisory` is the "sole accepted value today".
- **Nothing consumes severity at runtime.** `aidlc-sensor.ts` emits
  `SENSOR_FIRED` → `SENSOR_PASSED`/`SENSOR_FAILED` audit rows;
  `aidlc-runtime.ts` records them for reporting; **no code path halts on a
  failure.**
- **The fire hook is hardcoded non-blocking.** `core/hooks/aidlc-sensor-fire.ts`
  exits 0 unconditionally ("advisory always per G5").
- **It's already reserved.** `docs/reference/07-sensor-system.md` documents
  `blocking` as "reserved for the future ralph-driver work" — a sensor failure
  that "halts the gate rather than logging advisory telemetry."

So the design always anticipated this; it just isn't built.

## Proposed change

1. **Schema** — allow `default_severity: blocking` (relax the exact-`advisory`
   check to an enum `{advisory, blocking}`; default stays `advisory`).
2. **Enforcement point** — when a stage's `sensors_applicable` includes a
   `blocking` sensor that emits `SENSOR_FAILED`, the gate does **not** auto-pass.
   The natural home is the existing flow-altering hook (`aidlc-stop.ts`) and/or
   the conductor's approval gate, which would consult the per-stage terminal
   sensor results (already aggregated in `aidlc-runtime.ts`) and refuse
   completion until the blocking sensor passes (or the human explicitly
[...]

PR #432, opened 2026-06-24
feat(extensions): V2 extension mechanism + test-pro bundle (closes #430)

### LINKED ISSUE #430: [Feature]: V2 Fully featured extensions and gap remediation
## Summary

Layers 0–4 plus the §4 per-stage contribution seam make the extension mechanism
**functionally complete for a single, first-party bundle**: a bundle can add
stages/agents/scopes/rules/sensors, gate them with `when: {producer-in-plan}`,
and additively modify existing core stages — all projected as a byte-pinned,
drift-guarded delta with no edits to `core/`.

What is **not** yet solid is authoring a *complex* extension (a full operation
phase) or running **multiple independent third-party bundles** together. This
issue enumerates the concrete gaps, each with where it bites and a proposed
direction, so they can be triaged/sequenced rather than rediscovered.

Reference design: `docs/reference/18-extension-mechanism.md`. As-built status:
Layers 0–4 + §4 shipped (v2.0.7); §5 (multi-tenant guards) not started.

---

### 1. No cross-bundle conflict attribution (multi-tenant correctness)
**Status:** §5.3, not implemented.
Today the build validates one bundle at a time. Two bundles that both contribute
to the same stage, both claim overlapping number ranges, or where one consumes an
artifact another produces, are not validated *together* with bundle attribution.
The fragment-ordering comparator (`order`, bundle, anchor) and the dup-anchor
guard exist, but there is no "merge the fully-resolved multi-bundle graph and
report which bundle introduced an unsatisfiable edge."
**Bites:** the moment a second bundle is installed alongside the first.
**Direction:** a merged-graph validation pass keyed by bundle, surfacing
`bundle X requires artifact owned-by/none → unsatisfiable` with attribution.

### 2. Artifact namespacing is a hyphen prefix, not a real namespace
**Status:** shipped as `<bundle>-` prefix; doc sketched `bundle/artifact` slashes.
`ARTIFACT_SLUG_RE` forbids `/`, so contributed artifacts are validated as
`ops-min-foo`, a *convention* not an enforced namespace. Nothing stops a bundle
from colliding with a core artifact name that happens to start with its prefix,
and `core/*` is not a reserved namespace.
**Bites:** two bundles (or a bundle + core) with near-identical artifact names.
**Direction:** decide slash-namespacing (requires loosening `ARTIFACT_SLUG_RE`
+ registry changes) vs. keeping the prefix and adding a collision check across
the merged registry.

### 3. `requiresBundle` versions are parsed but not enforced
**Status:** §5.4 partial. `requiresBundle: ["compliance@^1"]` is checked only for
*existence* of the dep bundle; the `@^semver` constraint is split off and ignored.
**Bites:** a bundle that depends on a specific version of another bundle.
**Direction:** semver parse + compare against the dep's `version`; fail the build
on an unsatisfied constraint. Also: no topological ordering of bundle application
when one depends on another's contributions.

### 4. No real `bundle-active` predicate (activation is build-time only)
**Status:** intentionally deferred (Layer 4). There is **no runtime signal** of
which bundles are active — "active" means "the bundle's delta was overlaid at
install." So `when: {bundle-active: X}` cannot be evaluated and is parse-but-ignore
on contributions.
**Bites:** an author who wants a stage/contribution conditional on *another*
bundle being present, or a single install that toggles bundles without re-copying
files.
**Direction:** a runtime bundle manifest/registry (e.g. `aidlc-bundles.json` in the
install) the engine reads, plus a real `bundle-active` predicate in
`WHEN_PREDICATE_KEYS`.

### 5. Contribution fragments: limited anchors, no body-region targeting
**Status:** §4 shipped with `after-step:<n>` / `before-step:<n>` / `end-of-steps`.
There is no anchor for the `## Sensors` / `## Learn` compartments, no "after the
questions block," and no way to target a named subsection. A `kind: questions`
fragment was dropped (core stages have no `## Questions` H2); questions are
authored as a step fragment.
**Bites:** an author who needs to enrich a stage's questions, sensors compartment,
or learn ritual rather than just insert a step.
**Direction:** richer anchors (compartment-aware: `in:Sensors`, `in:Learn`), and a
decision on whether questions deserve a first-class structural home in stage
bodies.

### 6. Contributions cannot override or remove — only add
**Status:** by design (core immutability), but a real limit for complex
extensions. A bundle cannot change a core stage's `lead_agent`, relax a
`consumes[].required`, replace a step's prose, or remove a produced artifact.
**Bites:** an operation-phase extension that needs an upstream stage to behave
*differently*, not just *more*.
**Direction:** explicitly document the boundary; if override is ever needed,
design a separate, auditable patch layer (high risk to drift-guard + core
immutability — likely out of scope, but should be a stated non-goal).

### 7. `required_sections` enforcement is shallow
**Status:** §4 shipped — the sensor checks each named section exists as a `## ` H2.
It does not check ordering, nesting, non-emptiness, or content shape, and the
match is exact-string on the trimmed heading.
**Bites:** an author expecting "the section must contain X" or case/format
flexibility.
**Direction:** optional per-section predicates (non-empty, min length, regex) if
demand appears; document the exact-match contract clearly for now.

### 8. No authoring-time validation tooling for extension authors
**Status:** gap. An author only learns a contribution is malformed by running the
full `bun scripts/package.ts` and reading a build error. There is no
`aidlc-graph validate-extension <dir>` or lint that checks an extension in
isolation (target exists, anchors valid, artifacts namespaced, ranges claimed)
without a full harness build.
**Bites:** iteration speed for anyone writing a non-trivial bundle.
**Direction:** a standalone validate subcommand + a documented author workflow.

### 9. Test/count roster is not bundle-partitioned (Layer 5)
**Status:** §Layer 5, not implemented. Core counts (32 stages / 13 agents / 9
scopes) are hand-authored literals across ~15 test files. A bundle that adds
stages/agents can't be tested with the same "exactly N" discipline; today only
the delta byte-pin (t155/t156) guards bundle output.
**Bites:** building a *large* extension (many stages) with confidence.
**Direction:** computed-from-source rosters partitioned by bundle ("core exactly
N, each bundle exactly M").

### 10. Documentation: no author-facing extension guide
**Status:** gap. `docs/reference/18-extension-mechanism.md` is the *design/as-built*
record, not a "how to write an extension" guide. There is no walkthrough,
contribution-file reference, anchor catalog, or namespacing rules doc under
`docs/guide/` or `docs/harness-engineering/`.
**Bites:** any first external author.
**Direction:** an author guide with the `ops-min` fixture as a worked example.

---

## Doc-vs-reality divergences to reconcile (lower priority)

These were resolved during implementation by shipping reality; the design doc
(`18-extension-mechanism.md`) has been updated inline, but flagging for a clean
pass:

- Namespacing: doc sketched `bundle/artifact` (slash); shipped `<bundle>-`
  (hyphen). (See gap #2.)
- `when: {bundle-active}` on contributions: doc sketched it; shipped as implicit
  (parse-but-ignore). (See gap #4.)
- `fragments` referenced separate `file:` paths; shipped as inline
  `## fragment: <anchor>` body blocks in the contribution file.

---

## Suggested triage

- **Blocks multi-tenant (third-party) use:** #1, #2, #3.
- **Blocks a complex first-party extension (operation phase):** #5, #6 (boundary),
  #8, #9, #10.
- **Nice-to-have / future:** #4, #7.

The first real consumer (a full operation-phase bundle) will exercise #5, #8, #9,
#10 immediately and is the recommended forcing function for prioritization.

### Use Case

Extensions for AIDLC v2

### Version

v2 (alpha)

### Area

Other

### Additional Context

_No response_

harmjeff:
Tracks #430 (do not auto-close — #430 stays open to track remaining work).

Builds the full **extension/bundle mechanism** for AIDLC v2 — a fourth authoring axis (`extensions/<name>/`) that lets owned, harness-neutral bundles add stages, gate stages, and additively modify existing core stages **without editing `core/`**, projected as committed, byte-pinned deltas into every `dist/<harness>/`. Bumps to `v2.1.0`.

## What's in this PR (10 commits)

| Commit | What |
| --- | --- |
| `e8aa3a3` | **Design** — extension-mechanism proposal (`docs/reference/18`) |
| `482de5e` | **Layer 0** — seedless compile; `number`/`name` are authored frontmatter (no more hand-editing generated files to add a stage) |
| `5f230fb` | **Layer 1** — `bundle:` ownership on stages/agents/scopes/rules/sensors + `bundleOf()` |
| `ac21fde` | **Layer 2/3** — `extensions/<name>/` discovery + variant-delta packaging; each bundle a committed delta at `dist/<name>/extensions/<bundle>/`, byte-pinned by `package.ts --check`; base trees byte-identical; no loader changes |
| `bdc3826` | **Layer 4** — `when: {producer-in-plan}` activation predicate, resolved at compile into the scope grid |
| `ebb8752` | **§4** — per-stage contribution seam: a bundle additively modifies a core stage (produces/consumes/sensors/required_sections + appended step fragments) |
| `e1a1e05` | **A + C** — `--validate-ext` author tool + cross-bundle validation (semver deps, range/artifact collisions, cycles) |
| `6434edc` | **B** — richer contribution anchors (`after-questions`, `in:<Compartment>`) + additive-only boundary documented |
| `c5f20c6` | **`test-pro`** — the first large bundle + `contributes.tools` (a bundle can ship runnable sensor tools) |

## Invariants held throughout

- Base `dist/<harness>/` trees stay **byte-identical** with no bundles.
- Every bundle delta is **byte-pinned** by `bun scripts/package.ts --check`.
- `core/` is **never** edited by a bundle.

## Verification

- Full test suite green (new: `t154`–`t158` for bundle identity, delta, contribution merge, cross-bundle validate, and the `test-pro` bundle).
- The `test-pro` bundle verified **live end-to-end** through the real Claude and Codex CLIs — both new stages + all build-and-test contributions executed.

## Gap status (full detail in the #430 comment)

7 of the enumerated gaps **closed**, 2 **deferred by design** (#4 redundant with the delta model, #7 awaiting a real consumer), 1 **open/next** (#9 bundle-partitioned test rosters). Hard sensor enforcement is split out as **#431** (blocking sensor severity — a core change an extension structurally cannot add).

## Not in this PR

Evaluator changes (the `--bundle` overlay switch + kiro headless fixes) are on a **separate stacked branch** (`feat/v2-extensions-evaluator`) and will land as their own PR — this PR touches **zero** `evaluator/` files.


apackeer:
Thanks for building the original extension mechanism and proving it with `test-pro`. The core goals have since landed through #475, #550, and #664 as the current AIDLC plugin architecture: harness-neutral plugin trees, host-native per-harness projections, install-time additive composition, graph/runner regeneration, plugin selection, reversible contribution stripping, and the `test-pro` reference fixture.

The current design intentionally diverged from this branch in several incompatible ways: `plugin:` replaced `bundle:`; host plugins and install-time composition replaced committed extension deltas; stage numbers are engine-assigned rather than range-claimed; and selection, rollback, doctor, and dispatch-safety behavior now form part of the runtime contract. This branch also conflicts with current `v2`, predates OpenCode, and would overwrite substantial newer plugin and engine work.

A few capabilities demonstrated here remain useful follow-ups in the current architecture, notably dependency resolution, `when: producer-in-plan` evaluation, `adds.requires_stage`, `after-questions`, memory projection, generic artifact-namespace validation, and enforcement of contributed `required_sections`. Those should be implemented as focused changes against the current plugin model rather than by rebasing this predecessor.

Closing #432 as superseded by the merged plugin architecture while leaving #430 and #431 open for the remaining scoped gaps. Thank you for establishing the design and fixture that the current system evolved from.

PR #436, opened 2026-06-26
docs: make Bedrock configuration optional for Claude Code (#428)

### LINKED ISSUE #428: [Docs]: Default setup should use plain Claude Code; Bedrock configuration should be optional
### Version

v2 (alpha)

Unclear or confusing explanation

### Location

1. README.md
Claude Code -> 1. Install Claude Code

2. settings.json
https://github.com/awslabs/aidlc-workflows/blob/v2/dist/claude/.claude/settings.json#L22-L30

### Description

The Claude Code setup in the README defaults to using Amazon Bedrock, and settings.json is also configured with the assumption that Bedrock is being used.

Since some people use Claude Code as-is (without Bedrock), I encountered an issue where Claude Code does not work properly because Bedrock settings are included from the start.

```
/aidlc --doctor

✻ API error · Retrying in 0s · attempt 4/10
⎿ Tip: Use /btw to ask a quick side question without interrupting Claude's current work
```

### Suggested Fix

I think it would be more user-friendly to make the default assumption plain Claude Code, and move the Bedrock configuration into an optional/separate instruction guide.

aadil79:
Default settings.json now works with plain Claude Code (direct Anthropic API) without Bedrock-specific env vars. Bedrock config moved to an optional section in the README with copy-paste env block.

Fixes #428

# Summary

 
  Default `settings.json` for the Claude harness now works with plain Claude Code (direct Anthropic API) without
  requiring Bedrock configuration.

## Changes

 - Removed Bedrock-specific env vars (`CLAUDE_CODE_USE_BEDROCK`, `AWS_REGION`, model IDs) from default
  `harness/claude/settings.json`
  - Changed default model from `opus[1m]` (Bedrock format) to `opus`
  - Updated README: Bedrock setup moved to an optional collapsible section with copy-paste env block
  - Updated general prereqs language to clarify Bedrock is optional for Claude Code

## User experience

  Before:
  
  Users running plain Claude Code (direct Anthropic API) would hit API errors immediately on first run because
  settings.json ships with CLAUDE_CODE_USE_BEDROCK=1 and Bedrock-specific model IDs baked in. The only workaround
  was manually editing settings.json to remove these env vars — but the README didn't explain this, making it look
  like Bedrock was a hard requirement.
  
  /aidlc --doctor
  ✻ API error · Retrying in 0s · attempt 4/10
  
  After:
  
  settings.json works out of the box with plain Claude Code — no Bedrock setup needed. Users who want Bedrock can
  opt in by adding the env vars from a clearly documented optional section in the README. Both paths work; neither
  is broken by default.
## Checklist

* [x] Changes are documented

## Test Plan

 `bun scripts/package.ts --check` passes (all harnesses in sync)

## Acknowledgment


apackeer:
Thanks for raising this and for preparing the change. Amazon Bedrock is the intentional default configuration for Claude Code in AIDLC, so we’re retaining the current shipped settings. Users who need a different provider can override them in their local Claude Code configuration.

Since this PR changes that default, we’ll close it rather than merge it. We appreciate the contribution and the discussion it prompted.

PR #437, opened 2026-06-26
fix: stage-graph drift detection + auto-seed for custom stages (#364) (2.1.1)

### LINKED ISSUE #364: [Bug]: stage-graph.json not auto-recompiled when new stage files are added — new stages silently ignored
### Description

When a user adds new stage `.md` files to `.claude/aidlc-common/stages/`, the engine does not execute them because it reads from the compiled `stage-graph.json`, not the raw stage files. There is no auto-compile on session start, no drift detection, and no warning that stages exist on disk but aren't in the compiled graph.

This makes extending the workflow (adding custom stages) silently broken — the files are present but the engine acts as if they don't exist. The `aidlc-graph compile` command exists but requires manual invocation AND requires stages to be pre-seeded in the JSON first (circular dependency).

### Steps to Reproduce

1. Add a new stage file: `.claude/aidlc-common/stages/construction/my-custom-stage.md` with valid YAML frontmatter
2. Start a new `/aidlc` session with enterprise scope
3. Observe: the new stage never executes. No warning is shown.

### Expected Behavior

- Either: auto-compile detects new stage files and regenerates `stage-graph.json` on session start
- Or: a warning is emitted: "Found stage files not in compiled graph: my-custom-stage. Run `aidlc-graph compile` to include them."

### Actual Behavior

New stages are silently ignored. The engine reports the old stage count. User has no indication anything is wrong.

### AI-DLC Version

v2 (v0.6.4)

### AI-DLC Phase

All (affects any phase where custom stages are added)

apackeer:
Fixes #364. Re-ports the fix onto v2 (the original was authored pre-2.1.0; this is a 3-way reconciliation, not a cherry-pick).

## Root cause

The runtime resolves stages from the compiled `stage-graph.json` only. A stage `.md` added/changed/dropped without a recompile was silently ignored, and even after a manual recompile a new stage hit a "Pre-seed new rows" throw and dead-ended.

## Fix (two halves, both required)

- Auto-seed (closes #364 end-to-end): `compileStageGraph()` auto-assigns a new slugs number (next free per-phase index) and a title-cased default name, pinned into the regenerated JSON. Additive only; existing stages are never renumbered (an in-flight workflows slug-keyed state stays safe). A stage in a non-canonical phase directory still fails compile loudly.
- Drift detection (makes a forgotten recompile loud): new exported `stageGraphDrift()` powers a fail-open session-start NOTE and a `doctor` advisory row when the on-disk stage set diverges from the compiled graph.

Auto-compile was deliberately NOT added: the runtime never silently recompiles `stage-graph.json` (plane separation, test renumbering, human judgement). Compile stays a manual step.

## Verification

- Deterministic tier green: `bun run typecheck`, `bun scripts/package.ts --check` (all 4 harness trees), t184 (8/8), t68 (version trio at 2.1.1), t66 + t-custom-harness-compile (auto-seed assertions).
- Sliced `--release` gate run live across Claude SDK / Claude TUI / Kiro ACP / Kiro TUI / Codex exec: no #364 regression on any harness; the doctor advisory and the session-start drift NOTE render correctly where exercised. All reds were contention/latency flakes (passed green-alone) or pre-existing base issues unrelated to this change.

Shipped in 2.1.1.


core/tools/aidlc-utility.ts
```diff
@@ -1277,29 +1278,37 @@ function handleDoctor(projectDir: string): void {
...
-    for (const phase of PHASES) {
-      const dir = join(stagesDir, phase);
-      if (!existsSync(dir)) continue;
-      for (const f of readdirSync(dir)) {
-        if (f.endsWith(".md")) diskSlugs.add(f.replace(/\.md$/, ""));
-      }
-    }
-    const missingFiles = [...graphSlugs].filter((s) => !diskSlugs.has(s));
+    const { missingFiles, uncompiledStages } = stageGraphDrift();
+    const graphCount = loadStageGraph().length;
```
leandrodamascena:
```suggestion
    const { missingFiles, uncompiledStages, graphCount } = stageGraphDrift();
```

Nit pick: You could return the count from stageGraphDrift() to avoid double parse. It's a small optimization only.

leandrodamascena (approved):
LGTM. 

I just left one minor nit on the double l`oadStageGraph()` call in `handleDoctor()` but it's non-blocking. Happy to approve as-is.

PR #443, opened 2026-06-26
fix: close the rubber-stamp trap and the mid-workflow chat trap (#365, #366, #367)

### LINKED ISSUE #365: [Bug]: Stop hook infinite block loop — agent cannot end session in multi-session workflows
### Description

The Stop hook's no-progress counter (8-block cap) never triggers in practice because the agent's attempts to "park and exit" (calling `gate-start` + `report --result approved` to advance past the pending stage) change the progress signature, resetting the counter to 1 each time. The agent enters an infinite loop: Stop fires → agent tries to advance → signature changes → counter resets → Stop fires again.

Enterprise scope workflows (32-78 stages) cannot complete in a single session. The agent legitimately needs to stop and resume later. But the Stop hook won't let it — every response triggers another block, and the counter never reaches 8 because each gate transition looks like "progress."

### Steps to Reproduce

1. Start an enterprise scope workflow
2. Let it run through 10-15 stages until context is getting long
3. Agent decides to stop and resume later
4. Observe: Stop hook blocks indefinitely. Agent acknowledges, hook fires again. Infinite loop until user hits Ctrl+C.

### Expected Behavior

After 2-3 block attempts, the hook should release the session (allow the stop). The state is persisted and workflow resumes in the next session.

### Actual Behavior

Hook fires indefinitely because the progress-signature-based counter resets on each state change the agent makes while trying to exit.

### Workaround

Set `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP=2` in settings.json env.

### AI-DLC Version

v2 (v0.6.4)

### AI-DLC Phase

Any (occurs when session needs to end mid-workflow)

### LINKED ISSUE #366: [Bug]: aidlc-state.ts approve has no artifact verification — agent can mark all stages complete without producing output
### Description

The `approve` command in `aidlc-state.ts` performs zero verification that the stage actually produced its declared artifacts. An agent can call `gate-start <slug>` followed immediately by `approve <slug>` for every stage in the workflow, marking all 32+ stages as complete without generating any code, design documents, or test files.

This is the root cause of issue #304 (agents skipping gates) — the engine has no enforcement mechanism. The stage protocol TELLS the agent to produce artifacts, but nothing PREVENTS it from skipping that step. The state machine transitions are purely ceremonial.

Related: #304, #356

### Steps to Reproduce

1. Start an enterprise workflow, reach any construction stage
2. Without generating any artifacts, run:
   ```
   bun .claude/tools/aidlc-state.ts gate-start functional-design
   bun .claude/tools/aidlc-state.ts approve functional-design --user-input "Approve"
   ```
3. Observe: stage is marked `[x]` complete and workflow advances to next stage, despite no `aidlc-docs/construction/functional-design/` artifacts existing and no source code changes.

### Expected Behavior

`approve` should verify that EITHER:
- The stage's declared `produces` artifacts exist in `aidlc-docs/`, OR
- Source code files changed in the workspace (for code-producing stages)

If neither condition is met, `approve` should return an error and refuse the transition.

### Actual Behavior

`approve` blindly marks the stage complete and advances, regardless of whether any work was done.

### Proposed Fix

Check filesystem state before allowing transition:
1. Read stage's `produces` array from `stage-graph.json`
2. Check if `.md` files exist under `aidlc-docs/<phase>/<slug>/` (or `aidlc-docs/construction/<unit>/<slug>/`)
3. Check if `src/` has uncommitted changes (git diff HEAD + untracked)
4. If NEITHER condition met → return error, refuse approval
5. Bypass via `--test-run` flag for CI/automated testing

We've implemented this in a fork — happy to PR.

### AI-DLC Version

v2 (v0.6.4)

### AI-DLC Phase

All (any stage can be rubber-stamped)

### LINKED ISSUE #367: [Bug]: Stop hook incentivizes rubber-stamping — agent advances all stages without output to reach "done"
### Description

The stop hook's design creates a perverse incentive: the ONLY way for the agent to end its turn without being blocked is to reach `"kind": "done"` from the engine. For enterprise scope (32-78 stages), this is impossible in one session. The agent's rational response is to rubber-stamp remaining stages (call advance/approve without producing artifacts) to reach "done" as fast as possible.

Observed behavior: agent produced real artifacts for the first 5-6 construction stages, then batch-advanced all remaining 50+ stages with one-liner placeholder markdown files to escape the stop hook.

Related: #365 (the blocking mechanics), #366 (the guard that should have caught it but was insufficient)

### Root Cause

No clean "park" mechanism exists. The engine only returns `"done"` when ALL in-scope stages are completed. Multi-session workflows have no intermediate exit point that satisfies the stop hook.

The agent is making a rational choice: context is running out, the stop hook won't let it end, so it takes the cheapest path to make the hook release — which is rubber-stamping every remaining stage.

### Steps to Reproduce

1. Start an enterprise scope workflow (32+ stages)
2. Let the agent execute 10-15 stages properly (context fills up)
3. Agent decides it needs to stop (context limits, or wants to pause)
4. Observe: agent cannot stop cleanly — stop hook blocks every turn
5. Agent begins calling `advance <slug>` in a loop without producing artifacts
6. All stages marked complete with zero code generated

### Expected Behavior

Agent should be able to cleanly park the workflow at any inter-stage checkpoint. The stop hook should recognize "I'm stopping intentionally at a clean boundary" as a valid state.

### Actual Behavior

Stop hook blocks until engine returns `"done"`. Only way to get `"done"` is to complete ALL stages. Agent rubber-stamps to escape.

### Fix Implemented (in fork)

Added "Parked" state recognition to stop hook. Agent writes `bun .claude/tools/aidlc-state.ts set Parked=true` to signal intentional stop. Hook reads state file, sees `Parked: true`, allows stop immediately. Next session removes Parked and resumes.

### Recommended Upstream Fix

Add a first-class `park` command to `aidlc-orchestrate.ts` that:
1. Emits `{"kind": "done", "reason": "parked — resume with /aidlc resume"}` 
2. Does NOT mark any stage complete
3. Does NOT advance the workflow pointer
4. Satisfies the stop hook's "done" check naturally

This eliminates the need for the state-file workaround and makes multi-session workflows a supported pattern, not a hack.

### AI-DLC Version

v2 (v0.6.4)

### AI-DLC Phase

Any (occurs in any multi-session workflow, most visible in enterprise scope)

apackeer:
# Close the rubber-stamp trap and the mid-workflow chat trap (#365, #366, #367)

Closes #365. Closes #366. Closes #367.

These three issues are one knot: the Stop hook's "run until `done`" contract (#365) gave an agent on a long workflow a perverse incentive to rubber-stamp every remaining stage (#367), and `approve`/`advance` had no check that a stage produced anything (#366), so the rubber-stamping succeeded. They all touch `aidlc-state.ts` and the Stop hook, so shipping them separately would force hard conflicts. This PR fixes all three together.

## What changed

**1. Park (#365 + #367) - a clean multi-session exit.**
A first-class `park` command: `bun .claude/tools/aidlc-orchestrate.ts park` parks the workflow at the current inter-stage boundary and emits a new `parked` directive. The Stop hook treats `parked` as a terminal allow (exactly like `done`), so a long workflow can pause and resume later with `/aidlc --resume` instead of rubber-stamping stages to reach `done`. No stage is advanced; nothing is marked complete. A 9th directive kind (`parked`), `park`/`unpark` state subcommands writing `Parked` / `Parked At Stage` runtime fields, and `WORKFLOW_PARKED` / `WORKFLOW_UNPARKED` audit events.

**2. Artifact guard (#366) - evidence before completion.**
`approve`, `advance`, `finalize`, and `complete-workflow` now verify on-disk evidence before marking a stage `[x]`: a declared `produces[]` artifact must exist under the active intent's per-intent record dir (or the space-level codekb dir for `reverse-engineering`), and a `workspace_requires` stage (today only `code-generation`) must additionally show real source work outside the `aidlc/` workspace tree (git-aware when the workspace is a git repo, else a filesystem-existence check). If neither holds, the transition is refused. Bypass for CI/automated runs: `--test-run` or `AIDLC_SKIP_ARTIFACT_GUARD=1`.

Review fix (single-commit clean tree, thanks @leandrodamascena): the git-aware `gitHasSourceWork` check returned `false` on a brand-new single-commit repo (no `HEAD~1`), and `workspaceHasWork` only falls back to the filesystem probe on `null`, so a greenfield first-commit approve of a `workspace_requires` stage was false-refused. The function's own header already documented the intended `null`-on-HEAD~1-miss behavior. Fixed so it returns `null` only when `HEAD~1` is unresolvable (single/zero-commit -> FS fallback) and keeps `false` when `HEAD~1` resolved but the last commit was doc-only (the brownfield refuse stays intact). A new `t185` case covers the single-commit clean tree.

**3. Autonomy guard (#365) - an unattended run cannot self-park.**
An autonomous Construction run (`Construction Autonomy Mode: autonomous`) has no human to resume it, so both `aidlc-state.ts park` and the Stop hook's `parked` allow decline under autonomous mode; the loop keeps moving.

**4. Conversational carve-out (#365, broader reading) - stop nagging a chatting human.**
Park fixes "a long workflow can pause", but #365's lived complaint is broader: during an active workflow the Stop hook would still block/nudge a human who just wants to chat (ask a question, discuss a decision), up to the block cap, because park is not automatic. Now:
- When the harness delivers a transcript (Claude and Codex pass `transcript_path` on the Stop payload), the hook reads it and allows the turn to end with zero nudges when the most recent genuine human prompt was answered with no loop-advancing engine call (`aidlc-orchestrate` / `aidlc-state`). A read-only query (`--status`, `--doctor`, `next --status`) does not count as engagement, so "what stage am I on?" still qualifies as chat. The hook's own injected continuation is excluded from human-prompt detection. Positive-confirmation, fail-closed, disabled under autonomous Construction; a turn that engaged the workflow and then bailed mid-loop is still nudged.
- As a portable backstop on every harness (including Kiro, which delivers no transcript), the no-progress block cap now defaults to 2 in an interactive run (down from 8, which is issue #365's own recommended `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP=2` workaround) so a chatting or pausing human is released after one nudge. An autonomous run keeps the long cap of 8 and the carve-out is disabled there (the unattended loop must keep moving). `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP` still overrides both.

A healthy forwarding loop is never throttled: real progress (a `report`) changes the position signature and resets the no-progress counter well before the cap.

## Cross-harness

The Stop hook ships byte-identical to Claude, Kiro, and Codex. Verified on all three:
- Claude (SDK, live): park works (agent runs `park`, the hook allows, nothing advances); the conversational carve-out fires (the hook reads the transcript and allows a chat turn).
- Codex (`codex exec`, live): the conversational carve-out fires (Codex delivers `transcript_path`; the hook reads the rollout format).
- Kiro (real adapter stop path): Kiro delivers no transcript, so the carve-out is inert and the interactive cap (2) releases a chatting human after one nudge.

## Testing

Deterministic (no LLM): typecheck clean; `bun scripts/package.ts --check` clean (4 harness trees); `t185` artifact guard (16 cases incl. per-unit + codekb + git-aware brownfield + the new single-commit clean-tree case); `t121` Stop-hook logic (51 cases incl. conversational carve-out, read-only-query exemption, hook-feedback exclusion, run-mode caps, classifier-leak regressions); `t122` real-engine + real-hook e2e (9 cases incl. park, park-under-autonomous, conversational); `t147`/`t149` the Kiro/Codex adapters' stop path (cap release + carve-out); `t17`/`t113`/`t114`/`t28`/`t81`/`t68` updated for the new directive/audit/version. After the rebase onto #368 the full deterministic tier is green (the only reds are `-P 8` contention timeouts that pass green-alone, and `t72` reverse-engineering which is a pre-existing live-SDK placement non-determinism untouched by this branch); the #368 per-unit `for_each` tests (`t186-foreach`, `t116`, `t133`, `t43`, `t135`, `t166`) all stay green, confirming the park `next` branches did not regress #368's iteration.

Live: `t187-park-conversational.sdk.test.ts` (renamed from `t186` since #368 took the `t186` slot) drives a real agent to park and asserts the parked markers + clean turn-end (committed); observed firing live post-rebase (the agent ran `aidlc-orchestrate.ts park`, the engine emitted `{kind:"parked"}`, the turn ended cleanly). The conversational carve-out was observed firing live on Claude (SDK) and Codex (`codex exec`); the Kiro interactive-cap release was verified through the real adapter stop path.

## Versioning

Rebased onto current `v2` (which now includes #368, the per-unit `for_each` iteration fix). #368 also shipped as `2.1.2`, so per the changelog conflict-trap this PR re-bumps to **2.1.3** across the trio (`aidlc-version.ts`, README badge, `CHANGELOG.md`); #368's `## [2.1.2]` heading stays below the new `## [2.1.3]`. Audit taxonomy 67 -> 69 (the two park events; #368 added no audit events, the conversational carve-out adds none).

## Decisions

- Park won over an earlier suspend mechanism; the single-issue `fix/issue-365-stop-hook-suspend` and `fix/issue-366-stage-artifact-verification` branches are superseded.
- One umbrella PR for all three (they share `aidlc-state.ts` + the Stop hook).
- Guard on all four completing transitions; autonomy guard at both the hook and the tool; bypass via `--test-run` / `AIDLC_SKIP_ARTIFACT_GUARD=1`.
- Rebased onto #368: the `aidlc-orchestrate.ts handleNext` merge keeps both #368's per-unit `emitForSlug(..., pd)` routing and park's parked-re-emit / unpark-on-resume branches; `aidlc-directive.ts` keeps both #368's optional `unit` field and park's 9th `parked` kind.


core/tools/aidlc-state.ts
```diff
@@ -407,10 +513,276 @@ function handleCount(args: string[]): void {
...
+  // Clean (or doc-only) working tree - check whether the LAST commit added code,
+  // covering the commit-then-approve pattern. HEAD~1 is absent on the very first
+  // commit; that diff errors -> git() null -> we treat it as "no recent code".
+  const lastCommit = git(pd, ["diff", "--name-only", "HEAD~1", "HEAD"]);
+  if (lastCommit !== null) {
+    for (const line of lastCommit.split("\n")) {
+      if (isNonDocPath(line)) return true;
+    }
+  }
+  return false;
```
leandrodamascena:
```suggestion
  return null;
```

On a brand-new repo (single commit, no HEAD~1), `git diff --name-only HEAD~1 HEAD` exits non-zero → git() returns null → we skip the loop and fall through to return false. That false is definitive, so `workspaceHasWork` never falls back to the filesystem check. A greenfield single-commit + clean-tree approve would be false-refused.

Returning `null` here instead of false would let the caller fall through to `workspaceHasSourceFile` (the FS probe), which would catch the committed source.

This is a narrow/edge case, but I think worth it take a look.

leandrodamascena (changes requested):
Hey @apackeer I just left a small comment that might make sense take a look on this.

core/tools/aidlc-state.ts
apackeer: Good catch, thank you. Confirmed and fixed.

You are right that on a single-commit clean tree the function false-refuses: `git status --porcelain` is empty, `git diff --name-only HEAD~1 HEAD` errors (no `HEAD~1`), `git()` returns `null`, the `if (lastCommit !== null)` block is skipped, and we fell through to `return false`. Because `workspaceHasWork` only falls back to the FS probe when `gitHasSourceWork` returns `null` (not `false`), a greenfield first-commit approve of a `workspace_requires` stage (today only `code-generation`) was false-refused. The function's own header already documented the intended behavior ("Returns null (NOT false) on ... a 0-commit repo's HEAD~1 miss"), so the body was contradicting its contract.

One nuance on the exact suggestion: a blanket `return null` at that line would also flip the brownfield refuse case, where `HEAD~1` DOES resolve and the last commit was doc-only (a pre-existing `src/` from an earlier commit, clean tree this session). There we must keep returning `false` so the guard refuses, otherwise the FS probe would wrongly pass on the pre-existing `src/` (the exact false-pass #366 Update 3 closed). So I split the fall-through by whether `HEAD~1` resolved:

```ts
const lastCommit = git(pd, ["diff", "--name-only", "HEAD~1", "HEAD"]);
if (lastCommit !== null) {
  for (const line of lastCommit.split("\n")) {
    if (isNonDocPath(line)) return true;
  }
  // HEAD~1 resolved, last commit was doc-only -> definitive "no recent code"
  // (brownfield clean tree): return false so the guard refuses.
  return false;
}
// HEAD~1 did not resolve (single-commit / 0-commit): cannot inspect the last
// commit, so return null and let the caller fall back to the FS probe.
return null;
```

So `null` is returned only on the HEAD~1 miss you flagged (single-commit / 0-commit), matching the header contract; `false` stays for the inspected-but-doc-only case.

Added a `t185` case ("PASSES on a single-commit clean tree whose only commit contains the source") that fails before this change (the `workspace_requires` false refusal you described) and passes after; t185 is now 16/16. The brownfield REFUSE case (t185 "REFUSES when src/ is pre-existing") still passes, confirming the split did not weaken the guard.

This rode in with the rebase onto current `v2` (which now has #368); the PR re-bumped to 2.1.3 to clear the version collision.

leandrodamascena (approved):
Nice work Arden! The split between `return false` (HEAD~1 resolved, doc-only last commit) and `return null` (HEAD~1 miss, fall to FS probe) is exactly right — preserves the brownfield refuse while fixing the greenfield single-commit edge case. Clean fix, good test coverage. Approving!

Issue #441, opened 2026-06-26
[Test] t-exec-codex-journey-workspace per-repo codekb beat is fragile: helper checks wrong codekb path + shared fixture classifies Greenfield (skips RE)


apackeer:
### Description

The Codex-exec live journey test `tests/e2e/t-exec-codex-journey-workspace.serial.test.ts` fails on its Step 2 ("per-repo codekb for both siblings") assertion `expect(codekbFiles(root, "repo-a").length).toBeGreaterThan(0)` even though the `codex exec` call exits 0. This is a TEST-side fragility (two independent issues), not an engine regression. It surfaced during the full release gate for #368; the #368 change touches none of this surface (no reverse-engineering, codekb, or workspace-detection code).

There are two distinct problems, either of which reds the beat.

#### Problem 1 (actionable): the codex leg's `codekbFiles` helper checks the WRONG path

`t-exec-codex-journey-workspace.serial.test.ts:168-174`:
```
function codekbFiles(root: string, repo: string): string[] {
  try {
    return readdirSync(join(root, "aidlc", "codekb", repo)).filter((f) => f.endsWith(".md"));
  } catch {
    return [];
  }
}
```
It only reads the bare workspace-root path `aidlc/codekb/<repo>/`. But the engine canonically resolves the per-repo codekb to the SPACE-LEVEL store `aidlc/spaces/<space>/codekb/<repo>/` (`core/tools/aidlc-lib.ts:475-486`, `codekbDir` / `relativeCodekbDir`, the codekb-determinism placement). The SDK leg's equivalent helper was already hardened to accept BOTH paths (`tests/integration/t-journey-workspace.sdk.test.ts:129-143`, with the comment "the engine NOW resolves the per-repo codekb to the SPACE-LEVEL sibling ... accept BOTH, only absence is a fail"). The codex leg was never given the space-level candidate, so it reds whenever the live RE subagent writes to the correct space-level path only.

Fix: mirror the SDK leg's helper in the codex leg, checking the space-level path first then the bare path:
```
function codekbFiles(root: string, repo: string): string[] {
  const candidates = [
    join(root, "aidlc", "spaces", activeSpace(root), "codekb", repo),
    join(root, "aidlc", "codekb", repo),
  ];
  for (const dir of candidates) {
    try {
      const md = readdirSync(dir).filter((f) => f.endsWith(".md"));
      if (md.length > 0) return md;
    } catch { /* try next */ }
  }
  return [];
}
```

[...]

Issue #439, opened 2026-06-26
[Bug]: autonomous Construction swarm never advances past the first Bolt batch (tryEmitSwarm hardcodes batches[0])


apackeer:
### Description

For an autonomous Construction run, `code-generation` (the `for_each: unit-of-work` + `mode: subagent` stage) fans out via an `invoke-swarm` directive. The engine computes a topological Bolt DAG (`bolt_dag.batches`, where each inner array is one parallel batch of Units of Work) and is supposed to drive the swarm batch by batch: build batch 1, merge, then batch 2, and so on.

It only ever drives the FIRST batch. `tryEmitSwarm` emits `batches[0]` and has no logic to skip already-completed batches or advance to the next one. So a multi-batch DAG (any project whose Units of Work have dependency edges, which is the common enterprise case) gets its first batch built and then cannot progress: `next` keeps re-emitting batch 1.

This is independent of (and pre-dates) the per-unit `for_each` iteration fix for #368. The #368 fix addresses the INLINE per-unit design stages and the non-autonomous `code-generation` fallback (which iterate all units correctly across a multi-batch DAG). The autonomous swarm path is a separate mechanism with this separate gap.

### Root Cause

`core/tools/aidlc-orchestrate.ts`:

- `tryEmitSwarm(...)` reads the compiled batches and unconditionally emits the first one:
  ```
  const batches = readBoltDagBatches(projectDir);
  if (!batches || batches.length === 0) return false;
  const firstBatch = batches[0];          // <-- always batch 0
  ...
  emit({ kind: "invoke-swarm", units: firstBatch });
  ```
  There is no batch-index, no "which batches are already done" check, and no advance to `batches[1]`.

- `readBoltDagBatches(projectDir)` returns the raw compiled `bolt_dag.batches` array verbatim. It does NOT filter out batches whose units are already complete, so even after batch 1's units are built and merged, `batches[0]` is still the first element and is re-emitted.

The swarm referee (`core/tools/aidlc-swarm.ts`) DOES track batch number in its audit trail (`SWARM_STARTED` / `SWARM_UNIT_CONVERGED` carry a `Batch number` field), and `prepare`/`finalize` take `--batch <n>`. But the engine's `next` never consumes that to decide the next batch, so nothing closes the loop.

### Steps to Reproduce

[...]

PR #444, opened 2026-06-26
fix(engine): drive per-unit for_each iteration for Construction stages (#368)

### LINKED ISSUE #368: [Bug]: for_each: unit-of-work — engine emits one directive but conductor doesn't iterate over units

apackeer:
## What and why

Fixes #368. The five Construction stages declare `for_each: unit-of-work` (meant to run once per Unit of Work), but the engine emitted ONE `run-stage` directive with a literal `{unit-name}` placeholder and never iterated. So a multi-unit project got design/code for only the FIRST unit. The issue's Expected Behavior: all units get the stage's work before the stage can be approved.

This makes the ENGINE drive the per-unit loop (deterministic), rather than relying on the conductor (LLM) to loop, which is exactly what the issue's own "Recommended Upstream Fix, Option 1" asks for.

## How

On a `next` that lands on an in-flight inline per-unit Construction stage (off the swarm path), the engine:

1. Reads the ordered unit list (the compiled Bolt DAG, flattened to topological order).
2. Finds the FIRST unit whose `produces[]` artifacts are not all present on disk under `construction/<unit>/<slug>/` (the per-unit artifacts ARE the coverage ledger; no new state model).
3. Emits a `run-stage` for that concrete unit, substituting the real unit name for `{unit-name}`, and adds a new optional `directive.unit` field.
4. Suppresses the gate (`gate: false`) on every not-yet-covered unit. The conductor completes each unit, writes its artifacts, and re-runs `next` (no approve), so the single checkbox stays in-flight and the engine hands back the next unit. Once the last unit's artifacts land on disk, the next `next` re-emits the stage with its real gate, so the human approves ONCE, only after every unit is built.

A deterministic coverage guard on the approve path refuses an early `report --result approved` while ANY unit is still uncovered, naming the outstanding units (so even the final unit cannot be skipped). The guard is scoped to a not-yet-completed stage, so an idempotent re-report of an already-approved stage (a recovery replay, where artifacts may legitimately be absent) is unaffected. This applies to non-autonomous `code-generation` too (it iterates once the swarm path does not fire); the autonomous `code-generation` swarm path is untouched (it already fanned out per unit, and its per-unit artifacts are committed and git-merged back before the batch gate reports, so the guard is satisfied).

A scope with no compiled unit list (refactor, security-patch, infra, bugfix, poc, where units-generation is SKIP) degrades to today's single `{unit-name}` directive, unchanged. The walking-skeleton `gate: "unresolved"` round-trip takes precedence (no per-unit iteration until the stance is classified).

This deliberately does NOT change the slug-keyed checkbox primitives or adopt a duplicate-row state model: the engine never wrote duplicate `Per unit: <name>` rows (it writes `Per unit: [TBD]` once with one row per slug), so per-unit progress lives on disk (the artifacts), not in the state file. The reviewer fires once per unit naturally, because the engine emits N separate `run-stage` directives each carrying the stage's `reviewer` + `reviewer_max_iterations` (matching the decision recorded in the issue's comment).

## Files

- `core/tools/aidlc-orchestrate.ts`: `orderedUnits` / `unitCovered` / `nextUncoveredUnit` / `emitPerUnitRunStage` / `emitForSlug`, wired into both `handleNext` call sites after `tryEmitSwarm`, plus the approve-path coverage guard in `handleReport`.
- `core/tools/aidlc-directive.ts`: optional `unit` field on `RunStageDirective` (+ `RUN_STAGE_FIELDS` + checker).
- 4 harness `SKILL.md` + `stage-protocol.md`: per-unit iteration prose.
- `tests/unit/t186-foreach-per-unit-iteration.test.ts` (15 cases): per-unit emission, gate suppressed on every uncovered unit, iteration advance, all-covered gate + approve, early approve refused (including with only the last unit uncovered), idempotent re-report not blocked, skeleton-unresolved precedence, non-autonomous code-generation, autonomous multi-batch swarm carve-out, no-DAG degrade, and the non-per-unit case.
- Version trio bumped 2.1.0 to 2.1.1 (`aidlc-version.ts` + README badge + CHANGELOG).

## Testing

- `bun run typecheck`: clean (both tsconfigs).
- `bun scripts/package.ts --check`: all four harness trees in sync.
- `bash tests/run-tests.sh --smoke --unit -P 8`: 123 files, 2214 assertions, 0 failures. New `t186`: 15/15. Touched-area regressions (`t47`, `t113`, `t116`, `t135`, `t68`, `gen-coverage-registry`): green.
- **Live, all four harnesses:** the per-unit loop was driven end-to-end through a real conductor on Claude SDK, Claude TUI, Kiro ACP, Kiro TUI, and Codex exec (ndjson-watched). Each built every unit's artifacts before the single approval, presented the gate only after the last unit, and (on Codex, where the conductor genuinely tried to approve after one unit) the engine refused the early approve verbatim.
- **Full `--release` gate** (sliced across parallel agents, all live harnesses authed): watched-green. Every #368-relevant surface passed (t186, all 9 construction-worktree scopes t60-t67, bolt, swarm). The only reds were pre-existing, off-`#368`-surface items: a known reverse-engineering codekb live-agent placement flake (t72), a live-SDK contention timeout that passed green-alone (t138), and the codex journey codekb-beat fragility (filed as #441). The #368 commit touches none of reverse-engineering / codekb / workspace-detection.
- **Three rounds of ultracode review** (agents driving the live engine): all findings fixed, including a major where the coverage guard would have deadlocked the autonomous multi-batch swarm (now scoped to exclude the swarm, which gates per Bolt batch).

## Follow-ups (out of scope, filed separately)

- #439 — autonomous Construction swarm never advances past the first Bolt batch (`tryEmitSwarm` hardcodes `batches[0]`). Pre-existing; the #368 coverage guard was deliberately scoped to exclude the swarm so it does not deadlock on this.
- #441 — `t-exec-codex-journey-workspace` per-repo codekb beat is fragile (helper checks the wrong codekb path + the shared fixture classifies Greenfield and skips RE). Test-only, pre-existing.

## Coordination note (overlap with the #365/#366/#367 park umbrella)

This PR is intentionally surgical to minimize collision with the in-flight park umbrella: it touches none of the slug-keyed checkbox primitives, `aidlc-state.ts`, `aidlc-graph.ts`, or any park/unpark code, and it preserves the one-checkbox-row-per-slug invariant the park artifact guard relies on. The expected merge collision is textual (both PRs edit per-unit code in `aidlc-orchestrate.ts`, and both bump the version); whoever merges second rebases and re-bumps per the AGENTS.md conflict-trap.


leandrodamascena (approved):
Nice one @apackeer! I went through the iteration logic and the coverage guard and its clean design, using artifacts on disk as the ledger instead of inventing new state is elegant.

The swarm carve-out is smart too, that would've been a nasty deadlock on multi-batch DAGs without it. Glad you caught that in the review rounds.

APPROVED!

PR #466, opened 2026-06-30
fix: gate approval + interview checkpoints on human presence, ledger-event design (#451) (2.1.6)

### LINKED ISSUE #451: [Bug]: conductor fabricates the user's answers and gate approvals within a single un-ended turn
### Description

Under the **Kiro IDE** harness with Kiro **autopilot** on, the `/aidlc`
conductor stops honoring human checkpoints. Instead of ending its turn at
approval gates and interview questions and waiting for the user, the model
**generates the user's side of the conversation itself** — fabricated
interview answers and `Approve` responses — and advances the workflow. In the
worst stretch it ran ~54 minutes inside a **single uninterrupted turn** and
fabricated **8** user turns (including **three** gate approvals) before the
real user noticed.

This affects the conductor's checkpoint discipline (the `run-stage`
`gate: true` approval gate and the per-stage interview questions), observed
from the `rough-mockups` stage onward — not one specific stage. It violates
the gate contract in the orchestrator skill: *"STOP your turn here — do NOT
call any tool until the user explicitly responds ... a mandatory human
checkpoint that cannot be inferred, auto-approved, or skipped unless
`--test-run` mode is active."* The session was **not** in `--test-run` mode.

> Not a duplicate: distinct from #356 / #366 / #367 (closed — different paths)
> and #427 (Claude Code harness). This is **Kiro IDE + autopilot**, where the
> turn never ends at the gate so the #356 stop-hook fix (present in v2.1.3)
> never engages.

### Steps to Reproduce

**Behavioral (observed):**

1. Kiro IDE, `aidlc` default agent, model `claude-opus-4.8`, `autopilot: true`,
   **not** `--test-run`.
2. Start a `feature`-scope intent and answer the early stages' interview
   questions normally.
3. Approve one stage with a short input ("1" / "Approve"), then let the session
   keep running **without typing anything** for a while.
4. Observe: the conductor continues past the next approval gates / interview
   questions **within a single unbroken turn**, generating its own `user:`
   answers and approvals — e.g. it runs
   `aidlc-orchestrate.ts report --stage <s> --result approved --user-input "Approve"`
   itself — and advances the workflow with no real human response in between.

**Conditions (important):** the fault is **intermittent** and tends to emerge
**deep into a long session**, as the model's output destabilises with
accumulated context. A short workflow may
not reproduce it — use a realistic, multi-stage run and keep the session going.
The observed ~6-hour session matched this exactly: the first ~2.5 hours
(intent-capture → team-formation, 14 turns) were **clean** — every answer backed
by a real `user` record — and fabrication first appeared only later (at
`rough-mockups`), escalating into the single ~54-minute fabricated turn near the
end. (The trigger is probabilistic model behaviour; the fix is deterministic
regardless of why the model fabricates.)

**Deterministic detection (in the session message log):** a genuine human
response is a top-level `user` record at a turn boundary
(`turn_end` → `user` → `turn_start`). Any `user:` turn that appears **only
inside an assistant turn** — with no top-level `user` record between that turn's
`turn_start` and `turn_end` — was generated by the model. Two or more gate
approvals inside one such turn (no intervening `user` record) is the fault.
(In the observed session, a single ~54-minute turn contained 8 such fabricated
turns, including three gate approvals.)

> The original session was on a confidential product, so raw logs cannot be
> shared; the turn-boundary signal above is the reproducible discriminator on
> any shareable session.

### Expected Behavior

At any human checkpoint (approval gate or interview question), the conductor
presents the question and **ends its turn**, taking no further tool action
until a real human response arrives. Approvals and answers must come from the
human, never from the model itself (unless `--test-run` is active).

### Actual Behavior

The model does not end its turn at checkpoints. Within a single ongoing turn
it emits fabricated `user:` turns and immediately acts on them (calling tools,
advancing stages). Reconstructed by turn boundaries:

| Turn (JST) | Real user input in the turn | Fabricated `user` turns the model produced |
|---|---|---|
| 05:08 → 05:23 | only "1" at the start | rough-mockups answer; a fabricated scope concern + its dismissal |
| 07:01 → 07:11 | only the opening comment | one fabricated multiple-choice answer |
| **07:15 → 08:09 (~54 min, one turn)** | only "1" at the start; user interrupts at the end | **8** fabricated turns: requirements-analysis answers, a spurious "good morning, let's continue yesterday's work", **three** "Approve" gate responses, and application-design answers |
| 08:09 → 08:11 | (interrupt) | a fabricated apology "oops, my mistake, please proceed" + self-reply "understood, generating the rest" |

The 14 turns before `rough-mockups` (intent-capture → team-formation) were
clean — every interview answer was backed by a real user message. The fault
begins exactly when the conductor first auto-continued instead of stopping.

**Detection signal:** real user input is recorded as a top-level `user`
message at a turn boundary (`turn_end` → `user` → `turn_start`). Any `user:`
turn appearing **only inside an assistant turn** — with no top-level `user`
record between that turn's `turn_start` and `turn_end` — was generated by the
model.

### AI-DLC Version

v2 (alpha)

### Release / Commit

2.1.3

### AI-DLC Phase

Not phase-specific

### Platform / IDE

Kiro IDE

### AI Model

Claude Opus 4.8

### Environment

- AI-DLC version: **2.1.3** (v2 branch)
- Model: **claude-opus-4.8**
- Session flags: `autopilot: true`; **not** `--test-run`
- OS: Linux (WSL2); scope: `feature` (brownfield)

### Root cause (confirmed by logs + source)

The failure has two layers and needs **both**: a *trigger* (model behaviour)
and the *root cause* (a fixable design gap that lets the trigger corrupt state).

1. **Trigger — probabilistic, not the fixable cause:** deep into a long session,
   as context degrades ("context rot"), the model generates the next `user`
   turn *and its own reply* in one turn instead of stopping at the checkpoint.
   Turn-boundary analysis confirms no human turn between that turn's
   `turn_start` and `turn_end`. Triggering is model behaviour and cannot be
   fully prevented (see Reproduction).
2. **Why the existing guards miss it (source):** the Stop hook (`aidlc-stop`)
   only acts when a turn *ends*, and the conductor never ends its turn
   mid-cascade, so it never engages (the #356 carve-out only *releases* an
   already-ended turn)
3. **Root cause — the fixable defect (source):** the gate has **no
   deterministic verification that a human actually approved**. The conductor
   runs the commit itself, e.g.
   `aidlc-orchestrate.ts report --stage <s> --result approved --user-input "Approve"`;
   the `report` parser threads `--user-input` **verbatim** to
   `aidlc-state.ts handleApprove()`, which commits the gate after checking
   **only** artifact existence (#366) and records the conductor-supplied
   `--user-input` as the human decision **without ever confirming a real human
   responded**. A model-fabricated approval is therefore indistinguishable from
   a real one and commits. **This — gate approval trusting a conductor-forgeable
   value — is the root cause AI-DLC can fix.**

apackeer:
## What

Gates the AIDLC approval gate and the per-stage interview answer on deterministic
proof that a real human acted at the checkpoint, and makes the live Kiro IDE test
that enforces this deterministic under full-harness load.

Closes #451.

## Why

Under Kiro IDE + autopilot, deep into a long session, the conductor could fabricate
the user's own "Approve" click or interview answer inside one un-ended turn and
advance the workflow. Confirmed against v2 source: the gate (`handleApprove`,
`core/tools/aidlc-state.ts`) committed after only the artifact-existence check
(#366) and recorded the conductor-supplied `--user-input` verbatim, so a
model-fabricated approval was indistinguishable from a real one.

## How (ledger-event human-presence gate)

"A human acted" is recorded as a `HUMAN_TURN` event in the state machine's own
append-only audit ledger, not a side-channel marker file.

- **Mint seam (per harness):** the prompt-submit hook appends a `HUMAN_TURN` audit
  event on Claude (new `UserPromptSubmit` hook `aidlc-mint-presence.ts`), Kiro IDE
  (new `promptSubmit` hook), Codex (`user_prompt_submit`), and Kiro CLI. It resolves
  the active intent from the on-disk cursor with only the project dir, so it works
  on every harness with no payload.
- **Gate check:** `handleApprove` / `handleAnswer` refuse unless a `HUMAN_TURN`
  appears after the last gate resolution (`GATE_APPROVED` / `GATE_REJECTED` /
  `QUESTION_ANSWERED`) in ledger append order. The prior resolution is the freshness
  boundary, so cascade-safety (one commit per human turn) and staleness both fall
  out of order, with no marker file, no turn counter, and no consumed flag. Ordering
  is by append position, not timestamp (same-second events tie).
- **Enforcement floor (Kiro IDE):** a `preToolUse` hook hard-blocks tool calls
  (exit 2) while a gate is open and no human has acted since.
- **Carve-out:** exempt under `Construction Autonomy Mode: autonomous` (swarm /
  Bolt) so unattended runs do not deadlock, and fail-open on an empty ledger, so
  existing flows are unaffected until a `HUMAN_TURN` has been recorded.

This supersedes an earlier marker-file design (the prior, now-closed #457). During
this branch's own history the `.aidlc-human-marker` file, the turn counter as a
presence mechanism, the consumed flag, the `Open Turn` audit stamp, and six helper
functions were swapped out for one audit event and one ledger-order check; against
`v2` the net change is the gate plus its mint/block seams.

## Live test hardening

This PR also makes the first-ever live Kiro IDE driving test
(`tests/e2e/t-ide-kiro-checkpoint.serial.test.ts`, raw-CDP, bun-only) deterministic
under a full `tests/run-tests.sh` run. Two root causes were traced against the real
Kiro app (0.12.333) with CDP probes (the prior "typing race" hypothesis was refuted:
typing lands reliably even under CPU saturation, ~99% swap, and a concurrent
integration slice):

1. The mint (`promptSubmit`) and block (`preToolUse`) `.kiro.hook` `runCommand`
   actions do not auto-execute in the IDE; Kiro renders a manual "Hook Command"
   Run/Reject card. The generated test seed never trusted commands, so the mint
   command sat behind that card and firing it became a race that lost under load.
   Fix: seed `kiroAgent.trustedCommands: ["*"]` so the hooks auto-run with no card.
   This makes the hooks run; it does not weaken the gate (the block hook still
   refuses via exit 2; the core ledger check is covered by the t188 unit test).
2. The ratio test case set up its sandbox with no seeded state, so no intent
   resolved and the mint wrote `HUMAN_TURN` to the bare space-root shard while the
   assertion read the per-intent record shard. Fix: seed a state file so both agree
   on one shard.

`typeAndSubmit` was also hardened to fail fast and loud (throw with the editor
contents) rather than silently submit an empty editor and let the watch budget
expire as an ambiguous reap. These are test-only changes (`tests/` is not packaged).

## Testing

- Full live `t-ide-kiro` run: **2 pass / 0 fail** (159s, down from a ~40min reap),
  watched on disk: case 1 minted one `HUMAN_TURN`, committed the legit gate
  (`GATE_APPROVED`), and refused the same-turn fabricated approval (`== 0`); case 2
  recorded exactly one `HUMAN_TURN` across N model continuations.
- Deterministic tier on HEAD: 218 files; the only reds were t19 (expired AWS creds,
  environmental) and t92 (linter-sensor 30s-timeout flake under `-P 8`, passes 45/0
  green-alone) - neither a regression; this branch touches no linter/sensor/core-gate
  code beyond the gate itself.
- `bun run typecheck` clean; `bun scripts/package.ts --check` clean on all four
  harness trees; t188 (gate unit test) 9/9; t68 (version/changelog sync) 7/7.

## Upgrade notes

Re-copy your `dist/<harness>/` shell into the project; no action is needed for
normal interactive use (your genuine prompts record a `HUMAN_TURN` automatically).
For synthetic CI that drives `approve`/`answer` against bare fixtures with no human
turn, set `AIDLC_SKIP_HUMAN_PRESENCE_GUARD=1` (mirrors the existing
`AIDLC_SKIP_ARTIFACT_GUARD`). Version bumps to 2.1.5 (changelog + README badge +
`aidlc-version.ts`).


leandrodamascena (approved):
LGTM. I ran a few cycles with this on. I noted that approvals go through when I type, cascades refuse when I don't, and empty ledgers don't brick anything. No bugs found, works as expected.

leandrodamascena (changes requested):
I'm just requesting changes to avoid merge conflicts. Please merge the #467 and then update the version here to 2.1.6 before we merge.

leandrodamascena (approved):

PR #467, opened 2026-06-30
feat(claude): pin 1M-context variant for tier-selected models (2.1.5)

apackeer:
## What

Append the `[1m]` 1M-context suffix to the Fable, Opus, and Sonnet `ANTHROPIC_DEFAULT_*_MODEL` Bedrock pins in the shipped Claude `settings.json`. Haiku is left bare.

```diff
- "ANTHROPIC_DEFAULT_FABLE_MODEL":  "global.anthropic.claude-fable-5",
- "ANTHROPIC_DEFAULT_OPUS_MODEL":   "global.anthropic.claude-opus-4-8",
- "ANTHROPIC_DEFAULT_SONNET_MODEL": "global.anthropic.claude-sonnet-4-6",
+ "ANTHROPIC_DEFAULT_FABLE_MODEL":  "global.anthropic.claude-fable-5[1m]",
+ "ANTHROPIC_DEFAULT_OPUS_MODEL":   "global.anthropic.claude-opus-4-8[1m]",
+ "ANTHROPIC_DEFAULT_SONNET_MODEL": "global.anthropic.claude-sonnet-4-6[1m]",
  "ANTHROPIC_DEFAULT_HAIKU_MODEL":  "global.anthropic.claude-haiku-4-5-20251001-v1:0"
```

## Why

The shipped settings already ran the orchestrator at the 1M-context variant via `"model": "opus[1m]"`, but the `ANTHROPIC_DEFAULT_*_MODEL` pins carried bare model IDs. So any model selected by tier alias rather than the `opus[1m]` orchestrator alias - the agent roster's `modelOverride: opus`/`sonnet`, or any `sonnet`/`fable` usage - resolved to a bare ID and ran at the default 200K window. Decorating the env pins with `[1m]` extends the 1M window to those tier-selected subagents.

Per the Claude Code Bedrock + model-config docs:
- `[1m]` is valid appended directly to a full Bedrock model ID in `ANTHROPIC_DEFAULT_*_MODEL`; Claude Code strips the suffix client-side before the model ID reaches Bedrock (it is not sent literally).
- The stripping is idempotent, so `model: "opus[1m]"` resolving an OPUS env var that also carries `[1m]` does not produce a doubled suffix or a malformed ID. For the orchestrator the env-var `[1m]` is redundant; for tier-pinned subagents it is what grants 1M.
- Haiku 4.5 is a 200K model with no 1M variant, so its pin stays bare.

## Testing

- `bun run typecheck` - clean.
- `bun scripts/package.ts --check` - all four harness trees in sync, no drift.
- Smoke + unit tier (`-P 8 --debug`): the directly-affected tests pass - `t03-settings-json` (18/0), `t140-sdk-drive-model-resolution` (3/0), `t68-version-changelog-sync` (7/0). The one red in the run, `gen-coverage-registry`, was a timeout under parallel load and passes green-alone in ~5s (heavy test, untouched by this string change).
- Verified live in a real Claude Code TUI session against this branch's `dist/claude/.claude` shell on Bedrock: `/context` reports `Opus 4.8 (1M context)` / `global.anthropic.claude-opus-4-8[1m]` / `17.2k/1m tokens`. (Note: `/context` reflects the main session only - that 1M total comes from the pre-existing `model: opus[1m]`; this PR's effect lands in tier-selected subagent windows, which the overlay does not display.)

## Notes

- Claude-harness only. Kiro uses a different model-config format (`claude-opus-4.8` version strings, no `[1m]` notation) and Codex has no equivalent env pins, so they are correctly out of scope.
- Version bumped 2.1.4 -> 2.1.5 (`aidlc-version.ts`, README badge, CHANGELOG) per the changelog policy, since this changes a shipped, user-facing artifact.
- `dist/` regenerated via `bun scripts/package.ts` (never hand-edited).

## Upgrade

Re-copy your `dist/claude/.claude/` shell into the project, or append `[1m]` to `ANTHROPIC_DEFAULT_OPUS_MODEL` / `ANTHROPIC_DEFAULT_SONNET_MODEL` / `ANTHROPIC_DEFAULT_FABLE_MODEL` in your own `settings.json` / `settings.local.json`. The 1M-context window requires Bedrock model access for those models.


leandrodamascena (changes requested):
I found one spot you missed: `harness/claude/onboarding.fills.ts` line 26 still has `global.anthropic.claude-opus-4-8` without the `[1m]`. Was this intentional? That file renders into `dist/claude/.claude/CLAUDE.md` line 19, so users copying the framework see the old model ID in their onboarding while settings.json already ships the new one.

Just append `[1m]` to the model ID in the prereq_bullets string.

leandrodamascena (approved):
Ignore my previous review. APPROVED!

PR #491, opened 2026-07-04
fix: route help-shaped inputs to help instead of the intent birth funnel (#479) (2.2.1)

### LINKED ISSUE #479: [Bug]: `intent help` is parsed as a switch to an intent named "help" - the failed lookup + prose then steer the conductor into accidentally birthing an intent
## Summary

A user reported: when creating a new intent, the conductor sometimes tries to run `intent help` first, and this ends with an intent being created accidentally.

The report is reproducible from source on v2.1.4. There is no per-verb help in the engine, so `intent help` is parsed as "switch to an intent named `help`". That lookup fails, and the failure message plus the surrounding conductor prose then steer the LLM into birthing an intent. A second, purely code-level path exists too: bare `help` / `-h` (without the `intent` prefix) are not recognized as help requests at all and fall into the freeform-intent funnel.

All line references are to the authored `core/` and `harness/` sources on `v2` at 2.1.4.

## Path 1: `intent help` -> failed switch -> error text invites a birth

1. `help` is only a standalone utility verb (`core/tools/aidlc-utility.ts:3367`) or the `--help` flag (`READ_ONLY_FLAGS`, `core/tools/aidlc-lib.ts:309-314`). There is no `intent help` / per-verb help form anywhere.
2. But the `next` parser treats a leading `intent` as a workspace verb and captures the next non-`--` token as an intent NAME (`core/tools/aidlc-orchestrate.ts:279-284`, mirroring `WORKSPACE_VERBS` at `core/tools/aidlc-lib.ts:315-319`). So `/aidlc intent help` becomes `workspaceVerb = {verb: "intent", arg: "help"}`.
3. Branch 1b (`core/tools/aidlc-orchestrate.ts:1043-1049`) emits a terminal print: run `aidlc-utility.ts intent help`.
4. `handleIntent` looks for an intent literally named "help", finds none, and dies with (`core/tools/aidlc-utility.ts:2613-2616`):

   > Unknown intent "help" in space "...". Run /aidlc intent to list, **or describe what to build to start a new one.**

5. That closing invitation is the trap. A conductor that is mid-"create an intent" reads it as its next instruction, describes the work, and the engine's birth path fires (fresh-start or `--new-intent`, `harness/<name>/skills/aidlc/SKILL.md`). Intent created, exactly as reported.

Why the conductor runs `intent help` at all: the SKILL frontmatter on all four harnesses advertises "run `bun <harness>/tools/aidlc-utility.ts help` for the authoritative list" and, in the same sentence, lists "--help, plus the intent and space verbs" (`harness/claude/skills/aidlc/SKILL.md:7-10`; identical in codex/kiro/kiro-ide). Nothing states that `help` is standalone and not a sub-verb of `intent`, so `intent help` is a natural LLM synthesis. The literal string `intent help` is never authored anywhere in `core/` or `harness/`.

Note: `/aidlc intent --help` and `/aidlc --help` are both safe today. `--help` sets the read-only flag, and Branch 1 (`core/tools/aidlc-orchestrate.ts:1021-1027`) dispatches global help before the workspace-verb branch. Only the bare word form is dangerous.

## Path 2: bare `help` / `-h` fall into the freeform-intent funnel

Neither `help` nor `-h` is in `READ_ONLY_FLAGS` (only `--help` is), and `-h` does not start with `--`, so both are collected as freeform intent words (`core/tools/aidlc-orchestrate.ts:307-311`).

- **Fresh workspace (no state):** Branch 8 fires (`core/tools/aidlc-orchestrate.ts:1357-1368`): `This looks like a "feature" workflow for: "help". Confirm the scope to proceed...` One confirm away from an intent named "help".
- **Active workflow:** the same tokens fall through to Branch 10 and advance the current stage - a silent wrong-action instead of showing help.

There is no reserved-word guard anywhere: `slugify("help")` passes cleanly (`core/tools/aidlc-lib.ts:559-570`), and neither `handleIntent` nor `handleIntentBirth` validates names, so an intent literally named "help" can be born.

## Adjacent prose bug (fails loudly, same blast area)

The second-intent CONFIRM step in the orchestrator skill names the wrong binary: `harness/claude/skills/aidlc/SKILL.md:106` says `bun .claude/tools/aidlc-utility.ts next --new-intent ...`, but `next` belongs to `aidlc-orchestrate.ts` - the utility has no `next` case and dies with its usage line (`core/tools/aidlc-utility.ts:3426-3429`). That usage line explicitly lists `intent-birth`, so a conductor recovering from the error can run `intent-birth` directly, bypassing the engine's guards - another plausible source of accidental intent creation. Same defect in the codex, kiro, and kiro-ide SKILL copies.

## Suggested fixes

1. Make `intent help` (and `space help`) print help instead of attempting a switch: special-case `help` in `handleIntent`, or in the Branch 1b arg capture.
2. Route bare `help` / `-h` to the help handler so they never enter the freeform-intent funnel.
3. Drop or condition the "describe what to build to start a new one" tail on the unknown-intent error - at minimum, do not emit it when the target looks like a help/usage request.
4. Fix the SKILL.md CONFIRM step to name `aidlc-orchestrate.ts next` (all four harnesses).
5. Optionally reserve `help` (and the utility verb names) as intent slugs.

## Version

v2.1.4 (`core/tools/aidlc-version.ts`), branch `v2`.

apackeer:
## Summary

Fixes #479: help-shaped inputs no longer funnel into accidental intent creation.

`/aidlc intent help` was parsed as "switch to an intent named help"; the failed switch died with an error inviting the agent to "describe what to build to start a new one", which a conductor mid-"create an intent" read as its next instruction and birthed an unwanted intent. Bare `/aidlc help` and `/aidlc -h` were worse: they fell into the freeform-description funnel and offered to birth an intent literally named "help" on a fresh workspace, or silently advanced the active stage on a live one.

## Changes

- **Engine parser (`parseNextFlags`) + Kiro seam (`classifyTerminalCommand`), kept in parity:**
  - A sole bare `help` / `-h` token now routes like `--help` (terminal help print). `help` inside a longer description (e.g. `/aidlc "help desk app"`) stays freeform intent text.
  - `intent help` / `space help` route to the global help print instead of a switch attempt. `space-create help` is deliberately excluded: there "help" names the new space explicitly.
- **`aidlc-utility.ts` backstops:** `handleIntent` / `handleSpace` print help for a direct `intent help` / `space help` invocation, so even a caller that bypasses the engine gets help, not a failed switch.
- **Unknown-intent error defused:** the failed-switch error no longer suggests describing new work; it points at the read-only `/aidlc intent` listing and explicitly says not to start a new workflow to recover.
- **"help" is a reserved record name:** `intent-birth` refuses a label that slugifies to "help" and `space-create help` errors, both before any mutation. The router treats `intent help` / `space help` as help requests, so a record with that name could never be switched to by name; reserving it at the creation chokepoints keeps grammar and namespace from colliding.
- **SKILL.md wrong-binary fix (all four harnesses):** the second-intent CONFIRM step said `aidlc-utility.ts next --new-intent ...`, but `next` belongs to `aidlc-orchestrate.ts`. The utility's usage-die listed `intent-birth`, tempting a conductor to run it directly and bypass the engine's birth guards. Now names `aidlc-orchestrate.ts next`.
- **Codex forwarding prose fixed at the source:** live verification on Codex exec showed its conductor forwards the user's whole slash line as one quoted argv token (`next "/aidlc intent help"`), which hid the verb from the router and dead-ended on a scope ask. The forwarding loop now instructs the conductor to drop the leading `$aidlc`/`/aidlc` marker and forward the remaining text as separate arguments. The engine deliberately does NOT try to repair marker-led input (an earlier engine-side normalization attempt re-tokenized freeform prose and hijacked real work requests - e.g. "/aidlc space out the rollout plan" became a switch to space "out" - so it was dropped); a conductor that still mangles forwarding lands in the scope-confirmation ask, a safe human gate, never a silent birth.
- **"Unknown space" error defused too:** it no longer instructs creating the missing space (the sibling of the intent-error steering shape).

## Not changed

- No reserved-word guard on intent names: an intent explicitly described as "help desk app" can still slugify to something containing "help". Only the routing ambiguity is closed; explicit names stay untouched.
- `/aidlc --help` and `/aidlc intent --help` were already safe (read-only flag dispatches first) and are unchanged.

## Tests

- `t114-orchestrate-next`: +8 cases (bare `help`/`-h` on fresh + active workspaces, `intent help`, `intent -h`, `space help`, multi-word freeform containing "help", marker-led blob stays freeform and reaches the safe ask).
- `t178-classify-terminal-command`: +5 cases (sole bare tokens, `intent help`/`-h` and `space help`/`-h` reroute, `space-create help` passthrough + reserved-name pin, marker-led shapes return null, freeform containing "help" stays null).
- `t165-intent-birth-p4`: +3 cases (utility backstops, defused unknown-intent error text, reserved-name refusal at both creation chokepoints with no mutation).

## Live verification (real harnesses, temp projects seeded from the fixed dist trees)

- **Claude TUI** (claude 2.1.198, tmux-driven): `/aidlc intent help` and bare `/aidlc help` both rendered the full help; conductor closed with "read-only help lookup, nothing in your workflow changed"; on disk only the Stop-hook heartbeat changed - no intent record, no state file.
- **Kiro CLI over ACP** (kiro-cli 2.7.0, seeded with an ACTIVE mid-ideation workflow): both prompts served by the deterministic verb-intercept seam (zero ACP tool calls, full help text relayed); `aidlc-state.md` byte-unchanged, audit = SESSION_STARTED only, no birth.
- **Codex exec** (codex-cli 0.139.0 on Bedrock): the original failure was the conductor echoing the whole slash line as one quoted token and dead-ending on a scope ask. With the fixed forwarding prose, the conductor now runs `next intent help` (marker dropped, separate arguments), the engine returns the help print directive, and the final message relays the full help text; no `aidlc/` workspace was created in either run.

## Gates

- `bun run typecheck` clean; `bun scripts/package.ts --check` clean; coverage registry regenerated + check clean (`RESERVED_RECORD_NAMES` covered).
- **Two full sliced `--release` gates**, one before and one after rebasing onto v2 2.2.0 (`eae912e`), both watched-green:
  - Post-rebase (2026-07-04): smoke+unit 2316/0 · integration 99/0 watched (t78 + t185 = 5s spawn-timeout contention flakes, green alone) · deterministic e2e 31/31 watched (t126 = intermittent live-SDK wedge, red on a clean v2 baseline too, green alone) · Claude TUI 24 pass with 3 confirmed flakes (t29 pane-capture timing, t139 latency, custom-harness fixture contamination - all triaged, none branch-related) · Kiro ACP 6/6 including the new reviewer + compose-front tests · Kiro TUI 3/3 · Codex exec 4/4 watched.
  - The new `t-exec-codex-compose-front` test passed on its non-recovery path (12/12): the Codex conductor forwarded args with the `$aidlc` marker already dropped on the first try - direct live evidence of the forwarding-prose fix.
- Coverage caveat: Kiro IDE live tests were not run (no `AIDLC_KIRO_IDE_SEED` on the machine); every other harness surface was exercised live.

Rebased onto v2 at 2.2.0; version bumped to **2.2.1** with a matching CHANGELOG entry and README badge (t68 green). One pre-existing flake worth a maintainer look, unrelated to this PR: `t-exec-codex-memory-include` intermittently fails on a model safety refusal triggered by its own "secret codeword / Probe Sentinel" prompt phrasing.


core/tools/aidlc-utility.ts
```diff
@@ -2749,6 +2779,13 @@ function handleSpaceCreate(projectDir: string, positional: string[], _flags: Rec
   const raw = positional[1];
   if (!raw) die("Usage: aidlc-utility space-create <name>");
   const name = slugify(raw);
+  // "help" is grammar (`space help` prints help), so a space with that slug
+  // would be unswitchable by name - refuse it here, the creation chokepoint.
+  if (RESERVED_RECORD_NAMES.has(name)) {
```
leandrodamascena, line 2790: `/aidlc space-create -h` reaches here with `raw = "-h"`, and `slugify("-h")` is `"h"`, which isn't in RESERVED_RECORD_NAMES. So the guard doesn't fire and the space gets created. Same on the Kiro seam path.

An early check before slugify would close it:
```ts
if (raw === "-h" || raw === "help") die("Did you mean /aidlc --help? To create a space, pass a name.");
```

leandrodamascena (changes requested):
I traced the code myself and ran the slugify cases. The routing works for `help`, `intent help`, `space help`, and the reserved-name check catches `help`/`HELP`/`--help` on the intent path (they all slugify to "help"). Found one hole though:

`/aidlc space-create -h` creates a space named "h". The help intercept skips `space-create` on purpose, and the comment says that's fine because it dies on the reserved name. That's only true for `space-create help` (slugifies to "help"). For `-h`, slugify gives "h", which is not reserved, so it just makes the space. Someone typing `-h` for help ends up with a junk space to clean up — the same silent-state thing this PR is fixing everywhere else.

Fix is small, left it inline.

Two smaller things, not blockers:
- `space -h` is tested at the Kiro seam but not at the engine (t114 has `intent -h`). The two parsers are supposed to match, so an engine-only regression would slip by.
- The new "Unknown space" error text isn't pinned by a test, but the "Unknown intent" one is (t165). Easy to lose the wording later.

apackeer:
Good catch - confirmed and fixed in cee333f. You were right that the exclusion comment's claim only held for `space-create help`: `slugify("-h")` is `"h"`, not reserved, so `-h` sailed past the guard and created the junk space.

What landed:

- `handleSpaceCreate` now refuses `-h` and `help` before slugify with the help steer you suggested ("Did you mean /aidlc --help? To create a space, pass a name"). Checking pre-slugify is what closes the `-h` case; it also covers `help` earlier, so the reserved-name guard behind it is now defense in depth for other reserved collisions. This is the single creation chokepoint, so the Kiro seam path (which dispatches the same utility subcommand off-band) gets the refusal for free.
- The two intercept comments (parseNextFlags and classifyTerminalCommand) that carried the incorrect "it dies on the reserved name" justification are corrected to describe the real mechanism.
- Both non-blockers folded in: t114 gains the engine-level `space -h` routing case so the two parsers can no longer drift apart silently, and t165 pins the `space-create -h` refusal plus the "Unknown space" error wording the same way the "Unknown intent" one was pinned.

Full smoke+unit tier green (2317 assertions), package --check and typecheck clean.

leandrodamascena (approved):
Thanks for fixing the bug @apackeer! APPROVED!

Issue #468, opened 2026-07-01
[Bug]: v2 learnings surface returns 0 candidates — runtime-graph memory_path omits the per-intent record dir (write side)


rahib-Ahmed:
### Description

Under the v2 per-intent **workspace layout**, the `memory_path` written into `runtime-graph.json` for each stage **omits the per-intent record-dir segment** (`aidlc/spaces/<space>/intents/<slug>-<id8>/…`). `aidlc-learnings.ts surface` reads that string off the graph and opens it as a file (`memAbs = join(projectDir, row.memory_path)`); because the path is missing the `<slug>-<id8>` segment it points at a nonexistent file, so `raw` becomes `""`, `parseMemoryEntries("")` returns `[]`, and **every `surface` call yields 0 candidates**. Net effect: the §13 learnings ritual is a silent no-op across the entire workflow — no stage-diary entry can ever be surfaced for promotion to a practice.

The live engine builds the same path **correctly** for the run-stage directive (via `memoryPathFor(..., recordPrefix)`), so the defect is confined to the two secondary producers that write/emit `memory_path` without threading the prefix.

### Steps to Reproduce

1. Run any intent under the workspace layout to a gated stage, with a non-empty diary at `aidlc/spaces/<space>/intents/<slug>-<id8>/<phase>/<stage>/memory.md`.
2. `bun core/tools/aidlc-runtime.ts compile`, then inspect that stage's row in `runtime-graph.json`: `memory_path` is missing the `intents/<slug>-<id8>/` segment, while `memory_entries` on the **same row** is > 0.
3. `bun core/tools/aidlc-learnings.ts surface --slug <current-stage>` → `candidates: []`, `memory_entries_total: 0`, even though the diary has entries.

### Expected Behavior

`surface` reads the real diary and lists its Interpretation/Deviation/Tradeoff entries as candidates; the written `memory_path` matches the path the engine emits in the run-stage directive: `aidlc/spaces/<space>/intents/<slug>-<id8>/<phase>/<stage>/memory.md`.

### Actual Behavior

`memory_path` is `aidlc/spaces/<space>/intents/<phase>/<stage>/memory.md` (no `<slug>-<id8>`), which does not exist, so `surface` reports 0 candidates for every stage and the learnings ritual never fires.

### AI-DLC Version

v2 (alpha)

### Release / Commit

v2 @ 2.1.4 (`b61e0ed`) — confirmed present at branch HEAD.

### AI-DLC Phase

[...]

PR #492, opened 2026-07-04
fix: record and read the per-intent memory path in learnings surface (2.2.2)

### LINKED ISSUE #455: [Bug]: AI-DLC v2 `learnings surface` mis-extracts `phase` under the workspace layout — `memory_path.split("/")[1]` returns `"spaces"` instead of the phase
# AI-DLC Version
v2 (the `v2` branch) — observed on 2.1.1; code confirmed still present on the latest **2.1.4** (`core/tools/aidlc-learnings.ts:197` on the `v2` branch). Harness-independent (not IDE-only); first hit running on **Kiro IDE 0.12.333**.

# AI-DLC Phase
Any gated stage that runs the learnings ritual (§13 of `stage-protocol.md`) — the `learnings surface` step before a gate.

---

## Description

`aidlc-learnings.ts surface` derives the workflow `phase` from the stage's `memory_path` by taking the **second path segment** (`memRel.split("/")[1]`). That index is only correct for the **retired flat `aidlc-docs/` layout**; under the current **workspace layout** the second segment is `"spaces"`, so every `surface` call reports `phase: "spaces"` instead of the real phase (`ideation` / `inception` / `construction` / `operation`).

This is **harness-independent core logic** — it reproduces on any harness whose runtime-graph uses the workspace layout (Claude Code, Codex, Kiro CLI, Kiro IDE). We first hit it running on Kiro IDE 0.12.333.

## Root cause

`core/tools/aidlc-learnings.ts` on the `v2` branch (line numbers are latest **2.1.4**; 2.1.3 is the same code at `:219`):

```ts
const memRel = row.memory_path;               // from runtime-graph.json
...
const phase = memRel.split("/")[1] ?? "";     // <-- assumes aidlc-docs/<phase>/... at index 1
...
const out: SurfaceOutput = { schema_version: 1, stage_slug: slug, phase, ... };
console.log(JSON.stringify(out));             // phase ships to the conductor
```

`memory_path` is produced by `memoryPathFor()` in `core/tools/aidlc-orchestrate.ts`, which builds it from `relativeSpaceRecordPrefix()`. Under the two layouts:

- **flat-legacy** (retired): `aidlc-docs/inception/user-stories/memory.md`
  → `split("/") = ["aidlc-docs", "inception", "user-stories", "memory.md"]` → `[1] = "inception"` ✅
- **workspace** (current): `aidlc/spaces/default/intents/<id>/inception/user-stories/memory.md`
  → `[1] = "spaces"` ❌  (the phase is actually the **3rd-from-last** segment)

The phase is the third-from-last segment (`<phase>/<stage>/memory.md`) in **both** layouts, so a layout-agnostic fix is `memRel.split("/").at(-3)`.

`tests/integration/t99-learnings-gate-flow.test.ts` already seeds a **workspace-layout** `memory_path`:

```ts
const RP = `aidlc/spaces/${DEFAULT_SPACE}/intents/${DEFAULT_RECORD_DIR}`;
...
memory_path: `${RP}/inception/user-stories/memory.md`,
```

…but it never asserts the `phase` field of the `surface` output, so the wrong value (`"spaces"`) passes silently. `tests/unit/t112-learnings-distribution-guard.test.ts` uses the **flat** `aidlc-docs/...` path, which masks the bug entirely.

## Steps to Reproduce

1. Run any AI-DLC v2 workflow on a **workspace-layout** install (the default today — `aidlc/spaces/<space>/intents/<id>/...`).
2. Reach a gated stage and let the learnings ritual run, or call it directly:
   `bun .kiro/tools/aidlc-learnings.ts surface --slug <stage-slug> --project-dir <dir> --json`
3. Inspect the JSON output's `phase` field.

## Expected Behavior

`phase` reflects the real workflow phase, e.g. `"inception"` for an inception-stage slug.

## Actual Behavior

`phase` is `"spaces"` for every stage under the workspace layout (the literal 2nd path segment of `aidlc/spaces/...`).

## Suggested Fix

Extract the phase by position relative to the end of the path (`<phase>/<stage>/memory.md`), which holds for both layouts:

```ts
const segs = memRel.split("/");
const phase = segs.at(-3) ?? "";
```

and add a `phase` assertion to the existing workspace-layout fixture in `t99-learnings-gate-flow.test.ts` so the regression is covered.

## Additional Context

- Found while running on Kiro IDE 0.12.333; confirmed by reading the shipped core on 2.1.1 and again on the latest 2.1.4 source.
- The same third-from-last segment is the phase under both the retired flat layout and the current workspace layout, so the fix is layout-agnostic and changes no behavior for flat-layout callers.

apackeer:
Fixes #455 and #468: `learnings surface` was dead under the per-intent workspace layout - it surfaced zero candidates and reported the wrong phase. Two related defects, one on each side of the runtime graph.

## Write side (#468)

The runtime-graph row's `memory_path` was recorded without the active intent's record dir. Both `aidlc-runtime.ts compile` and the `aidlc-state.ts advance` transition called `relativeMemoryPath` with two args, so the `recordPrefix` defaulted to the bare space prefix (`aidlc/spaces/<sp>/intents`) and dropped the `intents/<slug>-<id8>` segment. `surface` then `join()`'d that path against the project dir and opened a nonexistent diary. Both call sites now thread `relativeRecordDir(projectDir)`, the same pattern the orchestrator already uses; a pre-birth shell with no active intent still falls back to the bare space root.

## Read side (#455)

`surface` extracted the phase as `memory_path.split("/")[1]`, which under the workspace prefix is `"spaces"`, never the real phase. `memory_path` always ends `<prefix>/<phase>/<stageSlug>/memory.md` regardless of prefix shape (per-intent record dir, bare space prefix, or the legacy flat `aidlc-docs` root), so the extraction now takes the third-from-last segment. The legacy flat layout keeps working, pinned by a test case.

## Tests

- New `tests/unit/t199-learnings-memory-path.test.ts` pins both sides through the shipped tools' CLI surface: the compile row carries the per-intent record dir, `surface` finds the diary and reports the correct phase under the workspace layout, and the legacy flat path still degrades correctly.
- `t100` pinned the old bare-space advance shape and is updated to expect the per-intent record dir.
- `t99` (integration, learnings gate flow) additionally asserts the surfaced `phase` is the real phase end to end under the workspace layout.
- Verified green: smoke+unit tier (130 files, t84 flake re-run green alone), integration `t99-learnings-gate-flow`, `bun run typecheck`, `bun scripts/package.ts --check`.

Version 2.2.2 with matching CHANGELOG entry and README badge. If a sibling PR takes 2.2.2 first, this re-bumps on rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.2.2] - 2026-07-05
+
+Fixes the learnings gate returning zero candidates (or the wrong phase) once a project moved to the per-intent workspace layout. The runtime-graph row's `memory_path` was recorded without the active intent's record dir, and the learnings surface read the phase from the wrong path segment, so `learnings surface` looked for the diary in the wrong place and mislabeled its phase. Both the write side (runtime-graph compile and the state advance transition) and the read side (surface's phase extraction) now use the per-intent record path. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `learnings surface` now finds a stage's captured learnings under the workspace layout instead of surfacing zero candidates: the runtime-graph `memory_path` it reads now includes the active intent's record dir.
+* The `phase` field in the `learnings surface` JSON is now correct under the workspace layout (it previously reported `spaces` for any per-intent project); the legacy flat layout still reports the right phase.
+
 ## [2.2.1] - 2026-07-04
 
 Stops help requests from accidentally creating intents. `/aidlc intent help` was parsed as "switch to an intent named help"; the failed switch died with an error inviting the agent to "describe what to build to start a new one", which a conductor mid-task read as an instruction and birthed an unwanted intent. Bare `/aidlc help` and `/aidlc -h` were worse: they fell into the freeform-description funnel and offered to birth an intent literally named "help" (or silently advanced the active stage). All help-shaped inputs now route to the help text. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.2.1-blue)
+![version](https://img.shields.io/badge/version-2.2.2-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-learnings.ts
+++ b/core/tools/aidlc-learnings.ts
@@ -194,7 +194,13 @@ function handleSurface(args: string[], projectDir: string): void {
   const raw = existsSync(memAbs) ? readFileSync(memAbs, "utf-8") : "";
   const entries = parseMemoryEntries(raw);
 
-  const phase = memRel.split("/")[1] ?? "";
+  // memory_path always ends `<prefix>/<phase>/<stageSlug>/memory.md` (see
+  // relativeMemoryPath), so the phase is the third-from-last segment regardless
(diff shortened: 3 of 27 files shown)


leandrodamascena (approved):
Traced this across the prefix shapes and it holds up — `.at(-3)` lands on the phase for the per-intent, bare-space, and legacy-flat paths alike, since they all end in `<phase>/<slug>/memory.md`. Both write sites (compile + advance) thread the record dir now, and the null/pre-birth fallback is already covered by t96. Nice fix.

One tiny thing, not blocking: the `compile writes runtime-graph.json under the per-intent record dir` test (t199:213) only checks the file exists — which t199:124 already reads at the same path — so it doesn't really pin anything the fix changed. Could drop it. Approving as-is.

PR #493, opened 2026-07-04
fix: spawn adapter children via process.execPath, not bare bun (2.2.3)

### LINKED ISSUE #454: Kiro adapter spawns bare `bun` — ENOENT when the hook PATH lacks ~/.bun/bin
## Summary

`aidlc-kiro-adapter.ts` re-spawns `bun` **by bare name** when dispatching the
inner hook scripts. The spawned child inherits the host hook's `$PATH`, which on
many setups does **not** include `~/.bun/bin` (the default bun install dir). The
result is a hard failure:

```
error: Executable not found in $PATH: "bun"
  path: "bun", errno: -2, code: "ENOENT"
  at runCore (.../aidlc-kiro-adapter.ts)
```

## Where

Both per-harness copies, same pattern:

- `harness/kiro/hooks/aidlc-kiro-adapter.ts:156`
- `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:168`

```ts
const r = Bun.spawnSync(["bun", join(HOOKS_DIR, hookFile)], { ... });
//                        ^^^^^ bare name — not resolved
```

(The generated `dist/kiro*/.kiro/hooks/aidlc-kiro-adapter.ts` copies carry it too;
they regenerate from the harness source via `bun scripts/package.ts`.)

## Repro (observed on Kiro IDE)

1. Install/open a project with the AIDLC `.kiro/` tree where the Kiro **server
   process** was started before `bun` was on `PATH` (or where the hook PATH is
   otherwise minimal).
2. Submit a prompt → `aidlc-session-start` / `aidlc-session-end` / `aidlc-stop`
   hooks fail with the ENOENT above. The workflow still limps along, but every
   lifecycle hook errors and no heartbeats are written to
   `aidlc-docs/.aidlc-hooks-health/`.

The Kiro **server snapshots `$PATH` at launch**, so a later `bun` install +
shell-rc fix does **not** reach already-running hook children — only a full
server restart re-snapshots. That environment dependency is exactly what makes
the bare-`bun` spawn fragile.

## Why it matters

The adapter is the dispatcher for every Kiro lifecycle hook, so a missing
`~/.bun/bin` on the hook PATH disables audit logging, session lifecycle, the stop
guard, and statusline sync. It is environment-sensitive: it "works" only when the
spawning shell happens to have bun on PATH, and silently breaks otherwise.

## Fix

Resolve the bun binary instead of spawning the bare name. The adapter is itself
running under bun, so the running interpreter is the obvious source:

```ts
// prefer the bun already running this adapter; fall back to PATH lookup, then
// the conventional install dir
const bunBin =
  process.execPath /* the bun running us */ ||
  Bun.which("bun") ||
  `${process.env.HOME}/.bun/bin/bun`;
const r = Bun.spawnSync([bunBin, join(HOOKS_DIR, hookFile)], { ... });
```

Apply to **both** `harness/kiro/hooks/aidlc-kiro-adapter.ts` and
`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts`, then regenerate dist
(`bun scripts/package.ts`). `process.execPath` is the most robust (no PATH or
home-dir assumption).

## Notes

- Found while spiking plugin distribution across harnesses; Claude/Codex didn't
  trip it because their probe shell already had bun on PATH, but the same
  bare-spawn pattern is latent on any host with a minimal hook environment.
- Discovered on the `v2` line / extensions work.

apackeer:
Fixes #454: the Kiro hook adapters re-spawn their inner children (the core lifecycle hooks via `runCore`, plus the off-band utility commands on Kiro CLI) by the bare name `bun`. The child inherits the hook environment's `$PATH`, which on GUI-launched apps and minimal server environments often lacks the bun install dir (`~/.bun/bin`), so the spawn fails with `Executable not found in $PATH: "bun"` and the whole hook layer dies: no audit logging, no session lifecycle, no stop guard, no statusline sync.

## Fix

Respawn via `process.execPath` - the absolute path of the bun binary already executing the adapter, which exists by construction and needs no `PATH` lookup. Three authored sites, one-line change each:

- `harness/kiro/hooks/aidlc-kiro-adapter.ts` - `runCore` plus the off-band utility respawn
- `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts` - `runCore`
- `harness/codex/hooks/aidlc-codex-adapter.ts` - `runCore` (identical inner-respawn pattern, same failure mode, fixed in the same pass)

Out of scope by design: the outer `.kiro.hook` / agent-JSON `"command": "bun ..."` registration strings (the adapter cannot fix the PATH of the process that launches it - and in the reported failure the outer bun did resolve; it was the inner respawn that died), and the core tools' bun spawns (they run under an environment where the outer bun already resolved and inherit it).

## Tests

- `t147` (Kiro) and `t149` (Codex) gain a functional case: the adapter dispatches `session-start` successfully with `PATH` stripped of every dir containing a `bun` binary (premise-guarded), which ENOENTs under the old bare-name respawn.
- Both suites also pin the shipped dist adapter sources: no `spawnSync(["bun"...` argv[0], `process.execPath` present, so a stale regeneration or hand-edit reds.
- Verified green: smoke+unit tier (t84 flake re-run green alone), `bun run typecheck`, `bun scripts/package.ts --check`.

Version 2.2.3 with matching CHANGELOG entry and README badge. Sibling PRs are in flight with adjacent numbers; whichever merges later re-bumps on rebase.


harness/kiro/hooks/aidlc-kiro-adapter.ts
```diff
@@ -376,7 +378,9 @@ function buildForward(): Forward {
 }
 
 function runCore(hookFile: string, input: Record<string, unknown>): { stdout: string; code: number } {
-  const r = Bun.spawnSync(["bun", join(HOOKS_DIR, hookFile)], {
+  // Reuse the exact bun binary running this adapter; the child must not depend on
+  // PATH containing bun (the hook environment often lacks the bun install dir).
+  const r = Bun.spawnSync([process.execPath, join(HOOKS_DIR, hookFile)], {
```
leandrodamascena, line 383: This fixes the adapter's own child, but the child (a core hook) inherits this env, including the stripped PATH — `runCore` passes no `env`. So when `stop.ts`/`sync-statusline.ts`/`runtime-compile.ts` then spawn bare `"bun"`, they ENOENT in the exact case this PR targets. If you want the stop guard and statusline to actually come back (as the CHANGELOG claims), you'd need to pass an `env` with bun's dir on PATH here, or thread `process.execPath` down. Fine as a follow-up — just don't want the CHANGELOG saying those layers are restored when they aren't.

leandrodamascena (changes requested):
The `process.execPath` swap is the right call. I checked and it's equivalent to `["bun", file]` but doesn't need bun on PATH, so it fixes the #454 adapter death. Didn't find anything that worked before and breaks now.

But the CHANGELOG says this brings back "no stop guard, no statusline sync" and I don't think it does. `runCore` doesn't pass its own `env`, so the core hooks it spawns inherit the same stripped PATH — and some of them still spawn bare `"bun"`: `stop.ts:778`, `sync-statusline.ts:56`, `runtime-compile.ts:122`, `sensor-fire.ts:207`. In the no-bun-on-PATH case, those still ENOENT. So session-start and audit-logger come back, but the stop guard and statusline sync don't.

Not a regression — those were already broken and just unreachable before, since the adapter died first. I'd just fix the CHANGELOG to say what actually works, and maybe a follow-up to pass env/execPath down to those spawns too.

apackeer:
Thanks - you're right on both counts. Verified against the hooks: `runCore` children inherit the same stripped `PATH`, and four core hooks spawn a further bare-name `bun` child of their own (`aidlc-stop.ts:778`, `aidlc-sync-statusline.ts:56`, `aidlc-runtime-compile.ts:122`, `aidlc-sensor-fire.ts:207`), so on a `bun`-less `PATH` those still ENOENT one level deeper. The stop guard fails open and the statusline doesn't update; only the spawn-free hooks (audit logging, session lifecycle, state validation, subagent tracking) come back with this fix.

Pushed 861ae67: the CHANGELOG now says exactly that - what's restored, plus a known-limitation bullet naming the four still-degraded hooks - instead of the "no stop guard, no statusline sync" overclaim. Kept it doc-only per your suggestion; threading `process.execPath` down through those four spawns is the follow-up (they're core hooks shared with the Claude harness, where `bun` on `PATH` is a documented prerequisite, so the fix wants its own look rather than a rider here).

leandrodamascena (approved):
Thanks @apackeer! Approved

PR #494, opened 2026-07-04
fix: reviewer Review sections source the date from the shell and name the reviewer persona (2.2.4)

### LINKED ISSUE #469: [Bug]: AI-DLC v2 reviewer `## Review` template lets the reviewer guess the `**Date:**` — the `[ISO timestamp]` placeholder carries no instruction to obtain a real timestamp
# AI-DLC Version
v2 (the `v2` branch) — observed on 2.1.1; template confirmed unchanged on the latest **2.1.4** (`core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md:60` and `core/knowledge/aidlc-product-lead-agent/reviewing.md:45` on the `v2` branch). Harness-independent — this is a documentation/prompt surface in shared core, so it applies to every harness's `dist/` build.

# AI-DLC Phase
Any gated stage that runs a reviewer (the review step in `stage-protocol.md`) — the reviewer emits a `## Review` block whose `**Date:**` field is a free-text placeholder.

---

## Description

The reviewer review-comment template instructs the reviewer to emit a `## Review` block that includes a `**Date:**` field, but the field is only a bare placeholder:

```markdown
**Date:** [ISO timestamp]
```

Nothing in the template tells the reviewer *how* to obtain that timestamp, and the engine never fills or validates it (see "Why the engine doesn't catch it" below). The `**Date:**` value is therefore whatever prose the model writes — i.e. a **guessed / hallucinated** date. A model has no reliable clock; the value it writes is unrelated to the actual review time and drifts from the timestamps that the CLI tools (`aidlc-state.ts`, `aidlc-audit.ts`) generate for the same stage.

This is a **prompt/documentation surface** in shared core (`core/knowledge/…`), so it projects into every harness's `dist/` (`claude`, `codex`, `kiro`, `kiro-ide`).

## Root cause

`core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md` and `core/knowledge/aidlc-product-lead-agent/reviewing.md` on the `v2` branch. Both carry the identical `## Review` template whose date line is:

```markdown
## Review

**Verdict:** READY | NOT-READY
**Reviewer:** <reviewer-agent>
**Date:** [ISO timestamp]          <-- placeholder only; no instruction to obtain it
**Iteration:** [1, 2, etc.]
```

`[ISO timestamp]` describes the *format* but gives no *procedure*, so the model fills it from its own (absent) sense of the current time.

Note this is **not** a permissions problem: the reviewer agent definitions already allow the command that would produce a correct value —

```jsonc
// harness/{kiro,kiro-ide}/agents/aidlc-{architecture-reviewer,product-lead}-agent.json
// and the mirrored dist/ copies — all four already carry:
"allowedCommands": ["bun \\.kiro/tools/.*", "date -u .*"]
```

`date -u` is already on the allowlist; the template simply never tells the reviewer to run it. The rest of the workflow already follows exactly this convention — e.g. `core/aidlc-common/protocols/stage-protocol.md` instructs authors to obtain timestamps with `date -u +"%Y-%m-%dT%H:%M:%SZ"`, and `core/aidlc-common/stages/initialization/state-init.md` does the same for the workspace start date. The reviewer template is the one such field that was left as a bare placeholder.

## Why the engine doesn't catch it

The `## Review` block is authored by the reviewer and written into the artifact as prose; nothing on the engine side synthesizes or validates the `Date:` value. Unlike `aidlc-state.ts` / `aidlc-audit.ts`, which auto-generate a fresh ISO timestamp per call (`stage-protocol.md`), the review `Date:` field has **no** auto-timestamp safety net — it is purely model-authored, so a missing instruction becomes a guessed value with nothing downstream to correct it.

## Steps to Reproduce

1. Run any AI-DLC v2 workflow to a gated stage that invokes a reviewer (e.g. an Inception or Construction review).
2. Let the reviewer emit its `## Review` block into the primary artifact.
3. Inspect the recorded `**Date:**` — it is a value the model wrote, not one obtained from the system clock, and it will not match the `date -u` timestamps the CLI tools recorded for the same stage.

## Expected Behavior

The `**Date:**` field reflects the real review time, obtained from the system clock (consistent with how the rest of the workflow sources timestamps).

## Actual Behavior

The reviewer writes a guessed date into `**Date:**`; there is no instruction to obtain a real one and no engine-side timestamp to correct it.

## Suggested Fix

Make the placeholder carry the same instruction the rest of the workflow already uses — a one-line, additive documentation change, no behavior change for the format:

```markdown
**Date:** [ISO timestamp — obtain by running `date -u +%Y-%m-%dT%H:%M:%SZ`; do NOT guess the date]
```

in both `core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md` and `core/knowledge/aidlc-product-lead-agent/reviewing.md` (which then projects into the four `dist/` copies via the normal core → dist build). No `allowedCommands` change is required — `date -u` is already permitted in all reviewer agent files.

### Scope / limitations of this fix

- **Instruction only.** This is a prompt/documentation fix. It makes a correct value *available and requested*; it does not force it — the engine still records whatever the reviewer returns.
- **Actual execution is harness-dependent and not asserted here.** Whether a given harness's reviewer sub-agent actually runs the `date -u` Bash (vs. inlining a value) can vary by harness; the fix removes the *missing-instruction* cause without claiming to guarantee execution on every harness. A stronger, engine-side fix would be to have the review-recording path stamp the `Date:` deterministically (as the CLI tools already do for their own timestamps) rather than rely on the reviewer — noted as the alternative not taken.

## Additional Context

- First hit while running on Kiro IDE 0.12.333, but the cause is a shared-core template with no engine-side timestamp, so it is harness-independent. Confirmed by reading the shipped core on 2.1.1 and again on the latest 2.1.4 source (`core/knowledge/{aidlc-architecture-reviewer-agent,aidlc-product-lead-agent}/reviewing.md`, `**Date:**` line unchanged).
- Exactly two core files carry a `**Date:**` review field, and both have the bare placeholder; the `date -u` allowlist entry is already present in all four reviewer agent `.json` files on the `v2` branch, so the gap is purely the missing instruction.

### LINKED ISSUE #496: [Bug]: AI-DLC v2 reviewing.md `## Review` template hardcodes the producer agent id in `**Reviewer:**`, not the reviewer's own
## Description

The `## Review` template that each reviewer is told to emit hardcodes a `**Reviewer:**` field, but the value names the **producer** agent, not the **reviewer** agent that actually owns the file and does the review:

- `core/knowledge/aidlc-product-lead-agent/reviewing.md:44` → `**Reviewer:** aidlc-product-agent`
  — but the reviewer is **`aidlc-product-lead-agent`**; `aidlc-product-agent` is the *producer* (the stage `lead_agent`).
- `core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md:59` → `**Reviewer:** aidlc-architect-agent`
  — but the reviewer is **`aidlc-architecture-reviewer-agent`**; `aidlc-architect-agent` is the *producer/support* agent.

Both files name the wrong side of the producer/reviewer pair — each points at the corresponding **builder** agent instead of at itself. Since the reviewer copies this template verbatim into the artifact's `## Review` section, the artifact records the wrong agent as its reviewer.

## Why this is a bug, not intentional

Each `reviewing.md` opens by explicitly separating the reviewer from the producer:

> "When invoked as a reviewer, your role changes. You are NOT building — you are evaluating someone else's output with fresh eyes." … "You did not produce this work. Judge the output, not the effort." (`## Stance`)

So the file's own stance says the reviewer is a distinct role from the builder — yet its `## Review` template stamps the **builder's** id into the `**Reviewer:**` line. The two agents are genuinely distinct and both real:

- `aidlc-product-agent` — stage `lead_agent` (producer) — vs. `aidlc-product-lead-agent` — stage `reviewer`.
- `aidlc-architect-agent` — producer/support — vs. `aidlc-architecture-reviewer-agent` — stage `reviewer`.

(Confirmed from `core/tools/data/stage-graph.json`: the `reviewer` field on the 11 reviewed stages is `aidlc-product-lead-agent` / `aidlc-architecture-reviewer-agent`; the wrong ids in the template are the `lead_agent` / `support_agents` of those stages.)

The `**Reviewer:**` value is also not a bracketed placeholder like the sibling fields (`**Date:** [ISO timestamp]`, `**Iteration:** [1, 2, etc.]`) — it is a concrete literal, i.e. the template says "write exactly this", and exactly-this is wrong.

## Impact

Low-severity but real: the `## Review` block that lands in the artifact attributes the review to the producer agent, so the on-disk provenance of the quality gate is wrong (a review done by `aidlc-product-lead-agent` is recorded as authored by `aidlc-product-agent`). Anyone auditing "who reviewed this artifact" from the file gets the builder's name. It also undermines any tooling that parses `**Reviewer:**` to attribute the review (e.g. a harness adapter that extracts the reviewer id from the review prose to log `SUBAGENT_COMPLETED` / an audit `Agent Type`): the field, if trusted, yields the wrong reviewer.

## Steps to Reproduce

1. Run any reviewed stage (e.g. `requirements-analysis`, reviewer `aidlc-product-lead-agent`).
2. Let the reviewer emit its `## Review` block using the template in `reviewing.md`.
3. Read the `**Reviewer:**` line in the artifact's `## Review` section — it reads `aidlc-product-agent`, not the reviewer that actually ran.

## Expected Behavior

The `**Reviewer:**` field names the agent that performed the review — the stage's `reviewer` agent.

## Actual Behavior

It names the corresponding producer agent (`aidlc-product-agent` / `aidlc-architect-agent`).

## Suggested Fix

Correct the literal in both templates to the reviewer's own id:

- `core/knowledge/aidlc-product-lead-agent/reviewing.md:44`
  `**Reviewer:** aidlc-product-agent` → `**Reviewer:** aidlc-product-lead-agent`
- `core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md:59`
  `**Reviewer:** aidlc-architect-agent` → `**Reviewer:** aidlc-architecture-reviewer-agent`

(These project into the four `dist/` copies via the normal core → dist build.) Optionally, to make it self-evidently correct and future-proof against another rename, replace the literal with a placeholder — `**Reviewer:** <this agent's id>` — matching the bracketed style already used for `**Date:**` and `**Iteration:**` in the same block.

## Additional Context

- Same `## Review` template, same two files, as #469 (the `**Date:** [ISO timestamp]` placeholder that carries no "obtain a real timestamp" instruction). #469 quotes the template but treats the `**Reviewer:**` line as `<reviewer-agent>` (correct-by-intent) and does not flag that the shipped literal is the wrong id. This issue is the narrow, separate follow-up: the value itself is wrong. A maintainer fixing #469 touches the same block and can correct both fields in one edit.
- Confirmed by reading the shipped core on the `v2` branch (`core/knowledge/{aidlc-product-lead-agent,aidlc-architecture-reviewer-agent}/reviewing.md`).
- Found while running on Kiro IDE 1.0.89, but the cause is a shared-core template, so it is harness-independent.

apackeer:
Fixes #469: the reviewer `## Review` template's `**Date:** [ISO timestamp]` placeholder carried no instruction to obtain a real timestamp, so the review-only agents (product-lead and architecture-reviewer) filled the `Date` field from memory - a guessed or stale date on the permanent review record.

## Fix

Both `reviewing.md` knowledge files (the only two copies of the template, verified by grep) now:

- change the placeholder to `[ISO timestamp from Bash]`, matching the convention the stage protocol already uses elsewhere, and
- add one instruction line after the template: obtain the timestamp by running `date -u +"%Y-%m-%dT%H:%M:%SZ"` in the shell and paste the actual output, never guess or infer the date.

Both reviewer agents have shell on every harness (Claude inherited, Kiro CLI agent JSON, Kiro IDE tools frontmatter grant, Codex default), so the instruction is executable everywhere with no harness-conditional prose.

## Tests

A new unit test (t200) pins the sourcing instruction in both shipped templates: exactly one Date line, the exact date command present, and an explicit prohibition on guessing, so a future template edit cannot silently regress it. Verified green: smoke+unit tier (129 files), `bun run typecheck`, `bun scripts/package.ts --check` (knowledge files regenerate into all four dist trees).

Version 2.2.4 with matching CHANGELOG entry and README badge. Sibling PRs are in flight with adjacent numbers; whichever merges later re-bumps on rebase.

## Also: Reviewer attribution fix

Fixes #496.

While fixing the Date field, a second template defect in the same two files: the `**Reviewer:**` line named the PRODUCER agent (`aidlc-product-agent` / `aidlc-architect-agent`) instead of the reviewer persona, so every `## Review` section on disk was attributed to the agent whose artifact was under review - misrecording exactly the maker-checker separation the dedicated reviewer personas (added in 3c4c1d9, "No self-review") exist to provide. Both templates now stamp the reviewer persona's own name, and t200 pins the attribution alongside the Date instruction. (#496 was filed independently while this fix was in flight and describes the same defect with the same producer-vs-reviewer analysis.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [2.2.4] - 2026-07-05
 
-The two review-only agents now record a real UTC timestamp on the `## Review` section they append, instead of a guessed date. The Review template's `Date` field was a bare `[ISO timestamp]` placeholder with no sourcing instruction, so reviewers filled it in from memory; it now instructs the reviewer to run `date -u +"%Y-%m-%dT%H:%M:%SZ"` in the shell and paste the actual output, matching the `[ISO timestamp from Bash]` convention already used elsewhere. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+The two review-only agents now record a real UTC timestamp and their own name on the `## Review` section they append. The Review template's `Date` field was a bare `[ISO timestamp]` placeholder with no sourcing instruction, so reviewers filled it in from memory; it now instructs the reviewer to run `date -u +"%Y-%m-%dT%H:%M:%SZ"` in the shell and paste the actual output, matching the `[ISO timestamp from Bash]` convention already used elsewhere. The template's `Reviewer` field also named the PRODUCER agent (the artifact's author) instead of the reviewer persona, so every review on disk was attributed to the agent whose work was being reviewed; it now names the reviewer persona. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
 
 * Reviewer `## Review` sections (product-lead and architecture-reviewer agents) now carry a real shell-sourced UTC timestamp in the `Date` field rather than an inferred date.
+* The `Reviewer` field in those sections now records the reviewer persona (`aidlc-product-lead-agent` / `aidlc-architecture-reviewer-agent`) instead of the producer agent whose artifact was under review.
 
 ## [2.2.3] - 2026-07-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.2.4] - 2026-07-05
+
+The two review-only agents now record a real UTC timestamp on the `## Review` section they append, instead of a guessed date. The Review template's `Date` field was a bare `[ISO timestamp]` placeholder with no sourcing instruction, so reviewers filled it in from memory; it now instructs the reviewer to run `date -u +"%Y-%m-%dT%H:%M:%SZ"` in the shell and paste the actual output, matching the `[ISO timestamp from Bash]` convention already used elsewhere. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* Reviewer `## Review` sections (product-lead and architecture-reviewer agents) now carry a real shell-sourced UTC timestamp in the `Date` field rather than an inferred date.
+
 ## [2.2.3] - 2026-07-05
 
 Fixes a fragile process respawn in the Kiro CLI, Kiro IDE, and Codex hook adapters. Each adapter dispatches the framework's core lifecycle hooks (and, on Kiro, the off-band utility commands) by spawning a child process named `bun` by bare name. That child inherits the hook environment's `$PATH`, which on GUI-launched apps and minimal server environments often omits the bun install dir (`~/.bun/bin`), so the spawn fails with `Executable not found in $PATH: bun` and the whole hook layer dies. The adapters now respawn via the absolute path of the bun binary already running them (`process.execPath`), so the hooks themselves run again without `bun` on `PATH`: audit logging, session lifecycle, state validation, and subagent tracking are fully restored. Known limitation: four core hooks spawn a further `bun` child of their own (the stop guard's engine consult, the statusline state sync, the runtime-graph compile, and the sensor dispatch), and those children still resolve `bun` off `PATH`, so on a `bun`-less `PATH` they stay degraded (the stop guard fails open, the statusline does not update, the runtime graph is not recompiled, sensors do not fire); a follow-up threads the running binary down to those spawns. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.2.3-blue)
+![version](https://img.shields.io/badge/version-2.2.4-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md
+++ b/core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md
@@ -56,7 +56,7 @@ Append a `## Review` section to the PRIMARY artifact file. Use this exact format
 ## Review
 
 **Verdict:** READY | NOT-READY
-**Reviewer:** aidlc-architect-agent
+**Reviewer:** aidlc-architecture-reviewer-agent
 **Date:** [ISO timestamp from Bash]
 **Iteration:** [1, 2, etc.]
 
--- a/core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md
+++ b/core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md
@@ -57,7 +57,7 @@ Append a `## Review` section to the PRIMARY artifact file. Use this exact format
 
 **Verdict:** READY | NOT-READY
 **Reviewer:** aidlc-architect-agent
-**Date:** [ISO timestamp]
+**Date:** [ISO timestamp from Bash]
 **Iteration:** [1, 2, etc.]
 
 ### Findings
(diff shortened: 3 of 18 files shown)


leandrodamascena (approved):
Both edits look right to me — Date now points at `date -u` with a no-guess line, and the Reviewer line names the reviewer persona instead of the producer, in both files. Works on all four harnesses. No issue with the fix itself.

Small thing on the test, not blocking: `t200:36` checks for "guess", but that word already shows up at line 10 of both templates ("implement without guessing"), so the assert would still pass even if you deleted the "Never guess or infer the date" line. The `date -u` string check on line 35 is the one actually holding it. If you want, tighten 36 to `toContain("Never guess or infer the date")`.

PR #497, opened 2026-07-05
fix: advance the autonomous swarm through every Bolt batch (2.2.5)

### LINKED ISSUE #439: [Bug]: autonomous Construction swarm never advances past the first Bolt batch (tryEmitSwarm hardcodes batches[0])

apackeer:
Fixes #439: under an autonomous Construction grant, the swarm never advanced past the first Bolt batch - `tryEmitSwarm` emitted `batches[0]` unconditionally, so after batch 1 converged and merged, `next` re-emitted batch 1 forever (a deadlock the engine comments previously documented as tolerated).

## Fix

The engine now walks the compiled Bolt DAG's topological batches in order and emits the first batch still carrying units the swarm referee has not verified converged, filtered to just those units (a partially failed batch re-fans only its baton-returned units, never its converged siblings). Once every unit of every batch has converged, `next` emits the stage's settle directive (run-stage, gate true, last unit - the same shape the per-unit engine presents on its all-covered re-entry) so the human approves once and the stage completes.

## The completion signal (differs from the issue's suggested fix)

The issue suggested reusing the per-unit artifact-coverage signal. Verified against the code, that cannot work for the swarm: `aidlc-bolt complete --merge` merges only metadata (state, audit delta, runtime-graph fragment) and the swarm flow never invokes the git worktree merge, so a converged unit's artifacts stay in its Bolt worktree and main-checkout artifact coverage never sees them. The reliable ledger is the SWARM_UNIT_CONVERGED audit event, which the referee's finalize writes into the main checkout's audit shard for each genuinely re-verified unit. Failed and baton-returned units get no converged row, so they re-fan; replays are idempotent.

## Conductor prose

The invoke-swarm row in all four harness SKILL.md files previously said finalize exit 0 means "report and continue the loop" - under the new contract that would let a conductor complete the stage after an intermediate batch. The row now says: exit 0 means re-run `next`; the engine emits the next unconverged batch or the settle gate, and only the settle directive gets the ritual plus `report --result approved`.

## Tests

New `tests/unit/t201-swarm-batch-advance.test.ts` (4 cases through the real engine spawn): batch 1 incomplete emits batch 1; batch 1 converged emits batch 2 only; all converged emits the settle run-stage (no invoke-swarm); a partial batch re-fans only its unconverged units. Fixtures seed audit-format-faithful SWARM_UNIT_CONVERGED blocks. Adversarially reviewed (engine logic, fallback path, guard preservation all confirmed sound; the SKILL prose fix above was the review's one required change). Deterministic tiers green: smoke+unit 130 files, e2e swarm referee, integration swarm slice, t181 SKILL-parity.

One known follow-up surfaced by the review, out of scope here: a unit that converges but then fails its metadata merge keeps its converged row (finalize exits 2 to halt-and-ask, so it is not silent), and the swarm flow never lands per-unit artifacts in the main checkout at all - the merge-back story for swarm artifacts deserves its own issue.

Version 2.2.5; sibling PRs hold adjacent numbers and whichever merges later re-bumps on rebase.

## Independent verification note

A separate verification pass confirmed t201's cases 2-4 fail against the pre-fix code (genuine discriminators, not tautologies) and all deterministic tiers green. One coverage observation for reviewers: every existing e2e/integration swarm fixture is single-batch (t135 seeds `[["a","b"]]`; t134 plays the conductor itself), so multi-batch advancement is proven at the unit tier only. A multi-batch integration fixture would be a worthwhile follow-up alongside the swarm artifact merge-back issue noted above.


core/tools/aidlc-orchestrate.ts
```diff
@@ -679,6 +682,30 @@ function readBoltDagBatches(projectDir: string): string[][] | null {
...
+// metadata (state + audit + runtime-graph fragment) back to the main checkout;
+// the unit's produced artifacts (code-generation-plan.md / code-summary.md and
+// the generated source) are NOT copied into the main record tree by the swarm
+// finalize flow. So unitCovered's disk check (the INLINE per-unit ledger) never
+// sees a swarm unit as covered, and the batch-advance signal must instead be the
+// `SWARM_UNIT_CONVERGED` audit rows `aidlc-swarm.ts finalize` writes from the
+// main checkout, one per genuinely-converged unit, each carrying `Unit name`.
+// Composes the same shard-concat + block-parse the other audit readers use; an
+// absent/empty audit yields the empty set (no batch has converged yet).
+function swarmConvergedUnits(projectDir: string): Set<string> {
```
leandrodamascena: This collects every `SWARM_UNIT_CONVERGED` row in the intent, keyed only on `Unit name`. The row (emitUnitConverged, aidlc-swarm.ts:259) carries just `Batch number` + `Unit name` — no stage/Bolt id — and the audit is append-only. So if the same per-unit construction stage re-runs with the same unit names (recompose / re-entry), the prior run's converged rows make the fresh batch look already-built and it gets skipped. If a per-unit stage can re-run in the current flow, scope this set to the current stage run (a stage id on the event, or a timestamp floor). If it genuinely can't re-run, a comment saying so would settle it.

leandrodamascena (changes requested):
The batch-advance logic works — I checked that the DAG unit name matches the `Unit name` written into `SWARM_UNIT_CONVERGED`, so the ledger signal actually lines up and the #439 deadlock is fixed. Cross-intent bleed is fine too (`readAllAuditShards` is scoped to the active intent).

Two things worth raising:

1. `swarmConvergedUnits` collects every `SWARM_UNIT_CONVERGED` row in the intent, keyed only on `Unit name` — and the row (emitUnitConverged, aidlc-swarm.ts:259) carries just `Batch number` + `Unit name`, no stage or Bolt id, on an append-only audit. If a per-unit construction stage ever re-runs (recompose, re-entry) with the same unit names, the old converged rows would make the fresh batch look already-built and skip it. Is re-running a per-unit stage possible in the current flow? If yes, this needs a stage/timestamp scope on the converged set; if it genuinely can't re-run, a comment saying so would settle it. Left it inline.

2. Related: `aidlc-swarm.ts:605` emits `SWARM_UNIT_CONVERGED` off `status === "converged"` without consulting `mergeFailures`, so a unit that converged but then failed `complete --merge` still gets a converged row — this PR makes that row the load-bearing completion signal, so it now reads as built when the metadata never landed on main. You flag the merge-back story as an out-of-scope follow-up and finalize exits 2 so it's not silent, but not emitting converged for a unit in `mergeFailures` would keep the signal honest. (aidlc-swarm.ts isn't in this diff so I can't inline it.)

Not from this PR: t186 case 9b fails, but it's the pre-existing 2.1.6 human-presence gate (fails the same on HEAD~1), not related here.

apackeer:
Both findings were real - thanks for catching them before this row became load-bearing. Pushed 664d9fb with fixes for each:

**1. Stale converged rows on a stage re-run.** You asked whether a per-unit construction stage can re-run with the same unit names. It can: a backward/redo jump resets completed stages' checkboxes to pending (`aidlc-jump.ts:314-334`) without touching the append-only ledger, and the jump never writes `Construction Autonomy Mode` (only `aidlc-bolt set-autonomy` does), so the autonomy grant survives and `tryEmitSwarm` fires on re-entry against the prior run's rows. (`recompose` can't reach it - the pending-only + behind-cursor + Running guards block a completed/in-flight stage - but the jump path is enough to make this real, and a re-init appending a second `WORKFLOW_STARTED` to the same shards is a second path.)

So instead of a comment, `swarmConvergedUnits` now floors the read at the stage's latest main-workflow `STAGE_STARTED` - the same boundary idiom `hasStageAuditEvent` already uses: rows tagged `Workflow: single-stage:<slug>` never move the floor, and with no qualifying row the floor degrades to counting all rows, never to excluding all. Both re-entry paths emit a fresh `STAGE_STARTED` (`aidlc-state.ts:979` on advance, `aidlc-jump.ts:420` on jump), no `STAGE_STARTED` fires between batches of one run, and audit-merge only appends the post-fork delta, so same-run rows from earlier batches are never orphaned. Same-second ties are included (`timestamp < since`) - excluding them would false-re-fan a just-merged unit into `prepare`'s existing-worktree error, the more dangerous direction. t201 grew five cases pinning this (stale rows ignored, post-floor rows advance, single-stage row inert, other-slug row inert, no-floor fallback).

**2. Converged-but-merge-failed units.** Fixed as you suggested: `finalize` now skips `SWARM_UNIT_CONVERGED` for any unit in `mergeFailures` (no `SWARM_UNIT_FAILED` either - the unit did converge; the envelope + exit 2 carry the merge outcome). The ledger then says the unit is still owed, and since re-fanning via `prepare` would hard-error on the preserved worktree, the recovery is a `finalize` retry scoped to that unit - re-verify re-derives the worktree path and `release-merge` is idempotent, so it's a pure re-invocation that emits the row once the merge lands. That's now documented in the exit-2 arm of all four harness SKILL files, and t134 case 14 pins the behavior (envelope converged, `merge_failures` named, zero converged/failed rows).

The envelope/`SWARM_COMPLETED` tally still counts the unit as converged - deliberate: the tally records the referee's verdict (the unit re-verified green), while the per-unit row is now the merge-conditional advance signal. `audit-format.md`, the state-machine and skill-system references, and the 2.2.5 CHANGELOG entry are updated to match.

Full deterministic tiers green on the branch (smoke+unit 130 files, integration 99 files, t134 13/13, t201 9/9). And agreed on the t186 case 9b red - it fails identically on the base commit, pre-existing 2.1.6 human-presence gate, untouched here.

core/tools/aidlc-orchestrate.ts
apackeer: Re-run is possible (backward/redo jump; autonomy survives), so this got a fix rather than a comment: the read is now floored at the stage's latest main-workflow STAGE_STARTED - see 664d9fb and the summary comment for the full reasoning. t201 cases 5-9 pin the floor.

leandrodamascena (approved):
LGTM

PR #498, opened 2026-07-05
fix: gate options name the computed next stage instead of a guessed one (2.2.6)

### LINKED ISSUE #406: [Bug]: Incorrect Approval Options Information during Construction Phase
### Description

During Construction phase, when I complete the functional design step. 
The approval option tell us to proceed to code generation (but actually go to NFR design).

Similar issue happen after complete NFR Requirements and Design 

Image

Image

### Steps to Reproduce

Run through the workflow til constrution phase (Functional and NFR Design) on Kiro-IDE

### Expected Behavior

Should tell the next steps correctly. 

Approve — Continue to NFR Requirements/Design/Infrastructure Design (Bolt 1: Foundation)
Request Changes — Provide revision feedback
Other — describe what you want instead

### Actual Behavior

Always get these 3 options

Approve — Continue to Code Generation (Bolt 1: Foundation)
Request Changes — Provide revision feedback
Other — describe what you want instead

### AI-DLC Version

v2 (alpha)

### Release / Commit

f3ce1b8e976942dfb7701eed7bca046d32c2a978

### AI-DLC Phase

### Platform / IDE

Kiro IDE

### AI Model

Auto

### Environment

MacOS Tahoe 26.5
Kiro IDE 1.0.0

### Additional Context

_No response_

apackeer:
Fixes #406: during Construction, the approval options said "Continue to Code Generation" regardless of the actual next stage (e.g. after functional-design, where the real next stage is nfr-requirements).

## Root cause

"Continue to [next stage]" was an LLM-filled placeholder in the stage protocol and each harness's question-rendering annex; the engine's directives carried no next-stage field, and the true advance target was computed only after approval by `nextInScopeStage`. Two independent sources of truth, so the rendered label drifted from reality - and "Code Generation" was the most salient guess.

## Fix

Run-stage directives now carry `next_stage: string | null` - the slug of the next in-scope EXECUTE stage after the emitted one, computed by the same `nextInScopeStage` call (same function, same inputs) the advance path uses, honoring state EXECUTE/SKIP overrides and completed/skipped checkboxes. `null` means the final in-scope stage; the conductor renders "Complete workflow". The stage-protocol question specs and all four harness annexes now bind the placeholder to the directive field: render it verbatim, never infer. The field is display-only - the approve path still recomputes the advance target authoritatively, so a stale label can never mis-route the workflow (and a recompose re-runs `next`, refreshing the directive).

Schema: the field is added to the directive interfaces, the field allowlist (the validator rejects unknown keys), and a new nullable-string check; the directive self-test covers both shapes.

## Tests

New `tests/unit/t202-gate-next-stage.test.ts` (4 cases through the real engine spawn): the reported bug pinned (functional-design carries nfr-requirements, explicitly not code-generation), SKIP-override honored, final stage carries key-present null, and the advance path names the successor. t113 gains string/null/invalid-type validator cases. Adversarially reviewed: APPROVE (correctness of the emitted value verified identical to the advance derivation; blast radius on gate:false per-unit directives confirmed harmless). Smoke+unit tier 130 files green; directive self-check exit 0.

Version 2.2.6; sibling PRs hold adjacent numbers and whichever merges later re-bumps on rebase.


leandrodamascena (approved):
Checked the real question — does the label match where the workflow actually goes. It does: `next_stage` is computed by the same `nextInScopeStage(node.slug, scope, stateContent)` the post-approval advance runs, so the "Continue to X" the human sees is the stage that actually runs on approve. The t202 case for the reported bug pins it — after functional-design the label now says NFR Requirements, not the old hardcoded "Code Generation" guess.

And even in the worst case it's safe: the field is display-only, the approve path recomputes the target itself, so a stale label could never send the workflow to the wrong stage. No issues.

Approved & good to merge.

Issue #438, opened 2026-06-26
[Bug]: project type defaults to Greenfield in the bugfix workflow


myakkala-acstp:
### Description

When the aidlc workflow kicked off in bug fix scope, the project type defaults to Greenfield instead Brownfield.

### Steps to Reproduce

1. Kick off the aidlc-worflow and select the scope - Bugfix.

### Expected Behavior

In the aidlc-state.md file the project type must be default to Brownfield.

### Actual Behavior

In the aidlc-state.md file the project type default to Greenfield.

### AI-DLC Version

v2 (alpha)

### AI-DLC Phase

Not phase-specific

### Platform / IDE

Claude Code

### AI Model

Opus 4.7

warren830:
This looks fixed on `v2`, and the specific ask was deliberately declined — worth recording so the next reader does not re-open it.

The root cause was not scope-based defaulting. Project type comes from a workspace scan (`core/tools/aidlc-utility.ts`: `projectType: brownfield ? "Brownfield" : "Greenfield"`), and the real defect was that `detectWorkspace` only scanned the repo root plus a fixed source-dir allowlist — so a project whose source sat one folder down scanned Greenfield. PR #499 (v2.2.7) added the depth-1 nested fallback; PR #507 (v2.2.10) added `.gitmodules` as a sixth brownfield signal.

The literal request — "in bugfix scope, Project Type must default to Brownfield" — was rejected on purpose, and the reasoning is in the code: an empty workspace genuinely has nothing to reverse-engineer, so routing is not overridden. Instead the incremental scopes (`bugfix`, `refactor`, `security-patch`) emit an advisory pointing at the fix. `tests/unit/t203-nested-workspace-detection.test.ts:219` pins exactly this (`bugfix on an empty workspace stays Greenfield, RE SKIP, and emits the stderr advisory`), so implementing the literal ask would now turn that test red.

Suggest closing as fixed in v2.2.7.

One residual gap, if you want it as a separate enhancement: the advisory goes to **stderr only**, which Claude Code often does not surface — so the person most likely to hit the misread is the least likely to see the note. A visible line in `aidlc-state.md` for incremental-scope-on-Greenfield would be ~1 line plus one t203 assertion. Happy to file that separately rather than reopening this.

apackeer:
[...]

PR #499, opened 2026-07-05
fix: detect nested projects in workspace detection (2.2.7)

### LINKED ISSUE #462: [Bug]: AI-DLC v2 workspace detection only scans top-level + a fixed source-dir list, so a project nested in an arbitrarily-named folder (e.g. `wordbook/`) is misclassified Greenfield and Reverse Engineering is skipped
# AI-DLC Version
v2 (the `v2` branch) — observed on 2.1.1; code confirmed still present on the latest **2.1.4** (`core/tools/aidlc-utility.ts` `detectWorkspace`, line numbers below on the `v2` branch). Harness-independent core logic (not IDE-only); first hit running on **Kiro IDE 0.12.333**.

# AI-DLC Phase
Initialization — `workspace-detection` (sets `Project Type`, which gates whether Reverse Engineering runs).

---

## Description

The deterministic workspace scanner classifies a project as Greenfield whenever its source lives one level down in an **arbitrarily-named** container directory. Concretely: a workspace root that contains a folder like `wordbook/` (or `backend/`, `my-app/`, `service/`, …) full of source code is classified **Greenfield**, because the scanner never descends into a directory whose name is not in a hard-coded list. As a result `Project Type: Greenfield` is written and Reverse Engineering (stage 2.1) is skipped, so no codebase understanding is produced for a project that clearly has existing code.

All five brownfield signals are effectively evaluated against the **workspace root top-level only**; none of them sees code inside a non-standard subfolder.

## Root cause

`core/tools/aidlc-utility.ts`, `detectWorkspace()` (latest **2.1.4** on the `v2` branch).

The source-file count has exactly two scan paths, and neither reaches an arbitrarily-named subdirectory:

```ts
// Top-level files only (no recursion at this depth)   — :1912
for (const entry of topSet) { /* counts extensions of FILES at the root only */ }

// Recurse into known source dirs if present (capped depth)   — :1932
for (const dirName of SCAN_SOURCE_DIRS) {
  if (topSet.has(dirName)) countFilesByLang(join(projectDir, dirName), langCounts, 6);
}
```

with the fixed allowlist:

```ts
const SCAN_SOURCE_DIRS = ["src", "app", "lib", "pages", "components", "tests"];   // :1745
```

The classification then ORs five signals (`:1977`):

```ts
const brownfield =
  hasSourceFiles ||      // langCounts non-empty — but langCounts is only filled by the two scans above
  hasFrameworkConfig ||  // detectFrameworks(topSet, ...) — top-level configs only
  hasNonDev ||           // root package.json with deps
  hasOtherManifest ||    // root requirements.txt / Cargo.toml / go.mod / pom.xml / ...
  hasAppSourceDir;       // SCAN_SOURCE_DIRS.some(d => topSet.has(d)) — fixed names only
```

So for a workspace laid out as `<root>/wordbook/**` (all code under `wordbook/`):

- top-level file scan: the root has no source files → contributes nothing.
- `SCAN_SOURCE_DIRS` recursion: `wordbook` is not in the list → never descended → `hasSourceFiles = false`.
- `hasAppSourceDir`: `SCAN_SOURCE_DIRS.some(d => topSet.has("wordbook"))` is `false`.
- `hasFrameworkConfig` / `hasNonDev` / `hasOtherManifest`: all read the **root** (`detectFrameworks(topSet, …)`, root `package.json`, root manifests) — the manifests/configs inside `wordbook/` are not seen.

All five signals are false → **Greenfield**, even though `wordbook/` is a full existing codebase.

This matches the shipped spec: `aidlc-common/stages/initialization/workspace-detection.md` lists the brownfield indicators as "Source code files exist", "Application source directories exist (src/, app/, lib/, pages/, components/)", etc. — all implicitly root-relative, with the source-dir indicator pinned to the same fixed names.

## Steps to Reproduce

1. Create a workspace root whose code lives in one nested, non-standard folder, e.g.:
   ```
   myproject/
     wordbook/          # a full existing app: package.json, src files, etc.
   ```
   (the root itself has no top-level source files and no `src/app/lib/pages/components/tests` directory).
2. Run `/aidlc` in `myproject/` and let Initialization classify the workspace, or call the scanner directly:
   `bun .kiro/tools/aidlc-utility.ts detect-scope --project-dir <myproject>` (and inspect `Project Type` in `aidlc-state.md`).
3. Observe `Project Type: Greenfield` and `Stages to Skip: 2.1 (reverse-engineering — greenfield)`, despite `wordbook/` containing real code.

## Expected Behavior

A workspace whose existing source is in a nested folder (regardless of that folder's name) is classified Brownfield, and Reverse Engineering runs.

## Actual Behavior

`Project Type: Greenfield`; Reverse Engineering is skipped; no codebase understanding is generated. The misclassification is decided entirely by whether the code happens to sit at the root or under one of the six hard-coded directory names.

## Fix we implemented and verified

We added a **nested-project fallback** to `detectWorkspace()` and verified it against the real engine. The change is +37 lines in `aidlc-utility.ts`, additive only:

1. A new `NESTED_SCAN_EXCLUDE` set (`docs`, `doc`, `examples`, `example`, `samples`, `sample`, `.github`, `.gitlab`, `scripts`) — top-level dirs that commonly hold sample/snippet/boilerplate code that is not the project's own source.
2. After the existing root + `SCAN_SOURCE_DIRS` scans, **only when no source signal was found at the root**, scan each arbitrarily-named top-level subdirectory one level in, skipping `SCAN_SOURCE_DIRS` (already handled), `NESTED_SCAN_EXCLUDE`, symlinks, and the existing `SCAN_EXCLUDE`:

```ts
if (Object.keys(langCounts).length === 0) {
  for (const entry of topSet) {
    if (SCAN_SOURCE_DIRS.includes(entry)) continue;          // already scanned
    if (NESTED_SCAN_EXCLUDE.has(entry.toLowerCase())) continue;
    const full = join(projectDir, entry);
    try {
      const st = lstatSync(full);
      if (st.isSymbolicLink() || !st.isDirectory()) continue;
      countFilesByLang(full, langCounts, 1);                 // one level in (shallow)
    } catch { /* ignore unreadable entry */ }
  }
}
```

The existing root-level fast path is byte-identical for normal layouts (the fallback never runs when the root already has a `src/`, a manifest, or top-level source).

We verified with a 9-fixture matrix (driver run): `root src/` → Brownfield (no regression); `wordbook/src/` → Brownfield (the fix); `backend/server/main.go` → Brownfield; empty root, `node_modules`-only, `.kiro`/`aidlc`-only, `.github`-only → Greenfield; `docs/examples/sample.py` → Greenfield (no false positive). Engine builds clean; the framework's own health check (32/32) and a 14-step orchestration e2e still pass.

### Limitations of this fix (deliberately scoped, not complete)

- **Depth 1 only.** It descends a single level into a container, so it catches `wordbook/src/**` and `backend/server/**` but **not** deeper nesting like `service/internal/core/x.rs` (still classified Greenfield). A deeper cap would catch more but raises false-positive and cost risk.
- **Name-based exclusion is a heuristic.** `NESTED_SCAN_EXCLUDE` is a fixed denylist; real source that happens to live under a folder literally named `examples/`/`scripts/` would be missed, and a non-source folder whose name is not on the list could still be counted.
- **Source-file count, not manifest-aware.** The signal is recognized source-file extensions (`LANG_BY_EXT`); a nested project that has only a manifest (e.g. a bare `go.mod`) with sources below the depth-1 cap would not be detected. A manifest-root probe (scan for `package.json`/`pyproject.toml`/`go.mod`/`pom.xml` one level down) would be more robust and is the alternative we did not take.
- **Only fires when the root signal is empty.** If the root has even one stray recognized source file, the fallback does not run by design (the root already classified Brownfield, so this is moot for classification, but it means the language list is taken from the root only).

A maintainer-side fix may well prefer the manifest-root approach over our extension-count heuristic; we are sharing the minimal, verified change that resolves the reported misclassification without changing the root fast path.

## Additional Context

- Found while running on Kiro IDE 0.12.333; the logic is in shared core (`detectWorkspace`), so it is harness-independent — any harness will misclassify the same nested layout.
- Confirmed on 2.1.1 and unchanged on the latest 2.1.4 (`SCAN_SOURCE_DIRS`, the two-path scan, and the five-signal OR are identical).

apackeer:
Fixes #462 and addresses #438: a project nested in an arbitrarily named folder (e.g. `wordbook/`) was classified Greenfield because `detectWorkspace` scanned only top-level files plus a fixed source-dir allowlist - so Reverse Engineering was skipped and the workflow treated an existing codebase as a fresh one. The bugfix scope hit this hardest (#438): its whole premise is existing code, yet a misread scan silently skipped RE.

## Fix

- **Nested-scan fallback**: the signal evaluation is extracted into a helper applied first at the root (byte-identical behavior for every existing layout, verified against the pinned scanner contract tests). When ALL root signals miss - the exact condition that used to mean Greenfield - the scanner re-applies the full five-signal set (manifests, framework configs, non-dev deps, source dirs, shallow source-file counts) one level down per non-hidden top-level directory, excluding dot-dirs, common non-source dirs (docs, examples, samples, scripts), and the aidlc workspace root. Any hit classifies Brownfield; multiple hits aggregate languages and frameworks. The winning subdirectories are recorded in a new optional `nestedRoot` field surfaced in the WORKSPACE_SCANNED audit event and `detect --json` (no state-file change).
- **Incremental-scope advisory** (#438): when the scan still reads Greenfield and the scope is bugfix, refactor, or security-patch, intent birth prints a one-line stderr note that the scope usually targets existing code, RE will be skipped, and how to correct it (fix the layout or edit Project Type in aidlc-state.md). Deliberately NOT a forced Brownfield or a routing override: an empty directory genuinely has nothing to reverse-engineer, and the existing scope-runner pins encode that semantics.

## Tests

New `tests/unit/t203-nested-workspace-detection.test.ts` (10 cases): nested project, deep container with no manifests, manifest-only nested project, empty dir unchanged, top-level unchanged, excluded-dirs-only stays Greenfield, two nested repos both recorded, symlink skipped, bugfix advisory fires on Greenfield, non-incremental scope does not. All named pins re-verified green including the live SDK detection pair (t70 greenfield / t71 brownfield) and t130 scope-runners. Adversarial review of the first cut caught a scope-override that would have broken those pins; the shipped version is the corrected design.

Known interaction, deliberately untouched: the workspace-journey test fixture seeds tiny stub repos at depth 1, so that fixture now classifies Brownfield - no journey test pins Project Type (verified), and the fixture's separate flake is tracked in #441.

Version 2.2.7; sibling PRs hold adjacent numbers and whichever merges later re-bumps on rebase.

### DIFF
--- a/core/tools/aidlc-utility.ts
+++ b/core/tools/aidlc-utility.ts
@@ -1764,6 +1764,9 @@ const LANG_BY_EXT: Record<string, string> = {
 };
 
 const SCAN_SOURCE_DIRS = ["src", "app", "lib", "pages", "components", "tests"];
+// Set view for the sweep-side skip in scanSignals: the depth-6 recurse there
+// is the SOLE counter for these dirs, so the file sweep must never enter them.
+const SCAN_SOURCE_DIR_SET: ReadonlySet<string> = new Set(SCAN_SOURCE_DIRS);
 const SCAN_EXCLUDE = new Set([
   ".claude",
   ".kiro",
@@ -1806,13 +1809,22 @@ const NESTED_SCAN_EXCLUDE = new Set([
   "example",
   "samples",
   "sample",
+  "demos",
+  "demo",
+  "reference",
+  "testdata",
+  "fixtures",
+  "templates",
   "scripts",
 ]);
 
+// skipDirs: directory names to skip at THIS level only (not propagated into
+// the recursion); the caller counts those dirs through a separate deeper call.
 function countFilesByLang(
   dir: string,
   counts: Record<string, number>,
-  maxDepth: number
+  maxDepth: number,
+  skipDirs?: ReadonlySet<string>
 ): void {
   if (maxDepth < 0) return;
   let entries: string[];
@@ -1833,6 +1845,7 @@ function countFilesByLang(
     // Don't follow symlinks — cycle protection.
     if (st.isSymbolicLink()) continue;
     if (st.isDirectory()) {
+      if (skipDirs?.has(entry)) continue;
       countFilesByLang(full, counts, maxDepth - 1);
     } else if (st.isFile()) {
       const dot = entry.lastIndexOf(".");
@@ -1959,8 +1972,11 @@ function hasNonDevDeps(projectDir: string): boolean {
 //       countFilesByLang(dir, counts, 0), same SCAN_EXCLUDE filter + symlink
 //       skip, files only) PLUS a depth-6 recurse into each present
 //       SCAN_SOURCE_DIRS entry.
-//     - a nested container: 6, scanning the whole container one level in so a
-//       project under `wordbook/src/**` or `backend/server/**` is seen.
+//     - a nested container: 1, sweeping the container's own files plus one
+//       level of arbitrary subdirs so a project under `wordbook/**` or
+//       `backend/server/**` is seen. A present SCAN_SOURCE_DIRS entry is
+//       skipped by the sweep: the depth-6 recurse below is its only counter,
+//       so files directly under it are never counted twice.
 interface DirSignals {
   brownfield: boolean;
   langCounts: Record<string, number>;
@@ -1980,9 +1996,14 @@ function scanSignals(dir: string, fileScanDepth: number): DirSignals {
   // Source-file count. countFilesByLang(dir, counts, 0) counts files directly
   // under dir (its recursion guard returns immediately at depth -1), matching
   // the base top-level file sweep. Any present known source dir is then
-  // recursed at the base depth cap.
+  // recursed at the base depth cap. The sweep itself never enters a
+  // SCAN_SOURCE_DIRS entry: at depth 1 (the nested-container scan) it would
+  // count the files directly under <dir>/src etc. that the depth-6 recurse
+  // below counts again, inflating that language and potentially flipping the
+  // reported primary. (At the root's depth 0 the skip is a no-op: the sweep
+  // never enters subdirs there.)
   const langCounts: Record<string, number> = {};
-  countFilesByLang(dir, langCounts, fileScanDepth);
+  countFilesByLang(dir, langCounts, fileScanDepth, SCAN_SOURCE_DIR_SET);
   for (const dirName of SCAN_SOURCE_DIRS) {
     if (entrySet.has(dirName)) {
       countFilesByLang(join(dir, dirName), langCounts, 6);
--- a/core/tools/aidlc-utility.ts
+++ b/core/tools/aidlc-utility.ts
@@ -1734,6 +1734,11 @@ interface ScanResult {
   languages: string;     // e.g. "TypeScript, JavaScript"
   frameworks: string;    // e.g. "React, Vite"
   buildSystem: string;   // e.g. "npm (package.json)"
+  // Comma-joined top-level subdirectory name(s) the nested-project fallback
+  // classified Brownfield from. Absent when the root itself decided the verdict
+  // (the common case). Surfaced only in the WORKSPACE_SCANNED audit event and
+  // the `detect --json` payload, never in the state file.
+  nestedRoot?: string;
 }
 
 const LANG_BY_EXT: Record<string, string> = {
@@ -1771,6 +1776,37 @@ const SCAN_EXCLUDE = new Set([
   ".next",
   "target",
   "vendor",
+]);
+
+// Package/build manifests that mark a directory as an application (not just
+// scaffolding). Shared by the root scan and the nested-project fallback.
+const SOURCE_MANIFESTS = [
+  "requirements.txt",
+  "pyproject.toml",
+  "setup.py",
+  "Cargo.toml",
+  "go.mod",
+  "pom.xml",
+  "build.gradle",
+  "build.gradle.kts",
+  "composer.json",
+  "Gemfile",
+];
+
+// Top-level directories the nested-project fallback never descends into: they
+// commonly hold sample/snippet/boilerplate code that is not the project's own
+// source. The harness/VCS/build dirs in SCAN_EXCLUDE and SCAN_SOURCE_DIRS
+// (already scanned at the root) are skipped separately. Lowercased for a
+// case-insensitive match.
+const NESTED_SCAN_EXCLUDE = new Set([
+  "aidlc",
+  "docs",
+  "doc",
+  "examples",
+  "example",
+  "samples",
+  "sample",
+  "scripts",
 ]);
 
 function countFilesByLang(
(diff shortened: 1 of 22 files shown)


core/tools/aidlc-utility.ts
```diff
@@ -1912,6 +1948,69 @@ function hasNonDevDeps(projectDir: string): boolean {
...
+    // dir doesn't exist yet (caller should scaffold first)
+  }
+  const entrySet = new Set(entries.filter((e) => !SCAN_EXCLUDE.has(e)));
+
+  // Source-file count. countFilesByLang(dir, counts, 0) counts files directly
+  // under dir (its recursion guard returns immediately at depth -1), matching
+  // the base top-level file sweep. Any present known source dir is then
+  // recursed at the base depth cap.
+  const langCounts: Record<string, number> = {};
+  countFilesByLang(dir, langCounts, fileScanDepth);
```
leandrodamascena: Double-count on the nested path. When called as `scanSignals(sub, 1)`, this line already counts files directly under `<sub>/src`, `<sub>/app`, etc. at depth 1 — then the loop just below recurses those same `SCAN_SOURCE_DIRS` dirs again at depth 6, so files directly under them are counted twice. The root call (`scanSignals(dir, 0)`) is immune because the depth-0 sweep never enters subdirs.

Effect: the `Languages` field can name the wrong primary language for a nested project (e.g. 3 `.ts` under `src/` doubled to 6 beats 4 top-level `.py`). Classification is fine (boolean OR), but that wrong value is written to aidlc-state.md, the WORKSPACE_SCANNED audit, and `detect --json`.

Fix: exclude `SCAN_SOURCE_DIRS` names from the depth-1 sweep so the depth-6 recurse is the only counter, mirroring the root path.

leandrodamascena (changes requested):
The root-scan refactor holds up: `scanSignals(projectDir, 0)` reproduces the old inline scan, and the nested fallback only fires when the root is already Greenfield, so it can't regress a Brownfield repo. Routing is safe.

One thing to fix first, in `scanSignals`: the nested path double-counts source files. `countFilesByLang(dir, counts, 1)` already counts files under `<container>/src` at depth 1, then the `SCAN_SOURCE_DIRS` loop recurses those same dirs again at depth 6 — counting them twice. The root path avoids this because its depth-0 sweep never enters subdirs. So a nested container with 4 top-level `.py` and 3 `.ts` under `src/` reports TypeScript as primary (3->6) when the same layout at the root correctly reports Python. Routing is unaffected (boolean OR on non-empty counts), but the wrong language lands in aidlc-state.md, the audit, and `detect --json`. Fix: skip `SCAN_SOURCE_DIRS` in the depth-1 sweep, same as the root path.

Smaller, non-blocking:
- The comment above the nested call says depth 6, but it passes 1 — worth fixing since that's what hides the double-count.
- `NESTED_SCAN_EXCLUDE` misses demo(s)/reference/testdata/fixtures/templates, and pluralizes example/sample but not demo — looks accidental.
- t203 checks languages with `toContain` only, so the double-count is invisible to the suite; refactor/security-patch aren't pinned for the advisory (only bugfix is).

apackeer:
All four items addressed in 6b44054:

- **Double count (blocking):** `countFilesByLang` now takes an optional skip set applied at the current level only; `scanSignals` passes `SCAN_SOURCE_DIRS`, so the depth-6 recurse is the sole counter for those dirs. At the root's depth-0 sweep the skip is a no-op (that sweep never enters subdirs), so root behavior is byte-identical.
- **Stale comment:** the "nested container: 6" note now describes the actual depth-1 sweep plus the source-dir skip.
- **NESTED_SCAN_EXCLUDE:** gains `demos`, `demo`, `reference`, `testdata`, `fixtures`, `templates` (the demo singular/plural gap was accidental, as you guessed).
- **t203:** languages pinned with exact equality instead of `toContain`, plus a dedicated regression case (4 top-level `.py` vs 3 `.ts` under `src/` must report `Python, TypeScript`; the doubled count reported `TypeScript, Python`). The exclude-dir case covers the new names, and `refactor`/`security-patch` now pin the greenfield advisory alongside `bugfix`.

Smoke+unit green (130 files), t203 13/13, typecheck + package parity clean.

leandrodamascena (approved):
LGTM

PR #500, opened 2026-07-05
fix: bound the compose-pending carve-out and guard recompose against autonomy (2.2.8)

### LINKED ISSUE #484: [Bug]: an orphaned .aidlc-compose-pending marker silently disables the Stop-hook forwarding enforcement - no janitor, no doctor probe
## What

A crash or abandoned session between "write the pending-compose marker" and "gate resolves" leaves `aidlc/.aidlc-compose-pending` on disk forever. Nothing ever cleans it up and `--doctor` does not know it exists.

## Why it matters

The marker is a Stop-hook carve-out: while it is present, the forwarding-loop enforcement allows any turn to end (`core/hooks/aidlc-stop.ts` - `isPendingComposeStop` returns the bare `existsSync` of the marker). The conductor is instructed to write it before presenting the in-flight compose gate and delete it the moment the gate resolves (`core/tools/aidlc-orchestrate.ts` dispatch print; the SKILL.md marker discipline). That discipline handles the happy path, but:

- a session that crashes or is killed mid-gate never deletes the marker,
- the next session inherits a permanently open carve-out: the Stop hook allows every stop, so the forwarding-loop enforcement is silently disabled for that workspace until someone hand-deletes a hidden file they have never heard of,
- `bun .claude/tools/aidlc-utility.ts doctor` (`handleDoctor`) has no probe for it, so there is no discoverable diagnosis.

## Proposed fix

Two small, independent pieces:

1. **Doctor probe**: `doctor` reports a present `aidlc/.aidlc-compose-pending` with its mtime and the fix line (delete it if no compose gate is actually pending). Cheap, read-only, no behavior change.
2. **Staleness bound in the Stop hook**: the carve-out only honors a marker younger than a generous TTL (e.g. 24h, or same-session via the session-id file); an older marker is ignored (and reported in the hook's transcript output) so an orphaned marker cannot disable the enforcement loop indefinitely.

The write/delete discipline stays with the conductor; this is a janitor and a tripwire, not a new owner for the marker.

## Origin

Surfaced by the adversarial review pass that hardened PR #477 before merge (2.2.0); pre-existing gap in the shipped in-flight compose flow, not a regression.

### LINKED ISSUE #485: [Bug]: recompose has no engine-side guard against autonomous Construction - the "never under autonomous" rule is prose-only
## What

The SKILL.md compose prose says plan-reshape "never runs under autonomous Construction" on every harness, but the `recompose` verb itself has no engine-side check: `handleRecompose` (`core/tools/aidlc-utility.ts`) never reads `Construction Autonomy Mode`. The rule is enforced only by conductor prose.

## Why it matters

The framework's design principle is determinism in tools, judgment in humans, knowledge in the LLM. Every other "never under autonomous" rule has a deterministic anchor:

- the Stop hook's pending-compose carve-out is autonomy-guarded in code (`core/hooks/aidlc-stop.ts` - `isPendingComposeStop` returns false under `Construction Autonomy Mode: autonomous`),
- park refuses under autonomy (`aidlc-state.ts park`),
- the approve/answer human-presence gate exempts autonomous Construction explicitly.

`recompose` is the exception: a conductor deep in an unattended swarm/Bolt run that drifts into "improving the plan" can call the verb directly and it will happily flip pending stages with no human at the gate. The strict validator catches starvation and anchor moves, but not the absence of a human.

## Proposed fix

`handleRecompose` reads `Construction Autonomy Mode` from the state file (it already reads `Status` and `Scope` there) and refuses with a clear message when it is `autonomous`, mirroring the park guard's shape. Mechanical follow-on: a t194 rejection case pinning it.

If there is a legitimate unattended-recompose story later, it can arrive as an explicit flag; the default should be the refusal the prose already promises.

## Origin

Surfaced by the adversarial review pass that hardened PR #477 before merge (2.2.0).

apackeer:
Fixes #484 and fixes #485 - two hardening follow-ups to the 2.2.0 Adaptive Workflows compose flow.

## Orphaned compose-pending marker (#484)

The conductor writes `aidlc/.aidlc-compose-pending` before presenting an in-flight compose gate and deletes it on resolution; the Stop hook honored the marker's bare existence as a carve-out from forwarding enforcement. A crash mid-gate orphaned the marker and silently disabled the Stop hook's enforcement forever.

- The carve-out now honors the marker only within a 24h freshness bound (mtime; the conductor rewrites the marker per gate, so the window tracks gate presentation). A stale marker is ignored, logged via the hook-drop channel, and best-effort deleted. Fail-open discipline preserved: any stat error falls into the existing catch and the hook proceeds to its cap-bounded default; the change can only ever reduce suppression.
- `/aidlc --doctor` gains a "Pending-compose marker" probe: a present marker is flagged with its mtime and age plus the remediation (delete it, or resolve the pending gate). Read-only.
- The TTL constant and marker path are now shared exports consumed by the hook, the doctor probe, and the staleness test, so the three cannot drift.

## Recompose autonomy guard (#485)

The "never re-shape the plan under autonomous Construction" rule was prose-only. `recompose` now refuses, inside the audit lock, when Construction Autonomy Mode is autonomous, with an actionable error (gate the run via `bolt set-autonomy --mode gated`, or let it finish). Mirrors the park guard's shape.

## Tests

t194 gains the autonomy-rejection case (state byte-unchanged after refusal) and its gated complement; t195 gains fresh-marker-honored and stale-marker-blocked-and-cleaned cases against the shipped dist hook; new `tests/unit/t204-compose-marker-doctor.test.ts` covers the doctor probe (absent silent / present flagged / age rendered). Adversarially reviewed: all six dimensions sound (fail-open verified path by path; the review's one blocker was the missing docs updates, now landed: orchestrator reference, CLI commands, hooks-and-tools recompose row, getting-started doctor table).

Two pre-existing gaps the review surfaced, left for follow-up issues: doctor never reads the hook-drop files despite the recorder's comment claiming it does, and `scope-change` has the same missing autonomy guard this PR closes for recompose.

Version 2.2.8; sibling PRs hold adjacent numbers and whichever merges later re-bumps on rebase.


core/tools/aidlc-utility.ts
```diff
@@ -762,6 +765,33 @@ function handleDoctor(projectDir: string): void {
...
+  // hook is the janitor for a stale one). No behavior change.
+  try {
+    const composeMarker = composeMarkerPath(projectDir);
+    if (existsSync(composeMarker)) {
+      const ageMs = Date.now() - statSync(composeMarker).mtimeMs;
+      const ageHours = Math.floor(ageMs / (60 * 60 * 1000));
+      const ageLabel = ageHours >= 1 ? `${ageHours}h old` : "under 1h old";
+      const staleLabel = ageMs > COMPOSE_MARKER_TTL_MS ? ", stale" : ", fresh";
+      results.push({
+        pass: false,
```
leandrodamascena: This pushes `pass:false` for any present marker, and a `pass:false` result flips doctor to exit 1 (line 1747). But a FRESH marker is the normal state at a live compose gate (written before the gate, deleted on resolve), so `/aidlc --doctor` run while a gate is legitimately open — a second terminal, or CI — reports a red row and non-zero exit on a healthy workspace. The `, fresh`/`, stale` label is already computed two lines up but ignored for `pass`. Suggest: `pass: ageMs <= COMPOSE_MARKER_TTL_MS` (fresh = advisory pass, only stale fails), matching how the other 'not a fault' rows (rule drift, leaked locks) are modeled as `pass:true`.

leandrodamascena (changes requested):
The two hardening fixes work. The Stop hook still fails open — the worst the TTL path does on an old marker is reap it and fall through to the cap-bounded block, no new trap. The recompose `die()` inside `withAuditLock` releases the lock through the `process.on('exit')` handler, the field-name check matches `readAutonomyMode`, and no engine/swarm path calls recompose, so the guard can't break a legit autonomous run.

One thing I'd change before merge: the new doctor probe pushes `pass:false` for ANY present compose marker, including a fresh one, and `pass:false` flips doctor to exit 1 (line 1747). A fresh marker is the normal state at a live compose gate — write-before-gate, delete-on-resolve per SKILL.md — so running `/aidlc --doctor` in a second terminal while a gate is legitimately open reports a red row and non-zero exit on a healthy workspace. The window is short and nothing in CI runs `--doctor` today, so the blast radius is small, but every sibling "not a fault" row (rule drift, leaked locks) is `pass:true` advisory, and this probe already computes the `, fresh`/`, stale` label at line 784 and then ignores it for `pass`. Make the fresh case `pass:true` (advisory) and only fail on stale — one line, and it matches the siblings.

Non-blocking:
- `scope-change` (handleScopeChange, ~2854) reshapes the same plan surface with no autonomy guard — the same hole this closes for recompose. PR body defers it; just track it so the rule is fully closed.
- t204 never seeds a marker past the 24h TTL, so the stale branch and the fresh/stale label go unverified. And t195/t204 re-hardcode the marker path instead of importing `composeMarkerPath`, so only the TTL constant is drift-locked to the tests.

apackeer:
Fixed in 50f53ac:

- **Fresh marker = advisory pass (blocking):** the probe's `pass` now follows the fresh/stale label it already computed: fresh renders as an advisory pass row (no fix text, no exit-code contribution, so `/aidlc --doctor` in a second terminal during a live gate stays green), stale renders as a fail row with the remediation and flips the exit code, matching the sibling "not a fault" rows. CHANGELOG bullet and the two doc mentions updated to match.
- **t204 stale branch:** now seeds a marker one hour past `COMPOSE_MARKER_TTL_MS` and asserts the stale FAIL row (label + remediation + exit 1) as well as the fresh advisory PASS row and the `, fresh` label on the 3h case.
- **Marker-path drift:** t195 and t204 import `composeMarkerPath` from the shipped lib instead of re-hardcoding the path, so the spelling is drift-locked alongside the TTL constant.
- **scope-change autonomy hole:** tracked as #515.

Smoke+unit green (130 files), t195 6/6 + t204 4/4, typecheck + package parity clean.

leandrodamascena (approved):
LGTM

PR #507, opened 2026-07-05
fix: detect git submodules as a brownfield signal in workspace detection (2.2.10)

### LINKED ISSUE #504: [Bug]: workspace detection is blind to git submodules - a multi-repo workspace with uninitialized submodules classifies Greenfield and skips Reverse Engineering
## AI-DLC Version

v2 branch, verified on HEAD `eae912e` (2.2.0). Harness-independent core logic.

## AI-DLC Phase

Initialization - workspace-detection (sets `Project Type`, which gates whether Reverse Engineering runs).

## What

`detectWorkspace()` (`core/tools/aidlc-utility.ts:1915-2005`) classifies brownfield from five signals: source files by extension, framework config, non-dev package.json deps, other manifests, app source dirs (`:1975-1990`). None of them knows about git submodules:

- `.gitmodules` is not a recognized extension, framework config, or manifest - it contributes nothing to any signal.
- An UNINITIALIZED submodule is an empty directory (no `.git` file inside until `git submodule update --init`), so it also fails `isGitRepoDir()` (`core/tools/aidlc-lib.ts:1515-1517`) and is not discovered by `discoverSiblingRepos()` (`aidlc-lib.ts:1534-1554`).

So a workspace whose code lives in submodules - the natural layout for "one AIDLC workspace orchestrating changes across several existing repos" - scans as `Project Type: Greenfield, Languages: Unknown` when the submodules are not yet initialized. Downstream, greenfield flips reverse-engineering to SKIP during state build (`core/tools/aidlc-utility.ts:2326-2338`, re-applied on scope change at `:2857-2863`), and `conditional_on: brownfield` consumes (architecture, component-inventory) are stripped from later stages (`core/tools/aidlc-orchestrate.ts:889-896`, `:914-938`).

Net effect: every design stage for what is actually brownfield work across existing production repos runs with zero code understanding, and the first agent that needs the real code (code-generation) has to discover and initialize the submodule itself mid-stage - unaudited, and too late to inform any design document.

## Steps to reproduce

1. Create a workspace repo with a `.gitmodules` declaring one or more submodules pointing at existing code repos; do not run `git submodule update --init`.
2. Run `/aidlc <any brownfield feature description>` from the workspace root.
3. Observe `WORKSPACE_SCANNED` reports `Project Type: Greenfield, Languages: Unknown`, and `aidlc-state.md` marks reverse-engineering SKIP with reason greenfield.

## Expected behavior

Layered, cheapest first:

1. **Detect**: `detectWorkspace()` treats a parseable `.gitmodules` at the root as a brownfield signal (it declares the existence of code even when the dirs are empty).
2. **Warn**: when `.gitmodules` entries exist whose paths are empty/uninitialized, the scan result (and `--doctor`) says so explicitly and tells the user to run `git submodule update --init --recursive` before proceeding - the single highest-leverage pre-flight for this layout.
3. Optionally: workspace-detection offers to initialize them (gated, since it touches the network).

## Related

- #462 - same function, sibling blind spot (source nested in an arbitrarily-named subdirectory). A fix for #462's directory walk does NOT cover this case: an uninitialized submodule dir is empty, so no walk finds anything - only the `.gitmodules` probe does.
- #438 - project type defaulting in the bugfix scope; same misclassification consequence from a different entry point.
- #27 (closed, v1-era) - the earliest instance of this failure family: missing file extensions made a brownfield project scan greenfield and skip reverse-engineering. Notably, v1's scan was model-executed prose (`aidlc-rules/aws-aidlc-rule-details/inception/workspace-detection.md` on `main`) that could in principle notice a `.gitmodules` in the listing; v2's deterministic scanner is the right trade (reproducible, auditable) but only sees what it is coded to see, so repo metadata must be an explicit signal. Each fix so far has patched one signal; #462 and this issue suggest `detectWorkspace()` needs a broader notion of "evidence of existing code" (extensions, nested dirs, and repo metadata like `.gitmodules`).
- #441 (test) - a shared test fixture classifying Greenfield and skipping RE; would benefit from whatever detection fix lands here/#462.

## Origin

Code inspection of `detectWorkspace()` and the greenfield->SKIP path while auditing how multi-repo workspaces classify.

apackeer:
Closes #504.

## What

Workspace detection now recognizes git submodules as evidence of existing code, so a multi-repo workspace whose code lives in submodules no longer misclassifies as Greenfield (which silently skipped Reverse Engineering).

- `detectWorkspace()` gains a sixth brownfield signal: a parseable `.gitmodules` at the workspace root with at least one valid `path` entry. New exported `parseGitmodules()` (line-oriented, drops absolute/`..`/pathless entries, malformed input degrades to no-signal, never throws) plus a submodule scan that probes each declared path for a `.git` entry.
- When declared submodules are uninitialized (path missing, empty, or lacking `.git`), the scan says so explicitly and carries the remedy: `git submodule update --init --recursive`. Surfaces: the `WORKSPACE_SCANNED` audit event (new `Submodules` field), the intent-birth stdout block, `detect --json` (additive `submodules` key), and a new advisory (never-failing) `--doctor` row.
- `workspace-detection.md` stage prose gains the sixth Brownfield bullet and a conductor note to relay the warning verbatim (no auto-init offer; initializing touches the network and stays a human decision).
- Projects without `.gitmodules` take a byte-identical code path (empirically verified: origin/v2 vs this branch produce identical birth stdout and `WORKSPACE_SCANNED` events on a bare project).

Version 2.2.10, CHANGELOG entry, README badge. Tests t211 (detection, warning payload, RE stays EXECUTE, malformed degrade, greenfield byte-stability, parser units) and t212 (all four doctor row states).

## Why

The brownfield signal list only looked for files (extensions, manifests, framework configs, source dirs). An uninitialized submodule is an empty directory, so a workspace declaring all its code via `.gitmodules` scanned as `Greenfield, Languages: Unknown`, reverse-engineering was flipped to SKIP during state build, and every design stage ran with zero code understanding. The first agent to need real code then had to discover and init the submodule itself mid-stage, unaudited and too late to inform design.

Sibling blind spots stay open and compose with this fix: #462 (source nested in arbitrarily-named dirs, a directory-walk fix) and #438 (bugfix-scope default). This change is additive metadata detection only.

## Verification

- Deterministic tiers green: smoke+unit (131 files, 2317 assertions, 0 failed), integration incl every `WORKSPACE_SCANNED`-asserting test (t18/t20/t70/t71/t111), t68 version sync, t174 legacy-refs.
- `bun scripts/package.ts --check` clean (all 4 dist trees), `bun run typecheck` clean, coverage registry regen produces zero diff.
- Adversarial verification pass: 15 hostile `.gitmodules` inputs (absolute paths, `..` escapes, CRLF, garbage, file-as-path) all degrade to no-signal without throwing; live end-to-end drive of birth/detect/doctor on a temp fixture.
- Known cosmetic edges (non-blocking): duplicate `[submodule]` sections double-count in the warning text; a quoted `path = "with spaces"` value reads as uninitialized (false warning only, classification unaffected); key matching is case-sensitive (degrades to no-signal).

Draft until the branch soaks a full pre-merge live gate alongside its four sibling PRs (2.2.9-2.2.13 series); whoever merges later re-bumps per the CHANGELOG conflict-trap policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.2.10] - 2026-07-06
+
+Workspace detection now recognizes git submodules. A workspace whose code lives in uninitialized submodules (empty dirs plus a `.gitmodules` file) previously scanned as Greenfield, so reverse-engineering was auto-skipped and every design stage ran with zero code understanding. The scanner gains a sixth brownfield signal: a parseable `.gitmodules` with at least one submodule path entry classifies the workspace Brownfield. When submodule paths are uninitialized, the scan warns and names the remedy (`git submodule update --init --recursive`) at birth, in the doctor report, and on `detect`. Languages stay as scanned (Unknown is truthful until the submodules are fetched). **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* Workspace detection classifies a workspace with a parseable `.gitmodules` (at least one submodule path entry) as Brownfield, so reverse-engineering is no longer skipped for submodule-based projects.
+* The `WORKSPACE_SCANNED` audit event gains a `Submodules` field (`N declared, M uninitialized`) when submodules are present, and its `Details` names the `git submodule update --init --recursive` remedy when any submodule path is uninitialized. Projects without a `.gitmodules` produce a byte-identical event.
+* Birth stdout prints a one-line warning when submodule paths are uninitialized, telling you to fetch them before proceeding so reverse-engineering can read the code.
+* `/aidlc --doctor` gains an advisory `Submodules:` row: no `.gitmodules`, all initialized, uninitialized (names the count, the paths, and the remedy), or present-but-unparseable. Advisory only - it never flips doctor's exit code.
+* The read-only `detect` utility gains a `submodules` key in its `--json` payload and a `Submodules:` line in its human output when submodules are present.
 ## [2.2.9] - 2026-07-06
 
 Per-unit Construction stages no longer demand CONDITIONAL artifacts before a unit counts as covered. A new optional stage-frontmatter key `optional_produces:` names artifacts a stage writes only when the unit needs them - `functional-design`'s `frontend-components` (only when the unit has a UI) and `infrastructure-design`'s `shared-infrastructure` (only when units share infrastructure). Those names moved out of `produces:` into `optional_produces:`, so a backend-only unit completes without an N/A stub file and the stage gate is reachable. The conductor still gets the artifact's write path in the run-stage directive when the unit does produce it, and the artifact name stays in the vocabulary registry. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
@@ -73,6 +82,7 @@ Stops help requests from accidentally creating intents. `/aidlc intent help` was
 ||||||| parent of ef64a3e (fix: detect nested projects in workspace detection (2.2.7))
 ||||||| parent of 48f8b61 (fix: bound the compose-pending carve-out and guard recompose against autonomy (2.2.8))
 ||||||| parent of 4fd686b (fix: exempt optional produces from per-unit coverage so conditional artifacts can be skipped (2.2.9))
+||||||| parent of 4ad1db9 (fix: detect git submodules as a brownfield signal in workspace detection (2.2.10))
 
 ## [2.2.0] - 2026-07-04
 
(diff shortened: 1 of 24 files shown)


leandrodamascena (approved):
Checked the parser, the byte-identical claim, and the doctor row. `parseGitmodules` never throws and drops any escaping path (`/`-absolute, `C:\`, `..`-segment) before it touches disk; no-`.gitmodules` projects stay byte-identical; and the doctor row is `pass:true` in all states, so it doesn't repeat the #500 exit-1 thing. No break or risk.

One test gap worth closing: the traversal-drop test only feeds `/etc/passwd` and `../../escape`, so the Windows drive-absolute branch never runs — add a `C:\evil` entry to that assertion so it can't silently break later.

Approved & good to merge.

PR #508, opened 2026-07-05
feat: preview stage and gate counts at scope confirmation (2.2.12)

### LINKED ISSUE #505: [Feature]: scope confirmation should preview the cost - stage count and gate count - before the workflow is born
## AI-DLC Version

v2 branch, verified on HEAD `eae912e` (2.2.0). Engine + orchestrator prose (all harnesses).

## AI-DLC Phase

Initialization - the scope routing ask before intent birth (and the composer's front moment).

## Summary

Since 2.2.0, cold-start freeform routing is much better than the old silent `feature` fall-through: a keyword hit asks a one-line confirm naming the matched scope, and rich/unmatched prose gets the compose offer (`core/tools/aidlc-orchestrate.ts:1530-1550`). But the confirm names ONLY the scope:

> `Starting a "feature" workflow for: "...". Confirm to proceed, name a different scope, or say "compose"...` (`aidlc-orchestrate.ts:1539-1541`)

Nothing tells the user what they are agreeing to. Verified absent from the confirm string, the compose offer (`:1544-1548`), `birthPrintDirective` (`:361-380`), and the docs: any statement of stage count, gate count, or expected ceremony. The data exists - the compiled grid knows `feature` = 32/32 stages (`renderScopeTable`, `core/tools/aidlc-utility.ts:3400-3418`; the scope table in SKILL.md) - it just never reaches the ask.

Why it matters: scopes differ by an order of magnitude in ceremony (poc = a handful of stages; feature/enterprise = the full 32-stage arc with 15+ human approval gates and per-unit design fan-out in construction). The `feature` scope is also the catch-all fallback for any unmatched description (`core/scopes/aidlc-feature.md:25-29`), i.e. the HEAVIEST practical scope is what you get by default when you do not know the scope vocabulary. A user confirming "feature" without knowing it means 32 stages is consenting to hours of generation and a dozen-plus gates they never chose. Judgement belongs to the human (the framework's own three-mechanisms rule), but the human cannot exercise judgement over a cost they are never shown.

There is precedent for this disclosure in the framework's own history: v1's Workflow Planning stage presented the full execution plan - an EXECUTE/SKIP decision per stage plus a workflow visualization - for approval before construction began (`aidlc-rules/aws-aidlc-rule-details/inception/workflow-planning.md` on `main`). v2's scope grid + composer replaced that stage with a better mechanism, but the up-front plan disclosure was dropped in the translation; this issue asks to restore it at v2's equivalent decision point, using the compiled grid instead of a hand-built chart.

## Proposal

1. **Confirm strings carry the grid numbers.** One line, from data already compiled: `Starting a "feature" workflow: 32 stages, ~17 approval gates, per-unit design in Construction. Confirm, name a different scope (poc = 5 stages, mvp = 21), or say "compose".` Exact numbers come from the compiled scope grid, never hand-maintained.
2. **The composer proposal already has a grid - surface its size.** The compose gate's proposal should lead with `N stages EXECUTE / M SKIP` next to the rationale.
3. Optionally, a rough magnitude note in the docs (docs/guide scopes page) about what the full arc entails, so the expectation exists before the first run. Verified absent today: no wall-clock or ceremony-size expectation anywhere in docs/guide or stage prose (the only acknowledgement is the parking note that "a long workflow ... need not finish in one session").

## Alternatives considered

- Rely on the scope-table docs: users in the flow never see them; the ask is the only surface guaranteed to be read at decision time.
- Wall-clock estimates: too environment-dependent to promise; stage/gate COUNTS are deterministic and already compiled, so start there.

## Related

- #421 - the cost this preview would disclose (inline front-loaded phases, compaction pressure) is exactly what users currently discover only after committing; a preview is the demand-side complement to that RFC's supply-side fix.
- #247 - the stage progress trail is the in-flight visibility twin of this pre-flight preview (where am I in the N stages vs what am I signing up for).
- #420 - cost governance steering rules; adjacent but about steering agent spend, not disclosing workflow size at the routing ask.

## Origin

Review of the 2.2.0 Branch 8 routing strings against the compiled scope grid, while auditing how users end up in the full 32-stage arc without choosing it.

apackeer:
Closes #505.

## What

The scope routing asks now disclose what the user is agreeing to, with numbers derived from the compiled grid (never hardcoded):

- The cold-start keyword confirm: `Starting a "bugfix" workflow ... - 7 of 32 stages, 4 approval gates, 1 stage repeats per unit of work in Construction.`
- The compose offer's example list carries counts: `bugfix = 7 of 32 stages, poc = 8, feature = all 32`.
- The intent-birth print and the scope-change gate line carry the same numbers.
- The composer proposal now leads with `N stages EXECUTE / M SKIP, G approval gates`, sourced from the deterministic validator: `ScopeValidation` gains a `summary` field populated by the new helper, and the composer contract plus all four harness SKILL.md composer blocks relay it verbatim (numbers come from code, not the LLM).

Mechanics: new `scopeCostSummary(scope)` / `gridCostSummary(grid)` helpers in `aidlc-lib.ts`; the per-unit stage constants moved from the orchestrator into the lib so per-unit is single-sourced; the gate count mirrors the engine's phase-derived gate rule (`computeGate`: initialization stages gateless, every other EXECUTE stage gated). Docs updated to quote the new strings, including a stale `bugfix = 8 stages` example corrected to 7.

Version 2.2.11, CHANGELOG entry, README badge. Tests: t213 (helper contract - expected values computed from the shipped grid JSON, cross-pinned against the scope table and the validator summary) and t214 (emitted strings on all routing surfaces).

## Why

Scopes differ by an order of magnitude in ceremony (poc = 8 stages / 5 gates; feature = 32 / 29), and `feature` is the catch-all fallback for unmatched descriptions - the heaviest practical scope is what you get by default. The confirm named only the scope, so a user saying yes to "feature" was consenting to 32 stages and 29 gates they were never shown. The compiled grid always knew the numbers; this change routes them to the one surface guaranteed to be read at decision time.

## Verification

- Adversarial verification pass independently recomputed all nine scopes' numbers from the shipped grid JSON and matched them against live engine output on a temp project; single-sourcing confirmed (no duplicate per-unit constants; all cost text flows through one helper).
- Full smoke+unit green (131 files, 2321 assertions, 0 failed) plus targeted integration slices; `package.ts --check` clean on all four dist trees; typecheck clean; t68 version sync green.
- Known non-blocking refinements (follow-up candidates):
  1. The per-unit clause shows on scopes that SKIP units-generation (bugfix/poc/refactor/security-patch/infra), where code-generation degrades to a single pass on a cold start - the clause overstates ceremony in the safe direction for consent, but could be suppressed when `units-generation !== EXECUTE`.
  2. The cold-start confirm shows the scope's nominal grid; greenfield projects drop reverse-engineering at birth, so effective counts run 1 stage / 1 gate lower for the 8 scopes that mark it EXECUTE (project type is unknown at confirm time; the scope-change surface already uses the adjusted grid).

## Addendum: codex compose-front test hardening (18724b7)

The combined gate surfaced `t-exec-codex-compose-front` failing reproducibly, including on clean v2 - a pre-existing fragility, root-caused here: codex's workspace-write sandbox carves the project-root `.codex/` out of the writable workspace (the same treatment as `.git/`), so the composer's sanctioned composed-scope write (scope file + grid entry) is EPERM-denied under headless `codex exec`, and the approve arc only ever completed when the live model improvised a recovery. The other codex tests pass because they only write to the workspace root.

Fix (test-only, no version bump): the test's scratch `config.toml` now grants `[sandbox_workspace_write] writable_roots = ["<proj>/.codex"]`, the codex-exec twin of the `.git` grant the shipped codex config already documents for headless runs. Beat 2 now asserts the REAL sanctioned arc (scope resolves through the on-disk registry; a custom-named scope must be authored fresh on the granted path). Proven: two consecutive live greens on the previously always-red arc (one custom-scope run, one stock-match run), plus a third green in the collateral pass; the other three codex tests green.


leandrodamascena (approved):
I ran the numbers against the grid myself and they hold up — the preview never shows fewer stages or gates than the engine actually runs, which is the direction that would mislead consent. The gate count uses `phase !== "initialization"`, and I checked the skeleton-gate branch it has to agree with: `resolveSkeletonGate` returns true for all three stances, so the closed form and the real `computeGate` line up.

Two spots where the count is off, both overstating (and both called out in the PR):
- Greenfield: the cold-start confirm shows the nominal grid, +1 stage / +1 gate for the RE-EXECUTE scopes once the project resolves greenfield. Project type isn't known at confirm time, so fair enough.
- The per-unit clause fires whenever `perUnitStages > 0`, so bugfix/poc/refactor/security-patch/infra say "1 stage repeats per unit" even though they SKIP units-generation and it runs once. Overstates, so consent isn't harmed, but it names a repetition that can't happen there. Cheap fix: gate the clause on units-generation being EXECUTE.

One test gap worth closing: nothing counts the gates `computeGate` actually emits per scope against `scopeCostSummary().gates` — t213/t214 both re-implement the same rule, so if `resolveSkeletonGate` ever returns false they'd move together and still pass while diverging from real gating. A compile-and-count guard would lock that coupling.

Issue #502, opened 2026-07-05
[Feature]: relevance-gate the per-unit construction design artifacts by unit kind (every unit currently gets the identical ~20-doc matrix)


apackeer:
## AI-DLC Version

v2 branch, verified on HEAD `eae912e` (2.2.0). Harness-independent engine + stage authoring.

## AI-DLC Phase

Inception (units-generation) + Construction (functional-design, nfr-requirements, nfr-design, infrastructure-design).

## Summary

Four construction design stages are `for_each: unit-of-work`, and each mandates a fixed artifact list per unit:

- functional-design: 4 files (`core/aidlc-common/stages/construction/functional-design.md:105-110`)
- nfr-requirements: 5 files, none conditional (`nfr-requirements.md:104-110`)
- nfr-design: 5 files, none conditional (`nfr-design.md:111-117`)
- infrastructure-design: 5 files (`infrastructure-design.md:123-129`)

That is roughly 20 mandated design documents per unit before any code is generated, and the engine applies the list to every unit identically: `resolveProduces` maps `node.produces` straight to paths with zero filtering (`core/tools/aidlc-orchestrate.ts:989-998`), and the per-unit coverage check requires each file to exist (`aidlc-orchestrate.ts:1791-1806`).

Nothing in the pipeline distinguishes unit kinds. units-generation captures name/description/boundaries, responsibilities, deployment model, and a complexity estimate (`core/aidlc-common/stages/inception/units-generation.md:100-105`), but no field is ever read by engine code to shape the artifact list. Verified absence: there is no unit-kind-driven pruning anywhere in `core/tools/` or the stage graph.

Previous behaviour was more selective on both axes: in v1's rules the construction design stages were "CONDITIONAL, per-unit" with explicit Execute IF / Skip IF criteria per unit (e.g. functional design ran only for units with new data models or complex business logic - `aidlc-rules/aws-aidlc-rules/core-workflow.md` on `main`, Per-Unit Loop), and the mandated matrix was about half the size (nfr-requirements and nfr-design were 2 files each, now 5 each). v2 made all four stages unconditional for every unit while enlarging the file lists.

The result scales as `4 stages x N units x ~5 files`. A 5-unit intent owes ~100 design documents before code-generation starts. And the matrix is indifferent to what a unit IS:

- a spec/contract unit (a schema or interface definition other units build against) still owes `scalability-requirements.md`, `reliability-design.md`, `monitoring-design.md`, ...
[...]

PR #509, opened 2026-07-05
fix: exempt optional produces from per-unit coverage so conditional artifacts can be skipped (2.2.9)

### LINKED ISSUE #501: [Bug]: per-unit coverage check requires CONDITIONAL artifacts to exist, so a unit can never legitimately skip them
## AI-DLC Version

v2 branch, verified on HEAD `eae912e` (2.2.0). Harness-independent engine logic.

## AI-DLC Phase

Construction: all `for_each: unit-of-work` design stages (functional-design, nfr-requirements, nfr-design, infrastructure-design).

## What

Two construction design stages mark an artifact CONDITIONAL in their stage prose:

- `core/aidlc-common/stages/construction/functional-design.md:110`: "**frontend-components.md** (CONDITIONAL - only if unit includes frontend/UI)"
- `core/aidlc-common/stages/construction/infrastructure-design.md:129`: "**shared-infrastructure.md** (CONDITIONAL - produce when multiple units share infrastructure resources)"

But the frontmatter `produces:` lists carry those artifacts flatly, with no conditional annotation (`functional-design.md:13-17`, `infrastructure-design.md:14-19`), and the compiled graph preserves them verbatim. The per-unit coverage check then requires EVERY `produces` entry to exist on disk before a unit counts as covered:

- `core/tools/aidlc-orchestrate.ts:1791-1806` (`unitCovered`): `for (const name of names) { ... if (!existsSync(abs)) return false; }`
- `core/tools/aidlc-orchestrate.ts:1816-1828` (`nextUncoveredUnit`): keeps returning the same unit while it is uncovered
- `core/tools/aidlc-orchestrate.ts:1868-1900`: the stage's real gate only fires on the all-covered re-entry (`pick === null`)

So a unit that follows the stage prose and skips `frontend-components.md` (a backend-only unit) or `shared-infrastructure.md` is never "covered": the engine re-emits the same unit on every `next`, and the stage cannot reach its gate. The only way out is to write a stub file whose body says "Not applicable", which contradicts both the CONDITIONAL marker and the depth guidance to skip the artifact (`core/aidlc-common/protocols/stage-protocol.md:648`).

The conditional wording is inherited from v1's rules ("If unit includes frontend/UI: Create ... frontend-components.md", `aidlc-rules/aws-aidlc-rule-details/construction/functional-design.md` on `main`; similarly "If shared infrastructure" in `infrastructure-design.md`), where it was honored in practice because nothing checked file existence. v2 kept the prose but added the coverage ledger that contradicts it - the bug is the interaction, not the prose.

Note the asymmetry: the `consumes` side already handles conditionality gracefully (`conditional_on:` entries are dropped at runtime, `core/tools/aidlc-orchestrate.ts:914-938`), but the `produces` side has no equivalent.

## Steps to reproduce

1. Run any scope that compiles a unit DAG (feature/enterprise/mvp) on an intent whose units include at least one backend-only unit (no UI).
2. Reach functional-design. Have the agent follow the stage prose and NOT write `frontend-components.md` for the backend-only unit.
3. Observe: the per-unit loop never completes for that unit; `next` keeps handing it back; the stage gate is unreachable until an N/A stub is written.

## Expected behavior

A CONDITIONAL artifact that a unit legitimately does not need should not block unit coverage. Either:

1. Carry the conditional marker into the compiled graph (e.g. `conditional: true` on the produces entry) and exempt it from the `unitCovered` existence check, or
2. Treat a unit as covered when all NON-conditional produces exist.

The stage prose, the depth guidance, and the engine's coverage check should agree; today the engine contradicts the other two and quietly trains runs to emit filler documents.

## Related

- #368 (closed, fixed in 2.1.2 via PR #444) introduced the per-unit `for_each` iteration + coverage ledger this bug lives in; this is a gap in that machinery (conditional produces were never exempted), not a regression of it.
- #502 - the unit-kind artifact-pruning feature generalizes the fix: kind-aware produces would subsume the CONDITIONAL exemption. This bug is the minimal engine-level correction either way.

## Origin

Surfaced by code inspection of the per-unit coverage path while auditing where construction doc volume comes from; the gap dates to when the coverage ledger was introduced.

apackeer:
Closes #501.

## What

Stage frontmatter gains an `optional_produces:` key (a second plain string list parallel to `produces:`, presence-gated in the parser so unannotated stages compile byte-identically). The two artifacts whose stage prose already marks them CONDITIONAL move into it:

- `frontend-components` in functional-design ("only if unit includes frontend/UI")
- `shared-infrastructure` in infrastructure-design ("produce when multiple units share infrastructure resources")

Effect on the engine, by construction rather than by new branching:

- `unitCovered` iterates `node.produces`, which now contains only required artifacts, so a backend-only unit that skips `frontend-components.md` counts as covered, the per-unit loop advances, and the stage gate is reachable. No code change in the coverage check itself.
- A missing REQUIRED artifact still blocks coverage exactly as before.
- `resolveProduces` concats both lists, so run-stage directives still carry the conditional artifact's output path for units that DO need it. No directive schema change, no conductor/SKILL.md changes on any harness.
- `producersOf`, the artifact registry, and the required-sections sensor's template eligibility union both lists, so the artifact vocabulary and consumer-coverage checks are unchanged.

Version 2.2.9, CHANGELOG entry, README badge, designer-export golden regenerated. Tests: t205 (frontmatter parse/emit/validate/compile, incl. the no-[]-stamped-on-other-stages pin) and t206 (engine behavior end-to-end through the real orchestrator: conditional-absent unit covered, required-absent blocks, gate placement, approve path).

## Why

The stage prose says the artifact is conditional; the depth guidance says skip it when not applicable; but the per-unit coverage ledger required every `produces` entry to exist on disk. A unit that legitimately skipped a conditional artifact was never "covered": the engine re-emitted it on every `next` and the stage gate was unreachable until an "N/A" stub file was written. The engine contradicted the framework's own guidance and quietly trained runs to emit filler documents.

The `consumes` side already handles conditionality (`conditional_on:` entries drop at runtime); this gives the `produces` side its equivalent, using the same parallel-list idiom as `required: false` consumes rather than restructuring `produces` into objects.

## Verification

- Adversarial verification pass confirmed: the t206 proof is non-vacuous (fails on origin/v2 two independent ways), every `node.produces` reader in core/tools + core/hooks was audited with no missed consumer, and presence-gating holds across all four dist stage-graphs (exactly 2 stages annotated, no empty lists stamped).
- Deterministic tiers green (smoke+unit 131 files / 2324 assertions / 0 failed; integration; deterministic e2e slice). `package.ts --check`, designer-export `--check`, and typecheck all clean. Live-flake claims for t126/t185/t78/t192 were all discharged by green re-runs; t126 is additionally structural (poc SKIPs both touched stages).
- Relationship to #502 (unit-kind artifact pruning): that feature generalizes this exemption; this PR is the minimal engine-level correction and composes with it.


leandrodamascena (approved):
Traced the produces readers and the coverage path. `unitCovered` reads `node.produces` only (so a backend unit skipping frontend-components is covered), and everywhere the path matters — `resolveProduces`, `producersOf`, registry, the required-sections sensor — both lists are unioned. Both touched stages keep 3-4 required produces after the move, and nothing downstream consumes the two conditional artifacts. No missed reader.

One behavior change worth a doc line: a unit that genuinely should write a conditional artifact (a UI unit omitting frontend-components) now counts as covered, and no deterministic gate catches it. Intended tradeoff — conditionality now rests on the reviewer + human gate, not the engine — just don't want someone assuming the engine still guards it.

Non-blocking: the validator doesn't reject the same artifact appearing in both `produces` and `optional_produces`; if that happened it'd stay coverage-required AND `resolveProduces` would emit its path twice (no dedup). Shipped config doesn't hit it.

PR #510, opened 2026-07-05
fix: backfill a skipped gate rejection at approve time (2.2.11)

### LINKED ISSUE #442: [Bug]: Conductor takes a gate decision but skips the engine verb that records it (state/audit under-records the action)
### Description

In live runs, the conductor sometimes performs the visible UI action a gate or step calls for but does not dispatch the underlying engine verb that records it, so the on-disk workflow state diverges from what the user just saw happen. The state machine and the tools are correct; the gap is the conductor (the LLM) skipping a tool call it was supposed to make.

The clearest instance is the **Request Changes** path. The stage protocol says selecting "Request changes" at an approval gate must run `aidlc-state.ts reject <slug> --feedback "..."`, which emits `GATE_REJECTED` + `STAGE_REVISING`, flips `[?]` -> `[R]`, and increments `Revision Count` (`core/tools/aidlc-state.ts` `handleReject`, and `core/aidlc-common/protocols/stage-protocol.md` §"Request Changes"). In a live Claude TUI run the conductor selected "Request changes" at the gate, reworked the stage in place, re-presented it, and reached `Completed` -- but never ran `reject`, so `Revision Count` stayed `0` and no `GATE_REJECTED`/`STAGE_REVISING` audit pair was written. The visible behavior (a revision happened) and the recorded state (no revision ever occurred) disagree.

This is a general class, not a single test: the same "did the visible thing, skipped the engine verb" pattern shows up on the Kiro ACP and Codex exec harnesses in the multi-repo Construction journey, where the conductor resolves the per-repo codekb target (`codekb-path --repo ...`) but then drifts to a bare `next` or runs the stage's question/memory ritual instead of writing the per-repo codekb artifacts, returning `rc=0` with the workspace mutation incomplete.

Related but distinct: #304 (agents skipping approval gates / tests) and #370 (conductor can bypass enforcement via direct `aidlc-state.ts` calls) are about *bypassing* or *skipping* gates; this is the conductor *taking* the gate decision but omitting the engine verb that persists it, so it is a fidelity/forwarding-reliability gap rather than a missing enforcement gate.

### Impact

- On-disk state and audit trail under-record what happened (e.g. a revision cycle leaves `Revision Count: 0` and no `GATE_REJECTED`/`STAGE_REVISING` events), which corrupts replay, outcomes packs, and any downstream consumer that reads the audit trail as ground truth.
- It is a live-conductor reliability gap (most visible under load / long context), so it surfaces as intermittent red on the live test tiers that assert the on-disk side-effect of a verb rather than just the terminal state.

### Steps to Reproduce

1. Run a workflow to an approval gate on the live Claude TUI.
2. At the gate, select "Request changes", let the conductor rework and re-present, then approve through to completion.
3. Inspect `aidlc-state.md`: `Revision Count` is still `0`, and the audit trail has no `GATE_REJECTED` / `STAGE_REVISING` pair, even though a revision visibly occurred.

(The committed live test `tests/e2e/t-tui-t139-revision-loop-idempotency.serial.test.ts` encodes exactly this as a vacuous-pass guard: `expect(revised.revisionCount).toBeGreaterThan(0)`. It reproduced 2/2 in a recent live run. The Kiro-ACP `t-acp-kiro-journey-workspace` and Codex-exec `t-exec-codex-journey-workspace` journeys show the multi-repo codekb variant of the same drift.)

### Expected Behavior

When the conductor takes a gate decision or a step that has a corresponding engine verb, it should dispatch that verb (e.g. `aidlc-state.ts reject` on Request Changes; the per-repo `codekb` write on multi-repo reverse-engineering), so the on-disk state and audit trail match the action taken. Ideally the framework should not rely on the LLM remembering: a deterministic backstop (a hook or an engine-side check) should detect "a revision was presented but no `reject` was recorded" / "the stage completed but its declared per-repo artifacts are absent" and either record the verb or refuse to advance, the same way the artifact guard (#366) makes stage completion contingent on evidence.

### Possible Directions (not prescriptive)

- A deterministic check that a `[R]`-revising transition was actually recorded before a stage can re-enter the gate and complete (analogous to the #366 artifact guard, which is the precedent for making a transition contingent on on-disk evidence).
- Tighten the conductor prose in `SKILL.md` / `stage-protocol.md` so the Request-Changes and per-repo-codekb steps are unambiguous "run THIS verb" instructions rather than describable-in-prose actions.
- Investigate whether the per-harness forwarding seams (Kiro ACP `$ARGUMENTS` handling, Codex exec) drop or mis-route the verb under load.

### AI-DLC Version

v2 (observed on the `v2` base, 2.1.x).

### AI-DLC Phase

Any gated stage (most visible at approval gates with Request Changes, and in multi-repo Construction reverse-engineering).

apackeer:
Fixes #442 (the Request Changes half): the conductor sometimes performs a revision round visibly - presents Request Changes, revises the artifact, re-presents the gate - but skips the `reject` verb that records it, so Revision Count stays 0 and the audit trail shows a clean single-pass approval for a stage the user watched get revised. The multi-repo codekb drift also described in the issue looks like a different mechanism and stays out of scope here.

## Detection

The audit ledger already carries the fingerprint. The PostToolUse audit-logger emits ARTIFACT_CREATED/ARTIFACT_UPDATED (with the file path) for every agent write under the record root on all four harnesses, and the presence hooks record a HUMAN_TURN per real human message. A revision-without-reject is exactly: a STAGE_AWAITING_APPROVAL for the stage, then a HUMAN_TURN (the human responding at the gate), then an artifact write matching one of the stage's declared produces files, with no GATE_REJECTED recorded anywhere after the gate opened.

The HUMAN_TURN pivot is what keeps the reviewer's legitimate `## Review` append out of the window: per the stage protocol the reviewer writes before the human responds at the gate, in either gate-start ordering. Learnings-ritual memory.md writes are excluded by the produces-file filter; human hand-edits in an editor emit no tool events at all; codekb stages are excluded entirely. Fail-open on an empty ledger, a missing gate-open anchor, or a missing human turn.

## Enforcement: reconciliation, not refusal

When `approve` sees the fingerprint (inside the same locked transaction, after the artifact and presence guards), it backfills the skipped pair: Revision Count increments, GATE_REJECTED and STAGE_REVISING are emitted tagged `Recovered: true` (the same tag `reject` already uses when it backfills a missing gate-open), plus a Recovered gate re-entry row, and then the approve proceeds normally.

Refusal was considered and rejected: forcing a retroactive `reject` writes a gate resolution, after which the immediate re-approve would be refused by the human-presence gate - the human would have to re-type an approval they already gave, and a false positive would hard-block a legitimate flow. Reconciliation never blocks; a residual false positive costs one visibly-tagged revision row.

Skipped under autonomous Construction (no human at the gate by design) and behind `AIDLC_SKIP_REVISION_BACKSTOP=1` (also set suite-wide in the test runner so no existing test changes behavior).

## Tests

New `tests/unit/t205-gate-revision-backstop.test.ts` (8 scenarios, seeded through the real audit CLI and the real dist audit-logger hook so the File field shape is production-faithful): the bug flow backfills with correct event ordering; a clean single pass does not fire; the reviewer-append false positive does NOT fire (the critical discriminator); a recorded reject suppresses; memory.md writes are filtered; autonomous mode and the off-switch are honored; a never-opened gate stays silent (the accepted false negative).

Adversarially reviewed: APPROVE across all six dimensions - predicate ordering matches the presence-gate idiom, no double-fire on approve retry (a half-emitted backfill self-suppresses via the recorded GATE_REJECTED), the synthetic rows cannot confuse later presence checks, and the endsWith pattern covers both the per-phase and per-unit artifact layouts. The review's one doc fix (the STAGE_AWAITING_APPROVAL emitter row) is folded in.

Version 2.2.9; other in-flight branches hold nearby numbers and whichever merges later re-bumps on rebase.


core/tools/aidlc-state.ts
```diff
@@ -157,6 +157,136 @@ function hasStageAuditEvent(
...
+  });
+  // 1. Anchor: the LAST gate-open for this slug (the currently open gate).
+  let anchor = -1;
+  for (let i = 0; i < events.length; i++) {
+    if (
+      events[i].event === "STAGE_AWAITING_APPROVAL" &&
+      events[i].stage === stage.slug
+    ) {
+      anchor = i;
+    }
```
leandrodamascena, line 286: This picks the LAST `STAGE_AWAITING_APPROVAL` for the slug with no filter on `Recovered`. That's the false-negative: when the conductor skipped the optional organic gate-start, the stage is `[-]`, so report auto-injects `gate-start --recovered` before approve (orchestrate.ts:2776) as a separate spawn. Its synthesized AWAITING row gets a fresh timestamp *after* the revision and human turns, becomes the anchor here, and the conjunct-3 scan from `anchor+1` then finds no HUMAN_TURN (`firstHuman === -1`) — so the predicate returns false and the revision goes unrecorded. Skip `Recovered=true` rows when choosing the anchor, or anchor on the real gate boundary (prior GATE resolution / STAGE_STARTED), so the pre-anchor human turn + revision stay in the window.

leandrodamascena (changes requested):
I tested this and found a false negative that misses the common case of the bug it's fixing.

The setup: a conductor revises the artifact at the gate but skips the `reject` verb. The backstop is meant to catch that at approve time — it anchors on the last `STAGE_AWAITING_APPROVAL` for the stage, then looks for a human turn + a revision write after it.

What I hit: the organic `gate-start` is optional, and a conductor that skips `reject` usually skips that too, so the stage sits `[-]` in-progress. When approve runs on a `[-]` stage, the report handler auto-injects `gate-start --recovered` right before it (orchestrate.ts:2776), as a separate process — so that synthesized `STAGE_AWAITING_APPROVAL` gets a fresh timestamp and lands *after* the revision and the human turns. It becomes the anchor, the scan after it finds no human turn, `firstHuman === -1`, predicate returns false. Revision Count stays 0, no GATE_REJECTED — exactly the unrecorded revision this was supposed to fix.

I ran it both ways on the real tools: no organic gate-start -> Count 0 / rejected 0; with one -> Count 1 / rejected 1. So the fix only fires when the conductor happened to run the optional gate-start before revising; in the skip-everything case it does nothing.

Fix direction: don't let a `Recovered=true` gate-start be the anchor (it's synthesized at approve time, not the real gate), or anchor on the real gate boundary so the human turn and revision stay in the window. Add a t205 case for the no-organic-gate-start flow.

Two non-blocking notes: t205's 8 scenarios all use a single shard where append order already matches timestamp order, so `events.sort()` never does anything — a two-shard case would actually exercise it. And the 3 backfill emits aren't atomic (a throw on the 2nd leaves an orphan GATE_REJECTED with Count 0), but it fails safe on retry and matches the existing audit-first pattern, so not from this PR.

apackeer:
Good catch - reproduced exactly as you described: the synthesized `gate-start --recovered` row postdates the human turns and the revision write, so it emptied the window and the backstop never fired on the skip-everything flow.

Fixed in the latest push (re-bumped 2.2.11 on rebase):

- The anchor now skips `Recovered: true` rows entirely: it is the LAST organic `STAGE_AWAITING_APPROVAL` for the slug, falling back to the LAST `STAGE_STARTED` (the current run's boundary) when the conductor never opened the gate at all.
- The stage-start fallback carries one extra guard: a produces write must also precede the first post-anchor `HUMAN_TURN`. Without a recorded gate-open, that pre-human write is the evidence separating "human reacted to produced work" from mid-stage coaching (human speaks before anything was produced, conductor then produces once) - coaching must not bump Revision Count.
- t205 scenario 9 covers your exact flow (STAGE_STARTED, produce, human, revise, human, `gate-start --recovered`, approve -> Count 1 / rejected 1); proven to fail on the pre-fix tool (Count 0). Scenario 10 pins the coaching guard.
- Your non-blocking note on the single-shard sort: scenario 11 now splits the ledger across two shards with a lexically-earlier clone shard holding the chronologically-later events, so the (Timestamp, position) sort is actually load-bearing; it also fails on the pre-fix tool.

The 3-emit atomicity note I left as-is per your assessment (fails safe on retry, matches the audit-first pattern).

leandrodamascena (approved):

PR #520, opened 2026-07-06
test: tolerate legitimate greenfield RE-skip in workspace-journey codekb beat

### LINKED ISSUE #441: [Test] t-exec-codex-journey-workspace per-repo codekb beat is fragile: helper checks wrong codekb path + shared fixture classifies Greenfield (skips RE)

apackeer:
Fixes #441 (the remaining half; the codekbFiles helper-path half already landed on v2 via the codekb-placement test fixes).

## What

The three workspace-journey legs (SDK integration, Kiro ACP e2e, Codex exec e2e) all assert their reverse-engineering beat wrote per-repo codekb for both sibling repos. That assertion assumes RE always runs, but the beat has two valid outcomes:

- Brownfield classification: RE executes and writes per-repo codekb; the existing asserts hold.
- Greenfield classification: the engine stamps RE SKIP at intent birth (recorded in the state file's Stages to Skip row), no codekb is ever produced by design, and the old asserts red on a correct skip. That is the flake: the fixture seeds each sibling repo with a tiny stub and live classification is non-deterministic, and the deterministic scanner only reads the workspace root (never the sibling repo dirs), so the live classifier is the deciding voice today.

Each leg now reads the born intent's state file via a shared greenfieldReSkip helper (Project Type is Greenfield AND the Stages to Skip row names the reverse-engineering slug) and branches: greenfield accepts the recorded skip, brownfield runs the two codekb asserts unchanged. A genuine brownfield RE failure still reds.

The ACP leg's driveCodekbUntilBothRepos poller also gains a second session/cancel arm: it fast-cancels the moment a greenfield RE-skip is on disk, so a legitimate skip no longer burns the full codekb budget (about 20 minutes of live Kiro credits) waiting for stores that will never appear.

## Interaction with the nested-workspace detection PR

The fix is valid in both merge orders relative to the depth-1 nested-scan fallback (PR #499): before it, the greenfield branch is the common path; after it, the fixture deterministically classifies Brownfield and the codekb asserts are the common path. Both branches stay live either way.

## Testing

Test-only change; no engine, fixture, or version edits (no version bump per the Changelog Policy).

- typecheck (both tsconfigs): clean
- smoke+unit (t177 fixture pin included): PASS, 2303 assertions, 0 failed
- deterministic e2e (filter ^t[0-9]): PASS, 129 assertions, 0 failed
- live SDK leg (t-journey-workspace.sdk): PASS, 1008s
- live Kiro ACP leg (AIDLC_KIRO_ACP_LIVE=1, t-acp-kiro-journey-workspace): PASS, 279s
- live Codex exec leg (AIDLC_CODEX_EXEC_LIVE=1, t-exec-codex-journey-workspace): PASS, 620s

### DIFF
--- a/tests/e2e/t-acp-kiro-journey-workspace.serial.test.ts
+++ b/tests/e2e/t-acp-kiro-journey-workspace.serial.test.ts
@@ -62,6 +62,7 @@ import { existsSync, readdirSync, readFileSync } from "node:fs";
 import { join } from "node:path";
 import {
   activeSpace,
+  getField,
   listIntents,
   readIntentRegistry,
 } from "../../dist/claude/.claude/tools/aidlc-lib.ts";
@@ -133,6 +134,29 @@ function activeRecordDir(root: string): string | undefined {
   return listIntents(root, activeSpace(root)).find((i) => i.active)?.dirName ?? undefined;
 }
 
+/** The RE codekb beat has TWO valid outcomes, and both keep the multi-repo journey
+ *  intact. Brownfield: reverse-engineering EXECUTEs and writes a per-repo codekb
+ *  store, so the codekbFiles asserts below hold. Greenfield: the engine stamps
+ *  reverse-engineering SKIP at intent birth (aidlc-utility.ts records it in the
+ *  Stages to Skip row), so no codekb is written and the asserts must not run (the
+ *  skip is the correct behaviour, not a flake). This reads the born intent's
+ *  aidlc-state.md and returns true only for that greenfield RE-skip: Project Type is
+ *  Greenfield AND the Stages to Skip row names the reverse-engineering slug. We match
+ *  the bare slug, never the row's human annotation (the engine writes it with an
+ *  em-dash phrase we deliberately avoid touching). A genuine brownfield RE failure
+ *  still falls through to the asserts and reds, as it should. */
+function greenfieldReSkip(recordDir: string): boolean {
+  let content: string;
+  try {
+    content = readFileSync(join(recordDir, "aidlc-state.md"), "utf-8");
+  } catch {
+    return false;
+  }
+  const projectType = (getField(content, "Project Type") ?? "").toLowerCase();
+  const stagesToSkip = getField(content, "Stages to Skip") ?? "";
+  return projectType === "greenfield" && stagesToSkip.includes("reverse-engineering");
+}
+
 /** Drive the reverse-engineering `--single` codekb turn and cancel it the moment
  *  BOTH repos' per-repo codekb has landed on disk — a DISK-CONDITION stop, not a
  *  tool-title one. A title stop (`/codekb/repo-b/…/`) is unreliable here: the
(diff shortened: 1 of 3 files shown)


leandrodamascena (approved):
APPROVE

Test-only, and the greenfield branch is a documented deterministic accept, not a coverage hole.

The key check: `Project Type` and `Stages to Skip` are written at intent birth by `aidlc-utility.ts` (`detectWorkspace`), not by the live LLM, so `greenfieldReSkip` reads a fixed verdict. A real brownfield run keeps Project Type=Brownfield → the else-branch runs the codekb asserts and reds on a missing store, so real failures still catch. The engine writes `2.1 (reverse-engineering — greenfield)`, and `.includes("reverse-engineering")` matches the bare slug with no other slug as a substring — no false positive. ACP poller cancels once (the `done` guard). No version bump, right for test-only.

Non-blocking:
- `expect(greenfieldReSkip(recordADir)).toBe(true)` inside its own `if` (sdk:254, acp:295, codex:280) is a tautology. Immaterial since the journey asserts run unconditionally, but if you want teeth there, assert the RE checkbox carries the SKIP suffix.
- `greenfieldReSkip` is copy-pasted across the 3 files — a shared home in `tests/harness/` would avoid drift. Follow-up.

PR #522, opened 2026-07-06
fix: quote $CLAUDE_PROJECT_DIR in Claude hook and statusline commands (2.2.19)

### LINKED ISSUE #519: [Bug]: Hooks break when the workspace path contains spaces (unquoted `$CLAUDE_PROJECT_DIR`)
### Description

| ID | Sev | Finding | Impact | Fix effort |
|----|-----|---------|--------|-----------|
| **F-2** | 🔴 High | Hooks use **unquoted** `$CLAUDE_PROJECT_DIR` — breaks when the workspace path has a space **and** hooks run under bash | Every hook can fail (`Module not found`); noisy, and may suppress sensor firings | Trivial — quote the var in 11 settings.json registrations |

### Steps to Reproduce

**Severity:** High — every hook can fail on turn events; the Stop hook errors on
every turn-end. Manifests when the project directory name contains a space
(this workspace: `AIDLC Harness Dryrun`) **and** the hook runner is bash — see
the shell-conditionality note below.

**Symptom (observed):**
```
Stop hook error: Failed with non-blocking status code:
error: Module not found "/Users/alok/Development/projects/AIDLC"
```
The path is truncated at the first space — the real path is
`/Users/alok/Development/projects/AIDLC Harness Dryrun`.

**Where:** `.claude/settings.json` — every hook command and the statusline are
registered as:
```
bun $CLAUDE_PROJECT_DIR/.claude/hooks/<hook>.ts
```
`$CLAUDE_PROJECT_DIR` is **unquoted**. Under a word-splitting shell, `bun` gets
`/Users/alok/Development/projects/AIDLC` as its module arg and the rest
(`Harness`, `Dryrun/.claude/hooks/…`) as stray args → "Module not found …/AIDLC".

**Affected registrations (grep-verified in settings.json):** **11 total** — 10
command strings (1 `statusLine.command` + 9 hook commands: session-start,
session-end, audit-logger, sensor-fire, sync-statusline, runtime-compile,
validate-state, log-subagent, stop) **plus** the 1 `permissions.allow` glob
`Bash(bun $CLAUDE_PROJECT_DIR/.claude/tools/*)`, which has the same unquoted shape.

**Why other tool calls still work:** the orchestrator's own Bash calls in this
session use a hard-coded absolute path via `cd "…/AIDLC Harness Dryrun"` with
quotes, so they're unaffected. Only the harness-invoked hook commands (which the
framework wrote unquoted) are exposed.

**⚠️ Shell-conditional (2026-07-06 validation, read-only):** running the *same*
unquoted invocation with the real space-containing path in each shell shows the
failure fires only under **bash**:
```
zsh 5.9 : arg=[/Users/alok/Development/projects/AIDLC Harness Dryrun/.claude/hooks/aidlc-stop.ts]   ← intact, no split
bash    : arg=[/Users/alok/Development/projects/AIDLC]  arg=[Harness]  arg=[Dryrun/.claude/hooks/…]  ← split → Module not found
```
This machine's `$SHELL` is zsh, which does **not** word-split unquoted variables
— so whether F-5 actually breaks hooks in a given run depends on the shell the
Claude Code hook runner uses, not just the presence of a space.

**Possible knock-on (conditional):** if hooks *do* run under bash, the failing
PostToolUse `aidlc-runtime-compile` hook would explain why the runtime-graph had
to be compiled manually (F-3) and the `0` sensor firings. This does **not**
explain F-1, which is proven independent (see F-1). Do not attribute all
anomalies to the folder space.

**Suggested fix (NOT applied — recorded only):** quote the variable in every
hook/statusline command in settings.json:
```
bun "$CLAUDE_PROJECT_DIR/.claude/hooks/aidlc-stop.ts"
```
and the permission glob `Bash(bun "$CLAUDE_PROJECT_DIR/.claude/tools/"*)`. The
framework's shipped settings.json should quote these by default so paths with
spaces are safe under every shell. Alternative workaround: use a workspace path
with no spaces.

### Expected Behavior

as above

### Actual Behavior

as above

### AI-DLC Version

v2 (alpha)

### Release / Commit

aidlc 2.1.4

### AI-DLC Phase

### Platform / IDE

Claude Code

### AI Model

_No response_

### Environment

**Framework:** `aidlc 2.1.4` · **Scope:** `mvp` · **Progress (as of 2026-07-06):**
12/22 stages, phase INCEPTION · **Platform:** macOS 26.5.1, arm64, shell **zsh 5.9**

### Additional Context

_No response_

apackeer:
Fixes #519

## Problem

A workspace path containing spaces broke every hook under bash-backed hook runners: `$CLAUDE_PROJECT_DIR` is unquoted in the shipped Claude `settings.json`, so `bun $CLAUDE_PROJECT_DIR/.claude/hooks/<hook>.ts` word-splits at the first space and fails with `Module not found .../AIDLC` (truncated path). The issue's shell validation shows zsh does not split but bash does, so the failure is shell-conditional - the shipped settings should be safe under every shell.

## Fix

Quote all 13 `$CLAUDE_PROJECT_DIR` registrations in `harness/claude/settings.json` (the authored source): 11 hook commands, the statusline command, and the pre-approved bun-tools permission glob. The glob keeps the `*` outside the quotes (`Bash(bun "$CLAUDE_PROJECT_DIR/.claude/tools/"*)`) so the permission matcher still globs.

`dist/` regenerated via `bun scripts/package.ts`; only `dist/claude/.claude/settings.json` and the per-harness `aidlc-version.ts` copies change (kiro/kiro-ide/codex settings do not carry `CLAUDE_PROJECT_DIR`).

## Tests

- NEW `tests/unit/t199-claude-project-dir-quoting.test.ts`: regression pin - walks command fields and permission entries in BOTH the authored and generated Claude settings and rejects any unquoted `$CLAUDE_PROJECT_DIR` (bug shape `/(^|[\s(])\$CLAUDE_PROJECT_DIR\b/`); also pins the exact permission-glob shape and the 13-reference count.
- `tests/integration/t40-settings-hook-config.test.ts`: T5 updated to expect the quoted bun-tools glob.

## Verification

- smoke+unit: 130 files / 2305 assertions / 0 failed
- t40, t199, t68 (version-changelog sync), t145 packaging parity, `package.ts --check`: all green
- Quoted form proven to survive bash word-splitting with a spaced path (`/tmp/path with spaces` resolves intact)

## Release mechanics

2.2.14 (open PRs claim through 2.2.13): `core/tools/aidlc-version.ts`, README badge, and CHANGELOG entry with the re-copy upgrade instruction, same commit per changelog policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.2.19] - 2026-07-09
+
+Hooks and the statusline no longer break when the project path contains spaces. Every `$CLAUDE_PROJECT_DIR` reference in the shipped Claude settings is now double-quoted, so the hook commands survive word-splitting shells (bash) as well as zsh. **Upgrade:** re-copy `dist/claude/.claude/settings.json` into your project (or re-copy the whole `dist/claude/` shell).
+
+* Fixed: all 11 hook commands, the statusline command, and the pre-approved bun-tools permission glob in the Claude `settings.json` quote `$CLAUDE_PROJECT_DIR` (`bun "$CLAUDE_PROJECT_DIR/.claude/hooks/<hook>.ts"`); previously a workspace path with a space made every hook fail with `Module not found` under bash-backed hook runners (#519).
+
 ## [2.2.18] - 2026-07-09
 
 Unit kinds prune the per-unit construction design matrix. Tag each Unit of Work in units-generation's edge block with an optional `kind:` (service, spec, ui, packaging, or library) and the four construction design stages (functional-design, nfr-requirements, nfr-design, infrastructure-design) now emit and require only the artifacts that apply to that kind: a spec unit owes no scalability doc, a packaging unit no business-logic model. A unit with no kind, a stage with no per-kind map, or an artifact left unannotated all behave exactly as before (the full matrix), so existing workflows are unchanged. A unit whose required set prunes to empty is covered by definition (the stage does not apply to it). **Upgrade:** re-copy your `dist/<harness>/` shell into the project; an older engine run against a kind-tagged edge block fails the units-generation gate loudly (re-copy dist to fix) rather than pruning wrong.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.2.18-blue)
+![version](https://img.shields.io/badge/version-2.2.19-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-version.ts
+++ b/core/tools/aidlc-version.ts
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.2.18";
+export const AIDLC_VERSION = "2.2.19";
--- a/dist/claude/.claude/settings.json
+++ b/dist/claude/.claude/settings.json
@@ -7,7 +7,7 @@
       "Read",
       "Edit",
       "Write",
-      "Bash(bun $CLAUDE_PROJECT_DIR/.claude/tools/*)",
+      "Bash(bun \"$CLAUDE_PROJECT_DIR/.claude/tools/\"*)",
       "Bash",
       "Glob",
       "Grep",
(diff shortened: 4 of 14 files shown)


leandrodamascena (changes requested):
The quoting itself is right — wrapping `$CLAUDE_PROJECT_DIR` in double quotes stops the bash word-split on a spaced path and doesn't change behaviour under any shell or for a space-free path. I checked the subtle case (quoting breaking the permission matcher) and it's fine: the glob became `Bash(bun "$CLAUDE_PROJECT_DIR/.claude/tools/"*)`, the quotes sit inside the pattern, and a real invocation still matches, so auto-approve still works. t199 fails on both authored and generated settings if either reverts to unquoted, so the pin is real.

One thing to fix before merge: the docs still show the unquoted form. `docs/guide/13-customization.md` has the old `Bash(bun $CLAUDE_PROJECT_DIR/.claude/tools/*)` on line 160 and in the prose on line 166, plus an unquoted hook command on line 138 (`bun $CLAUDE_PROJECT_DIR/.claude/hooks/aidlc-statusline.ts`). Anyone copying those reintroduces exactly the bug this fixes, and the repo Documentation Policy (AGENTS.md) says to grep `docs/` and update stale references in the same commit. Please quote all three and grep the rest of `docs/` for any other `$CLAUDE_PROJECT_DIR`.

Non-blocking: the new test file is `t199-claude-project-dir-quoting.test.ts`, but `t199-learnings-memory-path.test.ts` already ships on v2 (from #492). Nothing keys on the t-number so both run fine, but rename it for readability — the next free number is t218 (v2 is up to t212 and the open series PRs already claim t213-t217), not t213 as it might first look.

apackeer:
Both addressed in f2481ed5. Quoted all the doc examples: the three you flagged in docs/guide/13-customization.md plus three more instances a full docs/ grep turned up in docs/reference/14-claude-features.md (statusline, session-start, session-end examples), and added a sentence to the customization guide explaining why the `*` sits outside the quotes. Remaining $CLAUDE_PROJECT_DIR doc mentions are prose about the variable, not copyable commands, so left alone. Also renamed the test to t218-claude-project-dir-quoting.test.ts (confirmed t213-t217 are claimed by open PRs) and updated the t40 cross-reference. smoke+unit 130/130 green, t40 green, package.ts --check clean.

leandrodamascena (approved):

PR #525, opened 2026-07-06
fix: self-heal a missing bolt_dag from the units-generation artifact (2.2.14)

### LINKED ISSUE #506: [Bug]: AI-DLC v2 Construction silently builds only the first unit when the runtime graph has no `bolt_dag`
# AI-DLC Version
v2 (`v2` branch) — code confirmed on **2.2.0** (`core/tools/aidlc-orchestrate.ts`: `orderedUnits` at :1776 → `readBoltDagBatches` at :666; degrade branch in `emitPerUnitRunStage` at :1858-1862). Reproduced end-to-end on **Kiro IDE 1.0.89** (macOS Apple Silicon). The trigger on that harness is the empty-`tool_input` platform gap (kirodotdev/Kiro#6188, and the runtime-compile symptom I filed as #456), but the defect described here is in the **engine's failure mode**, independent of what caused the runtime graph to be stale.

# AI-DLC Phase
Construction — every per-unit stage (`functional-design`, `nfr-requirements`, `nfr-design`, `code-generation`) on a greenfield scope that ran `units-generation`.

---

## Summary

When a Construction per-unit stage runs but the runtime graph has **no `bolt_dag` node**, the engine does not fail, warn, or wait — it **silently degrades the per-unit loop to a single iteration**. On a multi-unit greenfield project this means only the **first unit** gets its design/code; every other unit is skipped, the stage gate fires as if the whole stage is done, and the workflow proceeds to completion. The user asked for a 4-unit feature and receives a walking skeleton, with **no error and no signal that units were dropped**.

This is a **silent-data-loss failure mode**, distinct from the hooks that populate the graph (#456) and from the platform payload gap (kirodotdev/Kiro#6188). Even after those are fixed, an engine that answers "no `bolt_dag` → build one unit and call the stage done" will silently truncate any multi-unit workflow whenever the graph is stale for *any* reason (a missed recompile, a compile error, a crash between `units-generation` and Construction, a hand-edited graph).

## Root cause

`emitPerUnitRunStage` reads the ordered unit list off the compiled `bolt_dag`, and treats an **empty list as a benign "no units" case**:

```ts
// core/tools/aidlc-orchestrate.ts  (2.2.0)

function orderedUnits(projectDir: string): string[] {
  const batches = readBoltDagBatches(projectDir);   // :1777  reads runtime-graph.json → bolt_dag.batches
  if (!batches) return [];                          //        no graph OR no bolt_dag node → []
  return batches.flat();
}

// ...in emitPerUnitRunStage:
// "No compiled unit DAG (a scope that SKIPs units-generation, refactor /
//  security-patch / infra / bugfix / poc, or a pre-compile moment): degrade to
//  today's single {unit-name} directive. Zero behaviour change off this path."
const units = orderedUnits(projectDir);             // :1861
if (units.length === 0) {                           // :1862
  emitRunStageForSlug(node.slug, ...);              // → ONE directive, {unit-name} unsubstituted, gate fires
  return;
}
```

The comment shows the degrade path was designed for scopes that **legitimately have no units** (`bugfix`, `poc`, a scope that SKIPs `units-generation`). For those it is correct and byte-identical. But the same `[]` is returned when the scope **did** run `units-generation` and *does* have a `unit-of-work-dependency.md` on disk — the graph just wasn't (re)compiled, so the `bolt_dag` node is absent. The engine cannot tell "legitimately no units" from "units exist but the graph is stale," and defaults to the silent-skip behaviour in both cases.

`readBoltDagBatches` returns `null` (→ `[]`) for a missing graph file, a missing `bolt_dag` node, **and** a malformed graph — all three collapse into the same benign-looking branch.

## Steps to Reproduce

1. Run a greenfield multi-unit workflow (e.g. an `mvp` scope) through `units-generation` so `inception/units-generation/unit-of-work-dependency.md` is written with ≥2 units.
2. Ensure the runtime graph is stale/missing a `bolt_dag` node at Construction entry. On **Kiro IDE 1.0.89** this happens on its own (the runtime-compile hook never fires — empty `tool_input`, #456). To force it deterministically on any harness: delete `runtime-graph.json` (or its `bolt_dag` node) after `units-generation` and before `functional-design`.
3. Enter Construction and run `functional-design`.

**Observed**: the `run-stage` directive has no `unit` field and its `produces` paths keep the literal `{unit-name}` placeholder. The stage produces artifacts for one unit only, the gate fires, and the next stage proceeds. `construction/` ends up with a single unit segment; the other units never appear. The workflow completes "successfully."

**Expected**: with `unit-of-work-dependency.md` present and describing N units, the engine must design/build all N units (per-unit iteration), or — if it genuinely cannot resolve the unit list — **stop with an error** rather than silently building one unit and reporting the stage complete.

## Why this is severe

- **Silent**: no error, no warning, no dropped-unit count. The audit log and gate look identical to a correct single-unit run.
- **Data loss**: the bulk of the requested work (all units after the first) is skipped. On my repro the app shipped with 1 of 4 units — add/list/persist only, with edit/delete, toggle, and filter/search/progress silently absent.
- **Determinism claim**: the per-unit fan-out is the engine's mechanism for covering a multi-unit plan deterministically. A failure mode that quietly reduces N units to 1 undercuts that guarantee — and it depends only on the runtime graph being stale, which can happen for many reasons beyond any one harness.

## Suggested fix (two independent options; ideally both)

1. **Distinguish "no units" from "units exist but uncompiled."** Before taking the degrade branch, check whether `inception/units-generation/unit-of-work-dependency.md` exists (or whether the scope executed `units-generation`). If it does and `bolt_dag` is absent, that is an error state, not a zero-unit scope — either recompute the DAG on the spot from the dependency file (it's a pure parse — `parseBoltDag` already exists), or emit an error/`ask` directive instead of silently degrading.
2. **Make the read side self-healing.** Have `readBoltDagBatches` (or `orderedUnits`) recompute the batch DAG from `unit-of-work-dependency.md` when the cached `bolt_dag` is missing, so the per-unit loop is independent of whether a hook refreshed the graph. This keeps the genuine zero-unit degrade (no dependency file → still `[]`) while removing the "units exist but graph is stale" trap.

Option 2 (recompute from the dependency file on cache-miss via the existing `parseBoltDag`) is sufficient on its own: with it, the per-unit loop covers all units on Kiro IDE 1.0.89 regardless of whether the hook refreshed the graph.

## Additional Context

- Trigger on my repro: kirodotdev/Kiro#6188 (empty `tool_input` on IDE hooks) → #456 (runtime-compile never fires) → stale graph → this silent degrade. But #456's fix (getting the hook to fire) only removes *one* way the graph goes stale; the engine's silent-skip response to a stale graph is the separate defect this issue is about.
- Related learnings-side symptom off the same stale graph: `aidlc-learnings.ts surface` fails with `stage "<slug>" not found in runtime-graph.json` because the graph's `stages[]` is also frozen (#455 territory).

apackeer:
Fixes #506.

## What

When runtime-graph.json was absent, malformed, or missing its bolt_dag node while `inception/units-generation/unit-of-work-dependency.md` existed on disk, the engine treated the workflow as unit-less: every per-unit Construction stage silently degraded to a single iteration, the stage gate fired, and a multi-unit plan shipped only its first unit with no error. The same stale graph also kept the autonomous swarm from ever firing and let an early `report --result approved` bypass the per-unit coverage guard.

The engine now resolves unit batches cache-first with a read-side self-heal (the issue's option 2): the compiled graph wins when present; on a miss the batches are recomputed from the dependency artifact via the same pure `parseBoltDag` the runtime compiler uses, so the heal is byte-equal to what compile would have written. All three consumers resolve through one helper:

- the inline per-unit loop (the issue's headline),
- the autonomous swarm trigger,
- the approve-time per-unit coverage guard (a bypass found during verification; a stale graph could previously let a partially built stage commit).

Three states, strictly separated:

- No dependency artifact on disk (bugfix, poc, refactor, security-patch, infra, or pre-2.7): the existing single-iteration degrade, byte-identical, no output.
- Artifact present and parseable: healed batches, plus a one-line stderr note so a chronically stale graph is visible (a heal mid-Construction means the runtime-compile hook chain failed).
- Artifact present but malformed or cyclic: a loud error directive naming the artifact and the parse failure, instead of healing to a wrong DAG or silently building one unit (mirrors the compile side's refusal to emit a wrong-but-valid node).

The heal never writes runtime-graph.json: `next` stays read-only and the compile hook repairs the cache on the next transition.

## Tests

New unit test (t215, 11 cases): heal on missing graph and on graph-without-node, byte-identical degrade with no artifact, loud cyclic and dangling errors, cache precedence over a malformed artifact (no stderr on the healthy path), approve-guard closure, swarm heal, healed iterate-and-settle across both units, and non-per-unit stages never triggering the read side. The test spawns the real engine, so it is registered in the coverage tooling's none-to-cli spawner pin per that pin's own instructions.

Verified green: typecheck; t215 (11/11, heal stderr notes observed in the per-file log); full smoke+unit (130 files, includes t68 version sync and the coverage-registry freshness gates); full integration (99 files; two parallel-tier live flakes re-ran green alone); `bun scripts/package.ts --check` clean on all four harnesses.

Version 2.2.14 with matching CHANGELOG entry and README badge. Sibling PRs claim 2.2.1 through 2.2.13 (2.2.9 twice); whichever merges later re-bumps on rebase.

## Cross-PR note

The swarm-batch-advance PR (#497) rewrites the same two lines in tryEmitSwarm this PR replaces and inserts its helper at the same location after readBoltDagBatches. Semantically clean in either merge order (its converged-batch walk consumes the resolved batches); whichever merges second resolves the adjacency by hand.

Implementation by gpt-5.5 via Codex CLI against a reviewed design; engine diff, tests, and all gates re-verified by the maintainer.


leandrodamascena (approved):
I ran a few construction cycles with a missing/stale bolt_dag and it heals the way #506 wants — malformed surfaces as a loud error at all three live consumers (`emitPerUnitRunStage`, `tryEmitSwarm`, the approve-guard), never a silent single-unit degrade. I went looking for the relocated bug (`orderedUnits` returning `[]` on non-ok) but that function has zero callers now, so it's unreachable. The heal recomputes via the same `parseBoltDag` the compile hook uses, and a non-empty cache short-circuits and wins — the right #506 boundary.

Two cleanup items, neither blocking:
- **Delete `orderedUnits` (~line 1831)** — dead code now, and it's the function still carrying the malformed->`[]` trap for a future caller.
- **Resolve the #497 merge deliberately.** This branch's `tryEmitSwarm` is the pre-#497 `batches[0]` version; v2 has #497's converged walk. Keep v2's version and just swap its batch read for `resolveBoltBatches` — don't take this PR's `tryEmitSwarm` verbatim. A botched merge fails typecheck (`pendingUnits` undefined) rather than silently reverting #497, so the gate catches it, but resolve it on purpose.

Issue #517, opened 2026-07-06
[RFC]: Bring product discovery into the Ideation phase: an opt-in design-thinking flow that explores and validates an idea before building


akash-doshi:
### Summary

Ideation today accepts unverified answers, validates them only after commitment, burdens reviewers at identical weight regardless of stakes, and starts its record after the asks that formed the initiative are gone. This RFC proposes a second implementation that embeds design thinking in the phase's native machinery of stages, artifacts, gates and skills. The flow builds an evidenced current-state picture, expresses the future state as at least three structurally distinct options, carries the thinking in two working records (open questions and assumptions), tests the riskiest assumptions first through cheap composed tests, lets each cycle's evidence reshape both pictures, and ends in commit, pivot or park over a decision pack that is a projection of the phase's own records, current without anyone assembling it. We extend the existing intent-capture stage in place, additively and for every scope that runs it (materials-first intake, source-labeled answers, mandate and decide-by capture, two new produced records), and ship four new compiled stages (current state, future state, experimentation, decision), one stock scope named `discovery`, framework-default artifact templates, and a portfolio view as a severable slice, almost entirely authored `core/` data. There is one intent stage: the flow enters through the stage every scope already runs, not through a parallel copy. The stages carry ordinary compiled frontmatter, so the 2.2.0 composer can EXECUTE or SKIP them per task. No stock scope gains or loses stages, and the six ideation files after intent-capture are untouched. Baseline: the v2 line at v2.2.10, HEAD 242953e. The implementation is PR #526 at version 2.2.11.

### Motivation

Every current Ideation stage has one shape: questionnaire, answers, markdown, approval. The questionnaire is valuable. The problems sit around it. The observations below come from a two-day retro with five enterprise teams on real work items, a small sample stated as such, and each problem is also visible in the stage files themselves.

1. **One knowledge instrument.** The phase never analyzes the code, design system, data or standards a team already has (`reverse-engineering` exists, reachable only in Inception). Trusted retro outputs came from existing assets. Untrusted ones were unsourced estimates, including an untraceable twenty-two week figure.
2. **Validation after commitment.** Every output is a document, so a claim first meets reality in Construction, the most expensive place to learn it. The first retro artifact that executed was wrong on first run, fixed only because running it exposed the error.
[...]

Issue #478, opened 2026-07-02
[RFC]: v2 - Author stage number/name in `core/`, no hand-editing per-harness `stage-graph.json`


teddyaryono:
### Summary

Adding or reordering a stage today can force a **manual, identical edit to
`stage-graph.json` in every harness tree** (`dist/claude/.claude/…`,
`dist/codex/.codex/…`, `dist/kiro/.kiro/…`, `dist/kiro-ide/.kiro/…`). This RFC
proposes moving the one datum that forces those edits — a stage's `number` (and
`name`) — into the hand-authored `core/` layer, so that adding a stage is
"drop the `.md` in `core/`, run `bun scripts/package.ts`, done" for **all**
harnesses at once.

### What already works

Contrary to first appearances, `scripts/package.ts` **already recompiles**
`stage-graph.json` for every harness. Step 3 of `buildTree()`
(`scripts/package.ts:319-338`) runs `bun tools/aidlc-graph.ts compile` inside
each assembled tree, regenerating both `stage-graph.json` and `scope-grid.json`
from the `core/` stage YAML. This is why `bun scripts/package.ts --check`
passes clean today even with a newly added stage: the four dist graphs *are*
what the compiler emits from `core/`.

So the **content** of a stage (its `produces`, `consumes`, `requires_stage`,
`scopes`, `sensors`, etc.) is already authored once in `core/` and fanned out
correctly. That part is not the problem.

### What doesn't work — the `number` / `name` seed

The problem is the **`number` / `name` seed**. `compileStageGraph()`
(`core/tools/aidlc-graph.ts:1215`) treats `number` and `name` as
*computed-not-authored*: on each compile it **harvests** them from the
*existing* `stage-graph.json` (`numberBySlug` / `nameBySlug`,
`aidlc-graph.ts:1223-1225`), and `seedCompiledData()` (`scripts/package.ts:237`)
reads that seed from **each harness's own committed dist JSON**. The auto-seed
path only ever **appends** a brand-new slug at the next free per-phase index and
**never renumbers** an existing stage (deliberate — runtime workflow state is
slug-keyed and must stay stable; see `aidlc-graph.ts:29-40`).

That design is correct for *appending* a stage at the end of a phase. It breaks
down for **inserting or reordering** a stage mid-phase, which is a legitimate
authoring act.

### Worked example — the `domain-glossary` insertion (the trigger for this RFC)

Adding `core/aidlc-common/stages/inception/domain-glossary.md` positioned at
[...]

PR #475, opened 2026-07-01
feat: AIDLC plugin mechanism (2.3.0)

harmjeff:
## Summary

Adds a first-class **AIDLC plugin mechanism** — an optional, owned set of stages, agents, scopes, method/rules, sensors, and *additive contributions to existing core stages*, authored in `plugins/<name>/` and emitted by the packager as a **real host plugin** per harness (`dist/plugins/<name>/{claude,codex,kiro,kiro-ide}/`).

A plugin never edits `core/`; with every plugin disabled an install is byte-identical to bare core. On install, a compose hook composes the plugin in — copying new stages and merging the contribution seam — then recompiles the graph so the orchestrator routes the new stages.

## How it works

- **Author once, harness-neutral.** `plugins/<name>/` mirrors `core/`'s shape (`stages/`, `contributions/`, `sensors/`, `tools/`, …) + a declarative `.aidlc-plugin/plugin.json`.
- **Emitted as a host plugin.** `bun scripts/package.ts` projects each plugin into `.claude-plugin/` / `.codex-plugin/` / `.kiro-plugin/`, with a harness-agnostic compose hook (`compose.ts`).
- **The contribution seam** additively modifies core stages: structural surfaces (`produces`/`consumes`/`sensors`/`required_sections`) set-union into the target stage; prose `fragments` splice into the stage body at named anchors. Additive-only, order-deterministic, idempotent, durable across recompiles. (Deferred surfaces — `adds.scopes`/`adds.requires_stage` merge, `agents`/`memory`/`knowledge` projection, `when:` evaluation — are documented as such in doc 18 §8.)
- **Hybrid distribution.** Install via the host's native `/plugin` commands where a store exists (Claude, Codex); folder-drop + explicit compose for Kiro. Trust is host-native — no AIDLC trust layer, no distribution infra.

## What's included

- **Schema:** `aidlc-stage-schema.ts` accepts `number`/`name`/`bundle`/`required_sections` (optional, shape-validated) + `when` (`producer-in-plan` predicate). Core stages omit them → byte-identical compile.
- **Packager:** discovers `plugins/<name>/` and emits a host projection per harness (claude, codex, kiro, kiro-ide), each derived from that harness's manifest; `--check` byte-parity drift-guards `dist/plugins/` too.
- **Reference plugin:** `plugins/test-pro/` (2 new stages + 4 contributions + 2 advisory sensors), validated across **all four harness projections**, doctor 0 failures.
[...]

PR #526, opened 2026-07-06
feat: bring product discovery into Ideation — an idea-to-decision flow that feeds Inception (2.2.11)

akash-doshi:
Implements the first slice of RFC #517 (Bring product discovery into the Ideation phase).

## Summary

Ideation today starts every initiative from a blank questionnaire, and nothing between the first answer and the first build checks whether the direction is right. This PR brings product discovery into the Ideation phase as an opt-in flow. It extends the existing intent-capture stage in place so every scope works materials-first, and adds four new ideation stages under a new `discovery` scope that explore and validate an idea before any commitment to build: evidence how things work today, frame what the future could look like as genuinely different options, test the riskiest assumptions first, and end in an explicit decision with a plain meaning — commit (build it), pivot (keep what was learned, reframe), or park (stop for now, keep everything, reopen any time).

## What changes, mechanically

1. **intent-capture (1.1), extended additively for every scope that runs it:** one intake question (asks, materials, mandates, decide-by dates) asked through the stage's own questions file; answers found in materials carry their source and stay unconfirmed until the person validates them at the gate; the intent statement gains three conditional sections; the stage produces two new records (`source-inventory`, `open-questions-record`). The clarifying questions themselves are unchanged; a stock run gains the intake question, the two additive records, and a confirm-unconfirmed-answers duty at the gate (the CHANGELOG states this plainly).
2. **Four new stages, discovery scope only:** `discovery-current-state` (1.8, ends in a playback the person confirms), `discovery-future-state` (1.9, at least three structurally different framings, the person chooses), `discovery-experimentation` (1.10, riskiest assumptions first, append-only evidence), `discovery-decision` (1.11, the decision pack and the decision).
3. **On commit the person chooses where the build happens, and the validated vision is wired into delivery rather than left to conductor initiative.** The decision stage appends a structured handoff contract to the pack (settled entries as pre-answers, exclusions as boundaries, open items with recommendations, promoted starting points, and a map from delivery's first questions to where the record answers them) and folds a "What Discovery Validated" section back into the intent statement. `requirements-analysis` and `user-stories` declare `decision-pack` as an optional consume (`required: false`), and `refined-mockups` declares `design-language-record` and `evidence-record` the same way — when the artifacts exist, the run-stage directive hands delivery their paths and upstream-coverage verifies they were referenced; when they do not (every stock run), the resolver drops the entries silently. Continue here relays the engine's own verbs (`scope-change`, `set-status`, `state skip` with a reason naming the decision pack) so the same workflow and record proceed into Inception — pinned by a deterministic test (t214) in both directions. Hand off completes the workflow and the pack travels to the receiving team. Both paths verified live.
4. One scope file (`discovery`, zero keywords, explicit selection only), nine framework-default templates (the first tier-2 templates to ship; `intent-statement` deliberately has none), a shared test-primitives knowledge file, a new user-guide chapter (Running a Discovery Initiative), glossary entries, and the full consistency surface (SKILL tables, stage graph 32 → 36, docs, tests). Version 2.3.11.

## Why this shape

One intent stage, not two — the same human moment should not have two implementations, and the shared entry is what makes native continuation into delivery possible. Materials-first because nobody should re-type what their documents already say; every borrowed answer carries its source and is confirmed by a person before it becomes fact. Gates where errors are cheapest. Three structurally different framings minimum, because variations of one idea are a rubber stamp. Riskiest-first testing, because the assumption most likely to kill the initiative should be the first money spent. The pack travels before the meeting. Opt-in with zero keywords, so nobody's routing changes by accident. The six questionnaire stages are untouched — replacement is a later, evidence-gated proposal (graduation criteria in the RFC), not a silent switch.

## Non-breaking evidence for the four stock scopes

- Every stock scope's compiled plan gains exactly four SKIP rows, with every other row byte-identical — nothing removed, nothing reordered (see the `mr9-parity` fixture diffs, the strongest single artifact in this PR). Stock walk fixtures are byte-identical.
- The engine's only hard artifact guard passes when any one declared produce exists, and sensors are advisory by contract, so no new hard-failure path exists.
- The clarifying questions, their artifacts' shapes, and the gate are unchanged. A stock run gains one intake question plus two additive records.
- `t213` pins the contract with exact pins: intent-capture keeps exactly its four stock scopes plus discovery, produces as an exact ordered list, the four new stages declare `scopes: [discovery]` only. `t214` pins the continuation verb sequence against a completed-discovery fixture (including a contrast test that `recompose` refuses the same state) and the optional-consume differential (pack present: the requirements-analysis directive carries its path; absent: no trace).

## Pairing with PR 2

This PR and the skills PR that follows are a pair: this one ships the flow's structure, the next ships the craft the agents load to execute each move deeply (framing generation from structurally different starting points, design-language capture, the five test moves as per-move playbooks, narrative generation). The structure runs on its own and is provably non-breaking, but the method's full depth arrives with the pair, so we ask that both be reviewed together and merged together — packaged however this repo prefers: one PR with two clean commits, two PRs with the second marked stacked, or a staging branch on your side.

## Change inventory and review guide

316 files; 187 are generated `dist/` copies (verify with `bun scripts/package.ts --check`, never by hand). The branch is five commits — the feature, then four review-response fix rounds — so the fixes from the first review are individually diffable. The authored surface, in review order:

| Read closely | Why |
|---|---|
| `core/aidlc-common/stages/ideation/intent-capture.md` | The one edit to a stage every scope runs. Diff it line by line: the four clarifying questions and gate text unchanged, every addition additive, every ask through the questions file |
| `tests/integration/t213-discovery-scope.test.ts`, `t214-discovery-continuation.test.ts` + `tests/fixtures/mr9-parity/*` diffs | The safety case |
| The four `discovery-*.md` stages | Frontmatter contracts; every named engine verb is real (`jump`, `park`, `scope-change`, `set-status`, `state skip`) |
| `core/scopes/aidlc-discovery.md` + nine `core/tools/data/templates/` | Template H2 sets are the sensor contract; `intent-statement` deliberately has no template |
| `core/knowledge/aidlc-shared/discovery-test-primitives.md` | The five test moves and their honesty caps |
| `docs/guide/18-discovery-walkthrough.md` | The user-truth check — what a first-time reader is promised |

Mechanical (verify only): the pinned stage numbers 1.8–1.11 in the four committed `stage-graph.json` files (per the pin-and-harvest practice in `docs/reference/15-stage-definition.md`), one-line scope tags on the three initialization stages, statusline/state-template/SKILL tables/welcome banner, the version trio (t68 guards it), comment-only count corrections in two sensor sources and `aidlc-graph.ts`. Docs: 29 files, mostly count/table consistency plus the new chapter and six glossary entries. Tests: 20 count-pin moves plus regenerated fixtures.

## Notes for maintainers

- We support RFC #478 (authoring stage numbers in core) — it would remove this PR's pinned-row edits.
- Found while building: the 2.2.0 CHANGELOG bullet saying a maintainer compile drops composed scopes appears stale — `mergeComposedScopes` preserves them (`aidlc-graph.ts:1255-1277`, wired at `:1505-1521`).
- The nine templates ride here because the sensor contract binds them to the stages; say the word and they split into their own PR.
- The RFC's portfolio view is deliberately absent here and severable.
- Small upstream gaps we noticed and did not touch (happy to file issues): the Phase Progress statusline can display stale after a completed stage; sensor outcomes are silent on the terminal (events land in the audit shard, findings under `.aidlc-sensors/`); `park` promises `/aidlc --resume`, which works but is missing from the help text; the state-file legend renders `[x]`-beside-SKIP confusingly after a scope change. One pre-existing doc error (`02-your-first-workflow.md` named a nonexistent artifact and five questions where the stage asks four) is fixed here since we touched the page anyway.

## Verification

`bun scripts/package.ts --check` green on all four harnesses. Deterministic suite: 329 files, 4,200 assertions, one known environmental failure (t92, a single type-check assertion, reproduced identically at the untouched v2.3.10 base in this environment; upstream's 2.3.10 eslint pin cured the other two). The three sdk-live tests that pin scope counts (t189, t192, t193) are updated for the tenth scope; they skip in this environment (no live credentials), so that fix is pin-correct by arithmetic rather than run-proven here — stated so this verification claim carries its environment scope. Coverage registry regenerated and fresh. Live smoke: discovery birth routes to intent-capture, the template-to-sensor round trip fires and fails correctly, a full run reaches the decision, the commit continuation proceeds into Inception in the same workflow, and the intent-statement write-back's sensor behavior was proven with a live hook run.


core/aidlc-common/stages/ideation/discovery-decision.md
```diff
@@ -0,0 +1,204 @@
...
+   `bun {{HARNESS_DIR}}/tools/aidlc-utility.ts set-status --stage <named stage>`.
+3. Mark every ideation stage the new scope expects but this run already
+   covered as skipped, so the record stays honest about why they will not
+   run: `bun {{HARNESS_DIR}}/tools/aidlc-state.ts skip <slug> --reason
+   "covered by the discovery run: see the decision pack and handoff
+   contract"` for each pending ideation stage.
+4. Hand control back to the forwarding loop (`next`). Delivery begins at the
+   stage the engine named, with the intent statement, stakeholder map, and
+   the whole discovery record already in place — the delivery path's entry
+   stages consume the intent statement this workflow produced at 1.1.
```
apackeer: This is the seam where the flow's central promise leaks. On the continue-here path the vision travels only as prose: no Inception stage consumes decision-pack, evidence-record, assumptions-record, or future-state (requirements-analysis declares only intent-statement and scope-document, both optional), and the intent statement being relied on here was written at 1.1, before any discovery ran - none of the four discovery stages writes back to it. So if the evidence narrowed or reframed the idea, delivery's entry stage is wired to read the day-one intent, and reaching the decision pack depends on the conductor's initiative. upstream-coverage cannot catch it because it only verifies declared consumes.

Two small fixes; either works, both is better:
- On commit, fold the chosen framing and recorded exclusions back into the intent statement before the scope-change, so the one artifact that does flow is fresh.
- Declare decision-pack as an optional consume on requirements-analysis (and ideally user-stories). Optional consumes are first-class since 2.2.9, and upstream-coverage would then verify Inception actually referenced the vision.

The hand-off path does not have this problem: the receiving run ingests the pack as supplied material at its own materials-first 1.1.

core/aidlc-common/protocols/stage-protocol.md
```diff
@@ -187,14 +187,14 @@ Then present the structured approval question as defined above.
 ### Part 4: Progress update (mandatory — after user approves)
 After the user selects "Approve", display a progress line before proceeding.
 
-**For enterprise and feature scopes** (all 32 stages active):
+**For enterprise and feature scopes** (the graph holds 36 stages; enterprise and feature execute all 32 delivery-path stages — every stage except the 4 discovery-only ideation stages):
 ```
 Progress: [N]/32 overall | [phase-N]/[phase-total] [Phase] stages complete. Next: [Next Stage Name]
```
apackeer: This template still says /32 while the enterprise example at line 210 below was updated to "13/36 overall", and the all-other-scopes template uses /36. The base was consistent (/32 everywhere); this PR introduces the split, and a conductor following the prose verbatim will render one or the other nondeterministically. Either this line should be /36 (matching the CHANGELOG's own claim that denominators move to N/36), or the example needs to change.

core/tools/aidlc-version.ts
```diff
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.2.1";
+export const AIDLC_VERSION = "2.2.2";
```
apackeer: 2.2.2 is already taken: v2 merged a different ## [2.2.2] and is now at 2.2.10, so this needs a rebase and re-bump (t68 will fail on the duplicate CHANGELOG heading otherwise).

One trap to catch during the re-bump: the literal string "as of 2.2.2" is baked into shipped prose that t68 does not guard - aidlc-sensor-required-sections.ts:47 and :163, aidlc-sensor.ts:404, aidlc-graph.ts:271, stage-protocol.md:745, and templates/.gitkeep:10, plus their dist mirrors. Recommend versionless wording ("the framework ships nine discovery defaults") so future re-bumps cannot strand them.

core/aidlc-common/stages/ideation/discovery-decision.md
```diff
@@ -0,0 +1,204 @@
...
+  - **Continue here** — the build happens in this workspace. The same workflow and the same record continue into delivery: proceed to Step 6 and the exit gate, then follow Step 8.
+  - **Hand off** — the build happens somewhere else (another repository, another team, a workspace that holds the code). Proceed to Step 6 and the exit gate; approval completes this workflow, and the decision pack's handoff contract is what the receiving team's workflow consumes.
+- **Pivot** — everything learned is kept, and the initiative returns to the framing. Do NOT proceed to the exit gate and never mark this stage completed: after the §13 learnings ritual, relay the engine's backward jump — `bun {{HARNESS_DIR}}/tools/aidlc-jump.ts execute --target discovery-future-state --direction backward`. The jump re-opens discovery-future-state and the stages after it so their artifacts can be revised; nothing is deleted, and the evidence record is append-only — it is never rewound.
```
apackeer, line 136: The orchestrator SKILL.md says never call the state tools directly (the engine's report dispatches them), but this stage instructs exactly that, here and in Step 8 - and no other stage body invokes aidlc-jump.ts or aidlc-state.ts directly. Mechanically the verbs exist and audit correctly (I verified each), so this is a contract inconsistency rather than a dead end, but a conductor obeying SKILL.md literally may refuse the move. One side needs an explicit carve-out.

core/aidlc-common/stages/ideation/discovery-decision.md
```diff
@@ -0,0 +1,204 @@
...
+When the person chose **Continue here**, the same workflow and record flow
+into delivery after the gate approves — no new workflow, no re-entry through
+the ideation questionnaire. Relay this sequence of engine verbs, in order
+(each is audited, and the whole sequence was chosen because every verb
+already exists — this stage adds no machinery):
+
+1. Ask which delivery scope fits the committed build (a structured question:
+   `feature` is the default, `mvp` and `enterprise` are the usual
+   alternatives, or a composed scope via `/aidlc compose`), then relay
+   `bun {{HARNESS_DIR}}/tools/aidlc-utility.ts scope-change --scope <chosen>`.
```
apackeer, line 172: This sequence has no test coverage, deterministic or live, and it works only because scope-change and next happen to lack the Completed-status guard that recompose already has (aidlc-utility.ts:3060 refuses non-Running state with a rationale that reads on exactly this situation). If that guard is ever extended to scope-change, this documented path breaks silently. Since this is the scope's advertised graduation path into delivery, one test pinning the sequence would make it load-bearing rather than accidental.

core/aidlc-common/stages/ideation/discovery-experimentation.md
```diff
@@ -0,0 +1,183 @@
...
+- Design language record from `<record>/ideation/discovery-current-state/` (if exists) — where absent, Build steps render mock-ups plainly and every verdict those mock-ups feed states the missing design language as a limit on what the reactions mean
+- Load guardrails from `{{HARNESS_DIR}}/rules/`
+
+### Step 3: Enter or Re-enter the Cycle Loop
+
+The loop is a bounded prose loop whose state lives in the artifacts on disk.
+On re-entry (a new session, a resumed workflow), this stage re-runs from its
+top per the standard stage ritual — but substep 1 reads what earlier cycles
+already recorded, so work continues from the record rather than starting over.
+At the artifact re-use question, the answer for `test-plans.md` and
```
apackeer: This pre-decides the person's answer to the protocol's artifact re-use question - stage-protocol defines it as a 3-option structured question, and Redo overwrites files. An executing agent will either skip the question (protocol violation) or ask it and risk a Redo that rewinds the append-only evidence record. Suggest rewording as "recommend Modify, and warn that Redo would rewind the append-only evidence record", or adding an explicit protocol carve-out for this stage.

core/aidlc-common/stages/ideation/discovery-current-state.md
```diff
@@ -0,0 +1,160 @@
...
+marked `unconfirmed` until the person validates it at this stage's playback
+gate. When an answer changes, every entry that depends on it reopens — record
+the reopening rather than silently overwriting. Ask the person ONLY what the
+materials and research could not answer, using the [Answer]: tag format from
+stage-protocol.md with the unified question flow (Guide Me / Edit File /
+Chat).
+
+### Step 4: Capture the Design Language Record
+
+Create or extend `<record>/ideation/discovery-current-state/design-language-record.md` with four parts:
```
apackeer: "Four parts", but the shipped design-language-record template carries five H2s (## Gaps is the fifth), and the required-sections sensor checks output against the template's full H2 set. An agent following this prose alone writes four sections and collects an advisory SENSOR_FAILED. Say five, or name Gaps here.

core/aidlc-common/stages/ideation/intent-capture.md
```diff
@@ -42,49 +53,143 @@ Load aidlc-architect-agent persona from `agents/aidlc-architect-agent.md` for te
...
+> C. A mandate — something already decided elsewhere that this must honor
+> D. A decide-by date (a steering meeting, a contract renewal, a budget cycle)
+> E. None of these — the description is the whole input
+> X. Other (please specify)
+
+Follow the unified question flow from stage-protocol.md section 3. Gather the
+selected items through follow-up questions written into the same questions
+file, each with a blank [Answer]: tag, before ending any turn that waits on
+the person. Record declines as declines; never nag. When E is the answer,
+skip ahead — the questionnaire path is unchanged.
```
apackeer: "Skip ahead" reads like "jump to Step 5", but Step 4 still requires creating source-inventory.md recording "None". An agent taking the skip literally never writes the file - no hard failure (the artifact guard passes on any one declared produce), but the "an empty inventory is itself a fact the record keeps" intent silently fails. Worth a clause: "skip ahead after recording the empty inventory in Step 4".

core/aidlc-common/stages/ideation/intent-capture.md
```diff
@@ -42,49 +53,143 @@ Load aidlc-architect-agent persona from `agents/aidlc-architect-agent.md` for te
...
-- **Initial Scope Signal** — Early indication of scope (enterprise, feature, mvp, poc, etc.)
+- **Initial Scope Signal** — Early indication of scope (enterprise, feature, mvp, poc, discovery, etc.)
+
+When the person brought asks, add:
+- **Where This Came From** — one row per ask: who asked, how often, what they
+  said (verbatim where short), and the source. The owner's reading from
+  Step 3, verbatim. Several asks can feed one initiative; an ask that turns
+  out to contain two problems is split into two rows and noted.
+
+When either exists, add:
```
apackeer: "When either exists" - the intended antecedent (a mandate or a decide-by date) is 70+ lines up in the intake options; the nearest conditional is "When the person brought asks" a few lines above, so a conductor can read "either" as "asks or materials" and emit empty Already Decided / Decide-By sections. Naming the two things explicitly removes the misread.

tests/integration/t199-discovery-scope.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    const graph = readGraph();
+    const node = graph.find((s) => s.slug === "intent-capture");
+    expect(node).toBeDefined();
+    // The four stock scopes that ran intent-capture before this scope landed
+    // still run it, discovery is the ONLY addition, and no stock scope lost
+    // or gained the stage.
+    expect(node?.scopes).toEqual(["enterprise", "feature", "mvp", "poc", "discovery"]);
+    // The two produces the extension added — every scope that runs the stage
+    // gets them; downstream discovery stages consume them by these names.
+    expect(node?.produces).toContain("source-inventory");
```
apackeer: toContain lets a sixth, unintended produce pass silently; an exact toEqual on the ordered array would pin the append-only contract fully. The rest of this file is genuinely falsifiable - this is the one soft spot. (Separately: the t199 number is already taken on v2 by tests/unit/t199-learnings-memory-path.test.ts, and --filter matches by filename regex, so please renumber on rebase.)

apackeer (changes requested):
## Verdict: request changes

The hypothesis-testing loop this PR builds is real, and the non-breaking case for the stock scopes genuinely holds (details at the bottom). The main ask is a design one: the flow's central promise - that discovery's validated vision sets up delivery - is only half-wired. After that come three test failures and a rebase collision, all mechanical.

### 1. The vision does not mechanically reach Inception (the main ask)

The tenet this PR implements: Ideation's output is a validated vision, which sets up Inception (the implementation plan), which sets up Construction (code). The discovery flow builds that vision honestly: future-state + assumptions-record (1.9) feed test-plans + evidence-record (1.10), riskiest-first and append-only, and the decision pack (1.11) is the vision artifact, with a machine-readable handoff contract on commit.

But on the continue-here path, no Inception stage consumes any of it. Grepping every inception and construction stage plus all agents: zero references to decision-pack, evidence-record, assumptions-record, or future-state. Inception's entry stage (requirements-analysis) declares only intent-statement and scope-document, both optional - and the intent statement is written at 1.1, before discovery ran; none of the four discovery stages writes back to it. Whether delivery ever reads the pack is left to the conductor LLM's initiative, nudged only by a skip-reason string. The upstream-coverage sensor cannot catch this because it only verifies declared consumes. The hand-off path is fine (the receiving run ingests the pack as supplied material at its own materials-first 1.1); it is the flagship continue-here path that leaks.

The fix is small - see the inline comment on discovery-decision.md Step 8 for two concrete options. Related: stage-protocol-recovery.md:60 loads resume context for ideation 1.1-1.7 only, so a run resumed inside or after discovery never reloads the pack either.

### 2. Blockers (mechanical)

- **Three deterministic test failures introduced by this PR.** t189-compose-dispatch.sdk, t192-compose-front-journey.sdk, and t193-compose-report-journey.sdk pin the stock scope count at 9 (t189:103,108; t192:89; t193:68,96,100; plus t192:73 STOCK_SCOPES lacking discovery). All three fail with `expected 9, received 10` at the PR head and pass at the merge base (A/B verified). The two e2e compose-front twins were updated for 9 -> 10; these three integration tests were missed. The PR body's "one known environmental failure (t92)" only holds in an environment where the sdk-live tests skip.
- **Version 2.2.2 collides.** v2 already merged a different 2.2.2 and is currently at 2.2.10, so the PR needs a rebase and re-bump (GitHub already reports it conflicting on CHANGELOG.md, README.md, the five aidlc-version.ts copies, and tests/.coverage-registry.json). See the inline comment on aidlc-version.ts for a trap: the literal version string is baked into six shipped prose sites that t68 does not guard.
- **Test number t199 is taken.** v2 already has tests/unit/t199-learnings-memory-path.test.ts; this PR adds tests/integration/t199-discovery-scope.test.ts. run-tests.sh --filter matches by filename regex, so t199 becomes ambiguous. Please renumber to a free slot on rebase.

### 3. Majors (inline comments carry the detail)

- stage-protocol.md Part 4 now contradicts itself: the enterprise/feature template says /32 while its own example says /36.
- The commit-continuation verb sequence has zero test coverage and depends on the absence of a status guard a sibling verb already has.
- discovery-decision.md instructs direct state-tool calls that the orchestrator SKILL.md forbids; no other stage body does this.
- PR body accuracy: the parity fixtures actually show four added SKIP rows per stock scope (zero removals), not "exactly one removed SKIP row" - the real property is stronger than the claim, but the sentence is wrong. And "with nothing supplied, the clarifying questions run exactly as before" oversells: a stock run always gains one intake question, two always-produced records, and a new confirm-unconfirmed-answers duty at the gate (the CHANGELOG states this correctly).

### 4. Minors

- discovery-experimentation.md pre-decides the protocol's artifact re-use question (inline).
- discovery-current-state.md says "four parts"; the template has five H2s (inline).
- Two intent-capture ambiguities an executing agent can trip on (inline).
- Stale refs the sweep missed: docs/guide/glossary.md:51 ("Ideation 7"), docs/guide/workshop-mode.md:59 ("1.1-1.7"), docs/guide/14-artifacts-reference.md:148 (heading says 1.1-1.7 above a table containing 1.8-1.11), core/aidlc-common/protocols/stage-protocol-recovery.md:60 (ships in dists).
- t199's produces pin uses toContain rather than an exact match (inline).

### What holds up well

- All 9 stock parity fixtures change by exactly four added SKIP rows with every other row byte-identical; walk fixtures untouched; no existing stage number shifts.
- The artifact guard passes on any one declared produce, so the added produces cannot create a hard-failure path; sensors stay advisory.
- The sensor/graph tool diffs are comment-only as claimed (verified byte-level); validateScope("discovery") returns clean with zero advisories.
- Every engine verb the stages name exists with the exact spelling and flags; keywords: [] makes the scope genuinely unreachable by inference.
- The ~18 count-pin test edits strengthen rather than weaken (t30 gains an exact discovery EXECUTE pin; t66's enterprise check became strictly stronger); the coverage ratchet only went up.
- bun scripts/package.ts --check is green on all four harnesses at the PR head, and settings.json changes announcement text only.

### Verification

Full deterministic suite (smoke+unit+integration, 229 files) run at the PR head: 6 red. t121/t145/t165 pass green-alone (parallelism flakes); t189/t192/t193 stay red alone and pass at the untouched merge base, so those three are introduced here. Packaging parity green on all four harnesses.

akash-doshi:
Cheers Arden, all items from the review are addressed on the updated branch (rebased onto v2.2.10, version 2.2.11). What changed and why, in the review's order:

**1. Wiring the vision into Inception (the design ask)**

The gap was real and ran wider than the continue-here path: no Inception stage declared any discovery output as an input, so validated content reached delivery only through conductor initiative. Changes:

- `requirements-analysis` and `user-stories` declare `decision-pack` as an optional consume (`required: false`). When the pack exists, the run-stage directive carries its path and upstream-coverage verifies the output referenced it. When it does not, the resolver drops the entry silently, so stock scopes are unaffected — parity fixtures unchanged.
- At commit, the decision stage folds a "What Discovery Validated" section into `intent-statement.md` (chosen framing, recorded exclusions, provenance in file-name form). Rationale: that file is the input requirements-analysis already reads, so after a commit it must reflect the validated direction, not the pre-discovery intent.
- The same leak existed for visual content one stage further. `refined-mockups` consumed nothing discovery produces, and its documented fallback designs from requirements and stories, which carry no visual content — so a validated design language and people-tested mock-up variants died at the phase boundary. It now declares `design-language-record` and `evidence-record` as optional consumes, with prose to design within the recorded language and treat promote-marked variants as starting points.
- The handoff contract is restructured to instruct its consumer rather than describe content: a consumption note (settled entries are pre-answers to confirm, never re-elicit; open items are questions to raise), a uniform entry shape (value, status, confidence, confirming gate, source), exclusions with evidence, open items with recommendations, promoted starting points, and a closing map from delivery's first questions to where the record answers them.
- `stage-protocol-recovery.md` now spans ideation 1.1–1.11 and names the pack as load-bearing resume context, closing the resume half of the same gap.
- t214 pins the consume differential in both directions: pack present, the requirements-analysis directive carries its path; pack absent, no trace of it in the directive.

**2. Blockers**

- t189/t192/t193: scope-count pins updated (9→10 for stock counts, 10→11 in the composed-on-top journeys, discovery added to STOCK_SCOPES). These tests skip without live SDK credentials, which is why local runs reported green here — the fix is arithmetic-correct rather than run-proven in this environment, and the PR body's verification statement now states its environment scope explicitly.
- Version: rebased onto 242953e, re-bumped to 2.2.11. The six shipped prose sites carrying a literal version string are now versionless ("the framework ships nine defaults"), so future re-bumps cannot strand them — the collision was routine, the baked-in strings were the recurring cost.
- Test number: renamed to t213. t203, the first rename, was also taken upstream in the interim.

**3. Majors**

- The /32 vs /36 template split was fixed in the branch's second commit, which predates the review. The numerator rule is now also disambiguated (plan SKIP rows count in neither the numerator nor the total; runtime `[S]` skips count in both) — the old wording let a careful reader compute contradictory numbers from the protocol and the guide examples.
- The continuation sequence is a pinned behavioral contract, not an accident of a missing guard: t214 drives scope-change → next → set-status → skip → next against a completed-discovery fixture, asserts each half of the no-guard behavior separately, and includes a contrast test proving recompose refuses the identical state. If a Completed-status guard is ever extended to scope-change, the test fails and Step 8 must be redesigned with it.
- The verb-relay carve-out is stated on both sides: the orchestrator SKILL rule (all four harness copies) names the stage-body relay exception, and the stage names the rule it excepts. Auditing that sentence surfaced a false claim of ours: set-status emits no audit event (jump, park, scope-change, and skip each do). Both surfaces now state the distinction precisely.
- PR body accuracy: the fixture sentence now states the actual diff (each stock scope's plan gains exactly four SKIP rows, nothing removed), and "exactly as before" is replaced with the CHANGELOG's wording — one intake question, two additive records, a confirm-unconfirmed duty at the gate.

**4. Minors**

All applied: the artifact re-use guidance recommends Modify and warns that Redo would rewind the append-only evidence record (a stage file cannot pre-decide a protocol question), the two intent-capture ambiguities are resolved by naming their referents (record the empty inventory in Step 4 before skipping ahead; a mandate or a decide-by date), workshop-mode reads 1.1–1.11, and the t213 produces pin is an exact ordered list so a sixth produce cannot pass silently.

**Corrections and disclosure**

- Two details in the review, for the record: `required: false` consumes predate 2.2.9 (2.2.9 added `optional_produces` on the produce side), and two of the listed stale refs (glossary phase count, artifacts-reference heading) were already fixed at the local head the stale PR could not show. Neither changes an action taken.
- One previously undisclosed change, now disclosed in the CHANGELOG: the welcome banner's scope table had two pre-existing cells disagreeing with the compiled grid (infra INCEPTION reads PD, RA; security-patch RE reads RE, RA). Both corrected.

**Other Things**

- Right now, we have two flows in the AI-DLC Ideation stage: 1/ the default workflow and 2/ this discovery phase that embeds stronger design thinking patterns. Am aware this is introducing more stuff in already a step-heavy workflow, and would like to first validate the capability of this workflow before making this new discovery phase the default ideation path.

core/aidlc-common/stages/ideation/discovery-decision.md
akash-doshi: Both options implemented, plus the same wiring one stage further where the leak repeated for visual content: optional `decision-pack` consumes on requirements-analysis and user-stories, the "What Discovery Validated" write-back before the scope-change, `design-language-record` and `evidence-record` consumes on refined-mockups, and the recovery protocol now spans 1.1-1.11. t214 pins the pack-present/pack-absent differential.

core/aidlc-common/protocols/stage-protocol.md
akash-doshi: Fixed: the template line reads /36, and the numerator rule is now disambiguated (plan SKIP rows count in neither the numerator nor the total, runtime `[S]` skips count in both). The old wording left it computable two ways.

core/tools/aidlc-version.ts
akash-doshi: Rebased onto 242953e, now 2.2.11. The six literal-version prose sites are versionless now, so future re-bumps cannot strand them.

core/aidlc-common/stages/ideation/discovery-decision.md
akash-doshi: Carve-out stated on both sides: the SKILL rule (all four harness copies) names the stage-body relay exception, and the stage names the rule it excepts. Auditing that sentence surfaced a false claim of ours: set-status emits no audit event (the other four verbs do). Both surfaces now state the distinction.

core/aidlc-common/stages/ideation/discovery-decision.md
akash-doshi: t214 now drives the sequence against a completed-discovery fixture, asserts each half of the no-guard behavior separately, and includes a contrast test that recompose refuses the same state. Extending the guard to scope-change now fails a test instead of silently breaking Step 8.

core/aidlc-common/stages/ideation/discovery-experimentation.md
akash-doshi: Reworded as suggested: recommend Modify, and warn that Redo rewinds the append-only evidence record.

core/aidlc-common/stages/ideation/discovery-current-state.md
akash-doshi: Fixed: the prose lists five parts with Gaps as its own heading, matching the template the sensor enforces.

core/aidlc-common/stages/ideation/intent-capture.md
akash-doshi: Now reads: skip ahead after recording the empty inventory in Step 4.

core/aidlc-common/stages/ideation/intent-capture.md
akash-doshi: Now names the referents: a mandate or a decide-by date.

tests/integration/t199-discovery-scope.test.ts
akash-doshi: Renumbered to t213 (t203, the first choice, was also taken upstream in the interim). The produces pin is an exact ordered `toEqual`, so a sixth produce cannot pass silently.

apackeer:
@raj-jain-aws @SiddhJog @spraja08 @leandrodamascena Can you guys take a look at this? I think directionally (and mechanically) it fits in with the model of Ideation produces an idea, vision, Inception produces a design or plan, Construction produces code etc.

leandrodamascena:
I ran a max-effort pass over this — nine angles, each finding reproduced or refuted against the PR tree, focused on what the four fix commits might have introduced. Short version: the code is healthy, all of Arden's findings landed, and what's left is purely rebase.

**Arden's findings — all fixed, I checked each:**
- Vision reaches Inception: discovery-decision Step 8 now writes the `## What Discovery Validated` section back into intent-statement.md, and requirements-analysis/user-stories declare `decision-pack` as an optional consume. I traced the chain and it holds — the record references the write-back emits match the slugs the upstream-coverage sensor greps for.
- The three test failures (t189/t192/t193): now pin 10 and include discovery in STOCK_SCOPES. Strengthened, not loosened.
- /32 vs /36: no `/32` remnant, and the progress arithmetic (a finished feature run reads 32/36) is right.
- t199 collision: renamed to t213/t214.
- The stale refs (glossary, workshop-mode, artifacts-reference, stage-protocol-recovery): swept. Recovery now loads discovery context 1.8–1.11.

**New bugs:** I could not find one. I ran the sensor greps against the templates, traced the commit/pivot/park continuation, diffed the discovery nodes against a core node, and checked the empty-keywords path (discovery is genuinely unreachable by inference). The stock scopes are byte-safe — the only fixture change is the four added SKIP rows, everything else identical.

**What actually blocks the merge — both rebase, neither is your code:**

1. **Version 2.2.11 collides (t68 will fail).** v2 has moved to 2.3.0 and already carries a `## [2.2.11] - 2026-07-08` heading (the #508/#510 backstop, a different feature). This PR adds a second `## [2.2.11]`. t68 asserts headings are unique, so two of them fail deterministically. This is the conflict-trap AGENTS.md documents — rebase and re-bump to the next free 2.3.x, and resolve the duplicate CHANGELOG heading.

2. **`bun scripts/package.ts --check` fails after the rebase, but not because of this PR.** v2 itself is currently broken: the #475 squash-merge dropped the `produces_kinds` parse call in `aidlc-lib.ts` and its `OPTIONAL_FIELDS` entry (`aidlc-stage-schema.ts:152`), while four construction stages still carry `produces_kinds` frontmatter — so `--check` fails on clean v2 today. This PR's head passes its own `--check` (it predates #511, no `produces_kinds` anywhere), but once you rebase onto v2 the dist regen inherits the broken tree. I'm filing this separately since it blocks every PR, not just yours — please don't treat it as a #526 defect.

**Non-blocking:**
- The PR body still says "exactly one removed SKIP row" (fixtures show four added, zero removed) and "clarifying questions run exactly as before" (a stock run gains an intake question, two records, and a gate duty). The CHANGELOG states it correctly — just fix the body.
- The version string is baked into a few shipped prose sites t68 does not guard; watch those on the re-bump.

Net: the discovery flow itself is in good shape. Rebase onto v2, re-bump off 2.2.11, and the `--check` will pass once the v2 `produces_kinds` regression is fixed (separate issue). I would approve once the rebase is done.

RachnaC1234:
I had opened same request for AIDLC 1.0 that was rejected earlier. I am curious to learn the drivers of the new phase in AIDLC 2.0. What are the key goals and what is driving including it as that determines the correctness and scope of the solution. My solution I proposed for AIDLC has landed eventually here as open source solution https://github.com/aws-samples/sample-ai-plc
Few learnings I had from running this as pre-req to AIDLC for workshops - 
1/ Your personas are not technical usually for discovery phase. So simplicity of the design is the key to success, something that is understandable and not overwhelming. 
2/ What works for one customer, does not work for other, can non technical personas extend what your provide or customize with Natural language ? 
3/ Output creation from the phase should be standardized, though the output can come from any implementations of Ideation phase done using various tools. ( Including Amazon Quick ) 
4/ Ideation brings together a new methodology and personas in the mix, could be more than PMs ..  think of strategists, CxO. This need to be accounted in the methodology and collab phase and how we the flow will work with the users. 
5/ Idea needs to be proven with prototypes, prototypes should be done simply, with least friction from technology so ideas can be validated before inception happens and user stories are written. 
6/ while it may first feel like that greenfield needs ideation, fact is most of the enterprises are not starting from scratch, they have existing products that they are extending, so the soln has to accomodate brownfield to be useful for enterprises.

akash-doshi:
Hi @RachnaC1234 thanks for your comment, please take some time to review this PR as keen to see your feedback on this approach - as it addresses most of the points you've described. You make great points around greenfield/brownfield and rapid prototyping to validate a decision, but also how stakeholders may differ in technical capability. It opens up a question where the start point and end point of AI-DLC workflows should be.

While this PR lands the technical parts components of introducing a product discovery scope within ideation, a few areas need to be evolved:
* More elaborated playbooks for design system harvesting and prototyping
* If we recognise that teams will have their own workflows outside of AI-DLC, more relevant to their stakeholders (eg: Product Ideation with Amazon Quick) - a handover contract between other frameworks to AI-DLC is required. So far, we've based this design off the work of @clairmit https://github.com/subcontext/agentic-product-discovery. We can extend this contract beyond just discovery --> inception and ensure external tools can leverage this same contract.

To help steer in the right direction, let's frame these feedback and changes into RFCs to get feedback on the mental model and bring your expertise into the AI-DLC Ideation Stage.

akash-doshi:
I've rebased onto v2 HEAD 064c1e37 (v2.3.10), re-bumped to 2.3.11, ready for re-review

RachnaC1234:
@akash-doshi  - I will review the PR closely. RFC is a good idea. Let us work on that.

RachnaC1234:
1/ See below the logs from my test run in Kiro. and the bolded outputs in code format. . A non-technical person describing their idea in natural, longer prose (RAI framework description I wrote) reliably trips the "no stock scope fits" fork. At that point they're asked to choose between "compose" (a term they've never seen, with no explanation of what it does) or "name a scope directly" with examples like bugfix = 7 of 36 stages, poc = 8, feature = all 32. That's a stage-count / engineering-workflow framing — meaningless to a business stakeholder. They'd likely freeze here.

**__{"kind":"ask","question":"No stock scope clearly fits: \"Responsible AI assessme
nt framework that can take any RAI framework - EU AI Act, AWS Responsible AI Well-Architected, or NIST AI RMF - and run an assessment per the framework the user chooses, producing a score. Solution for enterprises where decision makers need visibility into their responsible AI posture and compliance against EU AI Act rules and other best practices.\". I can compose a tailored plan for this task (recommended: reply \"compose\"), or you can name a scope directly (e.g. bugfix = 7 of 36 stages, poc = 8, feature = all 32; see /aidlc --help for all)."}        

Run Command Hook

aidlc-runtime-compile
Command executed successfully with no output.

Run Command Hook

aidlc-sync-statusline
Command executed successfully with no output.
```
The engine couldn't auto-match a stock scope to this description (fair — it's a fairly distinctive enterprise compliance tool) and is asking me to pick a path. Here's the choice, as it's presenting it:_

**_How should we scope this workflow?

A. Compose (recommended by the engine) — dispatches an adaptive composer that scans your description and proposes a tailored stage plan (which of the 36 stages to run, at what depth), which you then approve/edit before anything starts. B. feature — runs the standard 32/36 stages at Standard depth. Solid default for a real product build. C. enterprise — runs 32/36 stages at Comprehensive depth (more questions, deeper artifacts per stage). Given this has compliance/regulatory weight (EU AI Act) and enterprise decision-makers as the audience, this might actually fit better than feature. D. mvp — 22/36 stages, Standard depth, skips things like team-formation/rough-mockups/market-research — faster if you just want to prove the concept. X. Other — name a different scope, or tell me more about what you want cut/kept.

Given the regulatory/compliance nature and enterprise audience you described, I'd lean toward enterprise or compose (compose would specifically let you decide whether things like team-formation or market-research are worth running for this). Since this is a learning run, I'd suggest compose — you'll get to see the plan before committing, which is itself useful for understanding how the framework adapts.

Which do you want — A, B, C, D, or something else?_**_**
```

2/  The experience today is build for technical conductor (a developer or PM comfortable with dev tooling) running the process on behalf of business or non tech stakeholders, not for a non-technical person to drive solo end-to-end. Is this how you are envisioning how the ideation phase will be run with customers ? 

3/ We have better tools available for non tech personas to use to get to the same outcome. I suggest we create a agreed format for the artifacts these tools will generate that can then be added to the context in AIDLC development phase.

4/ we start with the capability model for what PMs need and then create various implementations that map to those capabilities instead of having various random implementations of same concept with same intent but using different terminology and scope , leaving no way for end user to compare the tools they could have leveraged.

akash-doshi:
Let me check into 1 for reproducibility and resolution. The rest of these feedback is super valid and worth a wider discussion, non-blocking to this PR. For 3, this is why I've suggested RFC#586.

From my perspective, the workflows have been adapted to get the most value out of the harnesses they support - however, there's more effort to be made to improve the UX and abstract complexity. As this isn't related to the PR, let's bring this up in a new issue/RFC

apackeer:
Thanks, Akash. The product-discovery work here is valuable, but before proceeding we need a clear recommendation on its ownership boundary.

This PR creates an alternative path through Ideation. Alternative definitions are valid, but they should be delivered as opt-in plugins. Bare core should continue to ship one canonical Ideation definition, and disabling the plugin should restore that path without residual graph or artifact changes.

Could you provide:

1. The minimal core Ideation or Inception contracts needed for alternative methodologies to hand work into delivery.
2. Which stages, artifacts, knowledge, and behavior belong in `plugins/product-discovery`.
3. The stage graph with the plugin enabled and disabled.
4. How this converges with #652 and how its handoff relates to #586 and the proposed `context-pack`.

An ownership table mapping each stage, artifact, and contribution to either `core` or `plugins/product-discovery` would make the recommendation actionable.

PR #541, opened 2026-07-09
fix: surface recorded hook drops in doctor (2.3.2)

### LINKED ISSUE #516: [Bug]: doctor never reads the hook-drop files that recordHookDrop writes for it
## Description

`recordHookDrop` (core/tools/aidlc-lib.ts) writes per-hook drop files to `aidlc/.aidlc-hooks-health/<hook>.drops` (one ISO timestamp per line), and its own header comment states the purpose: "Record drops to a per-hook counter file so `--doctor` can surface them." But `handleDoctor` never reads `.drops` files - it only reads the `.last` heartbeat files. Grep of the doctor handler for "drops" comes back empty.

So every hook that fail-opens and records a drop (audit emission failures, and now the stale compose-marker skip added in PR #500, which logs through this exact channel) is writing telemetry nobody surfaces. The silent-failure visibility the mechanism was built for - per its own comment, "silent failure was the whole point of the state-machine refactor" - stops one step short of the user.

## Expected Behavior

`/aidlc --doctor` gains a probe that reads `aidlc/.aidlc-hooks-health/*.drops` and reports, per hook, the drop count and most recent drop timestamp - pass row when empty/absent, fail (or advisory) row with the filename when drops exist, following the existing probe shape (the leaked-lock and compose-marker probes in the same handler are the precedent). Possibly with a staleness window so ancient resolved drops do not fail forever, or a documented "delete the file after investigating" remediation in the fix text.

## Environment

- v2 (2.2.x), core tool + hooks, all harnesses
- Pre-existing gap; surfaced by the adversarial review of PR #500 while verifying where the new stale-marker drop record goes

apackeer:
Fixes #516

## Problem

`recordHookDrop` (core/tools/aidlc-lib.ts) writes per-hook drop files to `aidlc/.../.aidlc-hooks-health/<hook>.drops` - one line per fail-open swallow (ISO timestamp, TAB, one-line reason) - and its header comment states the purpose: "Record drops to a per-hook counter file so `--doctor` can surface them." But `handleDoctor` never read `.drops` files; it only read the `.last` heartbeats. Every hook that fail-opened and recorded a drop (audit emission failures, the stale compose-marker skip, all 17 call sites across the 11 hooks) was writing telemetry nobody surfaced.

## Fix

Doctor gains a `Hook drops` probe directly after the heartbeat block, advisory always (pass row, never flips the exit code):

- Clean: `Hook drops: none recorded`.
- Drops present: `Hook drops recorded (advisory): <hook> xN (last <timestamp>), ...` with the remediation (inspect the named `.drops` file(s) under `.aidlc-hooks-health/`, delete once investigated).
- Unreadable dir or files are NAMED (`health dir unreadable` / `N .drops file(s) unreadable`) instead of absorbed into the clean row - EACCES is exactly the environment that produces drops, so a permission failure must not report a clean bill.
- A torn last line (no TAB - the disk-full case, since recordHookDrop fires under exactly those conditions) shows `(last unparseable line)` behind a timestamp-shape gate instead of splicing the raw fragment into the label.

Advisory rather than failing is deliberate: a drop records a PAST swallowed failure and the file persists until hand-deleted, so a failing row would pin doctor's exit at 1 (breaking CI consumers) long after the cause was fixed. Mirrors the submodules/heartbeat advisory pattern.

Also corrects the `recordHookDrop` header comment: the format carries a TAB-separated reason, not just a timestamp.

**Known limitation** (shared with the sibling heartbeat probe): the read follows the active-intent cursor, so drops recorded under a previously active intent are out of reach until that intent is active again. A sweep across sibling intent records would be a separate enhancement.

## Tests

- `tests/unit/t37.test.ts` cases 18/18b/18c/18d: none-recorded row; recorded drops (seeded through the REAL `recordHookDrop` writer so the reader is bound to the writer's actual format - a format drift now fails the test) with exit-code neutrality pinned against a clean twin project; empty file treated as none; torn last line shows the placeholder, never the raw fragment.

## Verification

- t37 27/27 + t68 green; full smoke+unit tier: 151 files / 0 failed
- `bun run typecheck` (both tsconfigs) and `bun scripts/package.ts --check` clean

## Docs

`docs/guide/12-cli-commands.md`: doctor's check table gains the Hook drops row and the example output shows it.

## Release mechanics

2.2.20: `core/tools/aidlc-version.ts`, README badge, and CHANGELOG entry in the same commit per changelog policy. The sibling PR (scope-change autonomy guard) also claims 2.2.20; whichever merges second re-bumps per the conflict-trap policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.3.2] - 2026-07-10
+
+`/aidlc --doctor` now surfaces recorded hook drops. Hooks fail open by design - an audit-emission failure or a swallowed error must never break your tool call - and each such swallow is recorded to `aidlc/.../.aidlc-hooks-health/<hook>.drops`, written specifically for doctor to surface. Doctor never read those files, so the silent-failure telemetry was invisible. It now reports them as an advisory row. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `/aidlc --doctor` gains a `Hook drops` row: `none recorded` when clean, or an advisory naming each hook with a drop count and the last drop's timestamp plus the remediation (inspect the `.drops` file for the reasons, delete it once investigated). Advisory only - it never flips doctor's exit code, so a long-fixed historical drop cannot pin CI red.
 ## [2.3.1] - 2026-07-10
 
 Authored agents no longer carry raw model pins: the authored contract on every agent is now a work-shaped `tier:` (`judgment` | `balanced` | `templated`), which the build projects into each harness's native model/effort keys. On Claude Code the nine judgment agents stop pinning Opus; the balanced/templated agents still project to a Sonnet-class model. The nine judgment agents (architect, aws-platform, compliance, composer, design, developer, devsecops, product, quality) become `model: inherit` with no effort pin on Claude Code, so a session running Fable (or any other model, at any effort) keeps it in delegated work - the framework never silently downgrades your session choices. The two balanced reviewers and three templated planners keep a Sonnet-class model; only the templated tier steps effort down (`medium`). To restore the old always-Opus behavior for one agent, edit the projected value in your installed copy (e.g. set `model: opus` in that agent's `.claude/agents/*.md`); to cap every agent when building from source, set `tier_cap:` in `core/memory/org.md`/`project.md` frontmatter or run the packager with `AIDLC_TIER_CAP=<tier>`. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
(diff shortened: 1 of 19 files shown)


leandrodamascena (approved):
APPROVE

The call I wanted to check was advisory-always, since this touches the F1 concern from the #475 review. I traced it: the failed runtime-compile drop is past-tense telemetry in an append-only file that persists until hand-deleted, and the runtime-graph mirror self-heals next transition — gating exit on it would pin doctor red in CI long after the cause is gone. Live breakage is still caught by the non-advisory probes. Before this PR the drop was written but never read, so it's invisible → visible-advisory with no exit regression. Advisory is right.

I ran the real `isoTimestamp()` against the probe regex — matches, and a torn line falls back to the placeholder. Writer and reader resolve the same `healthDir`.

One worth fixing (not blocking): t37 18b's `expect(r.status).toBe(doctor(clean).status)` doesn't pin exit-neutrality — the bare fixture already exits 1 on unrelated probes (the comment admits it), so it's `1===1` and would pass even if the drops row went `pass:false`.

Non-blocking: the two unreadable branches are untested; active-intent scoping means a drop under intent A reads "none recorded" under B (you flag it); 2.2.20 collides with #542, second to merge re-bumps.

PR #542, opened 2026-07-09
fix: refuse scope-change under autonomous Construction (2.3.3)

### LINKED ISSUE #515: [Bug]: scope-change lacks the autonomous-Construction guard that recompose now has
## Description

#485 (fixed in PR #500) added an engine-side refusal so `recompose` cannot re-shape the plan while Construction Autonomy Mode is autonomous - the rule was previously prose-only. The adversarial review of that fix surfaced that `scope-change` has the same hole and was left out of scope: `handleScopeChange` (core/tools/aidlc-utility.ts) flips stage inclusion (EXECUTE/SKIP suffixes) with no autonomy check, so an unattended autonomous run can still have its plan re-shaped through the scope-change verb, with no human at the gate.

Same rationale as #485: an autonomous swarm/Bolt run has no human present, so a mid-run plan mutation must be refused by the engine, whoever calls the verb. The fix shape is the one PR #500 established: at the top of the handler, inside the audit lock, refuse with a clear error when Construction Autonomy Mode is autonomous, naming the remediation (`aidlc-bolt set-autonomy --mode gated`, or let the run finish). The recompose refusal in PR #500 is the copy-paste-adjacent precedent, and t194's autonomy-rejection test shape (state-rewrite seeding, byte-unchanged assertion) transfers directly.

Worth checking in the same pass whether any OTHER plan-mutating verb shares the gap (jump with EXECUTE overrides, config-change) so the guard lands once, consistently.

## Expected Behavior

`scope-change` (and any sibling plan-mutating verb) refuses under autonomous Construction with a non-zero exit and an actionable message, leaving state byte-unchanged, mirroring the recompose guard.

## Environment

- v2 (2.2.x), engine-level, all harnesses
- Surfaced by the adversarial review of PR #500 (issues #484/#485)

apackeer:
Fixes #515

## Problem

#485 (fixed in PR #500, 2.2.8) added an engine-side refusal so `recompose` cannot re-shape the plan while `Construction Autonomy Mode` is autonomous - the rule was previously prose-only. The adversarial review of that fix surfaced that `scope-change` has the identical hole and was left out of scope: `handleScopeChange` flips stage EXECUTE/SKIP suffixes with no autonomy check, so an unattended autonomous run could still have its plan re-shaped through the scope-change verb, with no human at the gate.

## Fix

The recompose guard's twin in `handleScopeChange`, expressed through the exported `isAutonomousMode` predicate (per its contract in aidlc-lib.ts: new gate sites use the helper so the sites cannot drift; only the pre-existing open-coded sites are grandfathered). Refuses with exit non-zero and the same remediation text: switch to gated Construction first (`aidlc-bolt set-autonomy --mode gated`) or let the swarm finish.

Placement is fail-fast, before the same-scope early exit (matching recompose's posture): under autonomy even a no-op call is refused, so the conductor learns the rule on first contact. Verified unaffected paths:

- Starting a fresh workflow with `--scope` never reaches the guard (it dies on the missing state file first, and the engine routes fresh-workspace `--scope` to intent-birth).
- No autonomous caller exists: nothing in `aidlc-swarm.ts`, `aidlc-bolt.ts`, or the hooks invokes scope-change; the only engine emitter is the orchestrator's Branch-5 print when a human passes a differing `--scope`.
- An invalid scope name still wins over the autonomy refusal (validation order unchanged).

**Two adjacent gaps noted, deliberately out of scope** (same class, separate issues): `aidlc-state.ts skip` marks a pending stage `[S]` with no autonomy guard, and `config-change` rewrites Depth/Test Strategy mid-run unguarded. Same guard-shape applies; happy to file follow-ups.

## Tests

- `tests/unit/t36.test.ts` cases 8/8b (the t194 recompose-guard twins): refusal under autonomy with the remediation named, state byte-untouched, no `SCOPE_CHANGED` row emitted; gated mode proceeds exactly as before.

## Verification

- t36 9/9, t194 17/17 (recompose guard unaffected), t68 green; full smoke+unit tier: 151 files / 0 failed
- `bun run typecheck` and `bun scripts/package.ts --check` clean

## Docs

`docs/guide/12-cli-commands.md`: the `--scope` section documents the refusal and remediation.

## Release mechanics

2.2.20: version trio + CHANGELOG entry in the same commit. The sibling PR #541 (doctor hook drops) also claims 2.2.20; whichever merges second re-bumps per the conflict-trap policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.3.3] - 2026-07-10
+
+`scope-change` now refuses to run under autonomous Construction, closing the gap its sibling `recompose` closed in 2.2.8: both verbs re-shape the live plan's EXECUTE/SKIP stage inclusion, and an unattended autonomous run has no human at the gate to approve the new shape. Previously the "never re-shape the plan under autonomy" rule was engine-enforced for `recompose` but prose-only for `scope-change`. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `scope-change` (and `/aidlc --scope <name>` on a running workflow) exits non-zero with a clear error when `Construction Autonomy Mode` is `autonomous`, naming the remediation (`aidlc-bolt set-autonomy --mode gated`, or let the swarm finish). Gated and unset modes proceed exactly as before; starting a fresh workflow with `--scope` is unaffected.
 ## [2.3.2] - 2026-07-10
 
 `/aidlc --doctor` now surfaces recorded hook drops. Hooks fail open by design - an audit-emission failure or a swallowed error must never break your tool call - and each such swallow is recorded to `aidlc/.../.aidlc-hooks-health/<hook>.drops`, written specifically for doctor to surface. Doctor never read those files, so the silent-failure telemetry was invisible. It now reports them as an advisory row. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.3.2-blue)
+![version](https://img.shields.io/badge/version-2.3.3-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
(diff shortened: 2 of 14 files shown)


leandrodamascena (approved):
APPROVE

Same guard I reviewed on the recompose side (#500), now on scope-change — the twin makes sense, both verbs re-shape a live plan and both need a human at the gate.

I checked the one thing that would matter: does `isAutonomousMode` catch every "autonomous" value? It compares `=== "autonomous"`, and `set-autonomy` (bolt.ts:782) refuses to write anything but `autonomous` or `gated`, so there's no variant it misses. I also grepped for scope-change callers — the only emitter is the orchestrator's Branch-5, which fires when a human passes `--scope`. No swarm or hook calls it, so the guard can't break an autonomous path. It dies before any write, so state stays byte-identical, and t36 case 8 pins that.

Two same-class gaps the PR already flags for follow-up: `skip` and `config-change` are still unguarded. And #541 also claims 2.2.20 — second to merge re-bumps.

PR #550, opened 2026-07-10
feat: plugin content buckets, ownership carry-through, and install-time plugin selection (2.3.5)

apackeer:
# feat: plugin content buckets, ownership carry-through, and install-time plugin selection (2.3.5)

Completes the plugin mechanism (#475) with the surface a real plugin needs, proven at scale in a customer pilot before landing here: a pilot plugin shipping 23 stages, 18 agent personas, 90 knowledge files, and 4 scopes runs on this exact mechanism.

**Design principle: plugins add, the install selects.** A plugin never removes or overrides core; an install chooses which plugins' content its users see. Core itself is the implicit `aidlc` plugin and can be deselected - a plugin-only install shows only that plugin's commands and scopes while core stays installed and re-enableable.

## What's in here

- **`bundle:` renamed to `plugin:`** - **breaking for plugin trees authored against #475**: the old key is rejected everywhere it is read (schema validation errors with the fix named; the compose hook drop-skips a `bundle:` contribution; a `bundle:`-keyed plugin STAGE file is skip-and-dropped at compose copy time instead of landing and bricking the install's graph compile). There is no read-side alias. **Migration: rename `bundle:` to `plugin:` in your plugin tree.** "bundle" is reserved for a possible future collection-of-plugins concept and deliberately unused (see doc 18).
- **Plugin content buckets**: plugins ship `scopes/`, `agents/`, `knowledge/` alongside stages/sensors/tools/contributions - projected by the packager, composed additively (no-clobber, collision drop-logged). `memory/` stays deferred (real collision semantics vs the workspace method tree).
- **Ownership through compile**: `plugin:` carries onto compiled stage-graph.json nodes (absent on core nodes - shipped dist byte-identical). Invariants: plugin-owned slugs carry the plugin prefix; `plugin: aidlc` rejected; plugin names must not start with `aidlc-` (reserved for core - an `aidlc-<x>` plugin's runner dirs would land on core runner paths and silently clobber them; enforced at compile, compose, scope loading, and the packager); filename stem == slug enforced at compile; duplicate scope/agent names fail loudly naming both files.
- **Plugin-namespaced runner skills**: a plugin-owned stage/scope gets a runner named by its own slug (`/test-pro-integration`); core keeps `aidlc-<slug>` byte-identically. Scope-runner batch is data (`runner: true` frontmatter), not a code constant. Compose regenerates runners + the SKILL.md generated tables, so composed stages are typeable and visible immediately.
- **`select-plugins`**: transactional (snapshot + rollback - no torn installs), number-stable across round trips, closure-guarded (a selection stranding an enabled consumer of a disabled producer fails compile naming the plugin to enable), doctor-integrated (selection rows, hard-fail on selection/graph disagreement).
- **Disable removes what the plugin merged**: compose records the structural adds it actually applied to core stages in a per-plugin sidecar (`tools/data/plugin-contrib-<key>.json`); `select-plugins` strips a disabled plugin's merged produces/sensors/consumes/required_sections (sidecar) and spliced prose (fragment sentinels) from installed stage source, inside the same rollback transaction. Re-enable + next session start restores everything byte-identically (the plugin's compose hook re-merges). Compose no longer merges contributions for a plugin the current selection disables.
- **Active-workflow strand guard**: `select-plugins` refuses a change that would strand a running workflow (its scope, or a pending EXECUTE stage in its plan, owned by a plugin the new selection disables), naming each dependency and the remediation. Doctor hard-fails on a selection that already strands one, and lists selection-dropped `requires_stage` ordering edges as an advisory.
- **Generated surfaces**: the SKILL.md Stage Graph table is now generator output (`stage-table` verb, markers, `--check` drift guard; the hand-maintained table had already drifted in 2 cells); `state-template.md` documents the contract instead of hand-enumerating stages. Both follow the filtered graph under a selection.
- **Walking-skeleton decoupling**: `skeleton: on|off` scope frontmatter replaces the hardcoded six-name set - plugin scopes are first-class in the ceremony. `AWS_AIDLC_DEFAULT_SCOPE` naming a disabled scope falls back to the sole enabled plugin's first scope (a plugin-only install starts from plain `/aidlc`). A static probe (t225) prevents the name-coupling from regrowing.
- **Stage-file precheck at compose**: every plugin stage file is validated against the INSTALLED engine's schema before it lands - an invalid file (stale `bundle:` key, unknown fields, or a frontmatter-only empty body) is skip-and-dropped naming the file and the errors, the rest of the plugin composes, and the install's graph compile stays green (previously one bad copy bricked every later compile with no self-heal).

## Compatibility

- No selection = everything enabled; existing installs unaffected. Shipped dist data files (stage-graph/scope-grid/harness.json) are byte-identical.
- **Breaking**: `bundle:` frontmatter in third-party plugin trees no longer composes - rename the key to `plugin:`. The failure is safe (drop-logged, surfaced by `/aidlc --doctor`, install keeps compiling) but the plugin's content will not land until migrated.
- Breaking for CI/scripts: none.

## Test evidence

- New: t222 (runner naming + carry-through + aidlc- reservation), t223 (naming enforcement), t224 (selection journey incl. rollback, closure guard, contribution strip round-trip, strand guard, doctor rows), t225 (scope-name decoupling probe); t188 extended (buckets, stage-file precheck incl. bundle: stage + empty body + reserved names, compose-vs-selection advisory, bundle: rejection contract); t32 repointed to the generated region; t62/t64 pin the schema/parse rejection contract.
- Full deterministic tiers (smoke+unit+integration) green at tip; per-WP filtered slices green throughout.
- Live end-to-end drive on a fresh temp install: compose, select test-pro-only, verify runners/grid/table/doctor, re-enable with byte-identical stage numbers, unknown-name error, env-scope fallback under plugin-only selection, compose idempotence, disable-strip + re-enable restore round-trip.

## Version note

Claims 2.3.5 on top of v2@2.3.4 (rebased after #545 merged and took 2.3.4 + the t221 test slot; this branch's tests re-slotted to t222-t225).


apackeer:
Pushed the adversarial-review fixes as 8 commits on top of 29be05d5 - all 10 ranked findings plus the cleanup set. No version bump; the 2.3.4 CHANGELOG entry is amended where behavior changed.

**Severe findings**

- `c9189891` **version-skew guard**: composing a plugin onto a pre-2.3.4 engine (schema without the `plugin:` key) permanently broke that install's compile, with the retry marker re-failing every SessionStart. compose now probes the installed schema up front; on rejection it skips the plugin-owned stages/scopes/agents with a degraded drop naming the remediation and writes no retry marker for that cause. t188 gains an old-engine fixture.
- `82ff149c` **frontmatter-name collision guard**: a plugin file with a unique filename but a duplicate frontmatter `name:` bricked loadAgents/loadScopeMetadataAll (and with them compile, orchestrate, help, and the statusline). compose now checks each agents/scopes file's declared name against the installed roster and skips collisions with a drop naming both files; the statusline additionally fails open to slug display.
- `57cfc891` **composed-scope survival**: plugin selection destroyed composer-authored scopes twice over (filterScopeGrid stripped them when core was deselected; the transpose fabricated a flattened twin even when core stayed enabled). Composed entries (grid keys with no stage-frontmatter producer) are now excluded from transpose seeding and exempt from the selection filter - they round-trip byte-identical through any select-plugins run, and their runners survive the regen chain via written-grid membership.
- `1808bc78` **runner-prune safety**: pruning required only a command-signature match, so a user-authored skill quoting the runner commands was rm -rf'd silently. Generated runners now carry `generated-by: aidlc-runner-gen` and pruning requires it (one-release legacy rule for pre-marker installs; unmanaged dirs are listed on stderr, never deleted). An empty scope batch only prunes under an active selection - otherwise it warns and preserves. The eager `FIRST_BATCH` module const (import-time scope-dir I/O that could kill `write` on one malformed scope file) is gone; importers call `defaultScopeBatch()` at use time.

**Small findings**

- `2213b3e3`: select-plugins skips runner regen with a note on installs without a harness skills/ dir (Codex); the `AWS_AIDLC_DEFAULT_SCOPE` sole-plugin fallback announces the substitution on stderr; doctor counts an unparseable stage file as missing under an active selection (parse error in the fix message); t223/t224 `covers:` headers rewritten in the registry grammar - `subcommand:aidlc-utility:select-plugins` and `function:selectionAwareDefaultScope` (plus pluginsEnabled, mergeComposedScopes) flip to covered.

**Cleanup set**

- `4e684afc`: the stage-table/scope-table `--check` twins collapse into one parameterized helper and `replaceGeneratedRegion` shares the same marker validation; `stageEnabledBySelection` is exported once from aidlc-lib (compose keeps its self-contained copy with a sync comment); the scope-table BEGIN marker drops its em dash for the hyphen grammar (stale installed markers degrade to compose's existing advisory drop); the triplicated test scaffolding hoists into fixtures.ts `withEnvAndFreshCaches`; doctor reuses `artifactsRegistryFor` + `frontmatterBlock`; select-plugins emits a `PLUGIN_SELECTION_CHANGED` audit event (previous + new selection, registered and documented, event count 70 -> 71).
- `5871164d`: CHANGELOG amendments folding the user-visible changes into 2.3.4.
- `6c3fd8ae`: the fixtures hoist initially gave tests/harness/fixtures.ts a module-load-time core/ import, which broke t52's core-less sandbox (caught by the full integration tier); the cache resets now lazy-require core at call time.

**Verification**: full smoke+unit (155 files) and integration tiers green; per-WP filtered slices green before each commit; typecheck and `package.ts --check` green at every commit; live end-to-end drive on a temp install verified the skew fixture, the collision guard, composed-scope byte-identical round-trip + routing, decoy-skill survival through both prunes, the audit events, and the `bundle:` rejection drop.

leandrodamascena (changes requested):
CHANGES REQUESTED

This finishes the plugin mechanism from #475 and most of it holds up — I ran the selection journey and the compile carry-through and they work. But I found a few things that need fixing before this lands, and one is a real breaking change the PR description hides.

**The bundle → plugin rename is a breaking change, and the PR body says the opposite.** The description says a "read-side alias is kept" and that existing `bundle:` plugins "keep composing via the alias." I tested it: the shipped code rejects `bundle:` outright — `aidlc-stage-schema.ts:205` emits `bundle: was renamed; write plugin: for ownership`, no alias. So any third-party plugin written against #475 (which shipped `bundle:`) breaks. Worse, I traced it through compose: one `bundle:` stage file gets copied into the install and then aborts the *entire* graph compile, not just that stage, with no self-heal (`compose.ts:599`). The rename itself is fine, but the compat claim is false and the blast radius is unmitigated. Either keep the alias like the body claims, or drop the claim and add a migration note + skip-and-drop the bad stage instead of poisoning the whole compile.

**Disabling a plugin doesn't actually remove what it merged.** This is the orphan case I flagged on #475, still open. I composed test-pro into a fresh install, then ran `select-plugins aidlc` to disable it. The plugin's own stages/scopes go away, but its merged `produces`/`sensors`/`consumes` stay welded into the core stage source and the compiled graph. I traced why: `handleSelectPlugins` writes the selection and runs `aidlc-graph compile`, but never re-runs the compose hook that applied those contributions — so compile only filters the plugin's own nodes; the merged edits stay. `select-plugins` looks like it removed the plugin, but the contributions are still there. Either re-run a clean compose on select, or say plainly that disable hides the plugin's own content and keeps its merged contributions.

**select-plugins can strand an active workflow.** If I disable a plugin whose scope the current workflow is running under, the next `/aidlc` hard-errors and `--doctor` stays all green (`aidlc-utility.ts:444`). Needs a guard that refuses to disable a plugin the active workflow depends on, or a clear recovery path.

**Runner name collision.** A plugin named `aidlc-<x>` generates runner dirs identical to core runners (`aidlc-runner-gen.ts:88` returns the bare slug for plugin stages, `aidlc-<slug>` for core, so an `aidlc-`-named plugin lands on the same path), and `write` clobbers silently, so `/aidlc-<slug>` can route to the wrong stage. There's no naming guard — a plugin should not be allowed to claim the `aidlc-` prefix.

Smaller stuff, non-blocking:
- The `listOf` silent-drop I flagged on #475 is still there for produces/sensors — non-4-space YAML entries drop with no log, even though the consumes parser got that log in this PR (`compose.ts:654`).
- The closure guard only covers `consumes`; a `requires_stage` on a disabled core stage loses the ordering edge with no error (`aidlc-graph.ts:1462`).
- A malformed `plugin.json` still aborts the whole packager (no try/catch, `package.ts:843`).
- CHANGELOG says the empty-body stage guard is enforced by "the compose guard" — it's test-only, not a runtime check.
- Disabled-plugin `agents/` and `knowledge/` still load (the docs only promise stages/scopes/runners are filtered, so this is a doc-accuracy note).
- Version 2.3.4 collides with #545 — second to merge re-bumps.

Good parts I checked: t223 drives real rollback and the closure guard against the shipped code (not string literals), t224 is a real name-coupling probe, and the shipped dist data files are drift-clean. The selection transaction with snapshot/rollback is the right shape.

apackeer:
Thanks for the thorough pass - all four blockers were real, and all four are fixed, plus the non-blocking set. Rebased onto v2@2.3.4 first (#545 took 2.3.4 and the t221 slot), so this branch is now 2.3.5 with tests re-slotted to t222-t225.

**1. `bundle:` blast radius + the false compat claim.** You are right on both counts: the alias claim in the PR body was stale (the alias was removed deliberately mid-review - "bundle" should not exist as a live keyword anywhere), and the unguarded stage copy was a real brick: one `bundle:` stage file copied into an install aborted every later graph compile with no self-heal (the retry marker re-ran a compile that could never succeed, and no-clobber never replaces the bad file). Fixed at the copy chokepoint: compose now validates every plugin stage file against the INSTALLED engine's parser + schema before it lands - an invalid file (stale `bundle:`, unknown keys, or a frontmatter-only empty body) is skip-and-dropped with the file and errors named, the rest of the plugin composes, and compile stays green. The PR body and CHANGELOG now state the rename as breaking with the migration named (rename the key to `plugin:`); no alias. Pinned in t188 (bundle: stage file skip + compile-stays-green, empty-body skip).

**2. Disable now removes what the plugin merged.** The orphan case is closed for structural adds AND prose. Compose records what it actually added per target stage (never what was declared - so removal can never strip a value core already had) in a per-plugin sidecar at `tools/data/plugin-contrib-<key>.json`; prose fragments already carried their sentinel markers. `select-plugins` strips both on disable, inside the existing rollback transaction (stripped stage files join the snapshots). Re-enable + next session start restores everything - t224 pins the round-trip as byte-identical. Two design notes: the strip runs install-side off the sidecar because `select-plugins` cannot re-run compose (PLUGIN_ROOT is only known to the host harness when it fires the plugin's own hook, and compose is add-only anyway); and compose now skips contribution merges for a plugin the current selection disables, since merged contributions bypass the runtime selection filter and would have silently undone the strip on every session start.

**3. Strand guard.** `select-plugins` now enumerates every non-complete workflow across all spaces and refuses a change that would disable (a) the plugin owning a workflow's scope or (b) a plugin owning a pending EXECUTE stage in its plan - each dependency named, with the remediation (complete or park the workflow, or keep the plugin). Doctor gains a matching hard-fail row for a selection that already strands one (pre-guard or hand-edited harness.json). Pinned in t224 (scope case, pending-stage-under-core-scope case, completed-workflow non-block, doctor row).

**4. `aidlc-` prefix reserved.** Enforced at four chokepoints: graph compile rejects `plugin: aidlc-*` stage frontmatter, the compose stage precheck mirrors the same invariants (so a bad file drops instead of bricking compile), `loadScopeMetadataAll` rejects it in scope frontmatter (covers scope-runner dirs), and the packager refuses an `aidlc`/`aidlc-*` plugin directory at discovery and at `plugin build`. Pinned in t222 + t188.

**Non-blocking items**, all addressed:
- `listOf` produces/sensors now drop-log a parsed-N-of-M shortfall naming the 4-space rule, mirroring the consumes parser.
- `requires_stage` onto a disabled stage: deliberately NOT a closure error - the ordering edge is vacuous when the dependency never runs, and the shipped plugin-only flow legitimately runs plugin stages whose `requires_stage` names core stages (test-pro under a plugin-only selection would become un-selectable). Instead doctor lists every selection-dropped ordering edge as an advisory, so a surprising walk order is explainable. Happy to escalate it to a hard error if you disagree with the tradeoff.
- Malformed `plugin.json` now fails the packager with the plugin and file named, not a raw JSON.parse stack.
- CHANGELOG wording fixed: the empty-body guard is now a real runtime check (the compose stage-file precheck), and the entry says so accurately.
- Doc 18 now states explicitly that a disabled plugin's `agents/` and `knowledge/` stay loadable (not selection-filtered; inert since the dispatching stages are filtered), alongside the new disable-strip semantics.
- Version: rebased and re-bumped to 2.3.5; tests re-slotted to t222-t225 (t221 is #545's).

Full deterministic tiers green at the new tip; live disable-strip/re-enable drive green on a fresh temp install. PR body updated to match reality.

apackeer:
@leandrodamascena all four blockers plus the non-blocking set are fixed and replied to point-by-point above (branch rebased to 2.3.5, tiers green at the tip). When you get a chance, could you re-review? The one open judgment call is the requires_stage-onto-disabled-stage tradeoff (doctor advisory vs hard error) - happy to escalate it if you disagree.

leandrodamascena (approved):
Thanks for addressing the feedback @apackeer! I ran another round of review and couldn't find any bug.

PR #553, opened 2026-07-10
fix(codex): count explicit structured selections as human turns (2.6.44)

gadzbi123:
## Summary

- mint the existing HUMAN_TURN audit event after answered Codex request_user_input calls
- keep typed UserPromptSubmit handling unchanged
- pin the generated PostToolUse matcher and regenerate all distributions

This fixes approval and interview gates rejecting a structured selection until the user types a second acknowledgment.

## Testing

- bun test tests/unit/t150-codex-packaging.test.ts tests/unit/t132-hooks-doc-count-sync.test.ts tests/unit/t68-version-changelog-sync.test.ts
- bun scripts/package.ts --check
- full unit runner: 138/140 files passed; t140 and t142 could not load the optional @anthropic-ai/claude-agent-sdk dependency in this checkout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.44] - 2026-08-21
+
+Codex structured `request_user_input` selections now mint the same `HUMAN_TURN` evidence as typed prompts, without treating empty, cancelled, timed-out, auto-resolved, error, or malformed responses as human judgment. **Upgrade:** refresh `dist/codex/`, then re-run `bun scripts/package.ts codex trust --project "/absolute/project/path"` and replace the existing hook trust tables before starting a fresh Codex session.
+
+* The new PostToolUse registration routes `request_user_input` through the existing record-human-turn target only when `tool_response` is the Codex JSON-encoded `answers: Record<questionId,{answers:string[]}>` contract and every selected value is explicit, nonblank, and passes the shared cancellation-boilerplate floor.
+* Completed-looking `{}` payloads and cancellation metadata fail closed; substantive selections remain valid, and values such as `Abort` count when they exactly match an option the question actually presented.
+* Codex duplicate delivery replays the first response and emits exactly one `HUMAN_TURN`; projects without active workflow state remain untouched.
+
 ## [2.6.43] - 2026-08-21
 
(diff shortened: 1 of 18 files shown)


apackeer:
Thanks, Kacper. This addresses a real Codex gap: structured `request_user_input` selections still do not mint the human-turn evidence that typed prompts do, so the direction remains useful.

This path needs to preserve the framework's responsibility split: the LLM supplies the knowledge and context needed to present the question, the human supplies judgment by making a selection, and deterministic code verifies and records that decision. The current wiring collapses the last two responsibilities: `PostToolUse` proves only that the tool returned, but the adapter ignores `tool_response` and appends `HUMAN_TURN` unconditionally as though a human had decided.

Codex can return `{}` in headless mode, and #681 documents cancelled or automatically resolved questions arriving as completed-looking responses. Because `HUMAN_TURN` authorizes approval and interview gates, the adapter must require an explicit, nonempty human selection and reject empty, cancelled, dismissed, timed-out, auto-resolved, error, and malformed responses.

Adapter-level tests should cover those cases, a genuine selection, duplicate delivery minting exactly once, no active workflow state, and one real approval/interview path. #692 overlaps the downstream non-answer checks but does not add this Codex registration, so the two pieces need to land coherently.

The upgrade note also needs to tell Codex users to re-run the hook-trust step. There is no need to rebase this stale branch or resolve its old version slot; maintainers can port the corrected change and handle regeneration and versioning.

leandrodamascena:
Maintainer port pushed at `804910036`.

Codex structured selections now mint human-presence evidence without turning non-decisions into authority:

- Added a `request_user_input` PostToolUse registration using the existing `record-human-turn` target.
- The adapter accepts only the real Codex JSON-encoded response contract: `answers: Record<questionId, { answers: string[] }>`.
- Empty objects, malformed JSON, unsupported response shapes, blank answers, cancellation boilerplate, timeout/error/auto-resolved metadata, and partial multi-answer cancellations fail closed.
- Values such as `Abort` count only when they exactly match an option presented by that question.
- Question IDs remain opaque; an ID such as `error` is not mistaken for failure metadata.
- Duplicate Codex delivery emits exactly one `HUMAN_TURN`.
- Projects without active workflow state remain untouched.
- Typed UserPromptSubmit behavior remains unchanged.
- Codex hook wiring and trust hashes were regenerated.
- Upgrade guidance now explicitly requires regenerating and replacing the Codex hook-trust entries.
- The stale contribution was ported directly onto current `v2` and versioned as `2.6.44`.

Validation:

- Post-rebase focused slice: 93 tests, 1,794 assertions, 0 failures.
- Codex adapter, packaging, trust, and version slice: 50 tests, 362 assertions, 0 failures.
- The full deterministic profiles retained only the existing unrelated local timeout/baseline failures.

The PR is ready for CI and final approval.

leandrodamascena (approved):
The stale contribution has been safely ported onto current `v2` at `804910036`. Only explicit Codex structured selections mint `HUMAN_TURN`; non-decisions fail closed, duplicate delivery is idempotent, and trust/package validation is green. Approved pending CI.

PR #560, opened 2026-07-12
feat: aidlc single-command CLI - dark-launch dispatcher, grammar, and build (2.3.8-2.3.9)

apackeer:
## What this is

The aidlc single-command CLI, shipped dark: a noun-verb dispatcher (`aidlc <noun> <verb>`) that replaces path-typed invocations (`bun .claude/tools/aidlc-X.ts ...`), built as the user-facing half of the binary distribution plan. The grammar exists at `core/tools/aidlc.ts`, fully tested under bun and compiled/smoke-gated as a native binary - but no shipped string references it yet. The cutover (WP7) waits for a signed binary on PATH.

**Rebased 2026-07-14** onto v2 after the four-PR merge train (#550 2.3.5, #565, #562 2.3.6, #563 2.3.7): the base auto-retargeted to v2, this branch's versions shifted to 2.3.8/2.3.9, and two test slots moved (t230-handler-additions -> t231, t231-build-binaries -> t238; t232-t237 are claimed by merged and in-flight work). Previously rebased 2026-07-12 onto #550's review-fix head.

## The commits (landing order = the dark-launch safety order)

| Commit | What |
|---|---|
| 003ad5f | **Detectors first**: the stop hook and runtime-compile hook recognize BOTH command grammars (legacy regexes kept verbatim; new-shape branches inert until the grammar exists). 77-case pinned corpus (t226). |
| 6060be3 | **Entrypoint exports**: 19 tools export `main(argv)`, 11 hooks + 3 adapters export `run(input)`, zero side effects on import; spawned behavior byte-identical (t227/t228). |
| d0c6a49 | **Workspace verbs (2.3.8, user-visible)**: one shared two-token parser at both call sites (`space create teamB` now creates instead of switching to a space named "create"); verb names reserved at creation; doctor advisory; Kiro quoted-name fix (t229). |
| 056477c | **The dispatcher**: 29-route translating router, dev-spawn + compiled in-process modes, `help` / `help --all` with plumbing banner + alias table; 29 byte-parity cases old-vs-new shape (t230). |
| eb63cf9 | **Handlers (2.3.9, user-visible)**: `config get/list`, `plugin list/sync`, the `init` transition error (loud, one release, replaces the silent intent-birth alias), `upgrade` stub; emitted plugin hook probes aidlc-then-bun (t231). |
| 2e00daa | **Build script**: `scripts/build-binaries.ts` with mandatory smoke gates (run version+help from neutral cwd, empty-PATH version, bundle grep-gate); bytecode permanently banned with the reproduced failure documented (t238). |
| 7a59be1 | **Re-slot after the 2026-07-14 merge train**: versions 2.3.6/2.3.7 -> 2.3.8/2.3.9, handler test -> t231, build-binaries test -> t238, coverage registry regenerated. |

## User-visible changes (the complete list)

Everything else is byte-identical - the parity suite pins it. Two CHANGELOG'd releases ride along:

- **2.3.8**: workspace nouns gain `list/switch/create/birth` verbs (bare-name sugar preserved, legacy `space-create` accepted); `archive/rename/show` reserved; verb-shaped names refused at record creation; doctor flags pre-existing collisions; Kiro preserves quoted multi-word names.
- **2.3.9**: `config get/list`, `plugin list/sync`; `init` errors loudly for one transition release instead of silently creating an intent (its future meaning is data-tree scaffolding per the install-model decision); `upgrade` reserves its name.

## Evidence

- Post-rebase deterministic tiers at tip 7a59be1: smoke+unit **165 files / 0 failures** (stamp 2026-07-14T00-27-12Z); integration 102 files with 2 reds, both accounted for: t72 (green-alone rerun, latency flake) and t92 (pre-existing v2 baseline eslint/tsc environment issue, fails identically on untouched v2).
- 7 test files: t226-t231 + t238; `package.ts --check` clean; typecheck clean; coverage registry fresh.
- Old command shape stays a valid input forever (dev mode + third-party plugin manifests); the router translates new to old, so existing callers never break.
- The init meaning change was validated by its own blast radius: ~25 test files used the deprecated alias as their fixture driver and broke exactly as a stale caller would - all repointed to `intent-birth` in the same commit.

## Out of scope (later, per the plan)

WP7 cutover (rewrites the 67 authored + ~22-26 generated invocation strings, hook wiring, allowlists - only after a signed binary is on PATH) and WP8 docs. The sliced live pre-merge gate runs against the cutover PR; this PR ran the default deterministic tier per WP plus at tip.


leandrodamascena (changes requested):
CHANGES REQUESTED

I built the binary, ran the live engine, diffed the classifiers against base, and set up real workspaces. The dispatcher and new nouns are genuinely dark — nothing shipped invokes them, dev-spawn is byte-for-byte the legacy tool. But two things change live behavior, and one breaks a remedy this PR advertises.

One framing note: "no current behavior changes" isn't literally true. 2.3.8 is a deliberate, disclosed live workspace-grammar change (bare `intent list` now runs LIST instead of switching to a record named "list"). That's fine — the problem is the escape hatch it points to doesn't work.

**Required fixes (confirmed, live impact):**

**1. The `switch <name>` escape hatch is broken for verb-named records — `aidlc-lib.ts:571`.** `workspaceCommandUtilityArgv` returns `[noun, name]` for a switch, dropping the `switch` token, so the utility re-reads the name as a verb. This runs on the live orchestrate path and the Kiro adapter, not just the dark dispatcher. I reproduced it: `orchestrate next intent switch list` emits `intent list` (record unreachable), and `intent switch birth` emits `intent birth`, which *births a new intent* — a switch mutating state. On v2 you could create records named list/birth/create; after upgrade the advertised remedy fails for exactly those, while the doctor advisory and CHANGELOG promise it works. `archive/rename/show` survive (not utility verbs). t229 hides it — the "coverage" test calls the 3-token argv directly, which the engine never emits. Fix: emit `[noun, "switch", name]` (handlers already accept it) and test the engine path.

**2. The runtime-compile reject leg isn't inert — `aidlc-lib.ts:832`.** The PR adds `/\baidlc\s+runtime\b/` as a reject disjunct feeding the live PostToolUse hook. A legacy transition whose `--user-input` contains the words "aidlc runtime" flips base "fire" → PR "reject", suppressing the runtime-graph recompile — reachable today, no cutover. It self-heals (next clean transition catches up), so it's below #1, but it's a live matcher change contradicting "inert until cutover". Fix: keep the reject leg path-anchored, not a bare two-word substring.

**Majors (no current impact, but block the future binary cutover):**

**3. The compiled binary is broken for every delegate — `aidlc.ts:712`.** I ran `bun build --compile` on the shipped dispatcher: version/help work, but `doctor`/`state get`/`plugin sync` all throw `Cannot find module '/$bunfs/root/aidlc-utility.ts'`. The 15 tools are dynamically imported via a computed path `--compile` can't discover, so they never embed. No binary ships today — but this is the artifact WP7 would ship.

**4. The build gate is blind to #3 — `build-binaries.ts:383`.** `build-binaries.ts --target native` printed `native ok`, exit 0, while the artifact fails every delegate — the gates (version/help/grep/pathless) never delegate. Exactly the BYTECODE-1 "compiles but doesn't run" case it claims to guard. Add a gate that runs one real delegate against the compiled artifact.

**5. Plugin-compose hook couples to an `aidlc` on PATH — `package.ts:914`.** The emitted hook probes `command -v aidlc` and runs `aidlc plugin sync; exit $?` if found. Today it falls through to `bun compose.ts` (no-op, confirmed). But once a binary named `aidlc` is on PATH it runs plugin sync (exits 1 per #3) and the fallback is unreachable — plugin compose silently stops. Dormant.

**Non-blocking:** router drops `--project-dir` on workspace routes (post-cutover only; live channel is the env var). Stale init emitter left in `audit-format.md` (the PR fixed the parallel `12-state-machine.md`). Version 2.3.9 is consistent and correct, but open #572/#573 bump from a 2.3.7 base — coordinate merge order.

**Didn't fully verify:** whether the release names the binary `aidlc` on PATH (no `.github/workflows` in the repo) — that decides if #5 is dormant or live. Didn't run the Kiro/Codex adapters end-to-end (diffed the bodies; mechanical wrapper refactor).

The dark-launch discipline is careful and most holds. But #1 breaks an advertised remedy on the live path and #2 changes a live matcher, so land those first (plus the binary/gate fixes before any cutover).

apackeer:
Thanks for the thorough review. The requested fixes landed in f1f0b5f and were followed by runtime-data and dispatch hardening through the current head:

- Explicit `switch <name>` now preserves the `switch` token, while bare-name switch sugar retains its original two-token shape.
- The runtime-compile reject leg is anchored to a shell-command segment and no longer matches prose inside flag values.
- Compiled delegates use literal static imports so Bun embeds them.
- The binary build gate executes a real `plugin sync` delegate.
- Plugin compose falls back to Bun when the PATH binary returns nonzero.
- The stale audit-format init emitter was corrected.
- Compiled commands now resolve staged runtime data beside the executable, with doctor/runtime-data gates.

Regression coverage was extended across the detector, dispatcher, workspace, plugin-hook, binary-build, and doctor tests. Re-requesting review.

leandrodamascena:
Follow-up on my earlier CHANGES REQUESTED. I re-reviewed the reworked branch and my three original findings (workspace `switch` token, runtime-compile reject disjunct, dynamic-import bundling) are resolved. But I compiled the binary at `a6d8c3bc` and ran the delegated paths myself, and five independent regressions reproduce. Keeping the change-request open for these.

**P1 — the compiled binary is non-functional beyond version/help/doctor**

1. **Sensors are invisible in the binary.** `aidlc sensor list` is empty and `sensor describe linter` reports "Known ids: (none)", while the same commands under `bun` list all four sensors. Root cause: `sensorsDir()` at `core/tools/aidlc-graph.ts:337` resolves `join(__FILE_DIR, "..", "sensors")` — module-relative with **no executable fallback**, unlike its sibling `scopesDir()` (`aidlc-lib.ts`), which was given `resolveModuleOrExecutablePath(...)`. In a compiled binary `__FILE_DIR` is a `$bunfs` path and the sensor manifests aren't embedded (only `.ts` modules are), so `loadSensors()` returns an empty map. The exe-relative fallback that fixed scopes needs to reach `sensorsDir` too.

2. **Graph/asset resolution still falls to cwd-relative.** `aidlc graph compile` fails on a missing data/asset file. Even with the data dir co-located next to the binary, agent/scope resolution then fails on a cwd-relative path (`/private/tmp/agents`). `resolveDataDir` has the exe fallback, but not every asset resolver shares it, so the binary can't compile a graph outside a dir that happens to mirror the layout.

3. **Bolt/Swarm delegation re-invokes `$bunfs` paths.** `spawnSibling` (`core/tools/aidlc-bolt.ts:117`) spawns `bun run <fileURLToPath(new URL("./aidlc-state.ts", import.meta.url))>`. In the binary `import.meta.url` is `$bunfs/...`, so the worktree fork/merge steps shell out to `bun run /$bunfs/root/aidlc-state.ts` — which needs `bun` on PATH and points inside the binary's own FS. Compiled bolt/swarm can't delegate.

**P2 — grammar/parser gaps**

4. **Interleaved workspace flags misroute** (live, not binary-specific). `space --project-dir /tmp create teamB` **lists spaces instead of creating teamB** — the two-token workspace parser doesn't skip `--flag value` pairs before reading the verb. This one will bite at WP7 when the grammar goes live.

5. **`space-create` absent from the dispatcher.** `aidlc space-create teamB` → "unknown command or noun 'space-create'". The PR body says "legacy `space-create` accepted" — that's true on the old `bun .../aidlc-utility.ts space-create` path, but the new dispatcher grammar rejects it, so the claim is overstated for the CLI surface.

**Why t238 doesn't catch these:** the build gate's `realPluginSyncGate` copies a full `.claude` runtime tree into the project dir and sets `CLAUDE_PROJECT_DIR`, so the data files exist on disk right where a project-relative lookup finds them — which masks the exe-relative resolution bug. And the gate never runs `sensor list`, a graph compile outside that seeded layout, or a bolt/swarm delegation in the compiled artifact. So a binary that's broken for sensors/graph/bolt/swarm passes the gate green.

None of this is a production regression today — the binary is dark (not on PATH, WP7 out of scope). But shipping the build script + gate now, with the gate certifying a binary that's non-functional for its real delegated work, is the concern: the gate should exercise `sensor list`, an unseeded graph compile, and a bolt delegation in the compiled artifact before it can vouch for the binary. Happy to treat P1 as "accept dark, fix before WP7 cutover" if you'd rather scope it that way — but then the gate claims and the CHANGELOG's "routed commands reach their handlers" line should say so, and P2c (the interleaved-flag misroute) is a real parser bug worth fixing regardless.

leandrodamascena (changes requested):
My previous findings are resolved. I rebuilt and tested the current binary, but I found two remaining blockers:

1. Mutable commands can write into the packaged runtime. Running `plugin select` or `graph compile` against a project without an installed harness modifies files beside the executable, such as `harness.json` and `stage-graph.json`. Project commands should never modify the shared packaged runtime.

2. `--project-dir` is dropped for `hook`, `adapter`, and `statusline` routes. I confirmed that a hook from the current working directory runs even when `--project-dir` points to another project.

I also found some non-blocking issues:

- Relative `--project-dir` values break Bolt/Swarm re-entry.
- The runtime matcher still rejects quoted user input containing a separator followed by `aidlc runtime`.
- `__sensor-script-file` can execute any matching TypeScript file. This matters if the `aidlc` binary will be allowlisted as a trusted command.

The targeted tests, native binary tests, and package drift check pass. These issues are dormant while the binary remains dark, but the first two conflict with the claim that the packaged binary is safe and self-contained. I recommend fixing them before merge, or clearly scoping the binary as experimental until WP7.

apackeer:
@leandrodamascena I addressed the latest review findings and rebased the branch onto the current `v2` head in `4531b71`:

- Packaged runtimes are read-only fallbacks; mutable plugin/graph commands require an installed project harness.
- Hook, statusline, and adapter routes honor global `--project-dir`.
- Relative project paths survive Bolt/Swarm self-reentry.
- Runtime command matching is quote-aware.
- Sensor workers resolve registered IDs instead of caller-supplied TypeScript paths.
- Native binary gates cover these paths and packaged-runtime immutability.

The branch is now versioned `2.4.4`. Package drift, TypeScript, lint, focused dispatcher/plugin tests, and the native binary gate pass. Could you please re-review PR #560?

leandrodamascena (approved):
Approved. All five previous blockers have been addressed, and the implementation checks passed. The remaining t144 failure is a non-blocking test portability issue caused by `/tmp` resolving to `/private/tmp` on macOS.

PR #562, opened 2026-07-12
fix: advance the state file's Phase Progress rows at phase boundaries (2.3.6)

### LINKED ISSUE #556: [Bug]: The state file's `## Phase Progress` section is seeded once at init and never advanced — the code comment promises an Active/Verified flip that no code performs, so the section drifts stale against the Stage Progress checkboxes
## Summary

The state file (`aidlc-state.md`) carries a `## Phase Progress` section:

```
## Phase Progress
- **Initialization**: Active
- **Ideation**: Pending
- **Inception**: Pending
- **Construction**: Pending
- **Operation**: Pending
```

`aidlc-utility.ts` seeds this section once when the state file is created, and the code documents an intended lifecycle for it (`aidlc-utility.ts` around the `phaseStatus` helper):

> `// … Pending phases flip to Active on their phase-boundary advance and to Verified at phase completion.`

But no code performs that flip. I searched every tool for a writer that sets a phase row to `Active` or `Verified` after init — there is none. `handleAdvance` in `aidlc-state.ts` already computes `crossesPhaseBoundary` and emits `PHASE_COMPLETED` / `PHASE_VERIFIED` / `PHASE_STARTED` audit events on a boundary, but it updates only the `Lifecycle Phase` field and the Stage Progress checkboxes — it never touches the `## Phase Progress` rows.

So the section is a write-once field: it holds its init values for the whole run while the Stage Progress checkboxes advance underneath it. A few stages in, a reader sees every Ideation stage checked `[x]` but `- **Ideation**: Pending` right above — a visible contradiction. The comment describes behaviour the code does not implement (a code-comment drift), and the rendered section misinforms anyone reading the state file directly.

## Impact is display-only

I confirmed nothing consumes this section, so the drift does not affect routing:

- Engine routing reads the compiled `runtime-graph.json` plus the Stage Progress checkboxes; the phase it reads from state is the `Lifecycle Phase` field, not this section.
- `/aidlc --status` prints a "Phase Progress:" block, but it **recomputes** that block live from the checkboxes (`aidlc-utility.ts` `handleStatus`) — it does not read the `## Phase Progress` section.
- No hook, sensor, or skill parses the section.

So this is a correctness/clarity bug in a human-facing field, not a workflow-breaking one. But a state file that shows `Ideation: Pending` after Ideation is done reads as broken state to a user inspecting it.

## Two ways to fix

Either resolve the drift:

1. **Implement the documented flip** (make the field honour its comment). In `handleAdvance`, when `crossesPhaseBoundary` is true, flip the completed phase's row to `Verified` and the entered phase's row to `Active`. The existing `setField(content, "<Phase>", "<status>")` helper matches the `- **<Phase>**: …` lines directly and is a no-op when the label is absent, so it is a small additive change right after the `Completed` counter sync. Each phase name appears exactly once as a `- **<name>**:` field (only inside this section), so there is no collision with other fields.
2. **Or remove the section and the comment** if the field is not wanted — since nothing reads it, deleting it loses no behaviour and removes the drift surface.

I have taken approach (1) in my downstream IDE-format variant and verified it end-to-end: a non-boundary transition leaves `## Phase Progress` byte-identical, and a boundary transition flips exactly the completed phase → `Verified` and the entered phase → `Active`. (Note that a fix only corrects boundaries crossed *after* it lands; a run already mid-flow keeps its stale rows for the boundaries it already passed. A fresh run is correct from its first boundary.)

## Environment
- AI-DLC v2 2.3.3

apackeer:
## Summary

Fixes #556: the state file's `## Phase Progress` section was seeded once at intent birth and never advanced, so a few stages in it still showed `Ideation: Pending` above a fully-checked Ideation checkbox block. The `phaseStatus` comment promised an Active/Verified flip that no code performed. The section is display-only (routing reads `Lifecycle Phase` + the Stage Progress checkboxes; `/aidlc --status` recomputes its phase block live), so this is a correctness fix for the human-facing state file, not a routing change.

This takes the issue's option 1 (implement the documented flip) and closes two additional drift surfaces the issue did not cover: jumps and plan re-shaping.

## Changes

- **New `setPhaseProgress` helper** (`core/tools/aidlc-lib.ts`): flips one `- **<Phase>**:` row via the existing `setField` matcher. Silent no-op when the row is absent - an older or hand-edited state file must never fail a transition over a display row.
- **`advance` / `finalize`** (`core/tools/aidlc-state.ts`): at a phase boundary, completed phase -> `Verified`, entered phase -> `Active`, in the same atomic state write as the `PHASE_COMPLETED`/`PHASE_VERIFIED`/`PHASE_STARTED` trio. Non-boundary transitions leave the section byte-identical.
- **`complete-workflow`** (and a final-stage `finalize`): last phase -> `Verified`.
- **`jump execute`** (`core/tools/aidlc-jump.ts`): forward marks the source phase `Verified`, wholly jumped-over phases `Skipped`, target `Active`; backward returns reset phases (those with EXECUTE stages) to `Pending`. Jump already emits the phase-boundary audit trio, so the rows now track it.
- **`scope-change` / `recompose`** (`core/tools/aidlc-utility.ts`): re-derive only not-yet-reached rows (`Pending`/`Skipped`) against the new plan; `Verified`/`Active` rows record history and are never rewritten.
- **Birth seed made truthful**: birth completes every init stage and emits the init boundary trio itself, so the seed now reads `Initialization: Verified` + first post-init phase `Active` (previously `Active`/`Pending`, contradicting the audit events birth emits).
- `docs/reference/12-state-machine.md`: phase-machine prose + transition table updated to describe what the code now does (the table previously described this exact behaviour aspirationally).
- Version 2.3.7 + CHANGELOG + README badge; all four `dist/<harness>/` trees regenerated (`package.ts --check` clean).

Note for a run already mid-flow: rows stay stale for boundaries already passed and correct from the next boundary onward (as the issue predicted); a fresh run is correct from birth.

## Tests

- **New `tests/unit/t232-phase-progress-flip.test.ts`** (12 cases, spawns the real shipped CLIs): birth seeds for feature/bugfix scopes, byte-identical non-boundary advance, boundary flips for advance/finalize/complete-workflow, both jump directions, scope-change/recompose re-derivation, and missing-section tolerance.
- `t39` pin updated: `Initialization` at birth is now asserted `Verified` (was `Active`).
- t232 added to the coverage registry's manual `EXPECTED_NONE_TO_CLI` ratchet; registry regenerated.

## Verification

- Full deterministic tier (smoke+unit+integration, 254 files): the only failures were the coverage-registry honesty ratchet demanding the manual t232 entry (added; green on rerun) and 3 pre-existing `t92` linter/type-check sensor assertions that fail identically on the untouched v2 tip (verified via a detached baseline worktree).
- Rebased onto v2 @ 2.3.4 (#545); `t68` version/changelog sync, `t39`, `t232`, `package.ts --check`, coverage `--check`, and `bun run typecheck` all green at the new base.
- Every flip also exercised live against the `dist/claude` tools in a scratch project (birth -> 7 advances -> forward jump -> backward jump -> complete-workflow -> scope-change -> recompose).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.3.6] - 2026-07-12
+
+The state file's `## Phase Progress` section now advances with the workflow. Previously it was seeded once at intent birth and never touched again, so a few stages in it still showed `Ideation: Pending` above a fully-checked Ideation checkbox block - a visible contradiction for anyone reading `aidlc-state.md` directly (routing was never affected; the engine reads `Lifecycle Phase` and the checkboxes, and `/aidlc --status` recomputes its own phase block). The section now tracks every transition the tools already audit. **Upgrade:** re-copy your `dist/<harness>/` shell into the project; a run already mid-flow keeps its stale rows for boundaries it already passed, and corrects from its next boundary onward.
+
+* Intent birth seeds the section truthfully: `Initialization: Verified` (birth completes every init stage before handing off) and the first post-init phase `Active`; later phases stay `Pending`/`Skipped` by scope as before.
+* `advance` and `finalize` at a phase boundary flip the completed phase to `Verified` and the entered phase to `Active`; non-boundary transitions leave the section byte-identical. `complete-workflow` (and a final-stage `finalize`) flips the last phase to `Verified`.
+* Stage/phase jumps update the rows too: a forward jump marks the source phase `Verified`, wholly jumped-over phases `Skipped`, and the target phase `Active`; a backward jump returns reset phases to `Pending`.
+* `scope-change` and `recompose` re-derive only the not-yet-reached rows (`Pending`/`Skipped`) against the new plan; `Verified`/`Active` rows record history and are never rewritten.
+* A state file without the section (older installs) still transitions cleanly - the row update is a silent no-op, never an error.
+
 ## [2.3.5] - 2026-07-12
 
 Completes the plugin mechanism with **install-time plugin selection** and the full plugin content surface, proven at scale in a customer pilot. Plugins add, the install selects: a plugin can now ship scopes, agent personas, and knowledge alongside stages, every stage-enumerating surface is generated from the compiled graph, and a new `select-plugins` command lets an install choose which plugins' content its users see (core is the implicit `aidlc` plugin and can itself be deselected; a plugin-only install shows only that plugin's commands and scopes). The ownership frontmatter field is renamed from `bundle:` to `plugin:` (the old key is rejected with an error naming the fix). **Upgrade:** re-copy your `dist/<harness>/` shell and re-run any installed plugin's compose hook; a plugin tree still authoring `bundle:` must rename the key to `plugin:`. Existing installs without a selection are otherwise unaffected (an absent `plugins` key enables everything, and the shipped trees are byte-identical on core surfaces).
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.3.5-blue)
+![version](https://img.shields.io/badge/version-2.3.6-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-jump.ts
+++ b/core/tools/aidlc-jump.ts
@@ -21,6 +21,7 @@ import {
   type StageEntry,
   setCheckbox,
   setField,
+  setPhaseProgress,
   stageIndex,
   writeStateFile,
 } from "./aidlc-lib.js";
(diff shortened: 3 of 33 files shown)


leandrodamascena (approved):
APPROVE

I dug into this one. The Phase Progress section it fixes is display-only — routing reads Lifecycle Phase and the stage checkboxes, and `--status` recomputes live — so a wrong row here mis-displays but can't misroute. Nothing I found blocks the merge.

Here's what I actually checked. First I worried the `setField` matcher might rewrite the wrong line, so I read the state template: the five phase labels only show up as `- **<Label>**:` bullets under `## Phase Progress`, once each, and the regex is anchored with the `:` right after `**` — so `- **Construction**:` can't match `- **Construction Autonomy Mode**:`. No collision. Then I grepped all of `core/**/*.ts` for anything that reads those row values back, to make sure the "display-only" claim holds — zero readers for routing; the statusline computes from the checkbox blocks, not these rows. And I confirmed t232 isn't just asserting literals: no-op'ing setPhaseProgress flips exactly the row assertions to failing.

Two display inconsistencies I did reproduce, both worth a follow-up but neither blocking:

- **A double-boundary advance leaves the middle phase stale** (`aidlc-state.ts:1204`). advance flips only the phase you left and the one you entered. If a single transition steps over a whole phase, that middle phase keeps its birth `Pending` above its own all-`[S]` checkboxes — which is the exact row-vs-checkbox contradiction #556 is trying to kill, just relocated. The jump path handles this (it marks stepped-over phases `Skipped`), so the two paths disagree. No stock scope sandwiches an all-SKIP phase between two executing ones, so you won't hit it on a normal advance, but I reproduced it with `skip`: skip all of Inception, advance out of Ideation, and you land on Construction with `Inception: Pending`. I'd have advance/finalize mark fully-skipped middle phases the way jump already does, and add a double-boundary case to t232.
- **A forward jump marks the source phase `Verified` even when none of its stages ran** (`aidlc-jump.ts:382`). Jump forward from a phase's first stage and its row goes `Verified` above all-`[S]` checkboxes, while the phase you jumped clean over reads `Skipped` — same zero-completed state, different row, purely by where the cursor was. It's tracking the existing PHASE_VERIFIED audit event so I'm not calling it a bug, but `Skipped` would read truer.

Non-blocking housekeeping: version 2.3.7 collides with #560 (same version, same `## [2.3.7]` heading, both add a `t232-*` file). Different bases today, so no live conflict — whichever shares a lineage second re-bumps and renumbers. t232 also has no double-boundary case and I didn't trace the per-unit Construction swarm re-entry, both lower-risk and still display-only.

The mid-flow staleness the code comments already call out is accurate for the cases I traced. Good to merge.

Issue #529, opened 2026-07-07
[Feature]: debounce sensor firing - the Write/Edit hook re-fires the same sensor + stage + path tuple on every save, so doc-heavy stages accumulate hundreds of advisory fires per run


apackeer:
## AI-DLC Version

v2 branch, verified on HEAD `7c02d48` (2.2.6). Hook + dispatcher layer (all harnesses).

## AI-DLC Phase

Every phase whose stages write markdown; the pressure concentrates in Construction's per-unit design stages.

## Summary

The PostToolUse Write|Edit sensor hook dispatches every applicable sensor on every matching write, with no dedup or debounce: the per-entry loop spawns one synchronous subprocess per applicable sensor per matched file write (`core/hooks/aidlc-sensor-fire.ts:188-266`), and the hook's own comment notes the cost model: "Sync subprocess; user pays wall-clock per Write inside an active stage with applicable sensors" (`aidlc-sensor-fire.ts:195-196`). The audit-format documentation acknowledges the multiplication explicitly: "the PostToolUse Write/Edit hook fires the same sensor + stage + path tuple multiple times within a stage" (`core/knowledge/aidlc-shared/audit-format.md:164`), and a single tool call can fan out to four parallel fires, one per applicable sensor (`audit-format.md:166`).

Scale, from the shipped configuration alone: both document-shape sensors match every artifact write (`matches: "**/{aidlc-docs,intents}/**"` in `core/sensors/aidlc-upstream-coverage.md:8` and `core/sensors/aidlc-required-sections.md:8`), the four construction design stages are `for_each: unit-of-work` (frontmatter of `core/aidlc-common/stages/construction/{functional-design,nfr-requirements,nfr-design,infrastructure-design}.md`) with a ~20-document matrix per unit (#502), and a document is normally authored across several Write/Edit calls. N units x ~20 documents x several saves each x 2 matching sensors puts fires per run in the hundreds, exactly where the run is already longest.

Three costs, all structural rather than incidental:

1. **Wall-clock.** Each fire is a synchronous `spawnSync` the user waits on, per write, per sensor.
2. **Audit-ledger noise.** Every fire appends a `SENSOR_FIRED` row plus a paired terminal row (`audit-format.md:158`). Re-fires whose verdict has not changed add rows that carry no new information, and audit-walking consumers must already pair by `Fire id` precisely because the same tuple recurs (`audit-format.md:164-166`).
[...]

PR #563, opened 2026-07-12
fix: upstream-coverage matches real citation forms over the stage's deliverable set (2.3.7)

### LINKED ISSUE #557: [Bug]: `upstream-coverage` matches only the bare slug literal per single file, so directory-path / split-across-files upstream citations read as unreferenced — a construction run logs hundreds of false SENSOR_FAILED
# AI-DLC Version
v2 (`v2` branch), 2.3.3. `core/tools/aidlc-sensor-upstream-coverage.ts` + the dispatcher `core/tools/aidlc-sensor.ts` (harness-neutral — lives in `core/`, so all harnesses).

# AI-DLC Phase
Every phase whose stages declare `consumes:`; the volume concentrates in Construction's per-unit design stages (`functional-design`, `nfr-requirements`, `nfr-design`, `infrastructure-design`).

---

## Summary

`upstream-coverage` checks that a stage output references each artifact in the stage's `consumes:` list. Its match is a case-insensitive regex anchored on the bare slug: `\b<slug>\b` OR `[[<slug>]]` (`aidlc-sensor-upstream-coverage.ts`, the per-slug loop). The dispatcher fires it **per written file** (`aidlc-sensor.ts` `handleFire` threads a single `--output-path`), and the per-slug loop tests that one file's body.

Two structural mismatches between what the sensor demands and how this framework's artifacts actually cite upstream make it report a legitimately-referenced upstream as `unreferenced`:

1. **Citation form.** Artifacts cite upstream by **directory path** or **backticked filename** in their provenance header, not by bare slug. A representative `nfr-design` output header reads (paraphrased): *"basis/trace: `construction/api-service/nfr-requirements/`, `functional-design/`, `decisions.md` …"*.

   That header covers `performance-requirements`, `security-requirements`, `scalability-requirements`, `reliability-requirements` (all produced under `nfr-requirements/`) by naming the producing-stage directory once. The slug `performance-requirements` never appears as a standalone token, so `\bperformance-requirements\b` misses it and the sensor reports it unreferenced. (Note: `\b<slug>\b` *does* already catch a backticked `` `<slug>.md` `` — the `.` is a word boundary — so filename citations of the exact artifact are fine; the gap is the directory-path form and item 2 below.)

2. **Per-file vs per-stage.** A stage emits several artifacts and splits its upstream citations across them: in `nfr-requirements`, `business-rules.md` cites `business-logic-model` but `tech-stack-decisions.md` does not — and should not; a tech-stack decision doc has no reason to name the domain model. Because the sensor fires per file and demands the full `consumes:` set in each, every file that legitimately omits an upstream irrelevant to it fails. Coverage is a property of the stage's whole output, not of each file.

Neither is a bug in a specific artifact — they are how the framework's authoring conventions and the sensor's matching model diverge. The result is that the sensor's failures are dominated by false negatives.

## Evidence (one real Construction run, 2.3.3)

A greenfield MVP run that reached Construction logged, from its audit shard:

- **186 `SENSOR_FAILED`**, of which **178 are `upstream-coverage`** (the other 8: 7 `linter`, 1 `required-sections`). upstream-coverage fired-and-failed in 18 of the run's stages.
- Re-scoring only the **final** deliverables (excluding `*-questions.md` and `memory.md` scaffolding) with the shipped regex: of the slugs the sensor marked `unreferenced`, **~97% are still unreferenced in the finished artifact** — i.e. these are not mid-draft states that pass on the last save, they never pass.

I then re-scored the same deliverables against the two mismatches above (count a producing-stage directory path reference as coverage; evaluate coverage over the stage's sibling `*.md` union rather than the single fired file). The `unreferenced` slug count dropped ~81–89%, and the residue was dominated by genuinely-absent references (e.g. `business-logic-model` in `nfr-design`, which an NFR design legitimately may not cite) — i.e. the signal survives, the noise clears.

## Why this is distinct from #529

#529 (debounce sensor firing) addresses the **volume** of fires — the same tuple re-firing on every save — and proposes firing document-shape sensors once at the stage boundary. That is real and complementary, but it assumes the terminal verdict is correct and characterises the failures as "half-written draft … only passes on the last one." The evidence above contradicts that assumption for this sensor: the **final** state still fails ~97% of the time. Firing once at the boundary would therefore still emit a large batch of false `SENSOR_FAILED` at each stage's gate — the noise moves to the gate, it does not clear. This report is about matching **accuracy**, not firing frequency; the two fixes are orthogonal.

## Impact

`SENSOR_FAILED` is advisory (`aidlc-sensor.ts` header: "Sensor outcomes are advisory") so this does not block the workflow. The cost is:

- **Audit-ledger noise** — hundreds of `SENSOR_FAILED` rows (+ a findings detail file each) that are mostly false, which buries the minority of real coverage gaps this sensor exists to surface.
- **Latent risk if #431 lands.** If a future `blocking` severity lets a `SENSOR_FAILED` halt a gate, this sensor's current false-negative rate would block on citations that are in fact present. Accuracy should be fixed before (or with) making it able to gate.

## Fix directions (two independent root causes)

These are separable — either helps, both compose:

- **A. Broaden the match to the citation forms the framework actually uses.** In addition to `\b<slug>\b` / `[[<slug>]]`, count a reference to the artifact's **producing-stage directory** (e.g. `nfr-requirements/`) as coverage. The dispatcher already resolves each consume's producer (`producersOf` in `aidlc-sensor.ts` `presentConsumes`), so the producing-stage slug can be threaded to the sensor alongside the artifact slug without new graph plumbing.
- **B. Evaluate coverage over the stage's output, not one file.** Score the union of the stage directory's deliverables rather than the single fired file, so a citation in a sibling artifact counts. This can be done without changing firing granularity (the sensor reads sibling `*.md` at evaluation time), which keeps it independent of #529's firing-cadence change. **Caveat worth calling out:** the union should exclude scaffolding (`memory.md`, `*-questions.md`) — otherwise a slug that appears only in the Q&A file or the rule memory would falsely count as covered, trading a false negative for a false positive.

A design question underneath both: is a directory-path citation *intended* to satisfy coverage, or should artifacts be expected to name each consumed slug explicitly? That's a call for the sensor's owner — the two fixes assume the former (matching the current authoring convention), but if the latter is intended, the fix belongs in the stage authoring guidance / templates instead, and the sensor is behaving as designed (just louder than useful).

## Environment
- AI-DLC v2 2.3.3
- Evidence from a single greenfield MVP run reaching Construction, on the Kiro IDE harness (IDE 1.0.116). The harness is incidental here: the sensor + dispatcher are harness-neutral `core/` code, and the finding was reproduced by reading the run's audit shard and re-running the sensor over the on-disk artifacts, so the matching behaviour is identical on every harness and independent of the IDE version.

apackeer:
Fixes #557.

## Problem

`upstream-coverage` checked each consumed artifact with a per-file `\b<slug>\b | [[<slug>]]` grep against the single fired file. Two structural mismatches with how the framework's artifacts actually cite upstream made its failures mostly false (the reported Construction run logged 178 false `SENSOR_FAILED` out of 186, and ~97% of flagged slugs were still "unreferenced" in the finished artifacts):

1. **Citation form.** Artifacts cite upstream by producing-stage directory path in their provenance header (`construction/<unit>/nfr-requirements/`), never re-naming each slug, so `\bperformance-requirements\b` missed a legitimately-cited upstream.
2. **Per-file vs per-stage.** A multi-artifact stage splits its citations across sibling deliverables: in `nfr-requirements`, `business-rules.md` cites `business-logic-model` but `tech-stack-decisions.md` does not (and should not). Firing per file demanded the full `consumes:` set in each.

Plus one latent false positive: `\b` treats `-` as a word boundary, so the consume `requirements` matched *inside* the token `nfr-requirements`.

## Fix (both directions from the issue, composed)

- **A. Producer-directory citations count.** The dispatcher threads `--consumes` as `artifact:producer` pairs (producer resolved via the existing `producersOf`, no new graph plumbing). The sensor counts a citation of the producing stage's directory as a whole path segment (leading `nfr-requirements/` or trailing `../nfr-requirements`) as coverage of every artifact that stage produces. A longer sibling slug (`some-nfr-requirements/`) never matches; the producer as a bare prose token (no path form) does not count either.
- **B. Stage-level evaluation.** A new `--deliverables` flag carries the stage's produces union (`produces` + `optional_produces` through the same `templateEligibleArtifacts` filter required-sections uses). The sensor evaluates coverage over the union of those files in the fired file's directory, so a citation in a sibling deliverable counts. Firing granularity is unchanged, so this stays independent of #529 (debounce), which addresses volume, not accuracy.
- **Scaffolding excluded from both sides** (the caveat the issue calls out): a slug appearing only in `memory.md` / `*-questions.md` never counts as coverage, and a fire on a scaffolding file is judged against the deliverables. A scaffolding write before any deliverable exists passes vacuously with `reason: "no deliverables on disk yet"`.
- **Hyphen-substring false positive killed:** slug anchors are now lookarounds excluding `[\w-]`. Backticked filenames, wikilinks, and case-insensitivity behave as before.

Back-compat: a bare-slug `--consumes` without `--deliverables` (direct invocations, older dispatchers) keeps the previous per-file semantics. Sensor JSON gains `scanned_files`.

On the issue's design question (is a directory-path citation *intended* to satisfy coverage?): nothing in the stage protocol or memory rules ever instructs artifacts to cite consumed slugs by name, so the artifacts are following the only convention that exists and the sensor's model is what diverged. This PR matches the sensor to the convention rather than adding authoring guidance whose enforcement would depend on LLM compliance.

## Verification

- New unit test `t233-upstream-coverage-matching` (16 cases) spawns the shipped sensor at the process boundary (same pattern as t155): directory-citation forms positive and negative, sibling-union scoring, per-fire verdict convergence, scaffolding exclusion, vacuous fire, hyphen false positive, and all back-compat forms.
- `t92` Group M extended to pin the dispatcher argv (`artifact:producer` pairs + `--deliverables` with the scaffolding-filtered produces union).
- End-to-end via the real dispatcher against the shipped stage graph: the issue's exact `nfr-design` provenance-header shape now passes; on `nfr-requirements` a split citation passes while a genuinely-absent reference still fails, listing only the true gap.
- Full default tier (smoke+unit+integration): 254 files, green except three pre-existing `t92` linter/type-check environment cases that fail identically on the untouched `v2` baseline.

`SENSOR_FAILED` stays advisory; this also clears the accuracy prerequisite noted in #431 (blocking severity) and in the issue's "latent risk" section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.3.7] - 2026-07-14
+
+The `upstream-coverage` sensor now matches the citation forms the framework's artifacts actually use, clearing the hundreds of false `SENSOR_FAILED` audit rows a Construction run accumulated. It previously demanded each consumed artifact's bare slug in every single written file; artifacts instead cite upstream by producing-stage directory path in their provenance header, and multi-artifact stages legitimately split citations across sibling deliverables. Real coverage gaps still fail. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* A citation of the producing stage's directory (e.g. `nfr-requirements/` in a basis/trace header) now covers every artifact that stage produces; the producer slug must appear as a whole path segment, so a longer sibling slug never satisfies it.
+* Coverage is evaluated over the union of the stage's declared deliverables in the output directory, not the single fired file, so a citation in a sibling artifact counts. Scaffolding (`memory.md`, `*-questions.md`, `*-timestamp.md`) is excluded from both sides: a fire on it is judged against the deliverables, and a slug appearing only there never counts.
+* A scaffolding write that fires before any deliverable exists now passes vacuously with `reason: "no deliverables on disk yet"` instead of failing every consume against an empty body.
+* False-positive fix: a consume slug no longer matches inside a longer kebab token (`requirements` inside `nfr-requirements` previously counted as a reference).
+* Sensor JSON output gains `scanned_files`; the dispatcher threads `--consumes` as `artifact:producer` pairs plus a new `--deliverables` flag. Direct invocations with the old bare-slug `--consumes` and no `--deliverables` keep the previous per-file behavior.
+
 ## [2.3.6] - 2026-07-12
 
 The state file's `## Phase Progress` section now advances with the workflow. Previously it was seeded once at intent birth and never touched again, so a few stages in it still showed `Ideation: Pending` above a fully-checked Ideation checkbox block - a visible contradiction for anyone reading `aidlc-state.md` directly (routing was never affected; the engine reads `Lifecycle Phase` and the checkboxes, and `/aidlc --status` recomputes its own phase block). The section now tracks every transition the tools already audit. **Upgrade:** re-copy your `dist/<harness>/` shell into the project; a run already mid-flow keeps its stale rows for boundaries it already passed, and corrects from its next boundary onward.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.3.6-blue)
+![version](https://img.shields.io/badge/version-2.3.7-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/sensors/aidlc-upstream-coverage.md
+++ b/core/sensors/aidlc-upstream-coverage.md
@@ -3,13 +3,14 @@ id: upstream-coverage
 kind: deterministic
 command: bun {{HARNESS_DIR}}/tools/aidlc-sensor-upstream-coverage.ts
 default_severity: advisory
-description: Checks the output prose references the upstream artifacts the stage frontmatter declares it consumes
+description: Checks the stage's deliverables reference the upstream artifacts the stage frontmatter declares it consumes
 category: document-shape
 matches: "**/{aidlc-docs,intents}/**"
 input_schema:
   output_path: string
   stage_slug: string
   consumes: string[]
+  deliverables: string[]
 output_schema:
   pass: boolean
   unreferenced_artifacts: string[]
@@ -18,12 +19,27 @@ timeout_seconds: 5
 
 # upstream-coverage sensor
 
-Reads the stage frontmatter `consumes:` list and checks the output prose
-references each upstream artifact (by name or wikilink).
+Reads the stage frontmatter `consumes:` list and checks the stage's
+deliverables reference each upstream artifact. A reference counts in any
+of the forms the framework's artifacts actually use:
 
-Pure derivation from frontmatter — no per-stage config needed.
+- the artifact slug as a standalone token (not embedded in a longer
+  kebab token: `requirements` inside `nfr-requirements` does not count)
+- a wikilink (`[[<slug>]]`) or backticked filename (`` `<slug>.md` ``)
+- the producing stage's directory as a path segment (e.g. a provenance
+  header citing `nfr-requirements/` covers every artifact that stage
+  produces)
+
+Coverage is a property of the stage's whole output, not of each file:
+the check runs over the union of the stage's declared deliverables in
+the output directory (scaffolding - `*-questions.md`, `*-timestamp.md`,
+`memory.md` - is excluded from both sides), so a citation in a sibling
+deliverable counts.
+
+Pure derivation from frontmatter - no per-stage config needed.
 
 ## Failure mode
 
 Emits `SENSOR_FAILED` and writes detail listing artifacts declared in
-`consumes:` that don't appear anywhere in the output prose.
+`consumes:` that appear in none of the stage's deliverables in any
+accepted form.
--- a/core/tools/aidlc-sensor-upstream-coverage.ts
+++ b/core/tools/aidlc-sensor-upstream-coverage.ts
@@ -1,12 +1,19 @@
 import { existsSync, readFileSync } from "node:fs";
+import { basename, dirname, join, resolve } from "node:path";
 import { errorMessage } from "./aidlc-lib.ts";
 
 interface Result {
 	pass: boolean;
 	consumes: string[];
 	unreferenced: string[];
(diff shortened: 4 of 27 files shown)


leandrodamascena (approved):
APPROVE

I ran both new regexes at the process boundary myself, traced the dispatcher and fire-hook path, and checked the sensor against the real stage graph. No crash of the fire hook, no pass-everything over-correction, and no new false-negative on a real citation form. The sensor is advisory anyway, so nothing here blocks.

What I checked:

- **The hyphen false positive is dead, no new miss.** slug matcher is now `(?<![\w-])<slug>(?![\w-])|[[<slug>]]`. I ran it: `requirements` no longer matches inside `nfr-requirements` (the `-` is in the lookbehind), while every real form still matches — backticked `performance-requirements.md`, wikilink, plain prose, comma/paren/EOL adjacency, case-folded. The only things that stop matching are longer kebab tokens (`the-requirements-doc`, `performance-requirements-v2`), which are different artifact names, so that is correct rejection.
- **The producer-dir pattern matches a whole path segment and rejects siblings.** I ran `(?<![\w-])<p>(?=/)|(?<=/)<p>(?![\w-])` against the cases: leading `nfr-requirements/foo` matches, trailing `../U1/nfr-requirements` matches, mid `a/nfr-requirements/b` matches; longer siblings `some-nfr-requirements/x` and `x/nfr-requirements-v2/y` do not; bare prose `the nfr-requirements stage` does not. So a deliverable that cites its producer only by directory path now counts — the #557 false-positive class this targets.
- **Fail-soft and still discriminating.** Malformed `--consumes`/`--deliverables`, a missing sibling, a trailing-colon pair — all return exit 0 with valid JSON; the only exit-1 is a missing `--output-path`, guarded by the dispatcher's existsSync before spawn, and the fire hook records a non-zero sensor as a drop and still exits 0. And it is not pass-everything: an artifact cited nowhere still comes back `unreferenced`, and adding one producer-dir citation in a sibling deliverable flips it to pass. The scaffolding exclusion (memory/questions/timestamp) matches the dispatcher's `templateEligibleArtifacts` filter, and back-compat holds (bare-slug + no deliverables = old per-file behavior plus the hyphen fix).

Non-blocking notes:

- **The producer-dir premise is a widening, not an enforced convention.** I grepped the shipped stage bodies, knowledge files, and templates — nothing actually instructs an agent to write a provenance-path citation. So the new alternative accepts more forms (can only reduce false-positives), which is safe, but the design comment's "artifacts cite upstream by provenance path" is not something the shipped prose guarantees. If you want that convention to hold, it is a follow-up (authoring guidance).
- `producersOf(slug)[0]` threads only the first producer — unreachable today (all 122 artifacts have exactly one producer) and would only cause a false SENSOR_FAILED if hit, since the slug citation still counts independently.
- Transient false-fail before all deliverables land (a consume cited only in a not-yet-written sibling), self-clears on the next fire; advisory, and the code comment calls it out.
- Minor: `.` is not in `[\w-]`, so `requirements` matches inside `v1.requirements` — a false-pass direction, but not a regression (old `\b` did the same) and not a realistic citation form. t233 case 4 asserts `pass:false` without pinning `unreferenced`, but with one consume + one deliverable the array can only be the one slug.
- Version 2.3.8 is consistent and free today (open PRs sit at 2.3.7) — just confirm nothing grabs it before merge.

PR #568, opened 2026-07-13
feat: three-role ensemble - mode as communication topology (2.5.0)

apackeer:
# feat: three-role ensemble - mode as communication topology (2.5.0)

Closes North Star **Goal 1** ("Mimic what we practice in the real world"): a
stage may be executed by a configurable ensemble of agents - Owner,
Collaborator, Verifier - with the collaboration pattern a per-stage knob and
the semantics consistent across all four harnesses.

Built through an interactive design session with the maintainer (seven gated
decisions), live transport spikes on Claude/Codex/Kiro, and a three-angle
adversarial subagent review whose findings drove a rebuild of the
collaboration mechanics before this PR was cut.

## What changed

**`mode` is now the stage's communication topology** - who talks to whom
while the body runs:

| mode | topology | who writes |
|---|---|---|
| `inline` | voices (conductor adopts personas) | conductor (unchanged) |
| `subagent` | hub-and-spoke: supports become real, mutually-blind dispatched spokes | spokes write contribution files; lead integrates and owns `produces[]` |
| `pipeline` (new) | chain: links collectively author in declared order, each seeing all upstream work | links edit the artifacts directly; the final link leaves them complete |
| `mob` (new) | mesh as bounded rounds with the human in the room | round-1 collaborators write contribution files in parallel; lead integrates |
| `agent-team` | reserved (future native-bus transport; `mob` is the portable mode) | - |

Key properties, uniform across Claude Code, Kiro CLI, Kiro IDE, and Codex CLI:

- **Everyone writes, the owner collates** - the real-world working model.
  Each dispatched collaborator writes
  `<stage>/contributions/<agent-slug>.md` (identity-marker first line,
  Contribution + Positions); dissent is durable on disk, not ephemeral
  return text. The lead alone edits the stage's `produces[]` artifacts.
- **Deterministic completion evidence.** The engine refuses
  `report --result approved` on a mob (or subagent-with-supports) stage
  while a declared collaborator's contribution file is missing or lacks its
  identity marker - the tenet's determinism concern enforced at the state
  transition, not by prose compliance. Escape hatch:
  `AIDLC_DISABLE_ENSEMBLE_EVIDENCE=1`.
- **The human is a mob participant, not a post-hoc approver.** Unresolved
  objections are triaged: judgment calls (both positions legitimate)
  surface to the human mid-stage through the standard questions machinery;
  knowledge disputes get one confirm-or-maintain round. Autonomous
  Construction skips the mid-stage turn - dissent is recorded and surfaces
  at the final-batch gate; it never halts.
- **The conductor is the bus.** Only it delegates (the existing invariant,
  enforced on every harness); serialized dispatch on harnesses that cannot
  parallelize preserves the who-sees-what contract exactly.
- **The Verifier is untouched.** The 12a reviewer loop composes after the
  body on every mode; a NOT-READY re-invokes the lead alone (the ensemble
  convenes once).

**Shipped stage changes:**

- `user-stories` -> `mode: mob`, gaining developer + quality collaborators:
  the North Star's mob-elaboration ritual verbatim (PM Owner, Dev/QA/Design
  Collaborators, Product Leader Verifier).
- `reverse-engineering` -> `mode: pipeline` (from `subagent`): its body was
  always a 2-link chain (developer scans, architect synthesizes and writes);
  the mode now tells the truth. Zero behavior change - v2 already dispatched
  both agents.

**Kiro roster closure:** all 14 personas are delegation targets on Kiro
CLI/IDE (9 new agent JSONs per harness dir + conductor `trustedAgents` +
IDE grants); collaborator configs carry read+shell plus an
`aidlc/spaces/**`-scoped write grant for their contribution files.

**Swarm-trigger guard:** `aidlc-graph compile` emits a stderr `[advisory]`
when the per-unit Construction build stage carries a mode other than
`subagent` (that shape silently leaves the autonomous swarm path).

## Transport spikes (evidence-backed follow-ups, not in this cut)

Live spikes verified the native-transport upgrade paths this design leaves
open: nested subagent dispatch works on Claude (depth-2 Task chain observed)
and on Codex behind `[agents] max_depth = 2` (depth-2 spawn recorded in
session rollouts; depth-1 fail-closes); Claude's experimental teams primitive
hosts a true peer-messaging room with an extractable audit log. Kiro CLI
(2.12.1) REFUSES nested delegation at runtime - a delegated agent's config
may carry the `subagent` tool (validate accepts it) but the runtime withholds
it at depth 1 - so the conductor-simulated transport is the permanent Kiro
floor, which is exactly what v1 ships everywhere; Claude/Codex native
upgrades are documented follow-ups that change no stage file.

## Tests

- `t235-ensemble-modes` (new): enum + coupling validation, directive
  carry-through, compile-advisory behavior.
- `t236-ensemble-evidence-gate` (new): the approve-path refusal - missing
  files named per agent, malformed identity marker caught, escape hatch,
  idempotent replay never blocked.
- `t62`/`t65` (census: mob == user-stories, pipeline == reverse-engineering,
  no ensemble stage without supports, agent-team zero) / `t37` (inline
  supports stay voices) / `t148` (IDE grants derived from CLI JSONs) /
  `t66` (golden export regenerated) repinned to the new contract.
- Tier: smoke + unit + integration green (evidence block in the PR
  conversation).

## Upgrade

Re-copy your `dist/<harness>/` shell into the project; on Kiro CLI/IDE also
re-copy `.kiro/agents/` (nine new delegation-target configs + updated
conductor `trustedAgents`).


leandrodamascena (changes requested):
**CHANGES REQUESTED**

I found two failures in shipped execution paths and three additional correctness gaps affecting supported plugin/custom-stage flows.

1. **[P1] The Kiro developer collaborator cannot write its required contribution**  
   `core/aidlc-common/stages/inception/user-stories.md:9`

   `user-stories` now dispatches `aidlc-developer-agent` as a gate-checked mob participant, but its Kiro CLI configuration lacks an `fs_write.allowedPaths` entry. Without `--trust-all-tools`, unattended execution cannot create its contribution and the evidence gate refuses approval indefinitely.

   Grant it access to contribution directories. The existing design and quality grants should also be narrowed from `aidlc/spaces/**` to `aidlc/spaces/**/contributions/**`, preserving lead-only ownership of state and produced artifacts. Update the developer-agent description with its new role.

2. **[P1] `--single` bypasses the ensemble evidence gate**  
   `core/tools/aidlc-orchestrate.ts:3141`

   `report --single` returns through `handleSingleReport()` before this check. The generated and documented standalone `user-stories` runner can therefore complete without any collaborator contribution, although the same result is rejected in a normal workflow.

   Extract the evidence validation and invoke it from both reporting paths.

3. **[P2] Per-unit evidence is resolved from the wrong directory**  
   `core/tools/aidlc-orchestrate.ts:3150`

   This checks `<phase>/<stage>/contributions/`, while the protocol places per-unit evidence under `<phase>/<unit>/<stage>/contributions/`. A supported plugin/custom per-unit ensemble cannot pass with correctly placed evidence.

   Validate every unit contribution directory, or reject per-unit ensemble modes during graph compilation.

4. **[P2] The autonomous-swarm exemption is broader than swarm eligibility**  
   `core/tools/aidlc-orchestrate.ts:3144`

   `isAutonomousSwarm` checks only `mode: subagent` and persisted autonomous state. It omits Construction phase, `for_each: unit-of-work`, DAG availability, skeleton exclusions, and actual convergence. A plugin subagent ensemble can consequently bypass evidence without running as a swarm.

   Reuse the complete swarm eligibility predicate and require a converged settlement before applying this exemption.

5. **[P2] Plugin collaborator agents cannot be dispatched on Kiro**  
   `harness/kiro/agents/aidlc.json:53`

   Plugin composition copies an agent Markdown persona, but Kiro requires a native agent configuration and registration in `trustedAgents`. Plugin ensemble stages compile successfully but cannot dispatch their declared plugin agents on Kiro CLI or IDE.

   Generate the native Kiro projection and trusted-agent registration during composition, or reject this combination at compose time.

**Non-blocking notes**

- The first-line identity marker is structural evidence, not deterministic proof that dispatch occurred. The documentation should avoid claiming more unless it correlates contributions with stage-scoped completion receipts.
- `docs/reference/11-contributing.md:166` still documents `mode` as only `inline` or `subagent`; add `pipeline` and `mob`.
- `aidlc-architect-agent` lacks the same Kiro write grant, but its current ensemble role is pipeline-based and requires no contribution file.

apackeer:
Thanks for the thorough review - all five findings are addressed as of d81b128 (rebased onto v2 at 2.3.10).

**1. Kiro developer collaborator write grant** - `aidlc-developer-agent.json` (Kiro CLI and Kiro IDE) now carries the `fs_write.allowedPaths` grant, and its description names the collaborator role. One deliberate deviation from the suggested remediation: the design-agent and quality-agent grants were NOT narrowed from `aidlc/spaces/**` to `aidlc/spaces/**/contributions/**`. Both are lead agents elsewhere - `rough-mockups.md` and `refined-mockups.md` declare `lead_agent: aidlc-design-agent`, `build-and-test.md` and `performance-validation.md` declare `lead_agent: aidlc-quality-agent` - so they legitimately write produced artifacts outside `contributions/` on those stages, and the narrowing would have broken them. The developer agent has no lead role, so a broad grant was never needed there in the first place; it gets the same shape the other collaborators have.

**2. `--single` evidence bypass** - the check is extracted into `checkEnsembleEvidence()` and invoked from both reporting paths, including `handleSingleReport()`. The standalone `user-stories` runner is now rejected without collaborator contributions exactly like the workflow path. Covered in t236.

**3. Per-unit evidence directory** - the check now resolves each unit's own `<phase>/<unit>/<stage>/contributions/` directory and validates every unit the report covers, matching the stage protocol's placement.

**4. Autonomous-swarm exemption** - the evidence gate now reuses the real swarm eligibility predicate (Construction phase, `for_each: unit-of-work`, mode subagent, skeleton exclusions, resolvable Bolt DAG) and additionally requires a converged settlement, so a plugin subagent ensemble that is not actually running as a swarm no longer bypasses evidence. One nuance surfaced by the full tier: the per-unit coverage guard, which shares the eligibility predicate, must NOT require full convergence - it exists precisely to tolerate an in-flight multi-batch swarm (approving batch one must not block on batch two's units, pinned by t186). So the coverage guard uses plain eligibility while the evidence gate keeps the settled predicate.

**5. Plugin collaborator agents on Kiro** - compose now rejects a plugin stage that would dispatch a plugin-owned agent under mob, pipeline, or subagent-with-supports mode, with a drop naming the file and the remediation, since a native agent-v1 JSON plus trustedAgents registration cannot be derived from a Markdown persona. Rejection over projection per the smaller-and-safer option you offered.

Non-blocking items: `docs/reference/11-contributing.md` now documents `pipeline` and `mob`, and the identity-marker prose is softened to structural completion evidence rather than proof of dispatch.

Full default tier at the tip: 263 files, the only reds are the pre-existing t92 (typescript 7 exit-code drift, the unpin ships in #573) and one t197 live-agent latency flake that passed green-alone.

leandrodamascena (approved):
This is my re-review of #568 (2.5.0 — `mode` as stage communication topology + the deterministic ensemble evidence gate). My earlier CHANGES REQUESTED rested on one thing: the Kiro `aidlc-developer-agent.json`, a gated user-stories mob support, was missing its `fs_write` grant — a permanent Kiro-only wedge. Arden fixed it, and I re-checked it along with the engine lines that moved since.

**Verdict: APPROVE (with non-blocking notes).**

**What I checked:**

- **Developer-agent grant.** `harness/kiro/agents/aidlc-developer-agent.json` and the kiro-ide twin both now carry `fs_write.allowedPaths: ["aidlc/spaces/**"]`, matching design and quality; the dist copies match, and the description was updated to name the user-stories mob role. The gate reads `aidlc/spaces/<space>/intents/<slug>/inception/user-stories/contributions/<agent>.md`, which is under `aidlc/spaces/**`, so the grant covers the write — no prompt on an unattended Kiro run, the file lands, the gate passes.
- **Every gated mob support is granted.** Only `user-stories` is `mode: mob` (design/developer/quality — all three granted on both Kiro trees). `reverse-engineering` is pipeline (ungated). `code-generation` is the only `subagent` stage and has empty `support_agents`, so the gate short-circuits. No ungranted gated support anywhere. `architect` stays `null` — benign (pipeline only).
- **The evidence gate, including the ~34 moved lines.** The delta extracts the swarm predicate into `eligibleAutonomousSwarmBatches()` (loose, mid-run-tolerant) and `isSettledAutonomousSwarm()` (all-converged, used only to exempt a settled swarm). Both report call sites are scoped to not-completed stages, so idempotent replay is absorbed. The marker check survives BOM/CRLF/leading+trailing whitespace and rejects a wrong/malformed first line. The final commit reverted an intermediate version that wrongly used the strict predicate for the #368 coverage guard — the shipped code uses the loose one, which is correct.
- **No existing stage newly wedged.** `code-generation` still empty-supports; `reverse-engineering`'s subagent→pipeline move is load-bearing (as subagent-with-supports it would gate and wedge).
- **I mutation-tested the two pins myself:** neutering `checkEnsembleEvidence` to `return {ok:true}` flips 6 of 9 t236 cases to fail (the gate is a biting pin, not decorative); reverting the developer grant reds t148 (my blocker class is guarded for the current roster).
- Enum + coupling consistent; prose (§5/§11, conductor.md, user-stories.md, four SKILL.md) matches the gate byte-for-byte; `bun scripts/package.ts --check` clean on all four trees; version 2.5.0 free, t68 passes.

**Confirmed bugs: none.**

**Non-blocking notes:**

1. **Grant coverage is pinned by name, not derived from the graph.** t148 loops a hardcoded roster `[developer, design, quality]` — I confirmed it reds if you revert developer's grant, so my specific blocker is guarded. But no test derives "is a support on a mob/subagent-with-supports stage → must carry the Kiro `aidlc/spaces/**` grant." So the wedge *class* is unguarded: a future PR adding a new gated mob support whose Kiro JSON lacks the grant would ship green and wedge. Latent, not live (today only architect/composer lack the grant, and neither is a gated support). Worth a follow-up: derive the assertion from the graph's ensemble `support_agents`.
2. **Hollow floor (informational).** The gate is a one-line marker match — a stub file with only the marker passes. Intended structural floor per §5 (proves the collaborator convened, not that the contribution is substantive).
3. **Weak ALLOWED assertions in t236 (informational).** The three ALLOWED cases assert a negative (`not.toContain("ensemble must convene")`) that also passes without the gate; the refusal cases hold the line. If someone touches t236 later, assert the approve actually succeeded.

Net: the required fix is in and pinned, the moved engine lines are a contained refactor plus a stricter guard, and I found no live wedge, grant gap, corruption path, or enum throw.

leandrodamascena (changes requested):
**REQUEST CHANGES**

I re-reviewed the changes added after the previous approval at `d81b128`. The current tip is `cd15400`, with seven additional commits and a much larger workflow/state-machine delta.

I found five issues:

1. **[P1] Swarm convergence evidence is not scoped to a stage or run**
   `SWARM_UNIT_CONVERGED` records only the batch and unit name. The artifact and ensemble guards then accept matching rows after the latest `STAGE_STARTED`. A late result from an earlier run, or another swarm stage using the same unit names, can satisfy the current stage and bypass its artifact checks.
   References: `core/tools/aidlc-swarm.ts:266`, `core/tools/aidlc-orchestrate.ts:830`, `core/tools/aidlc-state.ts:928`
   Please include the stage and run/attempt identity in the event and require both consumers to match them.

2. **[P1] Revised artifacts bypass reviewer and learnings revalidation**
   The initial flow runs the reviewer, verification, and learnings before opening the approval gate. After a rejection, the lead can change the artifacts and report `revised`; the engine checks only structural completion evidence before reopening the gate. The revised artifacts can therefore be approved using an outdated READY verdict and outdated learnings result.
   References: `harness/claude/skills/aidlc/SKILL.md:77`, `core/aidlc-common/protocols/stage-protocol.md:970`, `core/tools/aidlc-orchestrate.ts:3703`
   Please rerun reviewer, verification, and learnings after material revisions before reopening the gate.

3. **[P2] Resume choices are accepted but not executed**
   `next --resume` offers four options: resume, redo, jump, or start fresh. `handleResumeReport()` only checks that the answer is non-empty and always returns “Resume choice accepted.” Selecting redo, jump, or start fresh therefore resumes the current stage instead of performing the requested action.
   References: `core/tools/aidlc-orchestrate.ts:1824`, `core/tools/aidlc-orchestrate.ts:3466`
   Please parse and route each supported choice, or offer only the resume action.

4. **[P2] Practices Discovery promotion remains valid after rejection**
   Promotion freshness is checked only against the latest `STAGE_STARTED`. If promotion succeeds and the gate is subsequently rejected, the existing `PRACTICES_AFFIRMED` receipt still authorizes approval after revised drafts are produced. Those revisions may never be promoted.
   Reference: `core/tools/aidlc-orchestrate.ts:2909`
   Please require the affirmation receipt to be newer than the latest rejection/revision boundary.

5. **[P2] OpenCode plugin composition can leave a dangling agent reference**
   The stage precheck treats a plugin agent as viable based on its file shape. Native OpenCode emission applies additional name-collision checks later. An agent can therefore satisfy the stage precheck but be rejected during emission, leaving the composed stage without its native dispatch target.
   References: `scripts/plugin-hooks-template/compose.ts:588`, `scripts/plugin-hooks-template/compose.ts:1145`
   Please apply the complete native-agent precheck before accepting stages that reference plugin-owned agents.

**Validation performed**

- `bun scripts/package.ts --check`: passed for all five harnesses.
- Standard smoke/unit/integration tier: otherwise green.
- `t188-plugin-compose`: passed 58/58 when rerun alone after one parallel timeout.
- Claude-dependent live tests could not run because the available AWS credentials were expired.

apackeer:
Thanks for the re-review - all five findings are addressed as of 450c5fbe. Each was independently re-verified against the code before fixing; two came back with scope corrections worth noting.

**1. [P1] Swarm convergence evidence scoped to the stage attempt** - fixed, with one design deviation from the suggested remediation. The codebase has no run/attempt id concept, so the attempt identity is the floor itself: each `SWARM_UNIT_CONVERGED` row now carries `Stage` plus `Run floor` (the stage's latest main-workflow `STAGE_STARTED` timestamp at emission, `core/tools/aidlc-swarm.ts`), and every consumer counts a row only when BOTH fields match the current attempt exactly. The duplicated scans in `aidlc-orchestrate.ts` and `aidlc-state.ts` are replaced by one shared reader (`swarmConvergedUnits` / `latestMainWorkflowStageStarted` in `aidlc-lib.ts`), so emitter and consumers compute the identity with the same function. Verification nuance: the floor was already per-stage-slug, so sequential cross-stage reuse was excluded by timestamp; the confirmed holes were a late `finalize` retry against a prior attempt's preserved worktree (current timestamp, stale work) and the empty-floor degrade failing open cross-stage - the exact-match closes both. Rows written by earlier versions fail closed (re-fan; finalize's re-verify makes that safe) - CHANGELOG carries the upgrade note. Tests: t201 cases 10-12 (stale-attempt row rejected, cross-stage row rejected, unstamped legacy row rejected), t134 asserts the emitted stamp.

**2. [P1] Revised artifacts re-validated before the gate reopens** - fixed for the reviewer half; the learnings half is deliberately declined. Stage-protocol Part 0, section 12a, conductor.md, and all five harness SKILL.md files now bind: when a rejection-driven revision changes a `produces[]` artifact and the directive carries a reviewer, re-run the 12a reviewer before reporting `revised` (fresh dispatch record, fresh `## Review` verdict replacing the stale one, same lead-alone loop and iteration budget). The learnings ritual stays once-per-stage: it is advisory and additive by design (section 13) and validates nothing about the artifact, so there is no stale result an approval can ride. Engine enforcement is not possible today - there is no reviewer-verdict audit event to check - so this lands as protocol binding pinned by t234; a `REVIEWER_VERDICT` event plus an engine check on the `revised` arm is the follow-up if you want it hardened. While fixing this we also found Part 0 itself contradicted the new section-13 audit ordering (it opened the gate before the completion message); Part 0 is rewritten so `report --result awaiting-approval` fires after the logged learnings answer, matching the QUESTION_ANSWERED-before-STAGE_AWAITING_APPROVAL proof.

**3. [P2] Resume choices routed** - fixed via per-choice directives (option: parse and route). `handleResumeReport` matches the answer against the engine's own question wording and returns the move: redo names `aidlc-jump.ts execute --target <current> --direction redo --scope <scope>`; jump instructs asking for the target then `next --stage <slug>`; start fresh routes through the second-intent flow (`next --new-intent`); an unrecognized answer errors listing the accepted choices. The engine still never mutates on `report` - it names the move and the conductor runs it, consistent with how scope-change/config-change/jump dispatch. All five SKILL.md ask rows now route every menu answer through the report. Tests: t118 SP4c (all four choices plus garbage, state byte-identical after each round-trip). Docs: reference 03-orchestrator resume options and the session-management guide corrected (the old "Start fresh archives the intent" claim never matched the engine).

**4. [P2] Practices promotion invalidated by rejection** - fixed. `hasFreshPracticesAffirmationReceipt` now floors on the latest of `STAGE_STARTED`, `GATE_REJECTED`, and `STAGE_REVISING` for practices-discovery, so a receipt minted before a rejection no longer authorizes approval of revised drafts; re-promotion after the rejection restores it. Stage prose and the recovery protocol state the invalidation explicitly. Tests: t236 - stale-receipt-after-rejection refused, re-promotion accepted, and an unrelated stage's rejection does not wedge the gate.

**5. [P2] OpenCode dangling agent reference** - fixed. `pluginShipsViableOpencodeAgent` now applies the full precheck including the name-collision axis against the installed native roster, so a stage referencing a collision-doomed persona no longer composes. Adjacent fix found while verifying this: `copyTreeNoClobber` ran the transform before the precheck, so one frontmatter-less plugin agent threw inside `emitOpencodeNativeAgent` and aborted the plugin's entire compose (knowledge/sensors/tools and the graph recompile all skipped) instead of the promised skip-and-drop; the precheck now runs first on the pre-transform text. Tests: t188 - colliding shipped agent drops its stage, malformed agent drops alone with later trees still composing, plus the negative control the Kiro/Codex twins already carried.

Also in 450c5fbe from a parallel re-review of the same delta: the guard's invocation regex had a residual O(N^2) on brace/paren runs (80k `{` measured at 10.5s, past Kiro's 15s hook cap) - the path-prefix class now excludes `{`/`(` and t242 pins 160k-run inputs; conductor.md's "do not consult the engine until the gate approves" sentence contradicted the revision-loop report binding and is rewritten; the opencode e2e cred skip-guard keys on the Bedrock MODEL prefix instead of `CLAUDE_CODE_USE_BEDROCK` (which configures Claude Code, not opencode); t73/t74 additionally require a learnings-flavored QUESTION_ANSWERED before the gate (the ordering pin alone passed if the ritual was skipped, since interview answers emit the same event).

Validation at 450c5fbe: typecheck, `package.ts --check` (all five dists), lint, t68, and the coverage-registry check all green; smoke+unit+integration tier 275/276 with the single red (t20 workspace scanner, a surface this commit does not touch) green solo on rerun; targeted runs t201 12/12, t236 25/25, t118 15/15, t134 13/13, t242 12/12, t234+t34+t37 156/156.

leandrodamascena (approved):

PR #569, opened 2026-07-13
feat(reviewer): enforce the §12a reviewer step on every completion route (2.5.5)

SiddhJog:
### What & why

The stage `reviewer` (stage-protocol section 12a) was compile-baked data but a prose-driven action: the run-stage directive carried reviewer + reviewer_max_iterations, but invocation was one bullet in a prose checklist - no audit trace, no gate dependency. A diligent model ran it; a rushed one silently skipped it, and nothing recorded or required it.

This is the tactical Track 1 of the reviewer-reliability work: make the reviewer step enforceable and observable without restructuring the directive.

### Changes

- New audit events `REVIEW_REQUESTED` / `REVIEW_COMPLETED` (framework event total 72 -> 74), emitted by a new tool actor:

  `aidlc-log.ts review --stage <slug> --reviewer <agent> [--unit <u>] [--iteration <n>] [--single] [--verdict <READY|NOT-READY>]`

  The `--verdict` form emits `REVIEW_COMPLETED`; without it, `REVIEW_REQUESTED`. `--single` tags the receipt with the synthetic `single-stage:<slug>` workflow id so an isolated stage-runner's review can never satisfy the main workflow.

- **All four completing transitions refuse a reviewer-bearing stage without a fresh terminal review.** `aidlc-state.ts approve`, `advance`, `finalize`, and `complete-workflow` (the same seams the artifact guard covers, same `!alreadyMarkedCompleted` gating) require a `REVIEW_COMPLETED` matching the stage AND the declared reviewer, recorded since the attempt floor. Hard on the review having happened, soft on the verdict - a NOT-READY-after-cap still lets the human approve at the gate with findings noted.

- **Freshness floors.** The receipt window is floored at the stage's latest `STAGE_STARTED` / `GATE_REJECTED`, plus stage-agnostic `WORKFLOW_STARTED` / `STAGE_JUMPED` floors (deliberately over-invalidating: a jump re-opens stages without emitting their own floor events, so fail-closed beats precise). On the approve path the check runs AFTER the gate-revision backstop, so a backfilled `GATE_REJECTED` (an unrecorded revision at an open gate) invalidates the pre-revision review and the approve refuses with the re-record directive. Under `Construction Iteration: unit-major` the `STAGE_STARTED` floor is skipped for per-unit stages (the protocol documents that those rows can land after the per-unit work).

- **Per-unit enforcement.** On `for_each: unit-of-work` stages the reviewer fires once per unit, so every applicable unit must carry its own receipt (`--unit`). The unit list resolves through the new shared `resolveBoltDag` (cache-first, self-heals a stale runtime graph from `unit-of-work-dependency.md`; malformed fails closed). Units whose kind prunes the stage's `produces` to empty owe no review - they never receive a directive.

- **Autonomous Construction is verified, not exempt.** `invoke-swarm` directives now carry `stage` / `stage_file` / `reviewer` / `reviewer_max_iterations`; the conductor reviews each unit inside its Bolt worktree and records receipts with `aidlc-log.ts review --project-dir <worktree>`; swarm `finalize` refuses to merge a claimed unit without a terminal `REVIEW_COMPLETED` after that unit's `BOLT_STARTED` boundary (the lying-conductor guard now covers reviews). `audit-fork` copies the clone token into the worktree so worktree receipts merge back into the main audit.

- **Deliberately NOT engine-enforced: `report --single`.** An isolated stage-runner run structurally cannot reach the four completing handlers (it commits a synthetic-id audit pair only and never touches the main pointer), and it has no gate to protect. Its reviewer step stays prose-driven (SKILL.md single-runner branch), and its receipts are `--single`-tagged so they cannot leak into the main workflow's guard. This is a design decision, not a gap.

- **Prose**: stage-protocol section 12a records dispatch (step 1) and verdict (step 3); all five harness SKILL.md gate branches (claude, codex, kiro, kiro-ide, opencode) carry the dispatch record, the verdict recording, the autonomous reviewer boundary, and the `--single` note.

- The superseded `readBoltDagUnits` / `readBoltDagUnitKinds` graph readers are deleted (replaced by `resolveBoltDag`).

### Behavior change

Reviewer-bearing stages now must record their review before ANY completing transition. The shipped conductor prose does this automatically; a workflow that never records a review hits a clear error naming the reviewer and the exact command to run. Stages with no reviewer are unaffected.

### Tests

- t115 R1-R11: refusal/unblock matrix incl. the four-handler seam (R9), persisted-autonomy Inception bypass (R10), single-receipt isolation (R11).
- t205: revision-backstop ordering (backfilled reject invalidates the pre-revision review).
- t208: kind-pruned (vacuous) units owe no review; mixed sets require only applicable units.
- t48 / t186: stale-DAG self-heal keeps N-unit enforcement.
- t209 test 9: unit-major reviews survive a later STAGE_STARTED.
- t135 / t201: swarm finalize refuses unreviewed claimed units; receipts merge back.
- t28 / t81: event-count pins 72 -> 74; audit-format registry reconciled.

Full default tier at the pushed tip: 276 files, 5631 assertions, 0 failures.

### Round-2 fix mapping (for re-review)

All five blockers from the 2026-07-16 review plus the autonomy-scoping blocker from the 2026-07-17 comment are addressed in the second commit ("close completion and autonomous bypasses"): completion seams -> four handlers; backstop ordering -> guard moved after backstop with state persisted on refusal; vacuous per-unit -> kind-pruned exemption; DAG self-heal -> shared resolveBoltDag; unit-major floor -> STAGE_STARTED skipped there + workflow/jump floors; autonomous exemption -> removed, replaced by in-worktree review + finalize receipt verification. The opencode harness (merged since this PR opened) is covered.

### Notes for reviewers

- `dist/<harness>/` changes are regenerated output; review the authored surfaces (`core/`, `harness/`) and spot-check the projections.
- Rebased onto v2 @ 2.5.1 (#595); version 2.5.5, re-check the slot at merge turn.


leandrodamascena (changes requested):
CHANGES REQUESTED

The direction is right and the audit events are worth having. But I found a few ways a reviewer-bearing stage can reach `[x]` with no fresh review — which is the exact thing this PR is trying to prevent. I traced the guard and reproduced these against the PR head.

**1. The guard is in the wrong place — a direct `approve` skips it.** The check sits in `report` (orchestrate.ts:3223), but `report` just shells out to `aidlc-state.ts approve`. The other three approve guards (artifact, human-presence, revision) all live in `handleApprove` on purpose — the comment at state.ts:745 says a guard only in `report` "is bypassable" because agents call `state approve` directly on recovery (issue #366). I reproduced it: at the gate with no review, `report --approved` errors correctly, but `aidlc-state.ts approve requirements-analysis` directly commits — GATE_APPROVED=1, REVIEW_COMPLETED=0. Fix: move the check into `handleApprove` like the other three.

**2. A stale review satisfies a re-approve.** `hasTerminalReview` (orchestrate.ts:985) greps the whole audit tail with no lower bound. Its sibling `swarmConvergedUnits` floors the same read at the latest STAGE_STARTED for this reason. I reproduced it: approve with a review, redo-jump the stage, approve again with no new review — passes, clearing the gate for a revised artifact nobody re-reviewed. Needs a STAGE_STARTED floor (plus a GATE_REJECTED anchor for the reject/revise case).

**3. Per-unit stages approve on one review.** The guard matches the slug only; `handleReview` has no `--unit`. Five reviewer-bearing stages are per-unit and all four SKILL.md files say "the reviewer fires once PER UNIT" — but one recorded review clears the whole stage, so units 2..N can be unreviewed. t186 case 9b bakes this in. Fix: record a Unit field and require one review per unit.

**4. The autonomous swarm halts at the reviewer gate.** code-generation is reviewer-bearing and the swarm stage. The settle reaches the guard, but the swarm-settle prose (SKILL.md:61) never records a review, and nothing in the swarm flow does. So an unattended run stops at the gate. It has an escape (the error prints the recovery command), so it's a halt, not a hard deadlock — but it's a regression and the PR's own t186 test 13 masks it. Fix: wire the review into the swarm settle, or exempt the swarm the way the coverage guard is.

Non-blocking: the enforcement is defense-in-depth (the tool writes the row from flags without checking a reviewer actually ran) — same shape as the human-presence mint, and the prose is honest about it. The real win is observability: a missing REVIEW_COMPLETED shows up in the audit. Also 2.3.5 will collide with the open #562/#563 on a shared lineage.

Blockers 1–3 each let a reviewer-bearing stage reach done with no fresh review, so I'd fix those before merge.

apackeer:
I traced the guard and re-verified @leandrodamascena's four blockers against the PR head - all four hold as described (I reproduced the reasoning for 1 and 2 against `core/tools/aidlc-state.ts:746-748` and the `swarmConvergedUnits` floor at `core/tools/aidlc-orchestrate.ts:770-790`). Rather than repeat those, here are findings NOT already covered by that review. None of these change the overall direction, which I agree is right.

**1. The recorded `Reviewer` field is never checked.** `hasTerminalReview` (`core/tools/aidlc-orchestrate.ts:990`) matches on the `Stage:` field only, so a row recorded with the wrong reviewer (a typo, or the conductor naming itself) satisfies the precondition. `node.reviewer` is already in hand at the check site (`:3223`), so also matching the `Reviewer:` field is nearly free and closes the self-certified-review hole. Worth folding in when the check moves into `handleApprove` per blocker 1.

**2. Stale event counts in docs this PR partially updates.** The PR adds the two new rows to the emitter table in `docs/reference/12-state-machine.md:220-221` but leaves the count above unchanged: `docs/reference/12-state-machine.md:176` still says "**71 events**, grouped below into 17 categories ... splits the same 71 into 19". Likewise `docs/reference/06-hooks-and-tools.md:340` still says "71-event taxonomy" and the Interaction row at `:352` still reads "4 | DECISION_RECORDED, GATE_APPROVED, GATE_REJECTED, QUESTION_ANSWERED". The repo's doc policy asks for these in the same commit. (t48 passes because it checks table rows, not the prose counts.) Also the updated t28 header comment says "72 events, 18 categories" (`tests/unit/t28-audit-event-sync.test.ts`, line ~24 of the file header), which matches neither the before state (71/19) nor the after (73/19).

**3. Coverage registry marks units this PR actually tests as UNCOVERED.** `tests/.coverage-registry.json` lands `REVIEW_REQUESTED`, `REVIEW_COMPLETED`, and `aidlc-log review` with `status: UNCOVERED`, yet t115's new R1-R5 tests exercise all three through the CLI. Adding `covers: subcommand:aidlc-log:review, audit:REVIEW_REQUESTED, audit:REVIEW_COMPLETED` to the t115 header and regenerating the registry would record that honestly.

**4. `Iteration` is documented as required but emitted conditionally.** `audit-format.md` lists `Iteration` under Required Fields for both events, but `handleReview` only sets it `if (flags.iteration)`. Either make `--iteration` required or mark the field optional in the registry (the `SCOPE_DETECTED` row already has an "(optional; ...)" notation to copy).

**5. `REVIEW_REQUESTED` is optional in prose and unused by the gate.** Step 1 emission is "(Optionally emit ...)" in stage-protocol section 12a and nothing consumes the row - the same "a rushed model skips it" weakness this PR exists to fix, so expect it to be rarely emitted in practice. Either drop it or make the SKILL.md step 1 require it; as-is it will mostly be dead registry weight.

**6. Duplicate regex-escape helper.** `escapeRegexLiteral` (`core/tools/aidlc-orchestrate.ts:996`) duplicates the exported `escapeRegex` in `core/tools/aidlc-lib.ts:4181`, which this file already imports from and which `findAllEvents` itself uses for the same audit-block matching.

**7. Test helper swallows failures.** `logReviewReady` in t186/t206/t208 ignores the `spawnSync` status; if the log call ever fails, the failure surfaces later as a confusing gate error in a different test assertion. One `expect(res.status).toBe(0)` inside the helper keeps failures local.

**8. Minor: CHANGELOG whitespace noise.** The diff adds stray blank lines around the 2.2.18 and 2.1.5 entry boundaries (e.g. a double blank line now precedes the 2.2.17 heading) - unrelated to the change, easy to drop while rebasing for the version re-slot.

For what it's worth I ran the touched test slice (`t115`, `t28`, `t81`, `t186`, `t206`, `t208`, `t68`) plus the `t48` emitter drift test on the PR head: all green, and `bun scripts/package.ts --check` confirms the dist trees are clean regenerations. The mechanism itself works as designed; the blockers already filed are about where it sits, not whether it functions.

SiddhJog:
All findings are addressed in the latest push (rebased onto v2 @ 2.4.0, version 2.4.1). Detail below, grouped by reviewer. Where a fix warranted a regression test I've named it.

**@leandrodamascena** — the four blockers
1. **Guard was in the wrong place (bypassable via direct state approve)**. Fixed. Moved the precondition out of orchestrate report into aidlc-state.ts handleApprove — the single seam every approve passes through, alongside the artifact/human-presence/revision guards, exactly as your state.ts:745 comment prescribes. A direct state.ts approve now hits it. New test t115 R6 reproduces your exact repro (direct approve, no review → refused, GATE_APPROVED=0).

2. **Stale review satisfies a re-approve (no lower bound)**. Fixed. The read is now floored at the later of the stage's latest STAGE_STARTED and any subsequent GATE_REJECTED — mirroring swarmConvergedUnits. A review from a prior run (redo-jump) or one recorded before a reject/revise no longer counts. New test t115 R7 (review → reject → revise → re-approve with no fresh review → refused; fresh review → unblocks). Note: my first cut used a timestamp floor, which broke on same-second events; switched to the position-tiebroken ordering idiom the codebase already uses.

3. **Per-unit stages approve on one review.** Fixed. aidlc-log review gains --unit (emits a Unit field), and the guard now requires one terminal REVIEW_COMPLETED per unit on per-unit stages, falling back to stage-level only when the unit list is unavailable. Prose in §12a + all 4 SKILLs updated to record --unit.

4. **Autonomous swarm halts at the reviewer gate.** Fixed. The guard is exempt under autonomous Construction — the same isAutonomousMode carve-out the human-presence guard uses. No conductor records a review in an unattended swarm, so requiring one would deadlock. t186 test 13's masking concern is moot now.

**Non-blocking (defense-in-depth)**. Acknowledged and left as-is by design: the tool records the row from flags without independently verifying a reviewer subprocess ran — identical in shape to the human-presence mint. The value is observability + the gate dependency, not cryptographic proof. Flagging explicitly rather than silently.

Version collision. Re-slotted through the moving target — now **2.4.1** on top of your 2.4.0.

**@apackeer**  — the eight findings

1. **Reviewer field never checked** — fixed; the guard matches Reviewer as well as Stage (test t115 R8: wrong reviewer name → refused).
2. **Stale doc counts** — 12-state-machine.md and 06-hooks-and-tools.md updated to 74 / 19; the t28 header comment corrected.
3. **Coverage registry marked tested units UNCOVERED** — added covers: subcommand:aidlc-log:review, audit:REVIEW_REQUESTED, audit:REVIEW_COMPLETED to t115 and regenerated.
4. **Iteration documented required but conditional** — marked optional in audit-format.md (copied the SCOPE_DETECTED notation); also added the Unit field there.
5. **REVIEW_REQUESTED optional & unused** — made it a required step-1 record in §12a + all SKILLs. Honest caveat below.
6. **Duplicate regex helper** — removed escapeRegexLiteral; the guard reuses the exported escapeRegex.
7. **Test helper swallows failures** — logReviewReady now asserts status === 0.
8. **CHANGELOG whitespace** — cleaned via the version re-slot.

leandrodamascena (changes requested):
Re-review of #569. My two prior findings are resolved — but verifying them, plus cross-checking a second reviewer's pass, surfaced five code holes. They're one family: **the reviewer precondition is right in isolation, but it isn't wired into every seam, ordering, and self-heal that its sibling guards already cover.**

**Prior findings — resolved:** the guard moved to `aidlc-state.ts handleApprove` (1709, before mutation; report no longer guards, cites #366), and it now floors at the latest STAGE_STARTED/GATE_REJECTED. I confirmed both by neutering each and watching the matching t115 case flip.

**Verdict: CHANGES REQUESTED.**

**Blocking:**

1. **Only `handleApprove` runs the guard — `advance`/`finalize`/`complete-workflow` don't** (`aidlc-state.ts:1709`). The sibling `verifyStageArtifacts` is on all four completing handlers (1290/1403/1519/1703) precisely because a direct `advance <active-slug>` is the #366 gate-skip seam; the reviewer guard was wired into one. Sharpest case: `complete-workflow` on a reviewer-bearing final stage (`code-generation` → `aidlc-architecture-reviewer-agent`) ships it with zero review. This re-opens the original report-only bypass, relocated. Fix: call it in the three, gated by the same `!alreadyMarkedCompleted` as the artifact guard.

2. **Revision backstop runs *after* the guard, with no re-check** (`aidlc-state.ts:1709` vs ~1734). The guard computes its floor at 1709 and passes on a review of the *pre-revision* artifact; the revision backstop then detects the post-review artifact write and backfills `GATE_REJECTED` at ~1760 — which logically invalidates that review — but nothing re-runs the guard, so it approves. Sequence: review → revise (unrecorded) → approve ⇒ backfilled reject + approve with a stale review. Fix: re-validate the reviewer precondition after the backstop backfills, or compute the floor including the backfilled reject.

3. **Per-unit branch lacks the vacuous-unit exemption** (`aidlc-state.ts:1130`). `producesArtifactsExist` exempts a unit whose kind prunes `produces` to empty (`allVacuous`, 852-859); the reviewer branch doesn't — so a `packaging` unit on `functional-design` (prunes empty → no per-unit directive → no review) is listed as "missing" and the approve wedges forever. The two guards disagree on identical state. `functional-design` ships in feature/mvp/enterprise, so this is a normal inline flow. Fix: mirror the kind-pruned exemption.

4. **DAG-stale fallback truncates per-unit to one unitless review** (`aidlc-state.ts:1131`). When `readBoltDagUnits` returns null the guard falls back to a single stage-level review. But `orchestrate.ts:796` has a self-heal: a stale runtime DAG with `unit-of-work-dependency.md` on disk is recomputed so the per-unit loop and coverage guard "never truncate a multi-unit plan because a hook failed to refresh." The reviewer guard skips that self-heal, so a stale DAG collapses N-unit enforcement to one review. Fix: use the same self-heal path the coverage guard uses.

5. **Unit-major floats the floor** (`aidlc-state.ts:1092`). stage-protocol.md:375 admits a stage's STAGE_STARTED "may land after that stage's per-Unit artifacts were written" under `Construction Iteration: unit-major`. So reviews recorded during the unit-major walk sit below their own STAGE_STARTED floor and the approve is refused on a fully-reviewed stage. Opt-in (default stage-major unaffected) and human-recoverable by re-recording — but a spurious re-review of already-reviewed work on a documented flow; t209 has no approve/GATE_APPROVED assertions, so this path is untested.

**Non-blocking:**

- **REVIEW_REQUESTED isn't recorded on the executable path.** The four SKILL.md gate branches carry only the `--verdict` (REVIEW_COMPLETED) form; the dispatch step that emits REVIEW_REQUESTED lives only in stage-protocol.md:883, not the SKILL.md. The guard only needs COMPLETED, but the promised observability (a dispatched-but-never-completed review being visible) doesn't happen on the main executable path. Add the dispatch record to the SKILL.md gate branches.
- **Self-cert floor:** the conductor can log a review it didn't run — a known design note (same bucket as #568's marker; hard on the review being recorded, soft on the verdict). Not blocking.
- **Version collision (mechanical):** origin/v2 already ships 2.4.1 (via #573); this PR bumps 2.4.1 too. On rebase, version/badge/CHANGELOG collide and t68 fails on the duplicate heading — routine second-to-merge re-slot (2.4.2 is taken by #571, so likely 2.4.3). PR title/body are also stale (still say 2.3.5).
- **Stale comment:** `aidlc-audit.ts:58-60` says the review gate is "enforced in aidlc-orchestrate report" — the exact seam this PR moved the guard out of; should cite `aidlc-state.ts`.

The direction is right and the two prior findings are closed. But 1–2 re-open the bypass/freshness class my original review raised (now through the sibling seams and the revision ordering), and 3–5 wedge or hollow legitimate per-unit runs, so I'd want these before approving.

apackeer:
**Additional blocker: the autonomous exemption is not scoped to Construction.**

`verifyReviewerPrecondition` returns immediately on `isAutonomousMode(content)` (`core/tools/aidlc-state.ts:1069`) without checking `stage.phase`. `Construction Autonomy Mode` persists in the state file, and `aidlc-jump.ts` preserves it on a backward jump. That makes this reachable:

1. Enter autonomous Construction.
2. Backward-jump to a reviewer-bearing Inception stage such as `requirements-analysis`.
3. Open and approve that stage without recording a review.

I reproduced the resulting state directly on this head: with `Construction Autonomy Mode: autonomous` present, `gate-start requirements-analysis` followed by `approve requirements-analysis` exits 0, marks the stage `[x]`, and emits no `REVIEW_COMPLETED`.

The documented carve-out is specifically "Autonomous Construction (swarm / Bolt)". At minimum, gate the return on the phase as well, e.g. `stage.phase === "construction" && isAutonomousMode(content)`, and add a regression case showing that an Inception reviewer stage remains guarded when the persisted Construction mode is autonomous. Otherwise a stale autonomy grant becomes a general reviewer bypass after a backward jump.

apackeer:
Pushed the round-2 fixes, co-authored with the PR author on this branch. The second commit's message body maps each blocker to its fix; the PR body above is refreshed to match.

Summary: the reviewer precondition now guards all four completing transitions; the revision backstop floors it; kind-pruned units are exempt per unit; the unit list self-heals from the dependency artifact; unit-major gets floor semantics that survive its documented STAGE_STARTED reordering (plus workflow/jump floors that close the redo-jump hole); and the autonomous exemption is gone entirely - swarm units are reviewed in their Bolt worktrees and `finalize` refuses to merge a claimed unit without its receipt.

Verified at the pushed tip: full default tier 276 files / 5631 assertions / 0 failures; `package.ts --check` green across all 5 harness trees; typecheck and the coverage-registry check green.

@leandrodamascena re-review when you have a chance.

leandrodamascena (changes requested):
I re-reviewed head `525812c4`. The blockers from the previous rounds appear fixed, but three completion-guard gaps remain:

### [P1] Validate the cached DAG against the authored dependency artifact
`core/tools/aidlc-lib.ts:5364`

A valid but stale cached `bolt_dag` is accepted without comparing it with `unit-of-work-dependency.md`. For example, if the cache contains only `alpha`, while the artifact contains `alpha` and `beta`, reviewing only `alpha` can unblock approval and silently omit `beta`.

The new stale-DAG test covers a graph without a `bolt_dag`, but not a valid, outdated one. Please detect this case and resolve the current unit set from the authored artifact.

### [P2] Invalidate reviews after relevant artifact changes
`core/tools/aidlc-state.ts:1303`

Review freshness is reset by workflow, stage, jump, and rejection events, but not by artifact updates. A review can therefore be recorded, the reviewed artifact can be changed, and the stage can still complete using the old review.

Please include relevant `ARTIFACT_CREATED` and `ARTIFACT_UPDATED` events in the review freshness calculation, scoped to the reviewed stage and unit.

### [P2] Require a terminal verdict
`core/tools/aidlc-state.ts:1323`

The guard accepts any matching `REVIEW_COMPLETED` row without validating its required `Verdict`. A malformed row without `READY` or `NOT-READY` can therefore satisfy the reviewer precondition.

Please require `Verdict` to be either `READY` or `NOT-READY`.

The main enforcement is otherwise much stronger: all four completion routes are guarded, stale reviews after rejection are rejected, and both stage and reviewer identities are matched.

apackeer:
Round-3 findings are addressed in the new tip commit ("fix(reviewer): authored unit DAG is authoritative; reviews go stale on artifact writes; verdicts must be terminal"), and the branch is rebased onto v2 @ 2.5.2 (#576), so the PR is conflict-free again. Fix mapping:

**[P1] Validate the cached DAG against the authored dependency artifact** - `resolveBoltDag` (aidlc-lib.ts) now parses `unit-of-work-dependency.md` whenever it exists and accepts the compiled `bolt_dag` only when its batches AND unit kinds match the artifact; on any mismatch the unit set is recomputed from the artifact (healed, one stderr note). Your exact repro is pinned: t208 test 5d2 seeds a valid cache holding only `alpha` beside an artifact declaring `alpha` + `beta`, reviews only `alpha`, and asserts the approve refuses naming `beta`. t215 test 6b covers the same heal on the per-unit iteration path.

**[P2] Invalidate reviews after relevant artifact changes** - the guard's event stream now includes `ARTIFACT_CREATED` / `ARTIFACT_UPDATED`: a post-review write to a declared `produces[]` / `optional_produces[]` file clears the matching receipt. On per-unit stages the `construction/<unit>/` path segment scopes invalidation to that unit; an ambiguous matching path fails closed by clearing every unit receipt. Pinned by t115 R12 (create + update both refuse until re-recorded), R13 (an unrelated artifact write does NOT invalidate), and t208 5f (only the touched unit's receipt is cleared; re-recording it unblocks).

**[P2] Require a terminal verdict** - a `REVIEW_COMPLETED` row now satisfies the precondition only with `Verdict` exactly `READY` or `NOT-READY`; missing or malformed verdicts are skipped. Pinned by t115 R14. This also brings the completion guard into agreement with the swarm-finalize receipt check, which already validated the verdict.

**One behavior change beyond the letter of the findings, flagged for your review:** with the artifact now authoritative, a malformed authored artifact fails closed even when a valid cached `bolt_dag` exists - t215 test 6 is deliberately flipped from its former "existing bolt_dag wins over a malformed dependency artifact" contract. Rationale: a malformed artifact next to a valid cache is the same silent-divergence hole as your P1 with a different trigger, and the pre-existing cache-wins behavior would have the guard trusting the cache most at exactly the moment the artifact is most suspect. The 2.7 required-sections sensor still surfaces a malformed edge block advisorily at authoring time; the engine now refuses to act on one at completion time instead of falling back, and the per-unit iteration error names the artifact as authoritative with the parse reason.

Also in this push: the rebase re-slots the version/CHANGELOG/badge trio above 2.5.2 (t68 green), and the dist trees + coverage registry are regenerated at the tip (`package.ts --check` and `gen-coverage-registry.ts --check` both green).

Full default tier at the pushed tip: 277 files, 5664 assertions, 0 failures.

leandrodamascena (approved):
I re-reviewed head `29c46e14`. All three findings from the previous round are resolved:

- The authored unit DAG is now authoritative over stale cache data.
- Relevant artifact writes invalidate existing review receipts.
- Only `READY` and `NOT-READY` verdicts are terminal.

The new regression tests cover each case, and the targeted suite, package drift check, typecheck, and coverage registry all pass.

Two non-blocking follow-ups remain:

- Same-second events across different audit shards can still be misordered because filename order is used as the tiebreaker.
- Workspace source changes after a `code-generation` review do not invalidate the receipt because `ARTIFACT_*` covers record/codekb files only.

These are broader ledger/workspace-freshness limitations and should be tracked separately rather than block this PR.

PR #571, opened 2026-07-14
feat: opencode harness support (2.4.6)

apackeer:
# opencode harness support (2.3.5)

Adds a fifth harness distribution: **opencode** ([opencode.ai](https://opencode.ai), verified live on 1.17.18). One core, many harnesses: `dist/opencode/` projects the same deterministic engine onto opencode's native surfaces - skills, subagents, a `/aidlc` command, and a hook-adapter plugin.

## Layout: the engine ships at `.aidlc/`, not `.opencode/`

opencode auto-imports every `*.ts` under `.opencode/tools/` and `.opencode/tool/` as custom tool definitions. A CLI-style engine script there (top-level dispatch, `process.exit`) crashes the session - live-reproduced: `aidlc-sensor-required-sections: --output-path is required` replaced the model's reply. So the engine tree ships at `.aidlc/` (harnessDir), which opencode never scans, and `.opencode/` carries only native emissions.

The shipped project-root `opencode.json`:
- registers `skills.paths: [".aidlc/skills"]` for skill discovery (live-verified),
- carries the method-tree `instructions` glob (`aidlc/spaces/default/memory/**/*.md`) - the native include, re-pointed by `/aidlc space <name>` via `repointHarnessIncludes` (live-verified with a magic-token probe),
- allowlists `bun .aidlc/tools/*` so engine invocations run without per-call prompts.

## The authored surface: `harness/opencode/`

`manifest.ts` + `emit.ts` (the `.opencode/` shell: 14 persona subagents with `mode: subagent` and projected tier keys, `command/aidlc.md`, `plugin/aidlc-opencode-adapter.ts`) + the orchestrator `SKILL.md` + question-rendering annex + onboarding fills + `opencode.json` + dot-gitignore.

## Hook mapping (plugin API, in-process)

The adapter plugin maps opencode hook moments onto the shared core hook bodies:

| opencode moment | core hook |
|---|---|
| `chat.message` | session-start (first message) + human-presence mint |
| `tool.execute.after` write/edit | audit emission + sensor dispatch (args key is `filePath`) |
| `tool.execute.after` bash | runtime-graph compile |
| `tool.execute.after` todowrite | statusline sync |
| `tool.execute.after` task | subagent logging |
| `experimental.session.compacting` | validate-state |
| `session.idle` | the Stop hook (advisory: a `block` verdict re-engages the loop by injecting a sentinel-marked nudge prompt that never mints presence) |

`SESSION_ENDED` is not emitted on opencode (no session-end hook moment); the swarm runs as task-tool fan-out only (`AIDLC_USE_SWARM=1` is a loud no-op). Full differences: `docs/guide/harnesses/opencode.md`.

## Tier projection

`TIER_PROJECTIONS` gains an `opencode` flavor: judgment omits both keys (session defaults win); balanced pins `amazon-bedrock/global.anthropic.claude-sonnet-4-6`; templated adds `variant: medium` (opencode's effort key is `variant`).

## Doctor and workspace

`/aidlc --doctor` on opencode checks the adapter plugin, the project-root `opencode.json`, and `.opencode/command/aidlc.md`. Workspace scan and artifact guards learn the `.opencode`/`.aidlc` dirs.

## Tests

- **NEW `t232-opencode-packaging` (unit)**: drift guard, byte-parity, shell shape, subagent frontmatter, `opencode.json` wiring.
- **NEW `t-run-opencode-status.serial` (live e2e)**: drives `/aidlc --status` through `opencode run --command aidlc` on the shipped tree, gated on `AIDLC_OPENCODE_RUN_LIVE=1` (run live-green; skips clean ungated).
- **Extended**: t151, t156, t157, t181 (pinned skill list), t217, t220 (EXPECTED + MD_TREES including the `.opencode` twins), t188 (five plugin projections), t148 leak guard, t-active-space-includes. Coverage registry regenerated.

## Docs

New guide chapter `docs/guide/harnesses/opencode.md` plus the harness-enumeration sweep across README, guides, and reference (primitives table, tier table, Stop-transcript note, glossary entry, open-set phrasings).

## Live verification

On a fresh copy of `dist/opencode/`: doctor 33/0, skill + subagent discovery (haiku probe), `/aidlc --status`, and a real intent-birth through `opencode run --command aidlc -- "..."` (the `--` is needed or opencode parses `--status` itself). Note: non-interactive runs auto-reject non-allowlisted bash; interactive sessions are needed for full workflows.


leandrodamascena (changes requested):
**Review: Request Changes**

The packaging work is solid, but the OpenCode adapter currently misses several runtime guarantees that AI-DLC relies on. These are functional and enforcement gaps, not documentation concerns.

### Blocking Findings

**[P1] Bash permission wildcard allows command chaining**  
`harness/opencode/opencode.json:8`

The rule `bun .aidlc/tools/*: allow` matches the complete shell command, including appended commands. For example, `bun .aidlc/tools/aidlc-utility.ts status && touch /tmp/example` is allowed without prompting. This bypasses the default `"*": "ask"` policy for arbitrary commands. The same applies to the hooks wildcard. Please replace these broad shell patterns with a boundary that validates the executable and arguments without permitting chained commands.

**[P1] Runtime graph compilation never recognizes OpenCode transitions**  
`harness/opencode/plugin/aidlc-opencode-adapter.ts:164-173`  
`core/hooks/aidlc-runtime-compile.ts:72-74`

The adapter forwards commands under `.aidlc/tools`, but the runtime compiler only accepts `.claude`, `.kiro`, or `.codex`. OpenCode transitions therefore exit at the command filter and do not refresh `runtime-graph.json`. This can leave status, memory snapshots, and graph-dependent routing stale. Please add `.aidlc` to all transition/runtime command filters and cover a real OpenCode transition in an adapter-level test.

**[P1] Reviewer read-scope enforcement is not installed**  
`harness/opencode/plugin/aidlc-opencode-adapter.ts:120-239`

The distribution includes `aidlc-reviewer-scope.ts`, but the adapter does not implement `tool.execute.before`. A reviewer can therefore read sibling construction units even though the current framework contract requires deterministic read-scope enforcement. I reproduced this against a sibling unit containing a marker file; the reviewer successfully returned that marker. Please wire the blocking pre-tool hook for file reads, searches, and shell commands, including the reviewer dispatch-record lifecycle expected by the core hook.

### Runtime Correctness

**[P2] `apply_patch` bypasses audit logging and sensors**  
`harness/opencode/plugin/aidlc-opencode-adapter.ts:151-163`

Only `write` and `edit` trigger `aidlc-audit-logger.ts` and `aidlc-sensor-fire.ts`. OpenCode exposes `apply_patch` as a normal editing tool, so changes made through it produce neither artifact audit events nor sensor execution. Please handle `apply_patch`, extract every affected path, and invoke audit then sensors for each changed artifact.

**[P2] Native subagents can still invoke `task`**  
`harness/opencode/emit.ts:26-31`

The emitter preserves `disallowedTools: Task`, but OpenCode treats this as an unknown provider option. Runtime inspection shows `tools.task: true` and no effective agent-level denial. The prose instruction is not an enforcement mechanism. Please translate this into the native per-agent task permission and add a test against the resolved agent configuration.

**[P2] Active-space switching silently fails for JSONC**  
`core/tools/aidlc-includes.ts:113-128`  
`core/tools/aidlc-includes.ts:207-216`

OpenCode supports JSONC and `opencode.jsonc`, but this implementation checks only `opencode.json` and parses it with `JSON.parse`. Comments or trailing commas cause the rewrite to be silently skipped while the active-space cursor still changes. The result is a split state where AI-DLC reports the new space but OpenCode continues loading the previous space memory. Please support both config filenames and use a JSONC-aware, non-destructive update.

**[P2] Session start is consumed before workflow state exists**  
`harness/opencode/plugin/aidlc-opencode-adapter.ts:129-139`

The adapter adds the session ID to `started` before running `aidlc-session-start.ts`. On a new workflow, the first prompt arrives before state creation, so the core hook exits without emitting `SESSION_STARTED` or binding the session to the intent. Later prompts never retry because the session is already marked started. Please mark the session started only after the hook can process an active workflow, or invoke session-start immediately after intent creation.

**[P2] Concurrent idle events can inject duplicate prompts**  
`harness/opencode/plugin/aidlc-opencode-adapter.ts:211-233`

There is no per-session in-flight guard around stop-hook evaluation and `client.session.prompt`. Two overlapping `session.idle` events can both observe a blocking decision and inject the same forwarding nudge. Please serialize idle handling per session and release the guard in `finally`.

**[P2] Generated agents reference a directory that is not shipped**  
`harness/opencode/emit.ts:63`

Token substitution produces references to `.aidlc/rules/`, but `dist/opencode` contains no such directory. The active method files are under `aidlc/spaces/<space>/memory/`. This leaves native agents with invalid knowledge-loading instructions. Please project these references to the actual active-space memory surface and add a packaging assertion that referenced paths exist.

### Validation

I ran the focused OpenCode/package tests, the package drift guard, runtime probes, and the CI profile.

- 29 focused tests passed.
- `bun scripts/package.ts --check` passed for every harness.
- Existing OpenCode packaging tests passed but do not execute the adapter lifecycle.
- The live OpenCode E2E remains opt-in and was skipped without `AIDLC_OPENCODE_RUN_LIVE=1`.
- Two resource-sensitive integration failures passed when rerun individually.
- Three unrelated tests could not load the missing Claude Agent SDK.
- The branch currently conflicts with `v2`; a synthetic merge confirms conflicts in versioning, plugin documentation, and test files.

Please address the runtime findings, add adapter-level regression coverage, rebase onto `v2`, and rerun the live OpenCode workflow before merging.

apackeer:
Thanks for the detailed review. The runtime and enforcement findings are addressed at 0f6daa4:

- Added a pre-execution shell boundary against chaining/redirection/substitution.
- Added `.aidlc` transition recognition through `KNOWN_HARNESS_DIRS`.
- Wired reviewer read-scope enforcement into `tool.execute.before`.
- Added `apply_patch` audit and sensor fan-out.
- Projected native `permission.task: deny` for subagents.
- Added non-destructive `opencode.jsonc` active-space rewriting.
- Made session-start retry until workflow state exists.
- Serialized idle handling per session.
- Corrected generated memory paths and asserted that they exist.

`t241-opencode-adapter.test.ts` now exercises the adapter lifecycle directly. The recorded follow-up validation is package generation plus drift check, typecheck, and 96 focused assertions. The opt-in live OpenCode E2E is not recorded as rerun in the follow-up commit. The branch is currently mergeable with `v2`. Re-requesting review.

leandrodamascena (changes requested):
**Review: Request Changes (round 2)**

Thanks for the fast turnaround on the round-1 findings. I re-verified the fix commit by reproducing each item against the actual code, and most of the runtime findings still hold. Several are new regressions introduced by the fixes themselves. Summary below, ordered by severity, each with the minimal fix.

### Blocking (P1)

**[P1] First-turn workflows still bypass Stop enforcement**
`harness/opencode/plugin/aidlc-opencode-adapter.ts` (session.idle handler + the `started` latch)

On a fresh project, `session-start` exits silently before workflow state exists (`core/hooks/aidlc-session-start.ts:63`), so the session is never added to `started`. If the workflow is created *during* turn-1 — the mainline `/aidlc init` flow — the turn-1 `session.idle` early-returns and Stop enforcement is deferred to turn-2. Reproduced: `stop after turn-1 idle = 0`, `after turn-2 = 1`. This is observably weaker than the siblings: Claude wires Stop unconditionally and Codex forwards the stop event verbatim, both letting `core/hooks/aidlc-stop.ts`'s own `existsSync(statePath)` be the only gate. The current t241 idle test stubs `session-start` to always return `additionalContext`, so it pre-latches the session and never covers first-turn birth.
*Fix:* don't gate idle Stop dispatch on the `started` latch for a fresh birth — either re-check state existence in the idle handler and latch then, or set a lighter "saw a human turn" flag in `chat.message` and let the core stop hook be the enforcement gate. Add a t241 case that creates state between the turn-1 `chat.message` and the turn-1 idle and asserts a nudge on turn 1.

**[P1] Bash boundary still permits unprompted arbitrary execution**
`harness/opencode/opencode.json` · `harness/opencode/plugin/aidlc-opencode-adapter.ts` (`aidlcBashBoundaryViolation`)

The boundary now blocks chaining/redirection/substitution, but it validates only command *syntax*, never the target filename against a shipped-entrypoint set. Executing the exported function confirms `bun .aidlc/tools/payload.ts` returns ALLOW. Since writing to `.aidlc/tools/` is itself ungated and the config allowlists `bun .aidlc/tools/*` / `bun .aidlc/hooks/*`, an agent can author `.aidlc/tools/payload.ts` and run it with no prompt while the ambient posture is `ask`. Round-1 asked to "validate the executable *and* arguments" — this fixed the argument half only.
*Fix:* emit a build-time allowlist of the actual shipped `.aidlc/tools/*.ts` and `.aidlc/hooks/*.ts` basenames and reject any invocation whose target isn't in it; as defense-in-depth, gate write/edit under those dirs to `ask`.

*(Note: I also found that the syntax check itself is bypassable — a backslash inside single quotes followed by a rebalancing `#'` swallows the closing quote and lets `;`/`&&` through, e.g. `bun .aidlc/tools/aidlc-utility.ts status 'a\' ; touch /tmp/x #'` returns ALLOW and executes under bash. Both the entrypoint gap and this parser gap need closing.)*

### Runtime correctness (P2)

**[P2] Relative `apply_patch` paths bypass real audit logging**
`harness/opencode/plugin/aidlc-opencode-adapter.ts` (`applyPatchPaths` + tool.execute.after) · `core/hooks/aidlc-audit-logger.ts:64`

`applyPatchPaths` forwards envelope paths verbatim, but the core audit gate requires an absolute path under `docsRoot()`; a relative path can never prefix-match, so the event is `process.exit(0)`-discarded. Reproduced end-to-end against the real hook: relative path → 0 audit entries; absolute form of the same file → full `ARTIFACT_CREATED`. `write`/`edit` escape this only because OpenCode supplies an absolute `filePath`; the codex adapter already resolves envelope paths (`isAbsolute(rel) ? rel : join(projectDir, rel)`) — the opencode port dropped that step. The t241 test stubs the hooks, so it verifies call ordering but never the real gate.
*Fix:* resolve each envelope path to absolute before forwarding (as codex does); harden t241 to exercise the real audit hook.

**[P2] JSONC active-space switching + doctor**
`core/tools/aidlc-includes.ts` (instructions regex, ~L114) · `core/tools/aidlc-utility.ts` (doctor check, ~L1120)

Two confirmed sub-issues: (a) the `/("instructions"\s*:\s*\[)([\s\S]*?)(\])/` regex is non-greedy and anchors to the *first* `"instructions"` token — with a commented-out block above the effective one, it rewrites the comment, reports success, and leaves the live glob unchanged (reproduced: `written=["opencode.jsonc"]` while the effective block stayed on the old space). (c) The doctor check hard-requires `opencode.json`, so a jsonc-only install — which the repointer itself explicitly supports — fails doctor. (b) Preferring `opencode.json` when both files exist is a latent ordering hazard I couldn't confirm without a live OpenCode run.
*Fix:* match the aidlc glob string directly (or greedy-capture to the last `]`) instead of the first `instructions` occurrence; make doctor accept either `opencode.json` or `opencode.jsonc`.

**[P2] Plugin agents are not emitted as native OpenCode agents**
`scripts/plugin-hooks-template/compose.ts` · `harness/opencode/emit.ts`

Core personas get a native `.opencode/agents` twin, but plugin composition copies agents only to `.aidlc/agents` (`compose.ts` has no `.opencode` awareness), and OpenCode discovers task-delegated subagents only under `.opencode/agents`. A plugin stage with `reviewer:`/subagent passes schema validation (roster is built from `.aidlc/agents`) but has no task-delegable twin and fails at runtime on OpenCode only. New gap from this PR's two-dir split; latent today because the test-pro fixture declares no such stage.
*Fix:* emit a native `.opencode/agents` twin for plugin agents at compose/package time; until then, add a doc caveat + a guard rejecting plugin reviewer/subagent stages on OpenCode.

**[P2] Child-session lookup failures cached as main sessions**
`harness/opencode/plugin/aidlc-opencode-adapter.ts` (`isMainSession`, ~L263-275)

`isMainSession` initializes `main = true`, and on a `session.get()` failure it caches that unconditionally with no negative-result invalidation. Reproduced: a child session whose lookup throws once is cached `main=true` permanently and never re-queried. A poisoned child then passes the `chat.message` gate and runs `aidlc-mint-presence.ts`, appending `HUMAN_TURN` to the shared ledger — which satisfies `humanActedSinceGate`, the exact defense against a model fabricating an approval with no human acting. Trigger is a transient, non-attacker-controlled failure, so it's a robustness degradation of a safety control (P2), not a one-shot exploit.
*Fix:* cache only confirmed lookups; default the transient-failure path to non-main (child) rather than poisoning the cache.

### Not blocking for this PR

**Reviewer read-scope is bypassable, but pre-existing.** Both vectors reproduce (`P=<sibling>; cat "$P"` passes literal-substring matching; the dispatch record is writable/deletable → fail-open). However `core/hooks/aidlc-reviewer-scope.ts` is byte-identical to the pre-fix base and unchanged by this commit, the fail-open is documented by-design for an advisory cost-guardrail (off-switch `AIDLC_DISABLE_REVIEWER_SCOPE_HOOK=1`), and it's cross-harness. Worth a separate hardening ticket, not a blocker here. (Separately, one real gap in the *opencode* wiring does belong to this PR: the `list` tool has no arm in `reviewerCalls()`, so a dispatched reviewer can enumerate a sibling unit's directory even though the core hook blocks the equivalent Claude `LS`.)

### Validation notes
- All findings reproduced against the actual checked-out code at the fix commit; the dist twins are byte-identical to the authored files, so each defect ships.
- **F2 (nudge/idle) needs a live OpenCode confirmation before we call it done.** The code defect is real (the `idleInFlight` guard is held across the whole `await client.session.prompt()`, so a forwarding loop can be capped at one continuation vs. the core cap of 8), but whether it bites depends on OpenCode's real idle-vs-prompt delivery order, which isn't reproducible statically. The defensive fix (release the guard before the await) is cheap and correct regardless and should land.

Please address the P1s + the four P2s, add adapter-level regression coverage that exercises the real hooks (not stubs), and rerun the live OpenCode workflow. Happy to re-review.

apackeer:
Round-two review findings are addressed in 227f6dc.

- First-turn workflow birth now reaches Stop enforcement through a human-turn latch, with the real Stop hook covered.
- The bash boundary validates against the exact packaged `.aidlc/{tools,hooks}/*.ts` set, rejects shell syntax and the quoted-backslash bypass, and prompts on engine-code edits.
- Relative `apply_patch` paths are resolved before real audit and sensor fan-out.
- JSONC repointing targets AIDLC globs directly, updates both config files when present, and doctor accepts JSONC-only installs.
- Plugin agents compose into native `.opencode/agents` twins with `mode: subagent`, `permission.task: deny`, valid memory paths, and no invalid bare model alias.
- Child lookup failures fail closed without cache poisoning; reviewer `list` maps to core `LS`; idle serialization releases before prompt delivery.
- The live rerun exposed one additional host constraint: OpenCode plugin modules must expose only the default factory. Named runtime exports were removed and the export shape is now pinned by `t240`.

Validation:
- `bun scripts/package.ts --check`: pass
- `bun run typecheck`: pass
- focused OpenCode packaging/adapter suite: 20/20 pass
- live `opencode run --command aidlc -- --status`: pass on OpenCode 1.17.18
- full CI profile: 262/263 files pass, including every live SDK journey. The sole failure is the base-identical `t92` expectation that TypeScript config failure exits 2; this environment returns 1, and the test, sensor implementation, package manifest, and lockfile are byte-identical to current `origin/v2`.

Re-requesting review.

leandrodamascena (approved):
Thanks for addressing all the feedback @apackeer! APPROVED!

PR #572, opened 2026-07-14
fix: pin the linter sensor's eslint version (2.3.10)

apackeer:
## What this fixes

The linter sensor invoked `bunx eslint` bare, so bunx's resolution order (project node_modules -> PATH -> registry) made sensor verdicts machine-dependent. Distro packages ship ancient eslint versions - Ubuntu's apt eslint is 6.4.0, installed as a transitive dependency of `apt install npm` - and pre-flat-config eslint cannot read `eslint.config.js`. On such a box the availability probe exits 127 and every linter fire silently degrades to a `Note=tool-unavailable` PASS, masking real lint findings. This surfaced as 2 red assertions in the release gate's t92 after a routine `apt-get install nodejs npm` put eslint 6 on PATH.

## The fix

`aidlc-sensor-linter.ts` pins `ESLINT_SPEC = "eslint@10"` on all three bunx call sites: the availability probe (`--version`), the config probe (`--print-config`), and the lint run itself. A versioned spec makes bunx bypass PATH resolution, so the sensor runs the same eslint everywhere. Verified live that the pin defeats PATH shadowing (bare `bunx eslint --version` resolves 6.4.0 on the affected box; `bunx eslint@10 --version` resolves 10.x).

Projects that need a different eslint can still override the sensor's `command:` in the linter sensor manifest.

## Version and tests

- 2.3.10 + CHANGELOG entry + README badge (re-slotted from the branch's original 2.3.9 after #560 took that slot).
- New `tests/unit/t237-linter-sensor-version-pin.test.ts`: source-pin guard that fails if any of the three call sites loses the versioned spec.
- Guards at tip: t68 + t237 green (stamp 2026-07-14T01-18-23Z), `package.ts --check` in sync.

## Related

- The other two release-gate reds from the same investigation (tsc 7 exit-code shift in t92 test 44, t139 TUI time-budget split) are test-only fixes on a separate branch (no version bump) - a follow-up PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.3.10] - 2026-07-14
+
+The linter sensor now pins the eslint version it runs (`eslint@10`) instead of resolving whatever `eslint` bunx finds first. A bare `bunx eslint` prefers a project-local node_modules copy, then any `eslint` on PATH, before fetching from the registry - and distro packages ship ancient versions (Ubuntu's apt eslint is 6.4.0, installed as a transitive dependency of `apt install npm`). Pre-flat-config eslint cannot read `eslint.config.js`, so on such a box every linter fire silently degraded to a `Note=tool-unavailable` PASS, masking real lint findings. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `aidlc-sensor-linter.ts` invokes `bunx eslint@10` for the availability probe, the `--print-config` probe, and the lint run itself; PATH-shadowed or node_modules eslint no longer changes the verdict.
+* Projects that need a different eslint can still override the sensor's `command:` in the linter sensor manifest.
+
 ## [2.3.7] - 2026-07-14
 
 The `upstream-coverage` sensor now matches the citation forms the framework's artifacts actually use, clearing the hundreds of false `SENSOR_FAILED` audit rows a Construction run accumulated. It previously demanded each consumed artifact's bare slug in every single written file; artifacts instead cite upstream by producing-stage directory path in their provenance header, and multi-artifact stages legitimately split citations across sibling deliverables. Real coverage gaps still fail. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.3.7-blue)
+![version](https://img.shields.io/badge/version-2.3.10-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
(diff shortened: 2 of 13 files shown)


leandrodamascena (dismissed):
APPROVE

I traced the change and it does what the PR says: `ESLINT_SPEC = "eslint@10"` replaces the bare `eslint` at all three bunx call sites (availability probe :164, config probe :190, lint run :288). The arg shapes are byte-identical to base — the only change is the literal-to-constant swap. There's no 4th invocation and one `spawnSync` import, so the probe and the lint run resolve the same eslint. The pin does kill the machine-dependence: a distro eslint 6 on PATH no longer gets picked up, because a versioned `bunx eslint@10` fetches/caches the spec instead of resolving PATH. t237 is a real guard, not a tautology — I reverted one site to bare `"eslint"` in a worktree and the assertion goes red. Version chain lines up (2.3.10 across version.ts/CHANGELOG/README, free, t68 green).

The sensor is advisory, so a wrong verdict is diagnostic noise, not misrouting — that's why the trade-offs below are notes, not blockers. Two are worth calling out because they're worse than a nit:

- **Offline / air-gapped with a vendored eslint now stops linting.** A versioned `bunx eslint@10` bypasses `node_modules` and forces registry resolution — the code comment itself says "bunx with a version spec bypasses node_modules too, which is the price of determinism." So an air-gapped box that linted fine with its vendored eslint now can't fetch, the probe exits 127, and the dispatcher degrades to a tool-unavailable PASS — findings silently stop surfacing. Bounded (advisory, override exists, narrow population), but the PR body doesn't call it out. Worth an offline caveat.
- **Legacy `.eslintrc`-only projects with a local eslint now stop linting.** eslint 10 dropped eslintrc, so `bunx eslint@10 --print-config` on an eslintrc-only project exits with "couldn't find an eslint.config" — which matches the sensor's no-config regex and degrades to tool-unavailable PASS. The same project under the pre-fix bare form (local eslint 8) would have surfaced real findings. So the pin masks errors on those boxes.

The escape hatch for both is the manifest `command:` override, but it's only discoverable in the source comment — the CHANGELOG mentions the override but not the flip side (offline / working-local-eslint). One sentence there would help the affected users.

Non-blocking nits: `eslint@10` pins only the major, so different boxes can cache different 10.x (the "deterministic resolution" claim holds at the major level, not exact — defensible, since eslint ships new rules disabled by default across minors). And the first fire on a cold cache eats an eslint@10 fetch inside the 30s budget — a one-time BUDGET_OVERRIDE on a slow link, warm after.

Net: this fixes a real machine-dependence bug and degrades gracefully where the pin doesn't fit. I'd approve, and I'd add the offline/eslintrc caveat to the CHANGELOG override note so the affected users know to set `command:`.

leandrodamascena (approved):
I traced the change and confirmed that all three `bunx` calls now use `eslint@10`. This fixes the reported machine-dependent behavior where an old distro ESLint on PATH could make the advisory linter sensor silently stop reporting findings.

The generated distributions, version files, and tests are consistent. The change does not block or reroute the workflow: the linter sensor is advisory, and unavailable tooling continues to degrade to a PASS by design.

Two non-blocking trade-offs are worth documenting:

- Offline projects with a working local ESLint may lose lint feedback because the versioned `bunx` call bypasses `node_modules` and may require a cached or downloadable ESLint 10.
- Projects using only legacy `.eslintrc` configuration may also lose lint feedback because ESLint 10 requires flat config.

One documentation detail should be clarified: changing the manifest `command:` cannot directly select `bunx eslint@8`; it requires a compatible `.ts` sensor adapter.

These affect local advisory diagnostics, not workflow execution, so I approve with a documentation follow-up.

PR #573, opened 2026-07-14
fix: count codekb-stage gate revisions + unpin environment-dependent test values (2.3.11)

apackeer:
## Summary

Two commits: an engine fix making Revision Count work for the Reverse Engineering gate, and test-only unpins for two environment-dependent assertions that turned red from toolchain drift.

### Engine fix: codekb-stage gate revisions were never counted (2.3.11)

A revised-then-approved reverse-engineering gate left `Revision Count` at 0 with no `GATE_REJECTED` row, proven live by the t-tui-t139 metamorphic journey (2026-07-13): the driver rejected the RE gate, the orchestrator revised architecture.md and re-presented the gate - all correct on screen - but the counter never moved. Two stacked gaps:

1. **The audit-logger hook only logged writes under the per-intent record root.** Codekb artifacts live at the space level (`aidlc/spaces/<space>/codekb/<repo>/...`), a sibling of `intents/`, so RE's writes emitted no `ARTIFACT_*` rows at all.
2. **The approve-time gate-revision backstop excluded codekb stages outright** - its produces matcher only knew the `/<slug>/<name>.md` suffix shape, which codekb paths don't have.

Fix: the hook's path gate accepts the space codekb root (breadcrumb `codekb > <repo> > <name>`), and `producesArtifactFile` gains a codekb arm matching `codekb/<repo>/<name>.md`. The reject verb itself was never the gap (it already counts RE rejects; stage-protocol.md line 163 already instructs it) - this closes the deterministic backstop behind the prose instruction. t205 scenario 12 pins the full flow.

### Test unpins (no behavior change)

* **t92 test 44**: the TS18003 exit-code assertion pinned tsc's own exit digit, which shifted 2 -> 1 in TypeScript 7 under `--incremental` (npm latest - this red spreads to any machine with a fresh bunx cache). Now matches the guarded shape `script-error: exit-<n>` instead.
* **t-tui-t139**: the 2400s ceiling split 50/50 starved the structurally-longer revised run (a healthy run blew the 1170s backstop mid-revision). Ceiling now 3600s; the clean run gets a fixed 40% backstop and the revised run inherits the actual remainder.

## Verification

* t205 (incl. new codekb scenario 12): 12/0. t92: 45/0. t68: 7/0. Dist drift check clean.
* Full smoke+unit+integration tier in the worktree: 258 files; the only reds (t17, t118, t52-drift-meta-validation) fail identically on unmodified v2 @ 3c76878 (A/B-proven pre-existing). The three slow SDK files (t72-stage-reverse-engineering, t183-codekb-placement-reverify.sdk, t-journey-workspace.sdk - the ones nearest this change) re-ran green.
* The live t139 journey that exposed the gap has not been re-run post-fix (~40 min live Bedrock run); t205 scenario 12 covers the mechanism deterministically.

## Related

* The eslint-pin fix that also surfaced in this investigation ships separately as #572.
* Version slotted 2.3.11: #560 claims 2.3.8-2.3.9, #572 claims 2.3.10.


leandrodamascena (changes requested):
I traced the production flow through the real audit hook, revision backstop, and `orchestrate report` path. The main fix works: valid codekb revisions are now audited and recovered correctly, without blocking approval or double-counting explicit rejects.

One repository-scoping issue should be addressed before merging.

With an active intent containing only `repo-a`, I edited `codekb/repo-b/architecture.md` through the real hook and then approved Reverse Engineering through `orchestrate report`, with all guards enabled. The PR recorded the `repo-b` write against the active `repo-a` intent, emitted a recovered `GATE_REJECTED`, and incremented `Revision Count` to 1. The base branch left the count at 0.

This does not block or reroute the workflow, but it writes incorrect revision history and state. Since this PR specifically extends revision recovery to the shared codekb tree, repository ownership should be part of that matching logic.

Requested changes:

- Scope codekb revision matching to repositories allowed by the active intent, while preserving the current multi-repo/workspace fallback.
- Add a negative test proving that a `repo-b` write does not count for a `repo-a` intent.
- Strengthen the behavioral test to use real repository resolution and the production `orchestrate report` path.
- Update the state-machine and audit documentation, which still says codekb stages are excluded or that only record-tree writes are logged.

The t92 and t139 unpins otherwise look reasonable. I am not asking for a broader audit-logger refactor; these changes can stay limited to the behavior introduced by this PR.

apackeer:
Thanks for the review. The requested changes are addressed at c594df5:

- CodeKB revision matching is scoped through `intentRepos()`, while preserving the empty/legacy workspace fallback.
- Scenario 13 proves a `repo-b` write does not count for a `repo-a`-only intent.
- Scenario 12 now records a real multi-repo intent and exercises the production `orchestrate report` approval path.
- The state-machine, audit-format, hook comment, and CHANGELOG text now describe CodeKB audit/revision behavior accurately.

The branch has also been rebased onto the current `v2` line and is currently mergeable. Re-requesting review.

leandrodamascena (approved):
**APPROVE.**

This is my re-review after you reworked the PR, and the blocker I filed last time — `docs/reference/12-state-machine.md` saying "codekb stages are excluded" while the code counted them — is cleared. That doc and `audit-format.md` now describe the codekb arm the way the engine behaves.

**What I re-checked myself, and it holds:**

- **The fix is live, not inert.** I traced `join(codekbDir(pd,"_"), "..")` through `node:path`: `.../codekb/_/..` collapses to `.../codekb`, so `underCodekb` matches a real write and the hook emits `ARTIFACT_UPDATED` for codekb writes.
- **The tests actually pin the fix.** I copied the PR's t205 onto clean v2: scenario 12 fails there with "Revision Count Expected 1 Received 0" and scenario 13 fails with "ARTIFACT_UPDATED Expected >= 1 Received 0" — the two halves of the bug. Both pass at head (13/13). Real A/B, not a tautology.
- **Repo-scoping doesn't re-introduce the uncounted-revision bug.** The recorded repo set and RE's write-path `<repo>` both come from the same `intents.json` `repos` array (written at birth, never mutated; names stored/read verbatim), so `recordedRepos.has(repo)` is true for correct usage. Empty set → any-repo fallback. No normalization divergence.
- **A clean first-approve is not over-counted.** RE writes its codekb artifact during the stage body, before the gate opens, so that write precedes the anchor and is outside the window. I re-created the pre-gate and pre-human-turn cases: Revision Count stays 0, no phantom `GATE_REJECTED`/`STAGE_REVISING`.
- **The non-codekb path and the separate artifact guard (`producesDirsForStage`) are untouched.** `--check` clean on all four trees; version 2.3.11 is free.

**Confirmed bugs: none.**

**Non-blocking notes (follow-ups, none blocking this PR):**

1. **Relative-path robustness in the audit-logger — worth a separate hardening pass.** One review raised that `codekb-path` prints a workspace-relative path and RE writes to it verbatim, so a relative `file_path` would miss the absolute `codekbRoot` gate. I checked all four adapters: Claude (native), Codex (`isAbsolute(rel) ? rel : join(projectDir, rel)`), and Kiro-IDE (`isAbsolute(rawPath) ? rawPath : resolve(projectDir, rawPath)`) all deliver an **absolute** `file_path`, so the write is logged and this fix works in real use — and t205's absolute fixture matches the hook's real contract, so it isn't masking anything. But the core hook doesn't normalize defensively, and the **Kiro CLI adapter is the only one that passes `ti.path` straight through with no `resolve`**. If Kiro CLI ever delivered a relative path it would break record logging framework-wide, not just codekb (so it'd be pre-existing, not from this PR). A one-line `isAbsolute(...) ? ... : resolve(projectDir, ...)` in the hook before the gate/statSync would close the gap and match what Codex/Kiro-IDE already do. Not for this PR — it's the shared hook, orthogonal to the codekb fix.

2. **Repo-scoping coverage gap (minor).** Both codekb scenarios seed non-empty repo sets, so the `recordedRepos.size === 0` legacy fallback is never exercised — deleting that clause would ship green. And there's no negative case for the matcher's structural guards. Worth a follow-up scenario; backstop fails open, so low risk.

3. **t139 not re-run live post-fix** — acceptable, since t205/12-13 are the deterministic proof and I verified them directly.

Ship it.

PR #576, opened 2026-07-14
feat: /aidlc --doctor --export — redacted diagnostic bundle (2.5.2)

### LINKED ISSUE #575: feat: /aidlc --doctor --bundle - generate a redacted diagnostic report
## Problem

When a workflow misbehaves, such as a gate that will not open, a stage that will not advance, or an approved report that is repeatedly refused, debugging often requires requesting the user's project directory.

That approach is:

- Huge: it includes source code, dependencies, build output, and unrelated files.
- Leaky: workflow records contain requirements, designs, decisions, questions, and other business data.
- Manual: maintainers must reconstruct the execution from state, audit events, runtime markers, and generated graphs.

Recent PR reviews demonstrate the need:

- #568: a Kiro collaborator cannot write its required contribution, causing the ensemble evidence gate to remain blocked.
- #571: runtime compilation can fail to run, leaving `runtime-graph.json` stale without an obvious user-facing explanation.

A diagnostic bundle should identify these conditions without exposing the project's work product.

## Proposal

Add `--bundle` to `/aidlc --doctor`.

The command runs a fresh doctor analysis and produces:

- A human-readable `report.md`.
- A machine-readable `report.json`.
- A redacted, normalized evidence set.
- A `manifest.json` describing versions, checksums, redactions, and included files.

The canonical output is a bundle directory. When a supported archive mechanism is available, the directory is packaged as a `.tar.gz`. If archiving fails or is unavailable, the directory remains usable and the command clearly reports its location.

The value is the diagnosis, not raw file collection.

## Shared Diagnostic Model

Refactor doctor findings into a shared structured model:

```ts
interface DoctorFinding {
  id: string;
  severity: "info" | "warning" | "error";
  summary: string;
  evidence: Record<string, unknown>;
  remedy: string;
  safeToAutomate: boolean;
}
```

The live doctor output and exported bundle must use the same findings. This prevents the command and bundle from developing separate diagnostic rules or remediation text.

Every bundle runs a fresh doctor analysis. Cached results are not used because stale diagnostics may hide the problem being investigated.

### Reconstructed Timeline

Build a timeline from available audit events:

- Stage entry and completion.
- Observed duration when both timestamps exist.
- Gate opened, approved, rejected, or currently unresolved.
- Revision count.
- Reviewer READY or NOT-READY outcomes and iteration count.
- Time between stages.
- Abnormally long or incomplete stages.

Missing evidence must be represented as `unknown`. The report must not infer events that were not recorded.

### Deterministic Diagnosis

Use fixed, versioned condition-to-remedy rules with no LLM involvement.

Initial conditions:

- Open or unresolved gates.
- Ensemble evidence missing or malformed.
- State and audit disagreement.
- Runtime graph older than its authored inputs.
- Frozen or missing hook heartbeats.
- Missing or malformed runtime markers.
- Reviewer loop exhausted or incomplete.

For ensemble evidence, report only structural results:

- Expected collaborator identifier.
- Contribution exists: yes or no.
- Identity marker matches: yes or no.
- File size and modification time.

Do not include contribution content or its first line.

Recovery-only remedies such as `AIDLC_DISABLE_ENSEMBLE_EVIDENCE=1` must carry a prominent warning that they bypass validation and are appropriate only when legitimate evidence was lost. They must never be marked safe to automate.

### Normalized Evidence

Export allowlisted fields rather than copying source files verbatim:

- Selected workflow state fields needed for routing and gate diagnosis.
- Selected audit event types and fields.
- Structural runtime graph data.
- Hook-health timestamps and statuses.
- Plan and recovery status.
- Turn counter and latch states.
- Sensor names and outcomes.
- Artifact, contribution, and question-file metadata.

Do not export raw `aidlc-state.md`, audit files, runtime graphs, artifact bodies, contribution bodies, question bodies, memory, environment variables, or command output.

## Redaction and Safety

Apply redaction before writing any bundle file:

- Home directories become `~`.
- Project root becomes `<project>`.
- Intent identifiers in filenames become stable short hashes.
- Unit and custom artifact identifiers are hashed when not required for diagnosis.
- Core stage and agent slugs may remain because they identify framework behavior.
- Error messages and audit fields are scanned for absolute paths and secret-like values.
- Symlinks are never followed.
- Individual files and total bundle size are capped.
- Bundle files are created with owner-only permissions where supported.

Add `manifest.json` containing:

- Bundle schema version.
- AI-DLC version.
- Harness name and version.
- Creation timestamp.
- Hashed intent identifier.
- Included files.
- Applied redaction rules.
- Per-file checksums.
- Truncation notices.

Tests must use secret canaries and verify that none appear anywhere in the generated directory or archive.

## Packaging

Do not implement a bespoke tar format in the first iteration.

Packaging order:

1. Generate and validate the canonical bundle directory.
2. Use a runtime-supported archive API when available.
3. Otherwise use an available system `tar` implementation.
4. If archiving is unavailable or fails, retain the directory and print clear manual-sharing instructions.

This keeps the diagnostic feature portable without introducing an archive parser or new package dependency.

## Output

Example:

```text
Diagnostic bundle created:
  aidlc-doctor-bundle-20260714T144321Z-a81f09.tar.gz

Findings:
  ERROR ensemble-evidence-missing
  WARNING runtime-graph-stale

No source files or artifact bodies were included.
```

## Explicitly Excluded

- Workspace source code.
- `node_modules` and build output.
- Artifact and contribution bodies.
- Question and answer bodies.
- Memory files.
- Raw environment variables.
- Raw command output.
- Unredacted state, audit, or runtime graph files.
- LLM-generated diagnosis.
- Automatic execution of recovery commands.

## Harness Support

The first implementation must support currently shipped harnesses:

- Claude Code.
- Kiro CLI.
- Kiro IDE.
- Codex CLI.

OpenCode support should be added when that harness is integrated into the target branch. The bundle model must remain harness-neutral so projection requires only a utility-handler wiring change.

## Acceptance Criteria

- [ ] `/aidlc --doctor --bundle` always performs a fresh doctor run.
- [ ] Produces `report.md`, `report.json`, `manifest.json`, and normalized evidence.
- [ ] Produces a timestamped `.tar.gz` when archive support is available.
- [ ] Falls back to the canonical bundle directory when archiving fails.
- [ ] Reconstructs stages, durations, gates, revisions, and reviewer iterations without inventing missing events.
- [ ] Diagnoses unresolved gates, ensemble evidence failures, state-audit drift, stale runtime graphs, and cold hooks.
- [ ] Uses one shared structured finding model for live doctor and bundle output.
- [ ] Every finding has a fixed remedy and `safeToAutomate` classification.
- [ ] Recovery bypasses are visibly marked unsafe.
- [ ] Exports normalized allowlisted fields instead of raw workflow files.
- [ ] Never includes artifact, contribution, question, or memory bodies.
- [ ] Redacts project paths, home paths, intent identifiers, and secret-like values.
- [ ] Does not follow symlinks.
- [ ] Enforces per-file and total-size limits.
- [ ] Creates owner-only files where supported.
- [ ] Includes schema/version metadata, checksums, and redaction details.
- [ ] Secret-canary tests verify that sensitive values do not appear in either directory or archive output.
- [ ] Works across all currently shipped harnesses.
- [ ] Adds OpenCode support when that harness is integrated.
- [ ] Introduces no package dependency.
- [ ] Updates the utility-command reference and troubleshooting guide.
- [ ] Updates the Utility Handler Checklist.
- [ ] Includes the required version bump and CHANGELOG entry.

leandrodamascena:
Closes #575.

## Problem

When a workflow misbehaves — a gate that will not open, a stage that will not advance, an approved report that is repeatedly refused — debugging today means asking the user for their whole project directory. That is huge (source, node_modules, build output), leaky (the record dir holds requirements, designs, decisions, questions), and unmastigated (we hand-reconstruct the run from state + audit + markers + graph).

Two recent reviews showed the shape of it: #568 (a Kiro collaborator can't write its contribution, so the ensemble evidence gate stays blocked) and #571 (runtime-compile fails to fire, leaving runtime-graph.json stale with no user-facing explanation). Both surface automatically from a diagnostic bundle.

## What this adds

`/aidlc --doctor --bundle` runs a fresh doctor pass, then writes a small, redacted, self-diagnosing bundle. The value is the diagnosis, not raw file collection.

- **report.md / report.json** — the reconstructed timeline (stage durations, gates, revisions, reviewer iterations, gaps, incomplete/long flags) plus findings. Missing evidence is `unknown`; no event is inferred.
- **manifest.json** — bundle schema version, AIDLC version, harness, hashed intent id, per-file sha256 checksums, applied redactions, truncation notices, and the excluded-content list.
- **evidence/normalized.json** — allowlisted, structural fields only.

Packaged as a timestamped `.tar.gz` when a system `tar` is available; otherwise the directory is retained with manual-share instructions. `--bundle-out <dir>` overrides the location (default `aidlc/diagnostics/`).

### Diagnosis is deterministic (no LLM)

A fixed, versioned condition→remedy table: unresolved gates, missing/malformed ensemble evidence, state/audit drift, stale or missing runtime graph, cold hook heartbeats, incomplete reviewer loops. Recovery bypasses (e.g. `AIDLC_DISABLE_ENSEMBLE_EVIDENCE=1`) are always reported unsafe-to-automate with a prominent warning. This is the deliberate line between "powerful" and "dangerous" — suggestion-grade output with check-grade safety.

### Shared finding model

Both the live `--doctor` report and the bundle draw from one `DoctorFinding` model, so the command and the bundle can never develop separate diagnostic rules or remediation text.

### Safety

Redaction runs before any file is written: home → `~`, project root → `<project>`, intent ids hashed, secret-like values scrubbed. Symlinks are never followed; per-file and total size are capped; files are owner-only where the platform supports it. The ensemble-evidence check is **structural only** — existence + identity-marker match, never the contribution body or its first line's content.

### Explicitly excluded

Workspace source, node_modules/build output, artifact/contribution/question/memory bodies, raw state/audit/runtime-graph files, environment variables, command output, and LLM-generated diagnosis. No new package dependency.

## Example

```text
Diagnostic bundle created:
  aidlc/diagnostics/aidlc-doctor-bundle-20260714T153357Z-f5a44b21.tar.gz

Findings:
  ERROR gate-unresolved
  WARNING runtime-graph-missing

No source files or artifact bodies were included.
```

## Implementation

The whole feature lives in a new module `core/tools/aidlc-doctor-bundle.ts` (~1340 lines) so the safety-critical `aidlc-utility.ts` takes only a small `--bundle` wiring branch in `handleDoctor`. Harness-neutral: it projects to all four shipped trees with no per-harness code (opencode wiring is a one-line follow-up when that harness lands).

## Tests

- **t237-doctor-bundle** (new): secret-canary test asserts an AWS key, a password, an absolute home path, and the raw intent slug appear in NEITHER the bundle directory NOR the produced `.tar.gz`; plus end-to-end (report/manifest/evidence present, gate-unresolved finding, real sha256 checksums) and unit coverage of redaction, timeline reconstruction, `isRecoveryBypass`, and `adaptLegacyResult`.
- Coverage registry regenerated (t237 added to the deterministic-spawner set).

## Verification

- Typecheck: 0 errors. `bun scripts/package.ts --check`: clean on all four trees.
- Full unit tier: net failures at/below base (the one failure, t115, fails identically on clean v2 and this PR touches neither orchestrate nor state).
- Version 2.3.12 + README badge + CHANGELOG heading agree (t68 green).

## Upgrade

Re-copy your `dist/<harness>/` shell into the project.


apackeer (changes requested):
Requesting changes for the following blocking issues:

1. **The documented user-facing command does not reach the exporter.** `parseNextFlags` recognizes `--doctor` as a terminal command, but the Branch 1 directive in `core/tools/aidlc-orchestrate.ts:1341-1345` emits only `aidlc-utility.ts doctor`; it drops `--bundle` and `--bundle-out`. The Kiro intercept does the same in `harness/kiro/hooks/aidlc-kiro-adapter.ts:138-142`, where `utilArgs` contains only the subcommand and optional workspace argument. A related detail: `parseNextFlags` does not consume the `--bundle-out` value token, so `/aidlc --doctor --bundle-out /tmp/x` leaks `/tmp/x` into the freeform intent words (harmless today only because Branch 1 wins first, but it will bite any refactor). The new test invokes `aidlc-utility.ts doctor --bundle` directly, so it bypasses both real routing paths. Reproduction:

   ```json
   {"kind":"print","message":"Run `bun .claude/tools/aidlc-utility.ts doctor` ..."}
   ```

   Please carry allowlisted doctor export arguments through `TerminalCommand`, the orchestrator directive, and the Kiro adapter, with a test that starts at `aidlc-orchestrate.ts next`.

2. **The share-safety contract is not currently met.** The PR says symlinks are never followed, but state and runtime graph inputs are read through `existsSync`/`readFileSync` in `core/tools/aidlc-doctor-bundle.ts:878-887`, and `readAllAuditShards` also follows symlinked shard files (`safeLstat` guards only the contribution check and the stage-source walk). In addition, custom stage identifiers are hashed only while rendering Markdown: raw `timeline` and graph structures are serialized into `report.json` and `evidence/normalized.json` at lines 926-946. The redaction context seeds only the active intent slug, despite the module header claiming "intent/unit ids" are hashed. Please reject symlink inputs and structurally sanitize/hash custom stage, agent, unit, artifact, and all intent identifiers before any JSON serialization. Add canaries for symlink targets and custom identifiers.

3. **The live doctor and exported report do not use one diagnostic model.** `handleDoctor` renders the legacy `results` array first (`core/tools/aidlc-utility.ts:2536-2555`). The new condition/remedy rules run later and only inside `buildBundle` (`core/tools/aidlc-doctor-bundle.ts:906-924`). A plain `--doctor` therefore cannot show `gate-unresolved`, `runtime-graph-stale`, or the other new findings, contrary to the issue and PR description. Please run one fresh analysis that returns a single explicit `DoctorFinding[]`, render the live output from it, and pass that same analysis to the exporter. Finding IDs and severities should be explicit rather than inferred from display labels.

4. **Timeline reconstruction is not reliable across shards or repeated runs.** `parseAuditEvents` preserves concatenation order even though `readAllAuditShards` sorts shard filenames, not events by timestamp. The module's own comment at `core/tools/aidlc-doctor-bundle.ts:254-255` claims `readAllAuditShards` "already merge-sorts", which contradicts the lib's contract: "cross-shard ordering by timestamp is the parsers' job" (`core/tools/aidlc-lib.ts:1598-1599`). Since per-clone shards (`audit/<host>-<clone>.md`) are the committed multi-machine design, this breaks exactly the multi-clone case a maintainer would be debugging. `reconstructTimeline` also chooses the first `WORKFLOW_STARTED` and groups all attempts by stage slug. Multi-host/worktree shards or a restarted workflow can therefore produce incorrect starts, durations, and gate outcomes. Please timestamp-sort with ledger position as the tie-breaker, scope the report to the latest workflow run (or explicitly group runs), and ensure an older approval cannot resolve a newer open gate.

5. **The ensemble-evidence rule diagnoses a mechanism that does not exist on `v2`, and false-positives on the shipped graph.** Rule 2 (`core/tools/aidlc-doctor-bundle.ts:525-576`) checks `contributions/<agent>.md` files with a `**Collaborator:** <agent>` first line and recommends `AIDLC_DISABLE_ENSEMBLE_EVIDENCE=1`. None of that exists on this base: no code writes contribution files, there is no `"mob"` mode (`VALID_MODES = ["inline", "subagent", "agent-team"]`, `core/tools/aidlc-stage-schema.ts:109`), and the only `AIDLC_DISABLE_*` env in the codebase is the reviewer-scope one. That whole contract is PR #568 (2.5.0, unmerged). Worse, the rule also fires on `subagent` stages with support agents, and the shipped graph has exactly one: `reverse-engineering` (subagent, supports `aidlc-architect-agent`). So on today's `v2`, every brownfield workflow that has run reverse-engineering gets a bogus ERROR `ensemble-evidence-missing` finding whose remedy names an env var that does nothing. Please either land this rule after #568 or gate it on the mechanism actually existing.

6. **The reviewer timeline signal is dead code.** `reviewerSignal` and the evidence allowlist read `**Review**` / `**Review Iterations**` audit fields (`core/tools/aidlc-doctor-bundle.ts:418-435, 799`), but nothing in the codebase ever writes them: the verdict lives in a `## Review` section appended to the primary artifact, and the iteration counter lives only in conductor context (`core/aidlc-common/protocols/stage-protocol.md:894-905`). `reviewerOutcome` is therefore always `none`, Rule 7 (`reviewer-loop-incomplete`) can never fire, and the PR description's "reviewer iterations" claim is unimplementable as written. Please either add a real audit emission for the reviewer verdict (a separate, deliberate change) or drop the advertised capability.

7. **Slot collisions on rebase.** `v2` already has `tests/unit/t237-linter-sensor-version-pin.test.ts` (merged in #572), so `t237-doctor-bundle.test.ts` is a duplicate t-number and must be renumbered (note #560 currently claims t238, and #573 holds 2.3.11). The 2.3.12 version, README badge, and CHANGELOG heading need re-slotting per the AGENTS.md conflict-trap when you rebase.

8. **The default output location gets committed into users' repos.** The default `aidlc/diagnostics/` sits under `aidlc/`, where the shipped `dot-gitignore` deliberately commits everything not explicitly listed. The PR adds no ignore rule, so bundle directories and `.tar.gz` archives are committed by default, which defeats the point of a share-then-delete diagnostic artifact. Please add `aidlc/diagnostics/` to every harness's shipped gitignore or default to a location outside `aidlc/`.

Non-blocking findings, several inside the safety-critical paths:

- `firstLineIsMarker` (`core/tools/aidlc-doctor-bundle.ts:740`) claims "Reads a bounded prefix so a huge file cannot blow memory" but does a whole-file `readFileSync`.
- `harnessTree()` (line 1337) reimplements `harnessDir()` and ignores the `AIDLC_HARNESS_DIR` override the lib honors (`core/tools/aidlc-lib.ts:184-190`), skewing `manifest.harness`, the shipped-graph path, and the stage-source mtime walk on env-overridden installs. The module already imports a dozen lib helpers; import this one too.
- Per-file truncation slices UTF-16 code units, not bytes (line 978), so the 512 KiB cap can overshoot severalfold on multibyte content, and `enforceTotalBudget`'s single-pass halving is not guaranteed to converge. Truncating `report.json`/`normalized.json` mid-stream also produces unparseable JSON with only a prose notice.
- `--bundle` followed by any non-flag token silently disables bundling: `parseArgs` (`core/tools/aidlc-lib.ts:4422-4429`) consumes the next token as the flag's value, so `flags.bundle === "true"` fails without an error.
- A same-second rerun silently overwrites the previous `.tar.gz`; intent slugs shorter than 4 characters are never redacted (the `id.length >= 4` guard at line 210).

I also recommend changing the user-facing name. This feature produces a diagnostic report with normalized evidence; it is not a raw log dump. `bundle` describes the archive container rather than the action. A clearer surface is:

```text
/aidlc --doctor --export
/aidlc --doctor --export --output <dir>
```

The resulting artifact can be named `aidlc-diagnostic-report-<timestamp>.tar.gz`. `--support-report` is a reasonable alternative. I would avoid bare `--report` because `/aidlc compose --report <path>` already uses it, and avoid `--dump`/`--logs` because those imply raw, complete data.

The redaction core and the secret-canary test are solid and worth keeping. The targeted `t237` tests and package drift check pass, but they do not exercise the broken user-facing route, and two of the seven advertised diagnosis rules are respectively wrong-on-`v2` (item 5) and unreachable (item 6). The PR is also currently conflicting with `v2` and will need to be rebased and regenerated after these changes.

leandrodamascena:
Thanks Arden — all four blocking points addressed, plus the rename. Pushed as a rebase over #572 (now 2.3.13).

**1. The export surface now reaches the exporter through the real path.** The problem was exactly as you traced it: `parseNextFlags` treated `--doctor` as terminal but the Branch 1 directive dropped everything after it, and the Kiro adapter's `utilArgs` did the same. Fixed by carrying an allowlisted trailing-arg set (`--export`, `--output <dir>`) through both deciders: `parseNextFlags` now captures `readOnlyArgs` and appends them to the directive, and `classifyTerminalCommand` returns `TerminalCommand.extraArgs` which the Kiro adapter pushes onto `utilArgs`. A fixed allowlist (`DOCTOR_EXPORT_FLAGS`) keeps the two deciders in parity so an arbitrary token can't ride the terminal path into the tool. New tests start at `aidlc-orchestrate.ts next --doctor --export --output <dir>` (asserts the directive names the export) and at `classifyTerminalCommand` (asserts `extraArgs`).

**2. Share-safety contract now met.** `safeRead` refuses to follow a symlinked input (state, runtime graph, plan, markers), and the audit trail is refused whole if any shard file is a symlink. All custom identifiers — every intent id across spaces, plus custom stage/agent/unit/artifact slugs — are seeded into the redaction context and hashed **before** any JSON serialization, so `report.json` and `evidence/normalized.json` are scrubbed too, not just the Markdown. Added canaries for a symlinked `runtime-graph.json` target and for a custom stage slug (with a positive control asserting the `<id:hash>` form is present).

**3. One diagnostic model.** `runDoctorAnalysis()` runs a single fresh pass returning an explicit `DoctorFinding[]` (ids/severities are explicit, not inferred from labels). `handleDoctor` renders the live report from it — a plain `--doctor` now shows a `Workflow diagnosis` section with `gate-unresolved` / `runtime-graph-stale` / cold-hook etc. — and hands that same analysis object to the exporter. They can't diverge.

**4. Timeline reliability.** `parseAuditEvents` now timestamp-sorts with ledger position as the tie-breaker; `reconstructTimeline` scopes to the **latest** `WORKFLOW_STARTED` run (earlier runs noted and omitted); `gateOutcome` honours event order so a re-opened gate reads `unresolved` and an older approval can't resolve a newer open gate.

**Naming:** took your suggestion — `--doctor --export` / `--output <dir>`, artifact `aidlc-diagnostic-report-<ts>.tar.gz`. Avoided `--report` (collides with `compose --report`) and `--dump`/`--logs`.

Verification: typecheck clean, `package.ts --check` clean on all four trees, t237 (11 tests incl. the symlink + custom-id canaries) + t178 (classifier `extraArgs` cases) green, full unit tier at/below base failures (the remaining failures are the pre-existing t115 set, unrelated).

leandrodamascena:
Thanks @apackeer — all four blocking points addressed, plus the rename. Pushed as a rebase over #572 (now 2.3.13).

**1. The export surface now reaches the exporter through the real path.** The problem was exactly as you traced it: `parseNextFlags` treated `--doctor` as terminal but the Branch 1 directive dropped everything after it, and the Kiro adapter's `utilArgs` did the same. Fixed by carrying an allowlisted trailing-arg set (`--export`, `--output <dir>`) through both deciders: `parseNextFlags` now captures `readOnlyArgs` and appends them to the directive, and `classifyTerminalCommand` returns `TerminalCommand.extraArgs` which the Kiro adapter pushes onto `utilArgs`. A fixed allowlist (`DOCTOR_EXPORT_FLAGS`) keeps the two deciders in parity so an arbitrary token can't ride the terminal path into the tool. New tests start at `aidlc-orchestrate.ts next --doctor --export --output <dir>` (asserts the directive names the export) and at `classifyTerminalCommand` (asserts `extraArgs`).

**2. Share-safety contract now met.** `safeRead` refuses to follow a symlinked input (state, runtime graph, plan, markers), and the audit trail is refused whole if any shard file is a symlink. All custom identifiers — every intent id across spaces, plus custom stage/agent/unit/artifact slugs — are seeded into the redaction context and hashed **before** any JSON serialization, so `report.json` and `evidence/normalized.json` are scrubbed too, not just the Markdown. Added canaries for a symlinked `runtime-graph.json` target and for a custom stage slug (with a positive control asserting the `<id:hash>` form is present).

**3. One diagnostic model.** `runDoctorAnalysis()` runs a single fresh pass returning an explicit `DoctorFinding[]` (ids/severities are explicit, not inferred from labels). `handleDoctor` renders the live report from it — a plain `--doctor` now shows a `Workflow diagnosis` section with `gate-unresolved` / `runtime-graph-stale` / cold-hook etc. — and hands that same analysis object to the exporter. They can't diverge.

**4. Timeline reliability.** `parseAuditEvents` now timestamp-sorts with ledger position as the tie-breaker; `reconstructTimeline` scopes to the **latest** `WORKFLOW_STARTED` run (earlier runs noted and omitted); `gateOutcome` honours event order so a re-opened gate reads `unresolved` and an older approval can't resolve a newer open gate.

**Naming:** took your suggestion — `--doctor --export` / `--output <dir>`, artifact `aidlc-diagnostic-report-<ts>.tar.gz`. Avoided `--report` (collides with `compose --report`) and `--dump`/`--logs`.

Verification: typecheck clean, `package.ts --check` clean on all four trees, t237 (11 tests incl. the symlink + custom-id canaries) + t178 (classifier `extraArgs` cases) green, full unit tier at/below base failures (the remaining failures are the pre-existing t115 set, unrelated).

apackeer:
Re-reviewed at d039141. The four points from my previous review were genuinely addressed: the export surface now reaches the tool through the real routing path on both deciders (`parseNextFlags` + `classifyTerminalCommand`, verified end to end by t237 tests 8/9), leaf-level symlinked inputs are refused (test 10), custom stage slugs hash in both the Markdown and JSON renders (test 11), and `handleDoctor` really does run one shared analysis. Typecheck is 0 errors, `package.ts --check` is clean, and the version trio (2.3.13 / CHANGELOG / badge) agrees.

That said, keeping **Changes requested**: two of the new diagnosis rules fire falsely on valid projects, and because plain `--doctor` now counts diagnosis errors into its exit code, that is a regression for existing users independent of `--export`.

## Must fix

1. **Rule 2 (ensemble evidence) is invalid on this base and false-errors every brownfield project.** `core/tools/aidlc-doctor-bundle.ts:560-611` requires a `contributions/<agent>.md` with a `**Collaborator:** <agent>` first line for subagent-with-supports stages. Neither that convention nor `AIDLC_DISABLE_ENSEMBLE_EVIDENCE` exists on v2 (`git grep Collaborator v2 -- core/` is empty; the env var appears only inside this new file) - that mechanism is #568, which is not merged. The shipped graph has exactly one matching stage, `reverse-engineering` (`core/aidlc-common/stages/inception/reverse-engineering.md:7-9`, mode `subagent` + one support agent), so every valid brownfield workflow that ran it gets a false ERROR, and the remedy names a bypass env var that does nothing here. Either gate this rule on the evidence mechanism actually existing, or drop it until #568 lands.

2. **Rule 3 (state/audit drift) scans the whole audit buffer and false-errors restarted workflows.** `core/tools/aidlc-doctor-bundle.ts:615` does `audit.includes("**Event**: WORKFLOW_COMPLETED")` over all runs, while `reconstructTimeline` (lines 316-327) was deliberately scoped to the latest `WORKFLOW_STARTED` in this same commit. A completed-then-restarted workflow (the exact scenario the scoping comment names) reports `state-audit-drift`. Scope this rule to the same latest-run slice.

3. **These false positives now break plain `/aidlc --doctor`, not just the export.** `core/tools/aidlc-utility.ts:2564-2574` counts diagnosis errors into `failed` and line 2628 exits 1 on any failure, so points 1 and 2 flip doctor's exit code on healthy projects, including any CI gate scripted on it.

4. **t237 slot collision.** `tests/unit/t237-linter-sensor-version-pin.test.ts` owns t237 (merged via #572, 2.3.10); this PR adds `tests/unit/t237-doctor-bundle.test.ts` beside it. Renumber - open PRs currently claim through t240, so t241 is the next free slot.

5. **`aidlc/diagnostics/` is untracked-output but not gitignored.** None of the four `harness/*/dot-gitignore` files exclude it, so the default export location lands git-tracked in user projects. Per the gitignore's own committed-vs-ignored rule, machine-local derived output should be ignored.

6. **Custom lead agents leak unredacted.** `seedCustomIdentifiers` (`core/tools/aidlc-doctor-bundle.ts:908-939`) seeds `s.slug` and `s.support_agents` but never `s.lead_agent`, and `evidence.graph.stages` (line 1001) serializes full `GraphStageLite` rows including `lead_agent`. A plugin or adaptive-workflow custom lead agent therefore appears raw in `normalized.json`, contradicting the CHANGELOG claim that every custom agent identifier is hashed. (`phase` rides along in the same rows; compiled graphs are constrained to the 5 core phases by `aidlc-graph.ts:1646`, but `readGraphStages` does no validation, so a hand-edited or corrupt graph - the very thing doctor inspects - can leak a custom phase string too. Seeding `lead_agent` the same way as `support_agents` covers the realistic case.)

## Remainder (should fix, non-blocking individually)

7. **Parent-directory symlinks are still followed.** `safeRead` (`core/tools/aidlc-doctor-bundle.ts:1394`) lstats only the final path component, and `readAuditSafely` (line 1424) checks entries inside the shard dir but not whether `audit/` itself is a symlink - `existsSync`/`readdirSync` traverse a symlinked directory happily. The field allowlist bounds the blast radius, but the "symlinks are never followed" guarantee in the module header, PR body, and docs is false as stated. Check each path component, or realpath-and-compare against the project root.

8. **Repeated stage attempts reconstruct incorrectly.** `reconstructTimeline` pairs `firstEvent(STAGE_STARTED)` with `lastEvent(STAGE_COMPLETED)` (lines 356-357) and only flags `incomplete` when no completion exists at all (line 384). `aidlc-jump.ts:455` re-emits STAGE_STARTED when jumping back to a completed stage, so a stage under active re-work reports completed with a stale duration. `aidlc-runtime.ts`'s `pairStartedCompleted` already does chronological pairing - reuse or mirror it.

9. **Reviewer diagnosis is unreachable.** `reviewerSignal` (lines 452-469) reads `**Review**` / `**Review Iterations**` audit fields that no emitter writes (grep across `core/hooks` + `core/tools` finds only `REVIEWER_SCOPE_BLOCKED`). Rule 7 and the timeline's reviewer column are dead code while `docs/guide/12-cli-commands.md:309-314` advertises reviewer-loop diagnosis. Either read the verdict from where it actually lives, or drop the rule + docs claim.

10. **Truncation produces invalid JSON.** `stage()` (line 1066) and `enforceTotalBudget` (line 1086) slice serialized JSON and append a plain-text notice, so a truncated `report.json`/`normalized.json` no longer parses - defeating the machine-readable artifact. Truncate by dropping whole elements (e.g. oldest audit events) and re-serializing, with the notice in the manifest only.

11. **The `assignment-secret` pattern misses JSON-shaped secrets.** The regex (line 191) requires the key followed directly by `:`/`=`, so serialized `"password": "hunter2"` never matches (the closing quote breaks `\s*[:=]`). Since whole JSON documents are redacted post-serialization, JSON-style assignments inside field values slip the scan. The canary test only exercises the `password=x` form - add a JSON-shaped canary alongside the regex fix.

12. **`--output` with a missing value creates a directory named `true`.** `parseNextFlags` pushes a bare `--output` (orchestrate lines 317-334) and `handleDoctor` uses `flags.output` verbatim (utility lines 2601-2603), which the flag parser defaults to `"true"`. Error out or fall back to the default location.

Nits, take or leave: the `parseAuditEvents` comment says unparseable-timestamp events "keep their ledger position" but they actually sort to the end (lines 273-277); the unsalted 8-hex `shortHash` of an intent slug is dictionary-confirmable (a per-bundle salt keeps correlation while blocking guessing); a 4+ char custom slug that happens to be a substring of core prose gets mangled by the split/join replacement (garbles output, does not leak).

Verification on my side: targeted unit slice (t237-doctor-bundle, t178-classify-terminal-command, t68-version-changelog-sync) 3/3 green; `bun run typecheck` 0 errors; `bun scripts/package.ts --check` clean on all four trees.

leandrodamascena:
Thanks @apackeer  — all six must-fix items are resolved, and I also took the six remainder items plus the two actionable nits. Details below; each has a regression test where it was testable. Pushed at `e1f3f840`.

### Must-fix

**1 — Ensemble rule (Rule 2) gated on the mechanism existing.** You were right that `contributions/<agent>.md` + the `**Collaborator:**` marker + `AIDLC_DISABLE_ENSEMBLE_EVIDENCE` + `mob` mode are all #568 and absent on `v2`. Rule 2 now short-circuits when the stage's `contributions/` dir doesn't exist ("mechanism not in use"), so `reverse-engineering` can't false-error; it becomes live automatically once #568 ships the dir. Dropped the dead env-var remedy string and the orphaned doc/CHANGELOG mentions (`isRecoveryBypass` stays generic for real `AIDLC_DISABLE_*` vars). Regression: no finding fires for a reached RE stage with no `contributions/`.

**2 — Drift rule (Rule 3) scoped to the latest run.** `reconstructTimeline` now exposes `workflowCompleted` computed from the same latest-`WORKFLOW_STARTED` slice, and Rule 3 reads that instead of the whole-buffer `includes()`. A completed-then-restarted workflow reads In-Progress with no drift; rule and timeline can't disagree. Two regressions cover it.

**3 — Plain `--doctor` exit code restored.** The workflow diagnosis is now advisory: findings render under `Workflow diagnosis (advisory):` but no longer `failed++`, so only the legacy environment/config checks drive the exit code. CI gates on `--doctor` are unaffected.

**4 — t237 collision.** Renamed to `t242` — not t241, since #571 reserves both t240 and t241 for the opencode packaging + adapter suites. Coverage registry regenerated.

**5 — `aidlc/diagnostics/` gitignored** in all four shipped `dot-gitignore` files; dist regenerated.

**6 — Custom `lead_agent` hashed.** `seedCustomIdentifiers` now seeds `s.lead_agent` alongside slug/support_agents. Regression exports a graph with a custom lead and asserts the raw slug is absent and the `<id:hash>` token is present in `normalized.json`.

### Remainder (all fixed)

**7 — Parent-dir symlinks.** `safeRead`/`readAuditSafely` now realpath-check that an input's real location stays under the project root (catches a symlinked *parent* dir, not just a leaf), and the `audit/` dir itself is checked. Module header + PR/doc claim corrected to match.

**8 — Repeated attempts.** Timeline pairs chronologically — the current attempt is the last `STAGE_STARTED`; a completion predating it is ignored, so a jumped-back stage reads incomplete instead of stale-completed. Regression added.

**9 — Reviewer dead code.** Removed the `reviewer-loop-incomplete` rule, the `**Review**` / `**Review Iterations**` audit fields, and the reviewer timeline column — no emitter writes them. Docs updated to drop the advertised reviewer-loop capability. (Reinstate alongside a real reviewer-verdict audit emission if we want it back.)

**10 — Truncation validity.** Over-budget `report.json` / `normalized.json` are replaced by a format-preserving placeholder (valid JSON carrying `{truncated, reason, originalBytes}`), never a byte slice; the manifest still records it. Regression forces a truncation and asserts both parse.

**11 — JSON-shaped secrets.** `assignment-secret` regex now also matches `"password": "x"` (optional quote before the separator).

**12 — Bare `--output`.** Now errors with a clear message instead of creating a dir named `true`; `--export` detection is presence-based so a stray token can't silently disable it. Regression added.

Also folded in the `harnessTree()` → lib `harnessDir()` nit (honors `AIDLC_HARNESS_DIR`) and corrected the `parseAuditEvents` unparseable-timestamp comment.

**Left as-is (the "take or leave" nits):** the unsalted `shortHash` for intent-slug correlation (a per-bundle salt would break the cross-file correlation the report relies on) and the 4+char-slug substring cosmetic mangling. Happy to do either if you feel strongly.

**Rebase:** the branch is already current with `v2` — the earlier conflict was against a pre-#572 `v2`, and `d0391417` was rebased over it; it fast-forwards cleanly and the version trio stays at 2.3.13.

Verification: `bun scripts/package.ts --check` clean on all four trees, typecheck 0 errors, coverage registry fresh, and t242 (18 tests incl. 7 new regressions) + t134 (mechanism-honesty) + t68 + t178 all green (46/0). The only unit-tier failures remain the pre-existing t115/t205 gate-revision set, unrelated to this PR.

apackeer:
Re-reviewed at `e1f3f84`. All six must-fix items from my previous review are genuinely resolved, and I verified the remainder items too: Rule 2 short-circuits when the stage's `contributions/` dir is absent, Rule 3 reads the latest-run-scoped `timeline.workflowCompleted`, the workflow diagnosis is advisory and no longer touches doctor's exit code, t242 has no slot collision (v2 tops out at t237; open PRs claim t238-t241), `aidlc/diagnostics/` is ignored in all four `dot-gitignore` files, and custom `lead_agent` is seeded. `package.ts --check` is clean on all four trees and typecheck is 0 errors on my box.

Keeping **Changes requested**: this round surfaced a new set, all reproduced against `e1f3f84`.

## Must fix

1. **The export omits the legacy doctor failures; `adaptLegacyResult` and `mergeFindings` are dead code.** `handleDoctor` builds its legacy `{pass,label,fix}` rows but never lifts or merges them: `core/tools/aidlc-utility.ts:15` imports only `buildBundle` and `runDoctorAnalysis`, and neither `adaptLegacyResult` nor `mergeFindings` is called anywhere in the repo. `buildBundle` receives only `analysis.findings` (`core/tools/aidlc-utility.ts:2615`), so a failing environment check (bun missing, a wired hook absent, settings.json stripped) never reaches `report.md`/`report.json`. Since the bundle exists precisely so the maintainer does NOT get the user's project, dropping the environment failures defeats it. This also contradicts the module header (`core/tools/aidlc-doctor-bundle.ts:884-887`, "merges these findings with its legacy check rows ... and hands this whole object to buildBundle") and the CHANGELOG's "the live report and the export share one analysis". Either wire the adapter + merge in `handleDoctor` and pass the merged set to `buildBundle`, or delete both exports and fix the header/CHANGELOG claim.

2. **A malformed runtime graph crashes plain `--doctor`, not just the export.** `readGraphStages` (`core/tools/aidlc-doctor-bundle.ts:1303`) casts `support_agents` to `string[]` without validating the elements; `seedCustomIdentifiers`'s `seed()` then calls `shortHash` on a non-string (its `id.length < 4` guard passes for a number because `undefined < 4` is false) and `createHash().update(42)` throws. Reproduced: `The "data" argument must be of type string ... Received type number (42)`. The call site matters: `runDoctorAnalysis` runs uncaught at `core/tools/aidlc-utility.ts:2542`, OUTSIDE the export try/catch at 2604, so a corrupt `runtime-graph.json` (valid JSON, wrong shapes) takes down the whole health check, the exact situation doctor exists to diagnose. Validate the elements in `readGraphStages` (filter or String() them) and/or wrap the analysis call so an analysis failure degrades to a finding instead of a crash.

3. **Path containment is broken on Windows, silently emptying the analysis.** `withinProjectRoot` (`core/tools/aidlc-doctor-bundle.ts:1421`) tests `real.startsWith(bundleRealRoot + "/")`. On Windows `realpathSync` returns backslash-separated paths, so every nested input fails the check, and `safeRead`/`readAuditSafely` return `""` for every source. No error is raised: the bundle is produced, it just contains an empty timeline and no state. Windows is a supported target here (`handleDoctor` has explicit `isWindows` branches). Compare with `path.sep`, or normalize both sides, or use `path.relative` and check for a leading `..`.

4. **The redaction bypass is only half-closed: JSON-ESCAPED secrets and punctuation-heavy values still leak.** Your round-2 item 11 covers the plain quoted form, but staging serializes with `JSON.stringify` BEFORE redaction (`stage()` at `core/tools/aidlc-doctor-bundle.ts:1072`), so any secret one nesting level down arrives as `{\"password\":\"hunter2secret\"}`; the escaping backslash between the key and the quote breaks `['"]?\s*[:=]` and the value survives (reproduced). Likewise `password=p@ssw0rd!` survives because `@`/`!` are outside the value class and the remaining run is under 6 chars. Suggest allowing `\\?["']` around the separator and widening the value class (or redacting values BEFORE serialization via `redactValue`, which is currently also uncalled).

5. **t242 test 17 is red at this head, and the cause is a real scaling bug, not a slow CI box.** On my machine it times out at 39.0s against its 30s limit (`tests/logs/2026-07-17T09-44-44Z`, 17 pass / 1 fail). Root cause: the test seeds ~12,000 custom stage slugs, and `redactString` runs an `includes` plus split/join over the FULL staged content per seeded id (`core/tools/aidlc-doctor-bundle.ts:221-225`), so staging is O(seeded ids x content bytes) across ~1.5 MiB, three times over. A long-lived real project with a big audit trail and custom stages hits the same wall at export time. A single pass with one alternation regex built from the seeded ids (longest-first, escaped) fixes both the test and the field behavior; bumping the timeout would only hide it.

6. **The branch is CONFLICTING against v2 again.** v2 is now at 2.4.3 (the 2.3.13 note in the thread predates the 2.4.x merges), so this needs a rebase plus a version re-bump; check the slot map at rebase time (2.4.4 is currently claimed by an open PR).

## Should fix

7. **`runtime-graph-missing` false-warns on a clean shell.** Rule 4's else branch (`core/tools/aidlc-doctor-bundle.ts:662`) fires "missing for the active workflow" when there is no workflow at all: reproduced on a fresh copy of `dist/claude/` (zero stages, no intent, warning still emitted). Gate it on some evidence a workflow exists (an intent resolved, a non-empty timeline, or a state file).

8. **Repeated-stage timelines render nonchronologically with negative gaps.** Stages keep first-seen order (`stageOrder`, `core/tools/aidlc-doctor-bundle.ts:349-358`) while showing latest-attempt timestamps (line 375), so an alpha -> beta -> alpha(jump-back) run lists the day-2 alpha attempt first and beta's `gapFromPrevMs` computes from alpha's later completion: reproduced, gap of -97,200,000 ms. Sort the rendered rows by the attempt's start time (or key rows by attempt rather than by slug).

9. **The "symlinks are never followed" claim still overstates.** `docs/guide/12-cli-commands.md:322` says never; by design `withinProjectRoot` follows symlinks that resolve INSIDE the project root, and `firstLineIsMarker` (`core/tools/aidlc-doctor-bundle.ts:767`) does a raw whole-file read with no containment guard (exposure is only a 1-bit marker-match oracle since `safeLstat` rejects a symlinked leaf, but the doc claim should say "inputs whose real path escapes the project root are refused" rather than "never followed"). The comment above it ("Reads a bounded prefix so a huge file cannot blow memory", line 766) is also false: it reads the entire file; `split("\n", 1)` limits the split, not the read.

Repro scripts for items 2, 4, 7, and 8 are small standalone bun scripts; happy to attach them if useful.

leandrodamascena:
Thanks Arden — all 9 addressed (reproduced first, each with a regression test), rebased onto `v2` 2.5.0, re-slotted to **2.5.1**. Pushed at `a761499d`.

**Must-fix**
1. Export now merges the legacy env checks into the bundle (`mergeFindings(results.map(adaptLegacyResult), …)`) — a broken `settings.json` now shows up as errors in `report.json`. `adaptLegacyResult`/`mergeFindings` no longer dead.
2. Malformed graph no longer crashes `--doctor` — `readGraphStages` filters non-string `support_agents`, `seed()` guards non-strings.
3. Windows: `withinProjectRoot` uses `path.sep`, not `/` (was silently emptying the analysis).
4. Redaction catches JSON-escaped (`{\"password\":\"x\"}`) and punctuation (`p@ssw0rd!`) secrets.
5. `redactString` is now single-pass (one alternation regex, cached) — fixes t242 test 17's timeout and a latent overlapping-id suffix leak.
6. Rebased + re-bumped; export args reconciled onto #568's `TerminalCommand.args` (verified the Kiro seam runs `… doctor --export --output <dir>`).

**Should-fix**
7. `runtime-graph-missing` no longer warns on a clean shell (gated on a workflow existing).
8. Timelines render chronologically — no negative gaps on jump-back.
9. `firstLineIsMarker` refuses escaping paths + bounded read; docs "never followed" claim corrected.

Rule 2 stays correct now that #568 merged (short-circuits when `contributions/` is absent; `reverse-engineering` is `pipeline` on v2).

Verification: `package.ts --check` clean on all 5 trees, typecheck 0 errors, t242 25/25 + adapter suites green. The only unit failures (t236, t115, t205) are **identical on clean v2** — pre-existing, not introduced here.

apackeer (approved):
Re-reviewed at `a761499d`. All 9 round-3 items are genuinely resolved - I re-ran my repros against this head: the JSON-escaped and punctuation secrets redact, the malformed-graph analysis completes end-to-end without crashing, the jump-back timeline renders chronologically with a positive gap, the clean shell is silent, `mergeFindings`/`adaptLegacyResult` are wired at `core/tools/aidlc-utility.ts:2860` (test 25 covers it), containment uses `path.sep`, `firstLineIsMarker` does a bounded contained read, and the export args are correctly reconciled onto #568's `TerminalCommand.args` through the Kiro seam. Version trio consistent at 2.5.1, typecheck 0 errors, `package.ts --check` clean on all five trees.

**Approving**, on the expectation that these land before merge:

1. **t242 test 17 is still red here: two consecutive runs at 81.2s and 79.3s against its 30s timeout** (round 3 was 39s, so the alternation made it slower, not faster). A 12k-branch alternation is O(branches) per scan position in JS regex engines; I benchmarked the pattern at ~9.4s per `redactString` call on this input, and the export stages three files. My box is Linux arm64, so if it's green on yours it's hardware-marginal at best, and real CI runners will hit it too. Since seeded ids are all slug-shaped, a single token-scan pass does the same job in ~10ms on the same input (benchmarked): match `/[A-Za-z0-9_-]{4,}/g` once and look each token up in `ctx.idHashes`, replacing on a hit. That also preserves the longest-match property for overlapping ids for free.

2. **Redaction can break the bundle's JSON validity.** The trailing `['"]?` in `assignment-secret` eats the JSON string's closing quote when a secret-like token ends an allowlisted field value. Repro: a state file with `- **Scope**: rotate the leaked token=abcdef123456` produces `"Scope": "rotate the leaked token=<redacted>` with no closing quote, and `evidence/normalized.json` no longer parses. Cleanest fix: redact values BEFORE serialization via the already-present (currently uncalled) `redactValue()`, which makes syntax corruption structurally impossible; alternatively stop the trailing quote match from consuming a quote that closes a JSON string. Same contract as your round-2 truncation fix: the machine-readable artifacts must always parse.

3. **t242 filename collision, one more time (sorry):** #568 merged `tests/unit/t242-state-transition-guard.test.ts`, so this PR's `t242-doctor-bundle.test.ts` now sits beside it. t243 is free (nothing on v2 or open PRs claims it).

Nit, take or leave: `harness/opencode/dot-gitignore` (the new fifth tree from #571) is missing the `aidlc/diagnostics/` line the other four have.

leandrodamascena:
Thanks — all 4 landed at `e6dde1c1`. Reproduced #1 and #2 first, both were real (and both were regressions I introduced in round 3 — good catches).

1. **test 17 speed.** You were right that the alternation made it slower, not faster. Replaced it with the token-scan you suggested: one `/[A-Za-z0-9_-]{4,}/g` pass, look each token up in `ctx.idHashes`. O(content), independent of id count, and it keeps the longest-match property for free (a whole `build-auth-extra` token is looked up whole). test 17 went **81s → ~2s** (the redaction work itself is ~10ms; the rest is bun startup).

2. **JSON validity.** Fixed at the root as you suggested: `report.json`/`normalized.json` now redact-then-serialize via the previously-uncalled `redactValue()` (`stageJson`), so a secret-scan replacement never sees — and never eats — a JSON quote. Also dropped the trailing `['"]?` from `assignment-secret` so `report.md` prose can't be corrupted either. Your exact repro (`- **Scope**: rotate the leaked token=abcdef123456`) now redacts and `normalized.json` parses; new test 26 covers it through the real export path.

3. **t242 collision.** Renamed to `t243-doctor-bundle.test.ts` (t243 free on v2 and open PRs).

4. **opencode gitignore.** Added `aidlc/diagnostics/` to `harness/opencode/dot-gitignore` — all five trees now carry it.

Verification: `package.ts --check` clean on all five trees, typecheck 0 errors across all three tsconfigs, coverage registry fresh, t243 26/26. Still a single commit on `v2` at 2.5.1.

PR #577, opened 2026-07-14
fix: packager contract - whole-root drift guard, manifest harnessDir, seed preservation (2.4.2)

apackeer:
## Summary

Treats each `dist/<harness>/` directory as one generated root, closing drift gaps the per-directory contract missed.

- Graph compilation and runner generation receive the manifest's actual `harnessDir` - a custom harness dir (e.g. `.foo`) no longer compiles or generates runners as `.claude`.
- An explicitly named harness without a manifest exits nonzero instead of being silently skipped.
- `bun scripts/package.ts --check` builds the complete tree into a temp root and diffs bidirectionally against the committed dist: modified, removed, and renamed project-root outputs (Kiro `AGENTS.md`, `.mcp.json`, `.gitignore`) are now caught, and outputs removed from a manifest surface as ORPHANs.
- Emitters (`emit.ts`) always write into the packager-provided dist root; the `EmitResult`/`ctx.check` self-check contract is removed - the whole-root diff subsumes it.
- `authoredExempt` is removed from the manifest contract: everything the packager writes is already in the built inventory, so the exemptions could only hide genuinely stale files.
- The compile seed survives a harnessDir rename via a canonical-seed fallback.

Byte-identical output preserved for all four shipped harnesses (`--check` green).

## Version

2.3.12. Slot ladder: 2.3.11 is held by #573, 2.3.13 by #576; the companion PRs for this effort take 2.3.14 (codex trust TOML) and 2.3.15 (documentation parity). Merge this first among the three.

## Testing

- `bun scripts/package.ts --check` green (all four harnesses)
- `bun run typecheck` green
- t145 regression suite extended with a test per contract case: synthetic-harness compile/runners, unknown-harness exit, modified root files, removed/renamed root files as orphans, seed preservation across a harnessDir rename, manifest-removed file as ORPHAN, single-root write isolation
- Full deterministic tier (smoke+unit+integration) green at the integration tip that unions this with its sibling PRs (single red: t92, pre-existing on v2, owned by #573)


leandrodamascena (changes requested):
I ran a deep pass and initially leaned approve — the whole-root drift guard is a real strengthening (I mutation-tested it: a stray root file above the harnessDir → ORPHAN, a modified `dist/kiro/AGENTS.md` → DIFFERS, `--check` non-destructive, build byte-identical). But two things I'd missed turned out to matter, and I verified both. One is a CI-red this PR ships.

**Verdict: CHANGES REQUESTED** — one blocking.

**Blocking:**

1. **`gen-coverage-registry.test.ts` fails at this PR's head** (`tests/unit/gen-coverage-registry.test.ts:932`, the "none→cli reclassification set" assertion). Extracting `spawnSync` into the new `runPackage()` helper (`tests/integration/t145-packaging-parity.test.ts:62`) makes the body-derived mechanism classify t145-packaging as a CLI spawner, but it wasn't added to `EXPECTED_NONE_TO_CLI`. I confirmed it directly: the test passes clean on origin/v2 (0 fail) and fails at PR head (1 fail), so this PR ships a red CI check. Same coverage-registry class as #576/#580 — the fix is one line (add `integration/t145-packaging-parity.test.ts` to `EXPECTED_NONE_TO_CLI`) plus regenerating the registry (`bun tests/gen-coverage-registry.ts`).

**Should-fix (not blocking):**

2. **Codex `emit.ts` still hardcodes `.codex`** (`harness/codex/emit.ts:254` `DCODEX = join(distRoot, ".codex")` and `:318` `process.env.AIDLC_HARNESS_DIR = ".codex"`, plus `ADAPTER_CMD` and the rules-rename). `EmitContext` now carries `harnessDir` and the PR's headline is "manifest-driven harnessDir" — `runTool` / graph-compile / runner-gen all honor it, but codex emit is the one path left ignoring it, so a codex-shaped harness renamed to `.foo` breaks (`Cannot find module .../dist/codex/.codex/tools/aidlc-runner-gen.ts`). Latent, not live: the codex manifest sets `harnessDir: ".codex"`, so the hardcode equals the manifest value today and no shipped harness is affected. But it contradicts the contract this PR establishes — worth threading `ctx.harnessDir` through `DCODEX`, `AIDLC_HARNESS_DIR`, `ADAPTER_CMD`, and the rules rewrite so codex matches the other paths.

**Rebase note:** v2 now ships 2.4.1 (via #573); this PR is still 2.3.12. Needs a re-slot on rebase, and the version/badge/CHANGELOG will conflict — routine second-to-merge bump.

**What I verified holds** (so the direction is right): the whole-root `--check` catches strictly more, bidirectionally (mutation-proven); it's non-destructive (build goes to a tmp root, tree stayed clean); removing `authoredExempt` + the emit self-check doesn't false-orphan (every formerly-exempt pattern is in the built inventory now); `harnessDir` threading through `runTool` is correct for the four shipped harnesses; the compile-seed fallback is inert for them; the unknown-harness exit is scoped to an explicitly-named target with no manifest. Once #1 is fixed the check itself is a net improvement over the old per-directory contract.

leandrodamascena (approved):

PR #578, opened 2026-07-14
fix: serialize Codex hook-trust entries with a real TOML library and validate trust arguments (2.4.3)

apackeer:
## Summary

`bun scripts/package.ts codex trust` previously built its `[hooks.state]` entries with ad hoc string concatenation, producing invalid or wrong TOML for any path containing spaces, quotes, or backslashes - essentially every Windows path.

- Trust entries are now serialized with smol-toml (pinned 1.7.0); tests parse the complete generated output back with the same standards-compliant parser.
- Unix paths, Windows drive-letter paths, UNC paths (both `\\server\share` and the forward-slash `//server/share` spelling, which is simultaneously posix-absolute and win32-absolute), spaces, quotes, and backslashes all round-trip exactly. The default `.codex/hooks.json` key is joined with the path implementation matching the project argument's spelling, so a Unix-invoked packager generates native Windows keys and vice versa.
- An explicit `--hooks-json` path is preserved verbatim as the trust identity - never joined or normalized.
- The trust command now rejects relative paths, missing values, duplicate flags, and unknown arguments with the reason plus usage on stderr, instead of printing trust keys that would never match.
- Normal Unix output stays byte-identical to the historical form (pinned in test); the shipped `dist/codex/.codex/trust-seed.toml` regenerates identically through the packager.

Codex harness docs updated for quoted paths and `--hooks-json`.

## Version

2.3.14. Slot ladder: 2.3.11 held by #573, 2.3.12 is #577 (packager contract), 2.3.13 held by #576. Merge after #577.

## Testing

- t150 extended: TOML round-trip cases per path family, full-CLI-stdout parse, explicit hooks-json preservation (project path must not leak), five malformed-invocation rejection cases, byte-parity pin for the Unix default
- `bun scripts/package.ts --check` green (proves the dist trust-seed regenerates byte-identically)
- `bun run typecheck` green
- Full deterministic tier green at the integration tip unioning this with its sibling PRs (single red: t92, pre-existing on v2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.4.3] - 2026-07-17
+
+Codex hook-trust entries are now generated with a standards-compliant TOML serializer, including paths with spaces, quotes, and backslashes on Unix and Windows. **Upgrade:** re-copy `dist/codex/`, run `bun install --frozen-lockfile` in the AI-DLC source checkout, regenerate the entries, and replace the existing entries for that `hooks.json` path in `$CODEX_HOME/config.toml`; appending a second copy creates invalid TOML.
+
+* `bun scripts/package.ts codex trust --project <abs-dir> [--hooks-json <abs-path>]` preserves both native Unix and Windows paths while emitting parseable TOML.
+* Forward-slash UNC project paths such as `//server/share/project` derive a native Windows UNC default for `.codex\hooks.json` instead of collapsing the server prefix.
+* The trust command now rejects relative paths, rootless Windows paths, incomplete UNC paths, missing values, duplicate flags, and unknown arguments instead of printing unusable trust keys.
+
 ## [2.4.2] - 2026-07-16
 
 The multi-harness packager now treats each `dist/<harness>/` directory as one generated root, closing drift gaps for project-root onboarding, configuration, and removed outputs. **Upgrade:** re-copy your `dist/<harness>/` shell into the project after regenerating it with this version.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.4.2-blue)
+![version](https://img.shields.io/badge/version-2.4.3-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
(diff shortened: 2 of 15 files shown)


leandrodamascena (changes requested):
I ran a deep pass and cross-checked a second reviewer's. The core of this PR — serializing the trust TOML with smol-toml — is correct: I ran `codex trust` on Unix, Windows-drive, UNC (both spellings), spaced, and quoted paths and parsed each back; every key round-trips to a runtime-matching form, the Unix output is byte-identical to v2 (existing users' trust isn't invalidated), the 5 arg rejections fire, `--check` is green, and smol-toml is a build-only exact pin. So the serialization does its job.

But three gaps around it survived, one of which ships a broken upgrade path.

**Verdict: CHANGES REQUESTED** — one blocking.

**Blocking:**

1. **The trust-seed upgrade instruction produces invalid TOML on a re-run** (`harness/codex/emit.ts:210`). The template says *"Append the command's complete stdout to the USER config.toml."* A user pre-seeding trust after an upgrade follows that and appends again — but the entries already exist, so the result has duplicate `[hooks.state."..."]` tables. I reproduced it: appending the output twice and parsing with smol-toml gives `Invalid TOML document: trying to redefine an already defined table or value`. This is the same class of defect the PR set out to fix (bad TOML), relocated to the re-run path, and it ships to every Codex upgrader. Fix: change the guidance to replace/remove the prior `[hooks.state]` entries rather than append (and ideally make the command idempotent or emit a replace-in-place note).

**Should-fix (not blocking):**

2. **Rootless Windows paths are accepted as absolute** (`scripts/package.ts:807`). `absoluteOnEitherPlatform` is `posix.isAbsolute(p) || win32.isAbsolute(p)`, and `win32.isAbsolute("\workspace")` is `true` — so `--project "\workspace"` passes and generates `\workspace\.codex\hooks.json`. On Windows that resolves against the *current drive*, so it's not the fully-qualified identity Codex computes (`C:\workspace\...`) and the trust key can mismatch. `--hooks-json` has the same hole. The validation should require a fully-qualified path (drive-letter or UNC), not a rootless one. Narrow (an unusual input), but it's exactly the Windows-path correctness this PR is about.

3. **t150 doesn't run on a bare checkout** (`tests/unit/t150-codex-packaging.test.ts:7`). The top-level `import { parse } from "smol-toml"` throws `Cannot find package 'smol-toml'` without a prior `bun install`. smol-toml is a devDependency so any dev has it after install, but it's worth noting against the "Bun-only deterministic tests" expectation — a dynamic import inside the TOML-parse cases, or just documenting the install step, would keep a fresh clone green.

**Rebase note:** v2 now ships 2.4.1 (via #573); this PR is still 2.3.14, and the "merge after #577" ordering means it also conflicts with #577 in `harness/codex/emit.ts` + the version files. Routine re-slot on rebase.

Everything else holds — the serialization, byte-parity, path round-tripping, arg validation, and the build-only dep are all fine. Fix the append guidance and this is a clear improvement over the string-concat form.

leandrodamascena (approved):

PR #579, opened 2026-07-14
chore: make bun run check green over all authored TypeScript with pinned Biome

apackeer:
## Summary

`bun run check` previously relied on an unpinned `bunx` Biome download and did not cover all repository code intended for static checking. This makes the check deterministic and closes the identified coverage gaps:

- `@biomejs/biome` is pinned exactly at 2.4.16 as a devDependency; lint invokes the local binary, never a floating `bunx` resolution.
- Lint covers `core`, `harness`, `scripts`, `plugins`, and `tests`. The only excluded test subtree is the deliberately failing linter fixture.
- `tsconfig.tests.json` covers ordinary fixture TypeScript. It excludes only the standalone React/Vite fixture, whose dependencies are not installed at the repository root, and the fixture that must produce a compiler diagnostic.
- `tsconfig.adapters.json` discovers emitted hook adapters with a glob instead of a hardcoded harness list. A comment documents why emitted adapters are checked.
- `bun run check` starts with `scripts/package.ts --check`, so stale generated adapters cannot produce a false green type-check.
- All diagnostics were fixed with behavior-preserving edits plus five narrow, rule-specific `biome-ignore` suppressions over intentional source text. No lint rule was weakened.

No version bump: this is contributor tooling with no user-visible runtime change.

## Testing

- `bun run check` exit 0 over 479 linted files
- all three TypeScript projects pass
- adapter glob resolves the Codex, Kiro, and Kiro IDE adapters
- `bun scripts/package.ts --check` passes for every harness
- smoke + unit tier: 159 files, 2,692 assertions, zero failures


leandrodamascena (approved):
**APPROVE.** Tooling-only, and nothing I could confirm changes engine, adapter, or template behavior; no suppression hides a bug; the pin, scope extension, and `tsconfig.adapters.json` all do real work. Two accurate coverage notes (below) qualify the "all authored TypeScript" claim but don't block.

**What I checked and it holds:**

- **core/tools edits are behavior-preserving.** I truth-tabled `depStage?.enabled !== false` vs `!depStage || depStage.enabled !== false` across all shapes — identical, and it's a boolean test (the value use re-fetches `depStage`). `Object.hasOwn`, `Date.now()`, `match?.uuid`, `let st: ReturnType<...>` all equivalent.
- **Codex adapter +12** = a comment + 10 bare `break;` after `process.exit()` — unreachable, added only for Biome's switch model. Zero runtime effect.
- **compose template +34** = literal-space→`{2}`/`{4}` regex (same match set), a `??=` split, an unused-param removal, `Object.hasOwn`. Behavior-identical.
- **The 5 suppressions** are single-line, rule-specific, over intentional code (shell-expansion `${...}`, Kiro's required `then:` field, tests asserting literal `${...}` source). No forgotten-backtick trap.
- **biome.json untouched**; pin exact `2.4.16`, local binary, scope really widened to `core harness scripts plugins tests`.
- **tsconfig.adapters.json** is a real, wired, strict check — and the authored adapters are **byte-identical** to their dist copies (no token substitution), so checking the dist copy is checking the authored source character-for-character.
- `bun run check`, all three typechecks, `package.ts --check`, and the suite are green in a worktree at the PR head.

**Confirmed behavior changes / broken checks: none.**

**Non-blocking notes:**

1. **`tsconfig.adapters.json`'s `include` is a hardcoded 3-file list.** The 3 authored adapters are excluded from the main `tsconfig.json` (`harness/*/hooks/*-adapter.ts`) and only their dist twins are checked here. Today that's fine — authored and dist are byte-identical, so a broken adapter in the current 3 *is* caught (and `package.ts --check` catches a stale dist). The gap is a *new* harness: its adapter wouldn't be type-checked until someone hand-adds it to this list — the same hardcoded-list-vs-discovered class #580's matrix just fixed for harness tests. Worth deriving this include from the harness set (or at least a comment noting a new adapter must be added).
2. **"All authored TypeScript" overstates it — `tests/fixtures/**` (18 TS files) is fully excluded** from both `biome.json` and `tsconfig.tests.json`. Some fixtures must fail on purpose (they're sensor/type-check test inputs, e.g. `failing-type-check/sample.ts`), so a blanket lint would break — but the exclusion is broader than needed (it also skips legit files like `brownfield-todo/vite.config.ts`). Either narrow the exclusion to the intentionally-broken fixtures, or soften the claim to "all authored TypeScript outside test fixtures."
3. `tsconfig.adapters.json` type-checks the committed dist copies, not the authored sources, so an adapter edit needs a repackage before `bun run check` reflects it; the t145 packaging drift guard closes that coupling. A one-line comment in the config noting this would save a future contributor some confusion.

PR #580, opened 2026-07-14
test: derive harness test coverage from a manifest-discovered matrix

apackeer:
## Summary

Tests that claimed to cover "every harness" hardcoded ~13 separate Claude/Kiro/Codex arrays, so a harness could be silently skipped and Kiro CLI vs Kiro IDE coverage was inconsistent.

- New `tests/harness/harness-matrix.ts` discovers shipped harnesses from `harness/*/manifest.ts` on disk and pairs each with an explicit `HARNESS_CAPABILITIES` record (reviewer-scope registration flavor, IDE agent tools, memory-include surface, root files, onboarding mode, plugin projection - judgments a manifest alone cannot express).
- `validateManifest` cross-checks every capability field against the real manifest at load, so the metadata cannot drift from the packager's truth; a discovered manifest without a capabilities entry fails loud at import with a fix-pointing error (deliberate: a new harness must declare its capabilities once, rather than being silently part-covered). The `ideAgentTools` check requires a `tools:` grant on an `agents/*.md` frontmatter addition specifically, not just any frontmatter addition.
- The ~13 hardcoded arrays are replaced with matrix iteration across onboarding, workspace shell, root files, skills/runners, reviewer-scope registration, tier projection, and plugin projection tests; Kiro CLI and Kiro IDE are distinct rows throughout. Floor guards prevent vacuous passes if a filtered set comes back empty.
- `setupWorkspaceJourney` resolves any discovered harness's dist root through the matrix. The matrix is loaded lazily there (it walks `harness/` and `dist/` at import time; fixtures.ts is imported by sandbox suites whose trees ship neither - a top-level import throws ENOENT inside the t52 sandbox).

Test-only; no version bump.

Note for #571 (opencode): once this merges, adding `harness/opencode/manifest.ts` will make matrix-importing tests throw until an opencode `HARNESS_CAPABILITIES` entry is added - one reviewed block, after which all matrix-driven tests cover it automatically.

## Testing

- `bun run typecheck` green
- 13 matrix-consuming test files green (1459+ assertions), including t52's sandboxed suite and the t188 plugin-compose guard
- Full deterministic tier green at the integration tip unioning this with its sibling PRs (single red: t92, pre-existing on v2)

### DIFF
--- a/tests/harness/fixtures.ts
+++ b/tests/harness/fixtures.ts
@@ -583,9 +583,9 @@ export function setupIntegrationProject(
 // its own git-init'd sibling repos makes toplevel == dirname(common-dir) hold
 // for each repo, so the guard passes (the same posture setupCodexProject takes).
 //
-// Harness-parameterized so every shipped distribution can reuse ONE fixture.
-// The three live logic drivers (Claude SDK, Kiro ACP, Codex exec) select their
-// respective CLI rows; the deterministic fixture test also covers Kiro IDE.
+// Harness-parameterized so every shipped distribution can reuse ONE fixture:
+// the matrix resolves any discovered harness's dist root, so live drivers and
+// deterministic fixture tests alike pick their row without touching this file.
 // ============================================================================
 
 export type JourneyHarness = ShippedHarnessName;
--- a/tests/harness/fixtures.ts
+++ b/tests/harness/fixtures.ts
@@ -40,10 +40,7 @@ import { hostname, tmpdir } from "node:os";
 import { basename, dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { seedCustomHarness } from "./custom-harness.ts";
-import {
-  harnessByName,
-  type ShippedHarnessName,
-} from "./harness-matrix.ts";
+import type { ShippedHarnessName } from "./harness-matrix.ts";
 
 const HARNESS_DIR = dirname(fileURLToPath(import.meta.url));
 const requireHere = createRequire(import.meta.url);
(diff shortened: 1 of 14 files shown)


leandrodamascena (approved):
**Does it resolve the silent-skip / kiro-vs-ide problem? Yes.** A harness that ships a `manifest.ts` can no longer be silently skipped, and Kiro CLI and Kiro IDE are covered as separate rows.

**Verdict: APPROVE.** Test-only, no version bump. Nothing rises to the CHANGES REQUESTED bar — the suite passes, no coverage dropped versus the old hardcoded arrays, and both guards are load-bearing (I proved it by mutation).

**What I checked and it holds:**

- **Discovery guard fails loud.** `discoverManifestNames()` uses the same `readdirSync(harness/).filter(has manifest.ts)` predicate as the packager, so "discovered" equals "what ships", and the top-level guard does an exact sorted compare against `HARNESS_CAPABILITIES` keys. I added `harness/faketest/manifest.ts` with no capabilities entry and ran t217: 0 pass / 1 error at import ("metadata must exactly cover discovered manifests"). Every matrix-importing test throws — no silent skip possible.
- **validateManifest binds capabilities to the manifest, not to themselves.** It cross-checks harnessDir, skillsRoot (skipRunnerGen XOR), onboarding mode/dist, rootFiles, the 3-way memoryInclude identity, kiroAgentJson, ideAgentTools (agent `tools:` frontmatter), and plugin kind/manifestDir. I flipped kiro-ide `ideAgentTools` true→false and t217 threw "ideAgentTools does not agree with manifest agent tools grants". So metadata can't drift from packager truth.
- **kiro vs kiro-ide stay distinct** — they share `.kiro` but row identity keys off the discovered name (dist/kiro vs dist/kiro-ide) and they differ on ideAgentTools + reviewerScopeRegistration. No dedup-by-harnessDir in any consumer.
- **Filters are floor-guarded** — every `HARNESS_MATRIX.filter(...)` in t148/t156/t220/t221 carries `expect(length).toBeGreaterThan(0)`, so an empty filter can't pass green.
- **t52 sandbox stays green** — fixtures.ts lazy-loads the matrix inside setup, not at module scope, so the sandbox (which ships neither harness/ nor dist/) doesn't trip the discovery walk. t52 e2e+integration: 7 pass / 0 fail.
- **Suite passes, no lost coverage** — the 12 matrix-consuming files run 1449 pass / 0 fail. The old hardcoded arrays map 1:1 onto the filters, and several tests gained coverage (t188 committed-vs-built inventory per harness; t157 exact root-file set; t156 §7 per-agent resources).

**Confirmed problems: none blocking.**

**Non-blocking notes:**

1. `reviewerScopeRegistration` is the one top-level capability `validateManifest` never binds to the manifest — today a mislabel is caught only because it empties one of t221's four buckets and trips the floor, an accident of the current 5-harness/4-category split. A future 6th harness mislabeled so all buckets stay non-empty would slip past. Consider binding it to a distribution fact.
2. `plugin.wiringFile` (and `plugin.{kind,manifestDir}` for claude/codex, whose manifests omit the plugin block) are unbound in validateManifest; correctness rests on t188 reading the built tree. Fine as defense-in-depth, but the comment slightly overstates coverage.
3. Weak floors without an exact count pin: t220/t156/t148 use `toBeGreaterThan(0)` rather than `toBe(2)`/`toBe(1)`. Backstopped by validateManifest + the discovery guard, so it catches nothing today, but a count pin would be stronger.
4. t181/t156 `skills.length === HARNESS_MATRIX.length` is tautological (array mapped from the matrix 1:1) — harmless, the per-path existsSync loop carries the real assertion.

PR #581, opened 2026-07-14
docs: reconcile documentation with implemented surfaces and pin it with derived anti-drift tests (2.4.4)

apackeer:
## Summary

A parity sweep aligning README, the guides, generated onboarding, and `/aidlc --help` with the implemented command, harness, agent, and audit surfaces - plus tests that derive those claims from the implementation so they cannot drift again.

- Audit event count claims corrected to the 72-event / 19-category registry (the hand-maintained code-comment literal is removed entirely; the count is derived).
- The 14-agent roster (11 domain experts, two review-only agents, the composer) reconciled across the agent docs; agent tool tables now describe expected Bash/WebSearch use rather than implying per-agent access grants - every persona inherits the full session toolset, now asserted from frontmatter.
- Kiro IDE activation corrected: the `/aidlc` skill is the conductor; `settings/cli.json` and the agent-v1 JSONs are CLI-only compatibility surfaces and do not select an IDE default agent. The shipped IDE SKILL.md and question-rendering headers no longer say "Kiro CLI harness".
- The IDE reviewer step no longer instructs writing the `.aidlc-reviewer-dispatch.json` record or claims hook enforcement: no IDE reviewer-scope hook exists (the IDE PreToolUse payload omits tool arguments), so the read-scope bound is declared prose-only on that harness and the dead record ritual is dropped. A CLI-only headless caveat is removed from the IDE doc.
- Kiro CLI covered in invocation guidance; Kiro IDE added to harness enumerations; `intent`, `space`, `space-create` documented and `codekb-path` reframed as a direct utility invocation (removed from `/aidlc --help`); stale package metadata corrected.
- New anti-drift test derives event count, harness roster, agent roster, engine subcommands, and workspace verbs from implementation sources (it caught the event count moving 71 to 72 when the plugin-selection merge landed); numeric event-count claims are pattern-matched so any wrong number fails, not just the historically stale one.

Historical CHANGELOG entries untouched.

## Version

2.4.2, rebased onto `v2` at 2.4.1.

## Testing

- 168 focused assertions green across t28, t68, t111, t221, and the new t239 documentation-parity suite
- `bun scripts/package.ts --check` green (all generated harness trees match authored sources)
- `bun run typecheck` green


leandrodamascena (changes requested):
This is larger than a documentation sweep, and it isn't merge-ready yet. The reconciliation work is solid where it lands — the agent-roster, event-count, and harness-enumeration fixes check out against the implementation — but three things block merge and two are doc-vs-doc inconsistencies the sweep left in place. I ran the tiers on a clean checkout, simulated the merge, and traced the CLI/harness claims to the router source.

**1. Version collision / merge conflicts.** The PR ships `2.4.2`, but `v2` already shipped a different `2.4.2` (`86bbbc74`, the packager-contract fix) and is now at `2.4.3` — exactly the conflict-trap the Changelog Policy documents. `git merge --no-ff` against v2 tip conflicts in `CHANGELOG.md`, `README.md`, and all five `aidlc-version.ts` copies, and GitHub reports the PR `CONFLICTING` / `DIRTY`. The internal guards miss it because they only check self-consistency — `t68` and `package.ts --check` both pass in isolation here. Please rebase onto the current `v2` tip, resolve the generated-file conflicts, and use the next free version (2.4.4).

**2. The existing t174 anti-drift test fails — and the PR's own new test proves which side is wrong.** Commit 2 rewrote the "Standing rules" row in `docs/reference/14-claude-features.md:25` so the Kiro IDE cell reads `runtime rules_in_context paths`, but the paired allowlist entry in `tests/fixtures/docs-legacy-refs.json` still pins the Kiro CLI value (`resources glob`) twice — the IDE column got a copy-paste of the CLI text. `git diff 32926b86..ab5b3d72 -- tests/fixtures/docs-legacy-refs.json` is empty: the doc cell moved, the pin never followed.

Reproduction: `bun test ./tests/unit/t174-docs-legacy-refs-gate.test.ts` → 1 pass, 2 fail (3 pass on base v2). Note the PR's *new* t239 asserts the same cell must contain `rules_in_context` and not contain `resources glob`, and passes (8/8) — so this PR ships two tests that disagree on one doc line; t239 confirms the doc is right and the fixture is the side to fix.

But don't only re-pin the fixture. The doc now says the IDE loads standing rules "via `rules_in_context`," and that mechanism isn't real: `rules_in_context` is a directive field (emitted at `aidlc-orchestrate.ts:1241`) that no conductor reads on any harness — I couldn't find a single `rules_in_context` reference in any SKILL, protocol, or conductor across `core/aidlc-common/` or `harness/*/skills/`. On the IDE the rules are actually loaded through the agent-JSON resource globs, and those are pinned to `file://aidlc/spaces/default/memory/**/*.md` — fixed to `default`, so they don't follow the active space. Please either wire and test the runtime behavior or document the actual mechanism (resource globs), and fix the `spaces/default` pin if the active space is meant to be followed.

**3. `/aidlc --help` advertises `select-plugins`, but it isn't routed.** The help tail lists `select-plugins [names]` under **Utilities:**, next to `intent`/`space`/`space-create`, under a `Usage: /aidlc [command]` header — so it reads as an `/aidlc` verb. But `WORKSPACE_VERBS` (`aidlc-lib.ts:420`) is only `{space, space-create, intent}`, so `/aidlc select-plugins test-pro` falls through to workflow-description auto-detection. This is the same trap the PR correctly fixed for `codekb-path` (moved to direct-only, and t239 even asserts it's absent from the help tail) — `select-plugins` was left in it. Please mark it clearly direct-only like `codekb-path`, or add router support.

**4. Kiro IDE reviewer instructions conflict.** The IDE `SKILL.md` now says not to create `.aidlc-reviewer-dispatch.json`, while the shared `stage-protocol.md:874`/`:899` it loads still says to create and delete that record and that "the reviewer-scope PreToolUse hook reads this record." SKILL precedence resolves it at runtime, and dropping the ritual was correct (there's no IDE reviewer-scope hook — the IDE agent JSONs carry `hooks:{}`), but two contradictory instructions in the material the conductor loads needs one harness-aware source of truth.

**5. Codex onboarding names the wrong agent directory.** `harness/codex/onboarding.fills.ts:40` says the agent TOMLs are "transposed into `.agents/`," but `emit.ts` writes them to `.codex/agents/` — the PR edited this exact line in the parity sweep and left the wrong token, so the rendered `dist/codex/AGENTS.md` lists `.codex/agents/` and then says `.agents/` in the same bullet.

**On the new parity tests:** they mostly validate selected strings and structure, and they don't cover the harness-specific `onboarding.fills.ts` slot bodies — t239 derives only from the shared `core/templates/onboarding.md`, and t151 only checks for leftover `{{SLOT}}` markers. So the Codex `.agents/` error in #5 slips through the very anti-drift suite this PR adds. Worth extending the derivation to the fills bodies. (Minor, same class: `stage-table` is a live `aidlc-utility.ts` subcommand still missing from the implemented-subcommands table this PR reconciled, though its sibling `scope-table` was added.)

Please land #1–#3 first — #1 blocks the merge, #2 turns CI red on merge (and the PR's own two tests disagree on it), and #3 ships an `/aidlc` command that doesn't route.

apackeer:
Addressed the requested changes in `9da41e7`:

- rebased onto current `v2` (`d1fbfd3`) and moved the release entry to `2.4.4`
- documented Kiro IDE standing rules as Kiro resource globs and added active-space repoint coverage
- removed direct-only `select-plugins` from `/aidlc --help` and documented it beside `codekb-path`
- made the reviewer dispatch-record protocol conditional on harness enforcement support
- corrected Codex agent TOML references to `.codex/agents/` and added rendered onboarding coverage
- added the omitted `stage-table` utility documentation

Verification: focused review/regression suite 117 pass / 0 fail; `bun run typecheck`; Biome on changed TypeScript; `bun scripts/package.ts --check`; `git diff --check`.

xg-gh-25:
This is the anti-drift enforcement pattern in action — deriving documentation claims from the implementation rather than maintaining them by hand. Three observations:

1. **The "deletion test" applies to docs too** — if you can delete a doc claim and no build breaks, it's ceremony. The new t239 suite that pattern-matches event counts / agent rosters / command surfaces inverts this: now a wrong number FAILS, not just the historically stale one. This is the "Beliefs as Compiler" pattern applied to documentation: the claim becomes a mechanical assertion.

2. **The agent tool table fix** (from "per-agent access grants" to "every persona inherits the full session toolset, asserted from frontmatter") resolves a common DDD smell: when the glossary implies privilege separation that the code doesn't enforce, readers build a wrong mental model. The frontmatter-asserted truth prevents that drift.

3. **The IDE reviewer caveat** (no hook exists because PreToolUse omits arguments, so the bound is prose-only) is exactly the kind of falsifiable claim that needs a mechanical check. If a future PR adds the IDE reviewer hook, the prose saying "no hook exists" should fail. Consider: add a `check_no_ide_reviewer_hook()` test that asserts the hook file doesn't exist — then when someone does add it, the test fails and forces them to update the "prose-only" doc.

### Precedent

`mattpocock/skills` uses a similar pattern in their glossary curation: inline `_Avoid:` synonym-banning during grilling (R23) so the DDD never says "we call this X" when the code says "Y". The difference here: your anti-drift tests PREVENT the divergence at commit time, not during code review.

**One nuance**: The README still says "72-event / 19-category registry" as a number literal. If a future PR adds event 73, the test will catch the drift in the code-comment location but the README number is still hand-maintained. Recommend: have the test also grep the README for "72-event" and fail if found (or template the README from the same derived count).

Overall: this PR is the structural fix for documentation rot. The tests make the docs a compilation artifact, not a maintenance burden.

---
*Discussed via [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Topic: [T-DDD](https://github.com/xg-gh-25/SwarmAI/discussions/TBD)*

leandrodamascena (approved):

PR #595, opened 2026-07-17
V2 odyssey tested

spraja08:
# Summary

This PR enhances the adaptiveness in the composer - it estimates implementation entropy (intent ambiguity, codebase structural uncertainty, verification entropy, risk, unresolved assumptions) and composes the minimum viable workflow, It grounds the structural estimates using CodeKB MCP call-graph and component analysis if CodeKB is configured. If not, it falls back to the file system repo scanning to get the code context.

## Acknowledgment


apackeer:
Thanks @spraja08 - entropy-scored composition grounded in structural evidence is a direction we want the composer to go, and the fold discipline and scoring rubric here are a real upgrade over the keyword-era prose.

Per our conversation (maintainer fixes welcomed), I have reviewed the PR, fixed everything I found, and pushed the result onto this branch. **Heads up: this was a force-push that rewrote the branch history** (the rebase onto `v2` required it), so before doing anything else locally please run:

```
git fetch origin
git checkout v2-odyssey-tested
git reset --hard origin/v2-odyssey-tested
```

Your original work is preserved as the first commit (`caa837a6`, your authorship intact); my two commits sit on top. The PR is now MERGEABLE against `v2`.

## What the review found

1. **Packaging was broken and Codex shipped the old composer.** The `tier: judgment` frontmatter line had been removed, so `bun scripts/package.ts --check` failed, and the dist trees were hand-edited rather than regenerated - the Claude copy lost its projected `model: inherit`, and `dist/codex/.codex/agents/aidlc-composer-agent.toml` (the binding surface on Codex) still contained the previous composer, so the new behavior did not exist on that harness.
2. **Hardcoded `.kiro` paths in the harness-neutral source.** `bun .kiro/tools/...` at three places works on Kiro but errors on Claude/Codex installs; the source needs the `{{HARNESS_DIR}}` token the packager substitutes per harness.
3. **The new scoring never actually ran.** A live trace showed the composer following the engine's dispatch prompt (which still described the old proposal shape) over the persona file: the proposal it wrote carried only `{scope, projectType, stages}` and zero ARS output reached the gate. The persona file alone cannot bind the flow - the dispatch print and the conductor SKILL.md blocks are the load-bearing surfaces.
4. **The automatic reverse-engineering fold starved downstream stages.** The persona claimed downstream stages query CodeKB at runtime; in this codebase they read the local RE artifact store (`aidlc/spaces/<space>/codekb/<repo>/`), which only the reverse-engineering stage writes, and the composer's MCP evidence is never persisted. (Also a naming collision: that local store is called "codekb" in this framework, unrelated to the CodeKB MCP server.)
5. **In-flight strictness regressed.** The `--strict` validation for in-flight proposals and the "an ADD whose required producer is skipped or behind the cursor is rejected" rule had been dropped, so an invalid flip could pass the gate and only fail afterwards at the recompose verb.
6. **Release metadata** (version/CHANGELOG/README) was unbumped, and the README note described a `v2-odyssey` branch while the PR targets `v2`.

## What the pushed commits do

**`f0093f59` - the mechanical fixes.** Restores `tier: judgment` and regenerates every `dist/<harness>/` tree with the packager (fixes `--check`, the two red test files t216/t220, `model: inherit`, and the Codex TOML); swaps the three `.kiro` literals for `{{HARNESS_DIR}}`; restores the in-flight `--strict` + ADD-producer rules; demotes the RE fold from automatic to gate-proposed (the SKIP rationale must disclose that downstream stages run without the local RE store; the human decides); bumps to 2.4.7 with a CHANGELOG entry and drops the README branch note. The branch is also rebased onto `v2` (2.4.6), which is why `dist/opencode/` now ships the new composer too - the previous base predated the opencode harness.

**`40dc14ec` - making the scoring load-bearing.** Wires the new proposal contract end to end: the dispatch print in `aidlc-orchestrate.ts` and the composer blocks in all five harness SKILL.md files now demand the full shape (`ars` block with `method: codekb|fallback`, `arsRationale`, grid, rationale, validator summary, plus two pre-rendered markdown tables) and a three-block gate render (summary line, ARS score table, stage-decision table, relayed verbatim). The persona's Step 8a is reframed so the tables are part of the composer's own returned proposal. Two prose corrections: the ARS composite is labeled an advisory heuristic (uncalibrated weights; stage selection keys off the component bands, nothing deterministic routes on the scalar), and Step 5's numeric decay product is replaced with evidence-based re-scoring (name what each completed stage actually resolved) since the per-stage reduction rates were uncalibrated constants. Also: the composer knowledge file is rewritten to match the new minimum-viable philosophy, the user/reference docs are updated (including a new CodeKB section covering per-harness setup and the fact that Kiro CLI/IDE run fallback-only as shipped, since their agent configs grant no MCP tools), and t189 now asserts the composer's returned proposal names the entropy components and evidence method - so the silent regression in finding 3 cannot pass green again.

## Verification at the new head

- `bun scripts/package.ts --check` green; typecheck clean (all three tsconfigs)
- smoke+unit: 170/170 files, 4143 assertions, 0 failed
- integration (live Claude SDK): 102/102 files, 0 failed - including t189/t192 compose journeys, with t189's new assertion confirming a live composer run returned the ARS components and evidence method in its proposal

## Follow-ups (filed separately, not blockers here)

- Move the ARS arithmetic into a deterministic `ars` subcommand with weights in `tools/data/`, so the numbers are reproducible and eventually tunable against a fixture corpus. The proposal schema shipped here is designed to stay frozen through that swap.
- Live compose verification on the remaining conductors (Kiro ACP/TUI, Codex exec) to observe the three-block render per harness.

Happy to walk through any of the changes. From my side this is ready for review toward merge.

apackeer (approved):
Approving on the basis of the full review documented in the comment above: all six original findings are discharged in the two fix commits on this branch, the entropy proposal contract is wired end to end (dispatch print, all five harness conductors, persona, docs), and the gates are green at this tip - package.ts --check, typecheck, full smoke+unit (170 files), and the live SDK compose journeys (t189/t192) including the new assertion that the composer's returned proposal carries the entropy components and evidence method. Version slotted 2.5.1 above the current v2 tip, t68 sync green.

PR #596, opened 2026-07-17
docs: add stage-by-scope matrix to the scopes guide (guarded by t244)

osataken:
# Summary

The User Guide states per-scope stage *counts* ("22 of 32") and skip prose, but the full stage × scope EXECUTE/SKIP view exists only as compiled JSON (`scope-grid.json`) or the runtime `scope-table` command. A reader deciding between scopes cannot see *which* stages a scope will walk through without joining 32 stage frontmatter blocks by hand. This PR adds a Stage-by-Scope Matrix to the scopes guide chapter and pins it to the compiled grid with a drift-guard test, so it cannot silently go stale.

## Changes

- **`docs/guide/05-scopes-and-depth.md`** — new "Stage-by-Scope Matrix" section directly after the Scope Routing Table: all 9 stock scopes × 32 stages, generated from the compiled `scope-grid.json` + `stage-graph.json` (no hand-typed cells), wrapped in `<!-- BEGIN/END scope-stage-matrix -->` markers. A footnote clarifies that ✓ marks scope *membership* (CONDITIONAL stages can still skip at runtime) and that composed scopes live in `scope-grid.json` only.
- **`tests/unit/t244-scope-matrix-doc-sync.test.ts`** — drift guard following t132's two-direction discipline: columns, row identity (numbers + names from `stage-graph.json`), every cell, the collapsed initialization row, and the totals footer, cross-checked against the compiled grid forward and reverse. Since t124 already pins the grid to stage `scopes:` frontmatter, the doc table is transitively pinned to the authored source. Mechanism: none (pure in-process reads, no spawn, no LLM).
- **`docs/guide/04-phases-and-stages.md`** — cross-links to the matrix from the stage-colors note and Next Steps.
- **`docs/reference/11-contributing.md`** — "Adding a Scope" step 8 now names the matrix and its guard, so scope authors update it in the same PR.

Doc + test only — no version/CHANGELOG bump per the changelog policy.

## User experience

**Before:** to answer "which stages will I see under `mvp` vs `poc`?", a user had to read skip prose scattered across nine scope descriptions, run `bun .claude/tools/aidlc-utility.ts scope-table` against an installed dist, or open 32 stage files and join their `scopes:` frontmatter by hand.

**After:** the scopes guide answers it at a glance — one table showing exactly which of the 32 stages execute under each stock scope, with per-scope totals, sitting next to the routing table where scope decisions are made. If a future PR retags a stage or adds a scope without regenerating the table, CI fails with the exact stage × scope cell that drifted.

## Checklist

* [x] Changes are documented

## Test Plan

- `bun test tests/unit/t244-scope-matrix-doc-sync.test.ts` — 8 tests, 402 assertions, all pass.
- Fault-injection verified: flipping a single ✓ cell in the doc table makes the guard fail (cell + totals checks), restoring it goes green again.
- `bun run check` — dist parity (`package.ts --check`), typecheck (all three tsconfigs), and Biome over the repo, all green.
- Coverage registry (`bun tests/gen-coverage-registry.ts`) unchanged.
- To verify the table content independently: compare any row against `dist/claude/.claude/tools/data/scope-grid.json`, or run `bun .claude/tools/aidlc-utility.ts scope-table` in an installed dist and check the per-scope totals match.

## Acknowledgment


apackeer:
Thanks for this - the matrix itself is genuinely useful, and the approach is right: generated cells pinned to the compiled grid, with the transitive chain through t124 back to the authored `scopes:` frontmatter meaning no re-parse is needed here. I verified the table content against current `v2`: `scope-grid.json` is byte-identical since this branch's base, and running the guard against today's compiled data passes 8/8 (402 assertions). So the content is accurate; the items below are about the merge state and gaps in the guard.

**1. [P1] Rebase needed - merge conflict with current v2**

GitHub reports the PR as CONFLICTING. The single conflict is in the Next Steps block of `docs/guide/04-phases-and-stages.md`: this branch is based before #581, which rewrote the adjacent agents bullet to "the 14-agent roster and its domain, review, and composition roles". Resolution: keep v2's 14-agent wording verbatim and add only the matrix link to the scopes bullet. The other two doc files auto-merge cleanly, and since the compiled grid hasn't changed, no table regeneration is needed.

**2. [P1] t238 is already taken on v2 - renumber to t244**

`tests/unit/t238-build-binaries.test.ts` landed on v2 via #560, and the suite is now up to t243. This PR's `t238-scope-matrix-doc-sync.test.ts` collides. Please renumber to **t244** on rebase - the filename, the reference in `docs/reference/11-contributing.md` step 8, and the BEGIN marker comment in `docs/guide/05-scopes-and-depth.md` (both name the test path).

**3. [P2] Row ordering is not guarded**

Test 4 sorts both sides before comparing (`t238-scope-matrix-doc-sync.test.ts:161`), and test 3 looks rows up via a `Map` keyed by stage number, which is order-blind. I reproduced this: swapping the 1.1 and 1.2 rows in the doc table still passes all 402 assertions - a fully scrambled table would go green. Fix by comparing document order directly:

```ts
expect(doc.rowNumbers).toEqual(bodyStages.map((s) => s.number));
```

`stage-graph.json` already lists body stages in document order (1.1 through 4.7), so this is a drop-in replacement for the sorted comparison in test 4.

**4. [P2] Initialization numbering is hard-coded**

`INIT_ROW_LABEL` (line 45) hard-codes the "0.1-0.3" label, and nothing compares it against `initStages.map((s) => s.number)` - line 135 pins only the count. Renumbering the three initialization stages without changing their count would leave the doc's row label stale and green. Suggested fix: derive the expected label from the graph, e.g.

```ts
const nums = initStages.map((s) => s.number);
expect(INIT_ROW_LABEL).toBe(nums[0] + "\u2013" + nums[nums.length - 1]);
```

(using the same `\u2013` escape the label constant already uses; alternatively assert each `initStages[i].number` is `0.${i + 1}` if the collapsed-range rendering should stay fixed).

With the rebase, the t244 renumber, and the two guard tightenings, this looks good to land.

osataken:
Thanks for the comment, I already rebased the and adjust the test number to 244 with recommended tighten guards.

leandrodamascena:
**Review Recommendation: Request Changes** — verified both P2s against the code; they're real. (This supersedes my earlier "looks good" pass, which only re-checked the four prior items and didn't scrutinize the adjacent legend or the parser deeply enough.)

**1. [P2] Incorrect ALWAYS definition — `docs/guide/04-phases-and-stages.md:135`** ✅ confirmed

The line this PR touched reads *"Green = ALWAYS (runs for every scope)"*, but the authoritative key at `docs/reference/03-orchestrator.md:709` says *"ALWAYS: Executes for all scopes that **include this stage**"*. The PR's own new matrix proves the "every scope" wording wrong: row `1.4 Scope Definition` (labeled ALWAYS) shows `✓ ✓ ✓` then blanks — skipped by 6 of the 9 scopes. `git blame` puts this line in commit `7452c7b1` of this PR, so it's in scope. Suggest matching the orchestrator wording, e.g. *"Green = ALWAYS (runs whenever the selected scope includes it). Yellow = CONDITIONAL (may skip based on scope, project type, or plan)."*

**2. [P2] Duplicate special rows bypass the guard — `tests/unit/t244-scope-matrix-doc-sync.test.ts:114`** ✅ confirmed

`parseDocTable` assigns `initCells` and `totals` as single variables and stores body rows in a `Map` keyed by stage number, so a duplicate initialization or totals row silently overwrites the prior value (last-wins) and the suite still passes — despite the parser promising duplicate detection. A duplicated special row would drift green. Fix by throwing on a second assignment (`if (initCells) throw …` / `if (totals) throw …`) and detecting a duplicate key before `rows.set`.

**Non-blocking:**
- PR **title** ("guarded by t238") and body still reference `t238-scope-matrix-doc-sync`; the implementation is now `t244`. Worth updating so the PR record matches.
- Biome: `t244…test.ts:138` (`initNums[0] + "–" + initNums[…]`) — prefer a template literal.

Both P2s are genuinely low-impact (the first is a legend sentence with no behavioral effect; the second needs an unlikely authoring error to bite), so P2 is the right severity — not blockers-of-principle, but the fixes are small and the ALWAYS contradiction sitting next to a table built for accuracy is worth correcting before merge. The matrix, the t124 transitive pin, and the four prior items all check out; with these two tightenings it's good to land.

leandrodamascena (changes requested):
Changes requested

osataken:
thanks for the comments, I made changes and updated the pull request.

leandrodamascena (changes requested):
The previous findings are fixed, but two gaps remain in the drift guard:

1. **[P2] Validate the Markdown table separator** (`tests/unit/t244-scope-matrix-doc-sync.test.ts:95`)

`parseDocTable()` assumes `lines[1]` is the Markdown delimiter and skips it without validation. I replaced it with:

```markdown
| this is not a Markdown table separator |
```

All 8 tests still passed, although the matrix no longer renders as a Markdown table. Please validate the delimiter shape and column count.

2. **[P2] Validate special-row labels and numeric syntax** (`tests/unit/t244-scope-matrix-doc-sync.test.ts:112`)

The guard validates the initialization cells but ignores its stage label. It also identifies the footer with `label.includes("Total")` and parses totals with `parseInt()`. Consequently, `Fabricated stage label`, `Not Total stages`, and `32 corrupt` all pass.

Please require the exact initialization label, exact `Total stages` label, and strictly numeric total cells.

Verification: smoke and all 164 unit files passed, including 3,208 assertions. Packaging and typecheck passed. The integration `t188` parallel timeout passed in isolation (60 tests); live integration was unavailable because the AWS credentials are expired.

osataken:
thanks for the comments 

**Addressed both P2s:**

- Separator validation — parseDocTable() now validates lines[1] is a well-formed Markdown table delimiter (each cell matches :?-{3,}:?) with the correct column count. Replacing it with arbitrary text will now throw.
- Special-row labels and numeric syntax — initialization row requires the exact label "Initialization (all 3 stages)" (derived from graph count); footer requires exact "**Total stages**" (strict equality, not substring); total cells must be strictly numeric (/^\d+$/ after stripping bold).

**Additional hardening on top:**

- Header labels validated (# and Stage) — renaming or reordering fixed columns throws at parse time
- Cell content constrained to only ✓ or empty — typos or wrong Unicode (e.g. ✔) throw immediately
- Every body row validated for correct column count — a missing or extra pipe no longer silently misaligns cells
- INIT_STAGE_LABEL derived from initStages.length so adding a 4th init stage auto-updates the expected label

All 8 tests pass (403 assertions).

leandrodamascena (changes requested):
The previous findings are fixed, but two issues remain:

1. **[P2] Correct the runtime semantics below the matrix** (`docs/guide/05-scopes-and-depth.md:153`)

The guide says users can “skip any stage at its approval gate.” The runtime does not support this: approval gates offer Approve/Request Changes, and `report --result skipped` is accepted only for an active `CONDITIONAL` stage. Please say that the matrix represents static scope membership, conditional stages may skip when their condition does not hold, and pending stages can be reshaped through an approved composer proposal.

2. **[P2] Ensure the matrix remains a rendered Markdown table** (`tests/unit/t244-scope-matrix-doc-sync.test.ts:76-116`)

Both of these mutations still pass all 8 tests:

- Insert a blank line between the header and separator. The table no longer renders correctly.
- Remove the closing `-->` from the BEGIN marker. The entire matrix becomes an HTML comment.

`parseDocTable()` discards non-pipe lines before validation, while the BEGIN marker is matched only by prefix. Please validate complete marker lines and one contiguous table block.

Non-blocking: the “Adding a Stage” instructions should mention the matrix and `t244`, since that table is now automatically validated.

Verification: both fault injections passed with 8 tests and 403 assertions on head `c5b2e04f`.

apackeer:
+1 to both of leandrodamascena's findings, independently confirmed at head c5b2e04f. Three additions:

- The same "skip individual stages at their approval gates" claim also appears later in the chapter (the "When in doubt, start with `feature`" line). It is pre-existing rather than introduced here, but since finding 1 rewrites the identical wording, worth fixing in the same pass.
- The contiguity fix should cover mid-table interruptions too, not just the header/separator seam: a blank line inserted after row 2.3 also leaves all 8 tests green (reproduced, 403 assertions). Requiring one contiguous pipe-line block from the header through the totals row covers both.
- Nit: the PR body still references `t238-scope-matrix-doc-sync` twice; the title was updated but the body wasn't.

For what it's worth, everything from the earlier rounds checks out at this head: I mutation-tested the hardened parser with 11 content corruptions (labels, cells, separators, duplicates, ordering, header renames, wrong-Unicode checkmarks) and it caught all 11, and smoke+unit (177 files, 4367 assertions), typecheck, and lint are green. With this round addressed the guard and the doc look ready.

osataken:
Thanks for the comments 

All findings addressed in 71cd1811:

**[P2] Runtime semantics — fixed both occurrences:**
- Matrix footnote: rewritten to say the matrix represents static scope membership, CONDITIONAL stages self-skip when their condition doesn't hold, and pending stages can be reshaped through an approved composer proposal. No mention of "skip at approval gate."
- "When in doubt, start with feature": same correction — now reads "CONDITIONAL stages will self-skip when their conditions do not apply to your project."

**[P2] Rendered Markdown table — both mutations now caught:**
- extractRegion() validates the BEGIN marker line ends with --> — removing the closing comment tag now throws.
- parseDocTable() requires all pipe-lines form one contiguous block (consecutive line indices). A blank line inserted anywhere — header/separator seam, mid-table after row 2.3, or elsewhere — now throws "table is not contiguous".

**Non-blocking:**
- "Adding a Stage" step 5 now references the matrix and t244.
- PR description updated to reference t244 (no more t238 mentions).

Rebased on latest v2 (257b43a3). All 8 tests pass, 403 assertions.

leandrodamascena (approved):
Thanks for working through the review rounds and addressing the findings. I re-checked head `71cd1811`, and the previous issues are resolved:

- Runtime skip and composition semantics are now described accurately.
- Table markers, continuity, separators, labels, duplicates, row ordering, cells, and totals are guarded.
- The matrix matches the compiled scope grid and stage graph.
- `t244` passes all 8 tests with 403 assertions.
- Package parity and CI checks are green.
- The branch merges cleanly with the current `v2`.

The matrix is useful, accurate, and does not change runtime behavior or methodology. I found no remaining issue that should block the PR. Approving.

PR #600, opened 2026-07-19
feat(kiro-ide): add v1 JSON hooks for IDE 1.0+ backward compat

osataken:
# Summary

Add v1 JSON hook format for Kiro IDE 1.0+ while preserving backward compatibility with pre-1.0 IDE installations. Ships new `.json` hook files alongside existing `.kiro.hook` files so both IDE generations work from the same distribution. We can remove the pre-v1 hooks once the majority of Kiro-IDE users upgrade to v1.0+

## Changes

- Add 9 new v1-format `.json` hook files in `harness/kiro-ide/hooks/` using the updated schema (`"version": "v1"`, PascalCase triggers, `matcher` regex, `action` object wrapper)
- Update `harness/kiro-ide/manifest.ts` `harnessFiles` to include the new `.json` entries for dist projection
- Regenerate `dist/kiro-ide/.kiro/hooks/` — both `.kiro.hook` (legacy) and `.json` (v1) files now ship side by side

**Trigger mapping (old → new):**

| Old `when.type` | New `trigger` |
|-----------------|---------------|
| `postToolUse` | `PostToolUse` |
| `preToolUse` | `PreToolUse` |
| `promptSubmit` | `UserPromptSubmit` |
| `agentStop` | `Stop` |

**Tool type mapping (old `toolTypes` → new `matcher`):**

| Old `toolTypes` | New `matcher` regex |
|-----------------|-------------------|
| `["write"]` | `^(fs_write\|fs_append\|str_replace\|delete_file\|code)$` |
| `["shell"]` | `^execute_bash$` |
| `[".*invoke_sub_agent.*"]` | `^invoke_sub_agent$` |
| (none/all) | (omitted) |

## User experience

**Before:** Only `.kiro.hook` files ship in `dist/kiro-ide/.kiro/hooks/`. Pre-1.0 IDE loads them. Kiro IDE 1.0+ (which expects the new v1 JSON format) cannot discover or register these hooks.

**After:** Both formats ship side by side. Pre-1.0 IDE continues to read `.kiro.hook` files (ignores `.json`). Kiro IDE 1.0+ reads the new `.json` files (ignores `.kiro.hook`). No double-firing — each IDE version recognizes only its own extension. Users copying `dist/kiro-ide/` get a distribution that works on both IDE generations without any action.

## Checklist

* [ ] Changes are documented

## Test Plan

1. Run `bun scripts/package.ts kiro-ide` — should regenerate without errors
2. Run `bun scripts/package.ts kiro-ide --check` — should report `OK` (no drift)
3. Verify `dist/kiro-ide/.kiro/hooks/` contains both `.kiro.hook` and `.json` files for all 9 hooks
4. Validate each `.json` file parses as valid JSON matching the v1 schema (`version: "v1"`, `hooks` array with `name`, `trigger`, `action`)
5. On a pre-1.0 Kiro IDE install: copy `dist/kiro-ide/` into a project, confirm hooks register and fire from the `.kiro.hook` files
6. On a Kiro IDE 1.0+ install: copy `dist/kiro-ide/` into a project, confirm hooks register and fire from the `.json` files

## Acknowledgment

### DIFF
--- a/dist/kiro-ide/.kiro/hooks/aidlc-audit-logger.json
+++ b/dist/kiro-ide/.kiro/hooks/aidlc-audit-logger.json
@@ -1,0 +1,16 @@
+{
+  "version": "v1",
+  "hooks": [
+    {
+      "name": "aidlc-audit-logger",
+      "description": "Logs artifact create/update events on file writes to aidlc-docs/.",
+      "trigger": "PostToolUse",
+      "matcher": "^(fs_write|fs_append|str_replace|delete_file|code)$",
+      "action": {
+        "type": "command",
+        "command": "bun .kiro/hooks/aidlc-kiro-adapter.ts audit-and-sensors"
+      },
+      "enabled": true
+    }
+  ]
+}
--- a/dist/kiro-ide/.kiro/hooks/aidlc-block.json
+++ b/dist/kiro-ide/.kiro/hooks/aidlc-block.json
@@ -1,0 +1,15 @@
+{
+  "version": "v1",
+  "hooks": [
+    {
+      "name": "aidlc-block",
+      "description": "Hard-blocks tool calls while an approval gate is open and no human acted since (human-presence floor).",
+      "trigger": "PreToolUse",
+      "action": {
+        "type": "command",
+        "command": "bun .kiro/hooks/aidlc-kiro-adapter.ts block"
+      },
+      "enabled": true
+    }
+  ]
+}
--- a/dist/kiro-ide/.kiro/hooks/aidlc-log-subagent.json
+++ b/dist/kiro-ide/.kiro/hooks/aidlc-log-subagent.json
@@ -1,0 +1,16 @@
+{
+  "version": "v1",
+  "hooks": [
+    {
+      "name": "aidlc-log-subagent",
+      "description": "Logs sub-agent completion to audit trail.",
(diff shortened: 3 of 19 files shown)


apackeer:
Thanks - this PR called the right approach. We've confirmed it empirically on live builds: 0.12.333 completely ignores v2 `.json` hook files (not listed, not fired, no error) while still running `.kiro.hook`, and 1.0.116 with both formats present runs only the v2 files (legacy shown as struck-through migration candidates, no double-firing). So ship-both-formats is safe with no dedup logic.

We're asking #614 to adopt this coexistence shape, since it also carries the docs, CHANGELOG, version bump, and test updates that a shipped change needs - closing this one as superseded by #614 rather than merging both. The side-by-side idea and the format mapping here fed directly into that outcome; if you'd like, a Co-authored-by line on #614 would be well earned.

leandrodamascena:
Thank you for the time and care you put into this contribution. The coexistence approach was valuable and directly informed #614, which now carries it alongside the required documentation, version/CHANGELOG updates, compatibility notes, and tests.

The remaining Kiro IDE 1.x stdin payload handling is tracked in #615. Since your work has been incorporated there, we recommend closing this PR without merging. We genuinely appreciate your contribution.

Issue #582, opened 2026-07-15
AIDLC Kiro harness will break under Kiro CLI v3 (KAS) — agent-schema, hooks, settings, and tool-layer contract all change


fanhongy:
## Context

Kiro CLI is moving v2 → v3 (v3 = "Kiro Agent Server" / KAS, a TypeScript ACP rewrite; currently early-access). AIDLC's Kiro harness (`harness/kiro/`, `harness/kiro-ide/`, and the generated `dist/kiro*`) is authored entirely against the **v2 agent/hook/settings contract**. v3 does not just relocate config — it re-interprets and in places silently drops it. This tracks the breakage so we can gate AIDLC support on v3 explicitly rather than shipping a config that silently degrades.

**Source:** internal engineering review "Kiro CLI v2 → v3 (KAS)" (18 Jun 2026, 40-doc deep review). Findings below are that review's §4/§9 (config re-interpretation) and §14 (execute_bash server guard) cross-referenced against AIDLC's shipped artifacts.

**Caveat:** v3 is early-access and not installed in this environment, so the impacts are **inferred from the review cross-referenced against the shipped harness**, not live-verified. A v3 early-access smoke test (see Next steps) is required to confirm severity per item.

## Confirmed dependencies in the AIDLC Kiro harness

- Agent JSON pinned to the v2 schema (`agent-v1.json`) with `tools`, `allowedTools`, and `toolsSettings.execute_bash.allowedCommands` — e.g. `harness/kiro/agents/aidlc-developer-agent.json` confines the developer agent to `bun \.kiro/tools/.*`.
- `.kiro.hook` files using triggers `postToolUse` / `preToolUse` / `promptSubmit` / `agentStop` and `toolTypes` `[write, shell, spec, .*invoke_sub_agent.*]` (audit-logger, stop, block, mint, log-subagent, runtime-compile, sync-statusline).
- `harness/kiro/settings/cli.json` — `chat.defaultAgent: aidlc`, `chat.modelDefaults`.
- The entire tool layer runs as `bun .kiro/tools/*` via `execute_bash`; tool names referenced are `fs_read` / `fs_write` / `execute_bash` / `invoke_sub_agent` / `thinking`.

### P0 — Per-agent permission model silently stops applying
[...]

Issue #555, opened 2026-07-11
[Bug]: The Kiro IDE harness (`harness/kiro-ide/`) ships CLI-format surfaces — JSON agents, `settings/cli.json`, and legacy `.kiro.hook` files — that Kiro IDE 1.0 does not read (agents/settings) or flags for migration (hooks)


wowzoo:
## Summary

`dist/kiro-ide/` is nearly a byte-for-byte copy of `dist/kiro/` (the CLI harness): it ships JSON agent configs, a `settings/cli.json`, and `.kiro.hook` files. But Kiro IDE 1.0 reads a different agent format (Markdown), activates agents differently (selector, not a settings key), and treats the `.kiro.hook` schema as legacy (migration-prompted). So an IDE install carries surfaces the IDE either ignores or warns about.

I maintain a downstream IDE-format variant and have confirmed each point below against the official IDE docs and a live 1.0.116 run. Filing this to align on scope before a PR.

## 1. Agents — IDE reads Markdown, not JSON (mismatch, factual)

The IDE custom-agent format is a Markdown file with YAML frontmatter (`tools:` tags, `permissions:`/`mcpServers:` inline); the JSON agent config is the Kiro **CLI** format. `dist/kiro-ide/.kiro/agents/` ships both `.md` and `.json`, plus the conductor as `aidlc.json`:

```
aidlc.json                            <- CLI conductor config; IDE does not read it
aidlc-architect-agent.json            <- CLI; the IDE reads the .md sibling
aidlc-composer-agent.json
aidlc-developer-agent.json
aidlc-product-lead-agent.json
aidlc-architecture-reviewer-agent.json
```

On the IDE these `.json` files are dead weight — the IDE resolves the conductor and every delegated subagent from the `.md` frontmatter. The IDE-correct shape is: conductor as `agents/aidlc.md`, and no agent `.json` at all.

### Code that must change when the conductor becomes `aidlc.md`

Two core checks assume the conductor is `aidlc.json`, so moving to `aidlc.md` requires an additive fix (both are otherwise byte-identical for a CLI tree):

[...]

Issue #543, opened 2026-07-09
[Bug]: 2.2.17's Kiro IDE adapter reads USER_PROMPT, empty on IDE 1.0.89 — audit/sensor (#417) and log-subagent (#459) regress


wowzoo:
**Apologies for the churn first.** My earlier reports (#417/#459, filed against 0.12.333) framed the payload channel as `USER_PROMPT`, and PR #471 (2.2.17) built the fix against exactly that — the Kiro IDE adapter reads `process.env.USER_PROMPT`. On the current IDE build (1.0.89) the channel is the opposite: the payload arrives on **stdin**, and `USER_PROMPT` is empty. So the shipped adapter reads `""`, and the audit/sensor and log-subagent hooks go silent again. The adapter is byte-identical from 2.2.17 through 2.2.19, so this is the state of the latest release.

I re-verified live on **2.2.19 / IDE 1.0.89** (probe hook + a full workflow run, `AIDLC_HOOK_DEBUG` on).

### Shared root cause

- **Channel = stdin JSON; `USER_PROMPT` is empty.** The stdin JSON channel is the IDE 1.x hook contract (1.0.52 onward per the Kiro docs); I tested on 1.0.89. The adapter's own `hookDebug` logs `hasUserPrompt=false` on every PostToolUse dispatch, so the shipped `process.env.USER_PROMPT` read (`aidlc-kiro-adapter.ts:73`) leaves `ide = {}` and every branch degrades to a no-op.
- **Field names are snake_case** `tool_name` / `tool_input` / `tool_response`.
- **`tool_input` is `{}`** on every event; the write path / agent identity live only in `tool_response` prose (`Created the <PATH> file.` / `Replaced text in <PATH>`), which `extractWrittenPath` (`:161-169`) already matches verbatim. So only the acquisition step reads the wrong channel.

### Fix 1 — acquisition (fixes #417 audit/sensor, and unblocks #459)

Read stdin first, fall back to `USER_PROMPT`, map onto the `{tool_name, tool_input, tool_response}` shape the downstream already parses:

```ts
// aidlc-kiro-adapter.ts — replace the USER_PROMPT-only acquisition block
let ide: IdeHookContext = {};
{
  // 1) stdin first (IDE 1.x). The IDE may open stdin but not close it,
  //    so race the read against a short timeout to avoid hanging.
  let stdinText = "";
  if (!process.stdin.isTTY) {
    try {
      stdinText = await Promise.race([
[...]

PR #603, opened 2026-07-20
fix(kiro-ide): migrate hooks to v2 schema + stdin context channel

csotomon:
## Summary

Kiro IDE >= 1.0.1xx no longer executes the legacy `.kiro.hook` format this harness ships, leaving `dist/kiro-ide/` installs with **no hooks layer at all**: no `HUMAN_TURN` minting (the human-presence floor), no `SESSION_STARTED`, no `ARTIFACT_CREATED/UPDATED` audit rows, no sensors, no context injection, no Stop enforcement. The workflow still advances (the engine writes the ledger via CLI), but every hook reinforcement is silently gone.

Verified live on Kiro IDE 1.0.165 (ACP client):

- A debug marker plus the adapter's unconditional entry logging showed the adapter was **never invoked** for any of the 9 legacy hooks (one step worse than the 1.0.89 era of #543, where the adapter WAS invoked with an empty `USER_PROMPT`).
- A ledger census across 5 sessions of a full intent: zero hook-origin events.
- A probe hook in the v2 schema (`{"version":"v1","hooks":[...]}`, PascalCase triggers) fired **in the same session, hot-loaded, without a restart**.
- The probe captured the 1.x context channel verbatim — **stdin**, snake_case fields, `tool_input` always `{}`, `USER_PROMPT` empty — confirming #543 unchanged on 1.0.165:

```json
{"session_id":"sess_…","hook_event_name":"PostToolUse","cwd":"/path/to/project","tool_name":"execute_bash","tool_input":{},"tool_response":"Output:\n…\nExit Code: 0"}
```

## Changes

**Part 1 — the nine kiro-ide hooks migrate 1:1 to the v2 schema** (`harness/kiro-ide/hooks/aidlc-*.json`; the legacy `.kiro.hook` files are removed, as #555 proposes). Same adapter targets, same semantics:

| Hook | Legacy trigger | v2 trigger | matcher |
|---|---|---|---|
| aidlc-mint | promptSubmit | UserPromptSubmit | — |
| aidlc-session-start | promptSubmit | UserPromptSubmit | — |
| aidlc-block | preToolUse | PreToolUse | — |
| aidlc-audit-logger | postToolUse [write] | PostToolUse | `fs_write\|str_replace\|fs_append` |
| aidlc-runtime-compile | postToolUse [shell] | PostToolUse | `execute_bash` |
| aidlc-sync-statusline | postToolUse [shell] | PostToolUse | `execute_bash` |
| aidlc-log-subagent | postToolUse [.*invoke_sub_agent.*] | PostToolUse | `invoke_sub_agent` |
| aidlc-stop | agentStop | Stop | — |
| aidlc-session-end | agentStop | Stop | — |

**Part 2 — stdin context acquisition in `aidlc-kiro-adapter.ts`** (the fix proposed in #543, with per-target gating as an improvement): the adapter reads the hook context from stdin (the 1.x channel), raced against a 2s timeout (the 0.12 IDE opened stdin without ever writing or closing it — a bare read hangs), falling back to `USER_PROMPT` so 0.12 keeps working. Both field spellings are accepted (0.12 camelCase / 1.x snake_case; 1.x carries no success flag, so only an explicit `toolSuccess: false` drops a write per #417's guard). The read is **gated to the two payload-dependent targets** (`audit-and-sensors`, `log-subagent`): every other target — including `block`, which now fires on EVERY PreToolUse — keeps its zero-latency, never-touches-stdin path.

Also: docs updated (`docs/guide/harnesses/kiro-ide.md`; `docs/reference/kiro-ide-hook-payload.md` now documents both context channels), version bump to 2.5.1 (re-bumped from 2.4.7 after upstream #568 cut 2.5.0, per the repo's changelog conflict policy) with a CHANGELOG entry (including the upgrade note to delete stale `.kiro.hook` files from existing installs), dist regenerated via `bun scripts/package.ts`.

## What was tested

Live on Kiro IDE 1.0.165 (same session, no restart), with this change applied to the installed tree:

- `block` firing on every real PreToolUse; `audit-and-sensors` extracting a real edit's path from `tool_response` via stdin with the record-root path gate passing; `runtime-compile`/`state-sync` self-gating as designed (`ide-audit-sync`).
- The next user prompt wrote real `SESSION_STARTED` and `HUMAN_TURN` rows to the ledger; `/aidlc --doctor` 34/34 (the hooks heartbeat went from "not yet fired" to populated).

In this repo:

- `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`: green.
- Smoke + unit tiers fully green (170 files / 4150 assertions), including 7 new t218 cases for the 1.x stdin channel: snake_case parsing, stdin-over-`USER_PROMPT` precedence, camelCase-on-stdin parity, and no-hang guards with a held-open stdin (a payload-independent target exits fast and never reads; a payload target resolves via the race and falls back to the env channel).
- Integration tier green except two environmental cases (t19 requires live AWS credentials; t188 passes serially and fails only under `--parallel 4`).

## Deliberately out of scope

- **The Kiro CLI tree (`dist/kiro/`)** is untouched: its hooks ride the agent-JSON `hooks` block. Empirical note: kiro-cli 2.13.0 fired neither `.kiro.hook` nor v2 `.json` files in a clean non-interactive probe (no aidlc agent active; non-interactive mode may suppress hooks) — needs re-validation in an interactive session before touching that tree.
- **The plugin-compose wiring** still emits the legacy `.kiro.hook` (a separate surface; its Kiro host-support caveat is already documented in the plugin chapters).
- **A real `SessionStart` trigger**: the v2 vocabulary has one, but this PR keeps session-start wired to `UserPromptSubmit` (per-prompt `SESSION_STARTED`, exactly as before) to stay a pure translation. Switching to the real trigger is proposed as a follow-up.

## References

- Addresses #555 (legacy hook format inert on IDE >= 1.0.1xx; live evidence: https://github.com/awslabs/aidlc-workflows/issues/555#issuecomment-5017268046)
- #543 (the 1.x stdin context channel; this ports its acquisition fix, adding the per-target gating)
- #582 (context for the hook vocabulary change)


csotomon:
Split into two PRs as requested: - **#614** — the bug fix: the nine hooks migrate 1:1 to the IDE's v2 hook schema (adapter untouched; the v2 JSONs invoke the exact same commands the legacy files did). Restores hook firing on IDE 1.x; 7 of 9 targets fully functional, the two payload-dependent ones fire but no-op safely until part 2. - **#615** (draft, stacked on #614) — the adapter enhancement: the 1.x stdin context channel (#543's acquisition fix with per-target gating), restoring payloads for `audit-and-sensors` and `log-subagent`. Together they reconstruct this PR's end state byte-for-byte (verified by diff, modulo the split CHANGELOG entries). Closing this one in favor of the pair.

PR #612, opened 2026-07-20
feat: step the balanced tier down to effort medium (2.5.13)

jstrunk:
Closes #607.

## What

Steps the `balanced` tier — the two review-only agents (`aidlc-architecture-reviewer-agent`, `aidlc-product-lead-agent`) — down to **medium reasoning effort on every harness**, instead of inheriting the session's effort. Models are unchanged (Sonnet-class everywhere).

| Harness | Before | After |
|---|---|---|
| Claude Code | `model: sonnet`, no `effort:` line (session effort) | `model: sonnet` + `effort: medium` |
| Codex | `model = "openai.gpt-5.4"`, no effort key | + `model_reasoning_effort = "medium"` |
| opencode | model pin, no `variant:` | + `variant: medium` |
| Kiro CLI/IDE | `chat.modelDefaults` sonnet-4.5 → `high` | sonnet-4.5 → `medium` (the balanced/templated collapse becomes a no-op — both tiers now pin medium) |

## Why

Field data from a 12-unit intent (2.2.17, Claude Code, xhigh session): a restarted nfr-design review spent ~36 of 39 minutes in time-to-first-token at ~140k context, and 11/12 nfr-requirements reviews returned READY — session-level effort was buying latency, not verdicts, on work the tier itself defines as "novel input judged against explicit criteria". Since 2.3.5 the review contract is checklist- and evidence-grounded, exactly the shape that does not benefit from maximum-effort deliberation. Full rationale and data in #607.

The `judgment` tier is untouched — the "never silently downgrade the session" contract still holds where output cascades downstream.

**Why keep `balanced` as a tier at all, when it now projects identically to `templated`?** The tier names classify the WORK, not the dial (`aidlc-tiers.ts` states this as a design rule): `templated` is pattern-following output whose methodology already lives in knowledge (delivery plans, CI/CD YAML, runbooks), while `balanced` is reviewer-shaped work — novel input judged against explicit criteria. Re-tiering the reviewers to `templated` would make the dials coincide by making the vocabulary lie, and a future contributor adding a review-only agent should be able to write `tier: balanced` without knowing today's model lineup. Two tiers coinciding in projection is already a designed-for state — balanced and templated have collapsed onto the same model on Kiro since 2.3.1 (the documented collapse rule); after this PR they coincide on every harness. Keeping the classification correct preserves the tuning seam: if a later lineup wants reviewer work on a different dial (a mid-size model at high effort, say), it's a one-line table edit rather than a re-tiering plus roster-doc rewrite.

[...]

PR #613, opened 2026-07-20
feat: reviewer turn backstop (maxTurns) + section-12a incomplete-attempt guard (2.6.8)

### LINKED ISSUE #609: [Feature]: reviewer turn backstop (maxTurns) + §12a missing-verdict guard — bound the reviewer's tool-call count and close the verdict-less-review branch
### Description

Nothing bounds how many tool calls a §12a reviewer makes. The read-scope bound (#534 → PR #538 + #539's PreToolUse enforcement) limits WHERE the reviewer reads, but not HOW MUCH it does inside that scope: field reviews ran **56–113 turns** each under an xhigh session. `maxTurns` — Claude Code's documented subagent frontmatter field ("maximum number of agentic turns before the subagent stops") — is the harness-native lever for that, but shipping it naively is dangerous, and the protocol has a latent gap it would expose:

**Probe-verified cutoff semantics** (maxTurns: 2 probe agent, 5-command task): the subagent is stopped mid-task with **no warning and no final-message turn** — the parent receives `(Subagent completed but returned no output)`. Since the reviewer writes its `## Review` section as one Edit at the END of investigation, an uninstructed cap silently produces verdict-less reviews. And §12a step 3 branches only on READY / NOT-READY — **a reviewer that returns without writing `## Review` is an undefined branch today**, reachable by any reviewer death (turn cap, crash, context exhaustion), independent of this proposal.

Three coordinated parts:

1. **`maxTurns: 40`** in the two balanced reviewer agents' authored frontmatter (`core/agents/aidlc-architecture-reviewer-agent.md`, `aidlc-product-lead-agent.md`). A hard backstop, not a target — recent field reviews ran 56–113 turns under xhigh; at the balanced tier's effort a review needs fewer. The packager's `projectTierFrontmatter` rewrites only the `tier:` line, so the key ships verbatim to the Claude dist (binding) and inertly in the kiro/opencode .md copies; it must NOT be added to Kiro agent JSON (fail-closed schema) and is not read by `emit.ts` for Codex TOMLs — on those harnesses part 2 is the bound.
2. **`## Turn Budget` persona section** in both reviewers: state the cap and its cutoff semantics; budget the investigation (~15 turns reading artifacts + contracts, ~5 validation tools, ~10 verifying top concerns, final ~5 RESERVED for writing `## Review` + the return summary); a verdict backed by fewer verified findings ALWAYS beats no verdict; never end the run with the primary artifact missing its `## Review` for this iteration.
3. **Missing-verdict guard** in stage-protocol §12a step 3 + one sentence in each of the five harness SKILL.md reviewer bullets: if the primary artifact has no `## Review` section for THIS iteration (check the iteration number on any existing section), treat the result as **NOT-READY** with the finding "review did not complete within its turn budget" and count the iteration against `reviewer_max_iterations`. The existing lead-alone revision loop and iterations-exhausted gate path apply unchanged.

### Use Case

Same 12-unit field intent as the data behind #534: 16 functional-design reviewer dispatches spanning >24h wall-clock, individual reviews 20–90 minutes, 56–113 turns each. The read-scope fix bounds the input set; this bounds the work loop itself, and — critically — closes the failure mode where any bounded or dying reviewer silently yields a verdict-less review that the conductor then misreads. With part 3, a cut-off review degrades gracefully: it consumes an iteration, the builder gets a concrete finding, and the gate is never presented on a silently-missing verdict.

This ran in the field (same project, 2.2.17 on Claude Code) from 2026-07-08 onward: reviews completed within the cap after the effort step-down, and no verdict-less review was observed post-change.

### Version

v2 (alpha)

### Area

Construction phase rules

### Additional Context

Implementation sketch (verified against v2 @ 2.5.0, commit 3f16023):

- `core/agents/aidlc-architecture-reviewer-agent.md`, `core/agents/aidlc-product-lead-agent.md` — frontmatter key + `## Turn Budget` section (keep the two numbers in sync; the section states the frontmatter is the enforcing surface on Claude Code and the budget is prose elsewhere).
- `core/aidlc-common/protocols/stage-protocol.md` §12a step 3 — the missing-verdict branch, worded additively: t221 pins the "Read verdict → delete `.aidlc-reviewer-dispatch.json`" ordering, and t217/t234/t200 pin step 1/2 phrases; add the branch after the record deletion, don't reword existing lines.
- `harness/{claude,kiro,kiro-ide,codex,opencode}/skills/aidlc/SKILL.md` — one sentence in the "Reviewer step (§12a)" bullet (kiro-ide's wording stays dispatch-record-free, as t221 pins).
- New unit pin test (t217's HARNESS_MATRIX pattern): frontmatter carries `maxTurns` on the two reviewers in core + the Claude dist; the missing-verdict sentence present in §12a and all five SKILLs.
- Docs: `docs/reference/04-stage-protocol.md` reviewer-invocation section; regenerate dists; CHANGELOG + version bump.

Design notes for maintainers:
- Part 3 stands alone even if the cap is rejected — the undefined branch exists today for any reviewer death.
- Codex/Kiro/opencode get the budget as persona prose only in v1 (no native per-agent turn-cap key is wired); extending `emit.ts` for a Codex-native bound is a possible follow-up.
- Numbers (40 / ~5 reserved) are field-derived starting points, not sacred; they live in exactly two places (frontmatter + persona section).

Related: #534 (read-scope bound — the WHERE axis; this is the HOW-MUCH axis), #539 (deterministic read-scope enforcement), and the companion issue proposing `effort: medium` for the balanced tier (independent; compounds the same wall-clock cost).

jstrunk:
Closes #609.

> Maintainer takeover per the thread below: rebuilt on current v2 (v2.6.2) by @apackeer, since the original branch predated the review-class engine and could not be rebased textually. jstrunk's design and field data drive the feature; the recovery model is re-derived on the current contracts.

## What

Bounds the section-12a reviewer's work and closes the undefined branch where a reviewer dies before writing its verdict.

- **60-turn hard cap** on the two review-only agents (`aidlc-architecture-reviewer-agent`, `aidlc-product-lead-agent`): authored once as `maxTurns: 60` frontmatter, projected per harness. Binding on Claude Code; renamed to the native per-agent `steps: 60` on opencode; prose-only on Codex CLI (the TOML emit rewrites the persona's frontmatter citation), Cursor, GitHub Copilot, and Kiro CLI/IDE (no per-agent cap key; the inert key ships on the tolerant .md surfaces and never enters the fail-closed kiro agent JSONs).
- **Harness-neutral `## Turn Budget` persona section** in both reviewers: plan for the worst-case cutoff, reserve the final turns for writing, exactly one `## Review` section with one canonical verdict, a thin verdict always beats no verdict.
- **Section-12a step 1** deletes any pre-existing `## Review` section before every dispatch (first entry, NOT-READY re-invoke, post-rejection re-review). "No current section" means "incomplete review" on every path; no stale pre-revision READY survives; nothing accumulates in artifacts or leaks into the claim-sources scan. Review history lives in the audit ledger.
- **Section-12a step 3** validates the verdict: exactly one current `## Review` section with exactly one canonical READY/NOT-READY. Anything else (missing, verdict-less, duplicated) is an incomplete attempt: re-dispatch the same unmatched request once via `--retry-pending` (consumes no review iteration, so an advisory single-pass budget survives a cutoff), then record a terminal `NOT-READY` receipt with the finding "review did not complete within its turn budget". The gate is never presented on, or deadlocked by, a silently missing verdict.
- **GitHub Copilot orchestrator catch-up**: its reviewer step predated the review-class engine; it now carries the `directive.review_class` branch, advisory single-pass terminality, terminal-receipt freeze wording, and the `--retry-pending` recovery contract. The protocol's reviewer-scope enforcement roster now names Copilot (its adapter forwards to the core reviewer-scope and review-freeze hooks), and copilot joins the t263/t266 harness test lists that previously skipped it.
- **t279-reviewer-turn-budget** pins every surface, including the step-3 terminal-receipt sentence that a previous rebase silently dropped.

## Why 60

Field timing data on #612 (132 observed reviews): at the medium-effort reviewer tier, legitimate reviews ran p90 40 / max 60 turns, with a clean gap to runaways. 60 is the top of the observed legitimate range: it catches runaways without truncating real reviews. The reviewer tier landed in 2.5.54, making the backstop meaningful (at inherited session effort there was no usable gap).

## Review-thread findings addressed

All prior review rounds are resolved in this rebuild: the stale-READY revision-path hole (now closed by delete-before-dispatch rather than the earlier supersede-rename, which leaked renamed reviewer prose into the claim-sources sensor and accumulated forever), the dropped terminal-receipt sentence (restored and pinned), iteration consumption on a cut-off review (now `--retry-pending`, no iteration consumed, aligned with the advisory one-pass budget), and canonical-verdict validation for partial or duplicated sections.

## Testing

- `bun scripts/package.ts --check` (7 harness trees, byte parity), `bun run typecheck` (3 tsconfigs), `bun run check` (biome) all green.
- smoke + unit + integration tier: 318/321 pass; the 3 reds are all pre-existing or environmental (t-memory-seed load flake, green alone; t145 /tmp module-resolution box artifact; t238 live-SDK red reproduced on clean v2), none touching this diff.
- Version/slots: 2.6.8 (v2 tops at 2.6.2; 2.6.3-2.6.7 are claimed by open PRs), test slot t279 (free on v2).


apackeer:
Reviewed at head 591f66a. Verdict: the design is sound and the external-harness claims all check out against primary sources. One real protocol gap on the revision path (finding 1 below) and the usual rebase/slot logistics (finding 2 below) before this can land; the rest are smaller. Findings ranked.

## Verified first (primary sources, not the PR text)

- `maxTurns` is a documented Claude Code subagent frontmatter field ("Maximum number of agentic turns before the subagent stops"). The official docs do NOT document the cutoff semantics, so the "no final message, caller receives no output" claim rests on the #609 probe, which is disclosed there. Acceptable evidence.
- The opencode claims verified against the shipped 1.17.18 binary: the agent config schema carries `steps` ("Maximum number of agentic iterations before forcing text-only response") and `maxSteps` annotated "@deprecated Use 'steps' field instead". So the emitter rename and the "final text-only turn cannot write the review" reasoning are both accurate, and the missing-verdict guard genuinely stays load-bearing there.
- The emitter change is well anchored: the frontmatter rename matches only a `^maxTurns: <n>$` line and the prose rename only backticked `maxTurns: <n>` citations, so it cannot munge arbitrary body text. `.opencode/agents/` is correctly the only translated surface; the `.aidlc/agents/` mirror keeps core shape, so each shipped tree's prose cites its own frontmatter key.
- The missing-verdict sentence is present in all five authored SKILL.md files plus dist copies; kiro-ide stays dispatch-record-free (t221's pin holds); the section 12a step 3 edit preserves the delete-record-then-read ordering.
- The iteration-number check binds to a real contract: the prescribed review format carries `**Iteration:** [1, 2, etc.]` plus "replace, don't append" in both reviewers' reviewing.md knowledge.
- At the PR head as authored: package drift check, typecheck, coverage registry, and the smoke (1117 assertions) + unit (3147 assertions) tiers all green.

## Findings

**1. P1 - the guard still accepts a stale READY on the Part 0 revision path.**

Sequence: a stage passes review READY at iteration 1; the human rejects at the gate; the builder revises a `produces[]` artifact; section 12a re-runs "as at first entry" with a fresh iteration budget, so the re-review is again iteration 1. If that re-review is capped before writing (the exact failure this PR targets), the artifact still carries the pre-revision section reading Iteration: 1, verdict READY. The new guard's "check the iteration number on any existing `## Review` section" matches, so the conductor reads a stale READY that predates the revision as current and presents the gate on it - precisely the silently-missing-verdict outcome the PR set out to eliminate, still reachable on the revision path. Note the iteration number in the template is reviewer-inferred; the conductor supplies no identifier to check against.

Two ways to close it: (a) the conductor removes or renames the stale section (e.g. `## Review (superseded)`) before every reviewer re-dispatch - then a missing section means an incomplete review on every path, and the iteration check becomes belt-and-braces rather than the sole discriminator (the "does not modify the artifact" constraint binds the reviewer, not the conductor, so no conflict); or (b) the conductor passes a unique attempt token into the dispatch and the reviewer echoes it in the section. (a) is the smaller change: one conductor-side sentence in section 12a step 1 plus the SKILL bullets. Whichever lands, please add a test that pins this failure path - t243 currently pins the wording of the guard, not its semantics.

**2. P1 - rebase + slot renumbering.**

The branch is CONFLICTING against v2, and both of its slots are now taken: v2 is at 2.5.2 (#576 merged, heading `## [2.5.2] - 2026-07-21`) and `tests/unit/t243-doctor-bundle.test.ts` occupies t243. The branch also predates 2.5.1 (the CHANGELOG diff inserts 2.5.2 directly above 2.5.0). Needed: rebase onto v2; re-bump version + CHANGELOG heading + README badge to the next free slot at merge time (coordinate with the other open claims: #617 has 2.5.3, #616 has 2.5.4, #569 has 2.5.5); rename the test to t244 (free on v2) including its covers: header; regenerate dists and the coverage registry; re-run the reported tiers. t68 will red on the duplicate `## [2.5.2]` heading immediately post-rebase, so this is mechanical but mandatory.

**3. P2 - the missing-verdict branch spends a full builder dispatch on a finding the builder cannot act on.**

"review did not complete within its turn budget" feeds the existing NOT-READY loop, whose first step re-invokes the lead alone to address the findings - but this finding is not an artifact defect; there is nothing for the builder to fix. Consider short-circuiting this specific finding straight to the reviewer re-invoke (still counting the iteration). Not blocking - the current shape is simple and bounded - but it wastes a whole subagent run per cutoff.

**4. P2 - the Codex TOML persona points at frontmatter that does not exist on that surface.**

`dist/codex/.codex/agents/*.toml` carry "the `maxTurns: 40` frontmatter above - keep the two numbers in sync" inside `developer_instructions`, which has no frontmatter and no cap key (Codex agent discovery reads only the TOML; the .md beside it is conductor-side prose). The harness-neutral stance is fine as disclosed, but this sentence is a concrete false pointer, and the keep-in-sync maintenance instruction is addressed to a model that cannot see the file it names. The opencode emitter already solved exactly this pattern with its prose rename; the Codex emitter could rewrite or drop the parenthetical when transposing to TOML.

**5. P3 - the agent-authoring guide omits the new key.**

`docs/harness-engineering/03-adding-an-agent.md` enumerates the authored frontmatter contract (name, display_name, examples, description, disallowedTools, tier, optional tools:) and now misses the optional `maxTurns`. One line there keeps the documentation policy satisfied.

**6. P3 - verify-item: the kiro-ide .md frontmatter parser.**

The kiro-ide dist personas now carry an unknown frontmatter key on the surface the IDE actually parses for subagent tool grants (the frontmatterAdditions seam in the kiro-ide manifest). Almost certainly ignored as an unknown key, but the PR's schema research covered the agent-v1 JSON, not the IDE's .md frontmatter parser - one sentence of confirmation from your kiro-ide probing would close it.

## On the test (t243, to become t244)

Good coverage shape: both sides of the number-sync contract, the worst-case-cutoff wording, the harness-name negative pin, the opencode translation (exactly one `steps: 40` per reviewer, no `maxTurns` leak, no cap on the other 12 agents), the section 12a branch ordering after t221's pin, and all five SKILL mirrors via HARNESS_MATRIX. The roster-count pin (14) will need a touch when an agent lands, consistent with existing pins. No asks beyond the renumbering and the stale-READY semantics test asked in finding 1.

jstrunk:
Thanks for the thorough review — addressed the P1 and rebased. Point by point:

**1. P1 — stale READY on the Part 0 revision path.** Confirmed the gap: the old guard's "check the iteration number on any existing `## Review` section" had no conductor-supplied identifier to check against, so a re-review cut off before writing could leave the pre-revision `Iteration: 1, verdict READY` section looking current. Implemented option (a): §12a step 1 now renames any existing `## Review` section to `## Review (superseded)` before **every** reviewer dispatch — first entry or re-dispatch alike — so step 3's check no longer infers staleness from the iteration number at all; "no *current* `## Review` section" now means "incomplete review" uniformly on both paths. Updated in `core/aidlc-common/protocols/stage-protocol.md` §12a (steps 1 and 3, plus the Part 0 revision-path paragraph) and mirrored in all five harness `SKILL.md` reviewer bullets. Added a dedicated test in the renamed `t248-reviewer-turn-budget.test.ts` that reproduces the stale-READY-after-revision scenario and pins the supersede-before-dispatch fix (plus a negative pin that the old iteration-number-inference phrasing is gone).

**2. P1 — rebase + slot renumbering.** Rebased onto current `v2` (`9f91454`). Used your pre-assigned numbers: version `2.5.2` → `2.5.14`, test file `t243-reviewer-turn-budget.test.ts` → `t248-reviewer-turn-budget.test.ts`. CHANGELOG heading is `## [2.5.14] - 2026-07-28`, README badge and all six `aidlc-version.ts` copies updated.

**3. P2 — wasted builder dispatch on a missing-verdict finding.** Fixed — small prose change. Step 3's missing-verdict branch now skips the lead-agent re-run and returns straight to re-invoking the reviewer (the artifact was never touched, so there's nothing for the builder to act on), still counting the iteration and still respecting `reviewer_max_iterations`. Ordinary NOT-READY with real artifact findings keeps the existing builder loop unchanged.

**4. P2 — Codex TOML citing nonexistent frontmatter.** Confirmed — Codex TOML personas have no frontmatter and no cap key (Codex agent discovery reads only the TOML). `harness/codex/emit.ts`'s `emitAgentToml` now rewrites the "the `maxTurns: 40` frontmatter above — keep the two numbers in sync" citation to name the actual state on that surface (no frontmatter, no native cap key, prose-only number), following the same precedent as the opencode emitter's own `maxTurns:`→`steps:` prose rename.

**5. P3 — agent-authoring guide missing `maxTurns`.** Added a line to `docs/harness-engineering/03-adding-an-agent.md`'s frontmatter-contract list documenting the optional `maxTurns: <n>` key and its per-harness enforcement shape.

**6. P3 — kiro-ide frontmatter parser verify-item.** Investigated, no code change needed. `harness/kiro-ide/manifest.ts`'s `frontmatterAdditions` seam only *appends* the `tools:` grant line to the persona `.md` frontmatter; there's no generic unknown-key validation anywhere in the packager for that surface. `maxTurns` ships there exactly like the pre-existing `disallowedTools` and `display_name` keys, which the IDE already tolerates as unrecognized-but-inert (field-proven on Kiro IDE 1.0.165+, per the 2.5.10 hook-registration work). Confirmed safe.

**Verification after rebase + fixes:**
- `bun scripts/package.ts --check` — clean, all five harness trees in sync.
- `bun tests/gen-coverage-registry.ts --check` — OK, ratchet held.
- `bun run check` (package drift + typecheck + biome lint) — clean.
- `bash tests/run-tests.sh --smoke --unit` — 179 test files, 4405+ assertions, 0 failures.
- Targeted: `t248-reviewer-turn-budget.test.ts` (9 tests), `t221-reviewer-scope-hook.test.ts`, `t68-version-changelog-sync.test.ts` — all green.

PR title updated to reflect 2.5.14. Ready for another pass.

leandrodamascena (changes requested):
Thanks for this, Jeff — the targeted tests, package parity, typecheck, coverage registry, and the opencode `steps: 40` resolution are all green, and I confirmed the supersede-before-dispatch rule genuinely closes the stale-verdict read. I saw a few semantic paths that the text-based tests don't exercise, though, and two of them can strand a workflow. Grouping by what I'd block on vs. what's worth fixing alongside:

## Blockers

**[P1] A cut-off review never records the terminal receipt.** `stage-protocol.md:993` dropped the old instruction to log `aidlc-log.ts review --verdict <READY|NOT-READY>`. On the missing-review path there's no verdict to "read," so nothing clearly emits a synthetic NOT-READY receipt — and after exhaustion `verifyReviewerPrecondition()` rejects the human's Approve because no fresh `REVIEW_COMPLETED` exists. Re-invoking just cuts off again, so the stage strands. Please record a synthetic NOT-READY receipt on every incomplete attempt. (The "reaches the human gate" guarantee itself is pre-existing — byte-identical at the merge base — so this is specifically about the new cutoff routing lacking the receipt that route needs.)

**[P1] A partial `## Review` leaves an undefined branch.** The guard only checks whether the heading exists. A reviewer can write `## Review` and then be cut off before writing a valid READY/NOT-READY verdict — that skips the missing-section branch (the heading is there) but satisfies neither verdict arm at :994/:995. Please require exactly one current section with exactly one canonical verdict; missing, malformed, multiple, or conflicting verdicts should all take the incomplete-review path.

## Worth fixing alongside

**[P2] The 40-turn default rests on the rejected effort step-down.** The field baseline under inherited `xhigh` was 56–113 turns; the evidence that reviews finish under 40 came only after the `medium` effort change from #612, which was rejected. With the #546 contract preserved, a 40-turn default may terminate otherwise-normal reviews most of the time. Could you recalibrate and validate the cap under inherited session effort, or make it opt-in/configurable?

**[P2] Reviewer knowledge conflicts with supersede-before-dispatch.** Both `reviewing.md` files still say to "Update the `## Review` section (replace, don't append a second one)" (arch :104, product-lead :82). But the conductor now renames the prior section to `## Review (superseded)` and expects the reviewer to preserve it and append one new current `## Review`. Please update those instructions and add coverage for the interaction.

## Nit

**[P3] The missing-review retry is bounded, but only implicitly.** The cutoff branch does count the attempt and the "(iterations-exhausted still applies unchanged)" clause does terminate it — so it's not an infinite loop. But the exhaustion condition lives in the normal-verdict branch and the cutoff branch reaches it only via that parenthetical. Worth branching explicitly: retry only while attempts remain, otherwise the unresolved-findings path.

Also minor, not code: the PR body cites `t243` (the file is `t248`; t243 is doctor-bundle) and "2.5.2" (shipped is 2.5.14) — and 2.5.14 collides with #661, so second-to-merge re-bumps.

apackeer:
Round 2, reviewed at head 3e88ec0f. All six round-1 findings are fixed and verified - the supersede-before-dispatch rule closes the revision-path hole cleanly, and t248's negative pins are exactly what I asked for. One new P1 that the rebase introduced, one P2 interaction with the claim-sources sensor, and the usual slot note. Findings ranked below.

## Round-1 findings: all verified fixed

1. **Stale READY on the revision path** - option (a) implemented: step 1 renames any existing `## Review` section to `## Review (superseded)` before every dispatch, mirrored in all five SKILL.md reviewer bullets; t248 pins the rule plus a negative pin that the iteration-number-inference phrasing is gone. Verified in `stage-protocol.md` and each harness SKILL.
2. **Rebase + slots** - on 9f91454, 2.5.14 + t248, CHANGELOG heading / README badge / all six version copies agree (t68 green).
3. **Missing-verdict branch** now skips the builder re-invoke and returns straight to the reviewer - confirmed in step 3 and all five SKILLs.
4. **Codex TOML citation** - `emitAgentToml` rewrites the frontmatter pointer to the actual state on that surface; verified in both shipped TOMLs, and the `.md` twin beside them keeps the frontmatter-accurate wording.
5. **`maxTurns` documented** in `docs/harness-engineering/03-adding-an-agent.md`.
6. **kiro-ide frontmatter** - agreed: the key ships beside `disallowedTools` and `display_name`, which that surface already tolerates as unknown-but-inert.

Independently re-verified at your head: `bun scripts/package.ts --check`, typecheck + lint, the coverage-registry check, and the smoke + unit tiers (179 files, 4405 assertions, 0 failures), plus t248 (9 tests) targeted.

## New findings

**1. P1 - the rebase dropped the terminal-receipt sentence from section 12a step 3, and the new cutoff branch never records a receipt.**

On current v2, step 3 reads "...then read the `## Review` section from the primary artifact. **Record the terminal receipt with the same `aidlc-log.ts review` command plus `--verdict <READY|NOT-READY>` (and the same `--unit` / `--single` fields)**, then branch on the verdict:". That sentence landed in 2.5.5 (#569), after this branch forked, and the step-3 rewrite here kept the pre-2.5.5 text - so the shipped protocol no longer instructs the conductor to write the terminal receipt at all.

This is load-bearing, not prose: `verifyReviewerPrecondition` in `core/tools/aidlc-state.ts` refuses every completion path (`approve`, `advance`, `finalize`, `complete-workflow`) for a reviewer-bearing stage until a fresh `REVIEW_COMPLETED` row exists. The SKILL bullets still carry their own copy of the receipt instruction, but it is phrased "after reading the verdict, run the same command with `--verdict ...`" - and on the missing-verdict branch there is no verdict to read, so no surface tells the conductor to record anything for a cut-off attempt.

Concrete failure: both iterations get capped; the new branch says iterations-exhausted applies unchanged; the protocol proceeds to the gate with unresolved findings noted; the human approves; the engine refuses the approve because zero `REVIEW_COMPLETED` rows exist for the stage. That is exactly the outcome the precondition's own comment wants to permit ("soft on the verdict: a NOT-READY-after-cap still lets the human approve") - but only when the receipt is recorded.

Asks: (a) restore the terminal-receipt sentence in step 3; (b) state explicitly on the missing-verdict branch that the conductor records the receipt with `--verdict NOT-READY` (the "after reading the verdict" phrasing does not reach this branch); (c) mirror in the five SKILL bullets if the wording changes; (d) add a pin - no test guards that sentence today, which is why the rebase could drop it silently. t248 is the natural home.

**2. P2 - `## Review (superseded)` sections lose the claim-sources exclusion and accumulate without bound.**

The claim-sources sensor skips exactly the H2 named `Review` (`REVIEW_HEADING = "Review"`, exact match in `core/tools/aidlc-sensor-claim-sources.ts`). After a supersede rename the heading no longer matches, so every untagged line of the old review body becomes a claim-sources finding. This fires on the very first NOT-READY loop of Intent Capture: iteration 2's dispatch renames iteration 1's section, and the sensor re-checks the deliverables on the next write. Advisory severity, so nothing blocks - but Intent Capture declares both this reviewer and this sensor, so the noise is guaranteed, not hypothetical. The renamed sections also accumulate across iterations and Part 0 revisions as duplicate identical headings.

Two ways out: teach the sensor to skip any H2 starting with `Review` (one line plus a manifest sentence), or have step 1 delete the stale section rather than rename it - the reviewers' own iteration contract is already "replace, don't append", so preserving superseded reviews in the artifact is new behavior rather than preserved history. Either works; deleting is smaller and keeps artifacts clean.

**3. P3 - version slot.** 2.5.14 is contested (#661 also claims it, and #617/#616/#667 hold 2.5.15/2.5.16). Standard policy applies: whichever merges second rebases and re-bumps the version, CHANGELOG heading, and badge. Flagging so it is not a surprise at merge time; the test slot t248 has a parallel collision risk for the same reason.

Everything else I re-checked is clean: the opencode roster carries `steps: 40` with the prose renamed and no `maxTurns` leak, the `.aidlc` mirror keeps core shape, the docs matrix accurately describes both the supersede rule and per-harness enforcement, and the CHANGELOG entry is well-formed with an upgrade instruction. Fix the P1 (mechanical - restore one sentence, extend it to the cutoff branch, pin it) and decide a direction on the P2, and this is ready.

jstrunk:
Re the request for a much higher turn limit: I posted field timing data on #612 (https://github.com/awslabs/aidlc-workflows/pull/612#issuecomment-5123822479) that sizes this cap, and it shows the two PRs really only work as a pair — I'd suggest deciding them together.

The short version: at reviewer effort medium there's a clean gap between normal and runaway reviews (p90 40 turns, max 60 across 132 observed reviews), so the 40–60 range is a real backstop. At inherited session effort there is no such gap — legitimate reviews reached 116 turns (p90 112), so 40 truncates a third of them and a "safe" ~120 cap protects against essentially nothing. If the `review` tier proposed on #612 lands, the cap in this PR should be 40–60; if it doesn't, `maxTurns` here buys little and I'd rather drop the cap than ship a meaningless number — though the §12a missing-verdict guard in this PR is worth keeping either way, since it catches silently incomplete reviews from any cause (crashes, opencode's forced text-only final step, model-routing bugs like anthropics/claude-code#81995), not just cap truncation.

apackeer:
Thanks, Jeff. I want to make sure we evaluate these two PRs against the combined goal you described. Medium reviewer effort is the primary latency control: in your field data it reduced the architecture reviewer's median from 42.0 minutes to 4.1 minutes and p90 from 102.7 minutes to 14.6 minutes. The turn limit is then a deterministic backstop for outliers, while the incomplete-review guard makes a cap, crash, or harness cutoff fail clearly instead of leaving an undefined verdict path.

We closed #612 because changing the existing `balanced` tier to medium would override its inherit-session-effort contract. Your subsequent proposal for a dedicated `review` tier is materially different: `balanced` would continue to inherit session effort, while the two review-only agents would use medium effort. That appears to address the reason #612 was closed and deserves an explicit decision before we dispose of this PR.

@leandrodamascena, could we decide the following here?

1. Do we want a dedicated `review` tier at medium effort?
2. If so, what backstop should accompany it? The submitted value is 40 turns, while Jeff's later data identifies 40-60 as the meaningful range.
3. Regardless of the tier decision, can we agree to carry forward the deterministic incomplete-review guard?

The current guard still needs to validate exactly one canonical verdict and record a terminal NOT-READY receipt for an incomplete attempt, but that is an implementation detail after the policy decision. Jeff, please do not rebase or revise this branch while we resolve that direction.

apackeer:
@jstrunk Reviewer tier has now been implemented. i will rebase and merge this. @leandrodamascena your request changes remains, ill request rereview when ready to merge

leandrodamascena (approved):

leandrodamascena:
@apackeer approved! You can merge.

PR #606, opened 2026-07-20
fix(kiro): remove model pins from agent surfaces - inherit session model

csotomon:
## Summary

Fixes #601. On the Kiro harnesses, every delegated agent spawn fails with `Sub-agent execution failed: Invalid model ID. Please select a different model to continue.` whenever the shipped model pins name a model that is not enabled on the user's Kiro install. This blocks the subagent stages, every `reviewer:`-bound gate, the 2.5.0 ensemble dispatches (pipeline/mob/hub-and-spoke), and any manual delegation.

Two stacked pin layers cause it (both in dialects/IDs Kiro may not resolve):

1. **Agent JSONs**: the authored conductor pin (`"model": "claude-opus-4.8"` in `agents/aidlc.json`, both Kiro harnesses) plus the tier-projected persona pins (`"model": "claude-sonnet-4.5"` on the balanced/templated agent JSONs).
2. **Agent `.md` frontmatter**: the same tier projection writes `model: claude-sonnet-4.5` lines, and Kiro IDE reads the frontmatter when spawning.

There is no universally safe pinnable value — whether an ID resolves depends on which models the user's install has enabled — so the fix makes Kiro agents **inherit the session model**, which the agent-v1 schema documents as the absent-`"model"` fallback ("If not specified, uses the default model").

Field verification (Kiro IDE, session on a model the pins never named): with both pin layers stripped, all 14 agents spawn and respond via `invoke_sub_agent`, and `--doctor` stays green.

## Changes

- `core/tools/aidlc-tiers.ts`: the kiro flavor of `TIER_PROJECTIONS` no longer pins a model for any tier (`balanced`/`templated`: `claude-sonnet-4.5` → `null`); `KIRO_TIER_EFFORT` is now empty. The packager consequently omits `"model"` from every projected Kiro agent JSON and every `model:` frontmatter line on both Kiro trees. The machinery (`kiroModelDefaults()`, the kiro projection slots) stays in place, dormant, should a per-install-resolvable pinning mechanism appear.
- `harness/kiro/agents/aidlc.json` + `harness/kiro-ide/agents/aidlc.json`: the authored conductor pin is removed — the conductor inherits the session model too.
- `settings/cli.json` (both harnesses) keeps ONLY the authored conditional effort default (`chat.modelDefaults["claude-opus-4.8"] → xhigh`): it applies when the session actually runs that model and is inert otherwise — and inert for spawns either way. The tier-derived `claude-sonnet-4.5 → high` entry no longer ships.
[...]

PR #614, opened 2026-07-20
fix(kiro-ide): migrate the nine hooks to the IDE's v2 hook schema

csotomon:
## Summary

Kiro IDE >= 1.0.1xx no longer executes the legacy `.kiro.hook` format this harness ships, leaving `dist/kiro-ide/` installs with **no hooks layer at all**: no `HUMAN_TURN` minting (the human-presence floor), no `SESSION_STARTED`, no `ARTIFACT_CREATED/UPDATED` audit rows, no sensors, no context injection, no Stop enforcement. The workflow still advances (the engine writes the ledger via CLI), but every hook reinforcement is silently gone.

Verified live on Kiro IDE 1.0.165 (ACP client):

- A debug marker plus the adapter's unconditional entry logging showed the adapter was **never invoked** for any of the 9 legacy hooks.
- A ledger census across 5 sessions of a full intent: zero hook-origin events.
- A probe hook in the v2 schema (`{"version":"v1","hooks":[...]}`, PascalCase triggers) fired **in the same session, hot-loaded, without a restart**.

This is the first of two PRs splitting #603 per review feedback (bug fix vs adapter enhancement). This PR is the **pure schema migration**: the nine hooks move 1:1 to the v2 format, the adapter is untouched.

## Changes

The nine kiro-ide hooks migrate 1:1 to the v2 schema (`harness/kiro-ide/hooks/aidlc-*.json`; the legacy `.kiro.hook` files are removed, as #555 proposes). Same adapter targets, same commands, same semantics:

| Hook | Legacy trigger | v2 trigger | matcher |
|---|---|---|---|
| aidlc-mint | promptSubmit | UserPromptSubmit | — |
| aidlc-session-start | promptSubmit | UserPromptSubmit | — |
| aidlc-block | preToolUse | PreToolUse | — |
| aidlc-audit-logger | postToolUse [write] | PostToolUse | `fs_write\|str_replace\|fs_append` |
| aidlc-runtime-compile | postToolUse [shell] | PostToolUse | `execute_bash` |
| aidlc-sync-statusline | postToolUse [shell] | PostToolUse | `execute_bash` |
| aidlc-log-subagent | postToolUse [.*invoke_sub_agent.*] | PostToolUse | `invoke_sub_agent` |
| aidlc-stop | agentStop | Stop | — |
| aidlc-session-end | agentStop | Stop | — |

Also: manifest + docs updated (`docs/guide/harnesses/kiro-ide.md` hook table, `docs/reference/14-claude-features.md` Automation row), version 2.5.1 + CHANGELOG entry (including the upgrade note to delete stale `.kiro.hook` files from existing installs), dist regenerated via `bun scripts/package.ts`.

**Known boundary this PR deliberately keeps:** on IDE 1.x the two payload-dependent targets (`audit-and-sensors`, `log-subagent`) fire but no-op safely, because 1.x delivers the tool payload on stdin and the shipped adapter still reads only the 0.12 `USER_PROMPT` env channel (empty on 1.x). The other seven hooks are payload-independent and fully functional. The stdin context channel lands in the follow-up PR (stacked on this one), so each change reviews on its own merits. Strictly better than today either way: currently NOTHING fires on 1.x.

## What was tested

- `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`: green.
- Full default suite (smoke + unit + integration, serial): 276 files / 5582 assertions — the only failure is t19 (requires live AWS credentials; failed with `ExpiredToken` in this environment, unrelated to the change).
- t218 (the IDE adapter contract) passes untouched against the old adapter — the v2 JSONs invoke the exact same commands the legacy files did.
- Live on Kiro IDE 1.0.165: v2-schema probe hooks fire hot-loaded in the same session (evidence in #555).

## Deliberately out of scope

- **The adapter stdin channel** (#543): follow-up PR, stacked on this one.
- **The Kiro CLI tree (`dist/kiro/`)**: its hooks ride the agent-JSON `hooks` block, not hook files.
- **The plugin-compose wiring** still emits the legacy `.kiro.hook` (separate surface, documented caveat).
- **A real `SessionStart` trigger**: this PR keeps session-start wired to `UserPromptSubmit` (per-prompt `SESSION_STARTED`, exactly as before) to stay a pure translation.

## References

- Addresses #555 (legacy hook format inert on IDE >= 1.0.1xx; live evidence: https://github.com/awslabs/aidlc-workflows/issues/555#issuecomment-5017268046)
- Replaces #603 (part 1 of 2) per review feedback to separate the bug fix from the adapter enhancement
- Version note: authored as 2.5.1; if another open PR merges first claiming it, this one re-bumps per the repo's changelog conflict policy.


apackeer:
Thanks for the careful split and the live evidence on #555 - the nine translations check out 1:1 against the deleted files, the adapter is untouched as advertised, and package --check / typecheck / lint / the targeted tier (t218, t221, t239, t68, t145) are all green at 47c4464.

One scope change request plus a short list of fixes.

## Rescope: ship both formats instead of deleting the legacy files

We want to keep supporting pre-1.0 IDE builds for a while (see @raj-jain-aws's point in #555), so please keep the nine `.kiro.hook` files and their manifest entries and ADD the v2 JSONs alongside - the approach #600 proposed. We ran a live probe spike to close the two unknowns that blocked this:

- **Kiro IDE 0.12.333** (last pre-1.0 patch): v2 `.json` hook files are completely ignored - not listed in the Agent Hooks panel, never fired, no error or migration UI. Legacy hooks fired normally (promptSubmit and postToolUse both verified). So the v2 files are dead weight there, not a hazard.
- **Kiro IDE 1.0.116** (the build #555 flagged as running legacy with a migration prompt): with BOTH formats present in the workspace, there is **no double-firing**. Only the v2 hooks executed; the legacy files were listed struck-through with a passive "Migrate legacy hooks to v1" banner and never ran. (Caveat: we did not test legacy-alone on 1.0.116, but for a distribution that always ships both together, both-present is the configuration that matters.)
- 1.0.165 behavior is already field-proven by your own evidence (legacy inert, v2 fires).

So coexistence is safe on every generation tested, and no dedup logic is needed. Concretely: restore the nine `.kiro.hook` files + manifest rows, keep your v2 JSONs, and flip the CHANGELOG upgrade note - users should NOT delete their legacy hook files, and (new finding from the spike) should NOT click the IDE's per-hook "Migrate" button on the aidlc hooks, since the v2 versions are already shipped and an IDE-generated twin could double-fire.

## Fixes needed on the current diff

1. **Rebase + re-bump.** v2 is at 2.5.2, so 2.5.1 is taken - re-bump per the changelog conflict policy (rename the heading too; t68 will catch a miss).
2. **Stop cannot block on the IDE.** Kiro's hook docs (https://kiro.dev/docs/hooks/) list `Stop` as "Can block? No", and stdout at exit 0 is ignored for Stop/PostToolUse. The core stop hook's `{"decision":"block"}` contract is therefore inert on this harness - a platform limitation, not something this PR could fix, but the "seven fully functional" claim (PR body, CHANGELOG) and the docs table row for `aidlc-stop` overstate it. Please downgrade stop to advisory/observability-only in the same known-boundary style you used for the payload-dependent pair. (Cross-checked: UserPromptSubmit stdout IS added to context and PreToolUse exit 2 DOES block, so session-start injection and the human-presence floor are fine.)
3. **Stale README lines.** `README.md:115` still says the IDE install registers hooks as `.kiro/hooks/*.kiro.hook`, and the repo-layout tree (~line 287) still lists only `.kiro.hook files` for `harness/kiro-ide/` - under coexistence both lines should name both formats.
4. **Guide's context-channel paragraph.** `docs/guide/harnesses/kiro-ide.md` ("The IDE delivers hook context through the USER_PROMPT environment variable") describes the 0.12 channel as current; on 1.x the payload arrives on stdin and USER_PROMPT is empty, which is exactly the known boundary this PR's body documents. Please add that caveat here too (or note it lands with #615).

Nice-to-have, fine to defer to #615: a small unit test pinning the v2 JSONs (version string, trigger vocabulary, adapter-target correspondence) so a `userPromptSubmit`-style typo can't pass silently, and the adapter's header comments still referencing the deleted `.kiro.hook` wiring.

Since #615 is stacked on this branch and also carries the deletions, the rescope flows into it as well.

csotomon:
Thanks for the thorough review @apackeer — all four fixes applied plus the rescope to coexistence. Force-pushed as `2e5a1c08` (rebased on upstream/v2).

**What changed:**

1. **Rescoped to coexistence:** legacy `.kiro.hook` files restored (both `harness/kiro-ide/hooks/` and `dist/`), manifest entries added back. v2 JSONs ship alongside. CHANGELOG upgrade note flipped: users should NOT delete legacy hooks, should NOT click the IDE's per-hook Migrate button.

2. **Re-bumped to 2.5.4** (2.5.1/2.5.2 taken on upstream, 2.5.3 claimed by #606). t68 green.

3. **Stop downgraded to advisory/observability-only:** docs table row, CHANGELOG known-boundary paragraph, and the seven-fully-functional claim replaced with five-fully-functional + two advisory (stop, session-end).

4. **README lines fixed:** line 115 and the repo-layout tree (~287) now name both formats.

5. **Context-channel caveat added** to `docs/guide/harnesses/kiro-ide.md`: notes that on 1.x the payload arrives on stdin and USER_PROMPT is empty, with a pointer to #615.

**Verification:** `bun scripts/package.ts --check`, `typecheck`, `lint` green; t68, t218, t221, t145, t239 all pass.

The nice-to-have (unit test pinning v2 JSONs + adapter header comment cleanup) I'll pick up in #615 since it's stacked on this branch.

leandrodamascena (changes requested):
The schema migration is directionally sound, but several issues prevent it from reliably restoring correct hook behavior.

## Blocking

1. **Rebase and version bump required**

   The branch declares `2.5.4`, while current `v2` is already at `2.5.6`. It conflicts in `CHANGELOG.md`, `README.md`, `core/tools/aidlc-version.ts`, and the generated version files.

   Please rebase onto `v2`, bump to the next available version, preserve newer changelog entries, update the date, and regenerate all distributions.

2. **The documented upgrade command installs hooks in the wrong directory**

   `README.md:108` and `docs/guide/harnesses/kiro-ide.md:34` prescribe:

   ```bash
   cp -r dist/kiro-ide/.kiro your-project/.kiro
   ```

   When `.kiro` already exists, as it does for the users this upgrade targets, this creates `.kiro/.kiro/hooks`. I reproduced this locally. Kiro therefore never sees the new hooks.

   Use content-copy semantics:

   ```bash
   mkdir -p your-project/.kiro
   cp -R dist/kiro-ide/.kiro/. your-project/.kiro/
   ```

## Correctness

3. **`aidlc-session-start.json:6` fires on every prompt**

   The v2 schema provides a real `SessionStart` trigger, but this hook uses `UserPromptSubmit`. The adapter labels every invocation as `source: "startup"`, so each prompt appends another `SESSION_STARTED` and reinjects the full workflow context.

   Please register it on `SessionStart`.

4. **`aidlc-session-end.json:6` records session termination after every turn**

   Kiro defines `Stop` as completion of an assistant turn, not closure of the conversation. This hook therefore appends `SESSION_ENDED` between prompts in the same session.

   Please remove this registration until a genuine session-end event is available, or otherwise prevent per-turn lifecycle corruption.

5. **`log-subagent` does not no-op with an empty 1.x payload**

   The changelog and guide state that both payload-dependent targets no-op safely on IDE 1.x. `audit-and-sensors` does, but `log-subagent` does not.

   With empty `USER_PROMPT`, `extractAgentIdentity("")` returns `"unknown"` and the adapter forwards a non-null event, producing a real `SUBAGENT_COMPLETED` row with `Agent Type: unknown`. No hook drop is recorded, so doctor cannot surface the degradation.

   Please return `null` when the result payload is empty, or correct the documented behavior.

## Coverage

6. **The nine new hook registrations have no structural contract test**

   No test parses the new JSON files and pins each hook's `version`, `trigger`, optional `matcher`, and adapter command. Packaging parity only proves that generated files reproduce the authored files; a valid-JSON typo could silently disable a hook while the suite remains green.

   Please add a table-driven test covering all nine registrations and their generated copies.

## Documentation

- `aidlc-stop.json:7` says the hook “Enforces the forwarding loop,” while this PR correctly documents that IDE `Stop` cannot block.
- `README.md:115` says `settings/cli.json` selects the IDE default agent, while the IDE guide correctly identifies it as CLI-only.

The relevant tests, packaging check, and typecheck pass on the current head. The remaining lint warning is pre-existing and outside this diff.

csotomon:
All 7 items addressed at head d0005eda. Rebased onto v2 at 2.5.7 (#639).

**Blocking**

**1. Rebase + re-bump to 2.5.8.** 2.5.7 was taken by #639 (which also touched the adapter). CHANGELOG heading, badge, and version.ts all set to 2.5.8 explicitly (not from conflict-resolution fallout). t68 passes.

**2. Install command nesting.** All Kiro install blocks (README Kiro IDE + CLI, kiro-ide.md, kiro-cli.md) now use content-copy semantics: `mkdir -p your-project/.kiro your-project/aidlc && cp -R dist/kiro-ide/.kiro/. your-project/.kiro/`. Added an explanatory paragraph in kiro-ide.md warning about the old `cp -r` form. Claude/Codex/opencode blocks use a different form (`cp -r src/ dst/` with trailing slash) that already works correctly on existing targets, so left untouched (can file a follow-up to unify if desired).

**Correctness**

**3. session-start fires on SessionStart.** `aidlc-session-start.json` trigger changed from `UserPromptSubmit` to `SessionStart`. The adapter comment updated to document both paths (v2 SessionStart vs legacy per-prompt). The legacy `.kiro.hook` file keeps its pre-1.0 per-prompt behavior unchanged.

**4. session-end v2 registration removed.** The JSON file is deleted, the manifest entry removed (with a comment explaining why), kiro-ide.md hook table drops the row and gains an explicit note that session-end has no v2 registration and why. The Session-audit-events comparison row now says `SESSION_STARTED only on IDE 1.x`. The legacy `.kiro.hook` stays for pre-1.0 (where agentStop behaved differently).

**5. log-subagent no-ops on empty payload.** When `toolResult` is empty (the IDE 1.x case), the adapter now calls `recordHookDrop()` and returns `null` instead of forwarding a fabricated `SUBAGENT_COMPLETED` row. A non-empty payload without an identity marker keeps the documented `unknown` fallback (t218 S3 still passes, confirming). `--doctor` can now surface the degradation.

**Coverage**

**6. `tests/unit/t244-kiro-ide-hook-registrations.test.ts`** added: table-driven contract test covering all 8 v2 JSONs (version, trigger, matcher, adapter command), the absence of `aidlc-session-end.json`, no unexpected extras, and all 9 legacy `.kiro.hook` files present. Runs against both authored and dist trees. 29 assertions per tree.

**Documentation**

**7a.** `aidlc-stop.json` description reworded: "Forwarding-loop audit on agent stop (advisory-only: the IDE's Stop trigger cannot block; enforcement relies on the conductor's own Stop protocol)."

**7b.** README:115 reworded: `cli.json` is now called out as a Kiro CLI-only compatibility surface that the IDE ignores.

**Verification:**
- `bun scripts/package.ts --check`: all 5 harnesses in sync
- `bun run typecheck`: green
- Smoke + unit: 176 files, 4350 assertions, 0 failures
- t68 (version sync): 7/7; t244 (hook registrations): 29/29; t218 (adapter): 30/30

**Scope note:** the `cp -r` nesting bug also applies to the Claude/Codex/opencode blocks, but those use a trailing-slash form (`cp -r dist/claude/.claude/ your-project/.claude/`) which already copies contents, not the directory itself — so they're correct as-is. Filed nothing because behavior matches intent; if the reviewers want unification across all 5 blocks I'm happy to do it in this PR.

leandrodamascena (changes requested):
Two documentation inconsistencies remain:

1. `harness/kiro-ide/onboarding.fills.ts:35` says Kiro IDE emits `SESSION_ENDED`, but this PR removes the IDE 1.x session-end registration. The incorrect claim propagates to `dist/kiro-ide/AGENTS.md:44`.

2. `docs/reference/kiro-ide-hook-payload.md:9-61` presents `USER_PROMPT` and dead stdin as universal behavior. Please distinguish pre-1.0 from IDE 1.x, where payloads arrive on stdin.

The seven previous findings are otherwise resolved. Before merge, please also rebase onto `v2`, bump beyond current `2.5.9`, resolve generated version conflicts, and coordinate the `t244` collision with PR #596.

csotomon:
Round-4 addressed at head 3d4433ec:

**1. `onboarding.fills.ts:35` SESSION_ENDED claim.** Fixed: now reads "`SESSION_STARTED` is emitted on IDE 1.x (via the `SessionStart` v2 hook); `SESSION_ENDED` is NOT emitted on 1.x (the IDE's `Stop` trigger is turn-scoped, not session-scoped, so there is no safe registration for it)." Propagates to `dist/kiro-ide/AGENTS.md` via the packager.

**2. `kiro-ide-hook-payload.md` rewrite.** The doc now distinguishes both channels up front: pre-1.0 (USER_PROMPT env, stdin dead) vs IDE 1.x (stdin with snake_case JSON, USER_PROMPT empty). Explains the adapter's race+fallback strategy. The per-event section's "stdin is dead" line updated to "stdin is dead on pre-1.0; on 1.x it carries the payload."

**3. Rebase + version.** Rebased onto v2 at 2.5.9 (from #f32bad80); bumped to **2.5.10** explicitly in all three places. t68 passes.

**4. t244 collision with PR #596.** Renamed to **t245-kiro-ide-hook-registrations.test.ts** (PR #596 takes t244 for scope-matrix-doc-sync).

**Verification:**
- Drift guard: 5/5 OK
- Typecheck: green
- t68 + t245: 36/36 pass

csotomon:
**Update:** force-pushed `fbff8c5a` (amends the round-4 commit at `3d4433ec`). Additional doc fix found in the sweep:

- `docs/guide/harnesses/kiro-ide.md` "How hooks work" section: rewrote the context-channel description to present 1.x (stdin) as primary and pre-1.0 (USER_PROMPT) as legacy, rather than the old framing that presented USER_PROMPT as universal with 1.x as a "known boundary" callout. Now matches the structure of `kiro-ide-hook-payload.md`.
- `docs/reference/kiro-ide-hook-payload.md` "Consequences" section: `session-end` reference updated to note it has no v2 registration.

All other claims verified clean in an exhaustive sweep of `SESSION_ENDED`, `cp -r`, `UserPromptSubmit`/`SessionStart`, `stdin`/`USER_PROMPT`, and hook counts across the full tree. Drift guard + typecheck + t68 + t245 still green.

leandrodamascena (changes requested):
One documentation inconsistency remains:

`docs/reference/kiro-ide-hook-payload.md:12-15,60` says the adapter reads IDE 1.x stdin and normalizes snake_case payloads. The implementation still discards `input` and reads only `USER_PROMPT` (`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:71,78-89,405-410`). A direct 1.x-shaped stdin probe produced no `SUBAGENT_COMPLETED` event.

Please document the current limitation consistently with the Kiro IDE guide and changelog, or implement and test the stated dual-channel behavior.

Non-blocking: `tests/unit/t245-kiro-ide-hook-registrations.test.ts:1` still identifies itself as `t244`.

The previous findings are otherwise resolved; focused tests and `bun run check` pass. Once this final inconsistency is fixed, we will proceed with the merge.

csotomon:
Fixed at head 1bfe9505:

**The inconsistency:** `kiro-ide-hook-payload.md` and `kiro-ide.md` claimed the adapter reads stdin and normalizes both channels. That's the follow-up PR (#615), not this one. The current adapter reads ONLY `USER_PROMPT`.

**What changed:**

- `docs/reference/kiro-ide-hook-payload.md`: rewritten to clearly state "the shipped adapter reads ONLY the `USER_PROMPT` environment variable (the pre-1.0 channel). It never reads stdin." The 1.x stdin section is retained as an observed-payload reference (for the follow-up implementation) but now carries an explicit "> Note: The adapter does not yet consume the 1.x stdin channel" callout. The "Critical limitations" item 2 now reads "Stdin is not read by the current adapter." The Consequences section explicitly says audit-and-sensors and log-subagent no-op on 1.x.

- `docs/guide/harnesses/kiro-ide.md`: the "How hooks work" section no longer presents the dual-channel behavior. It says: "The adapter reads hook context from the USER_PROMPT environment variable (the pre-1.0 channel). On IDE 1.x, where context arrives on stdin and USER_PROMPT is empty, the two payload-dependent targets fire but no-op with a visible hook drop. The stdin context channel is planned as a follow-up enhancement."

- Both docs are now consistent with each other, with the CHANGELOG ("Known boundary: on IDE 1.x audit-and-sensors and log-subagent fire but no-op"), and with the implementation (`aidlc-kiro-adapter.ts` lines 71, 78-89, 405-410 which only read `USER_PROMPT`).

**Non-blocking:** `t245` comment header fixed (was still identifying as `t244`).

**Verification:** drift guard 5/5, typecheck green, t68 + t245 pass.

leandrodamascena (approved):
The final documentation inconsistency is resolved. The Kiro IDE payload reference now accurately describes the current `USER_PROMPT`-only adapter behavior and the IDE 1.x stdin limitation. The `t245` identifier is also consistent.

Validation passed:

- 75 focused tests
- Packaging parity
- Typecheck
- Lint
- Conflict-free merge against current `v2`

All previous findings are resolved. Approved and ready to merge.

PR #3, opened 2025-11-22
Reorganized rules files to take up minimal space in context window

raj-jain-aws:
*Description of changes:*

Reorganized rules files layout such that we load only `core-workflow.md` as persistent additional context. Rules details files are dynamically loaded as needed.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1, opened 2025-11-20
Kiro CLI support and multi-platform compatible architecture


scoropeza:
## Issue Type
Enhancement / Feature Request

### Current Limitation
The AI-DLC workflows currently support Amazon Q Developer IDE plugin, and the now **old** Q CLI with rules structured specifically for the `.amazonq` directory. This is not compatible with the recent **update and rebrand of Amazon Q CLI to Kiro CLI**.

On top of it, a few files mention specific tool names for Amazon Q that just got renamed: 

Source: https://kiro.dev/docs/cli/migrating-from-q/#one-time-migration-during-kiro-cli-installation

```text
Changes -> fs_read to read, fs_write to write , use_aws to aws, execute_bash to shell, report_issue to report
```

I propose using tool/platform agnostic instructions and avoid mentioning tools/commands that may not be available in other AI coding assistant platforms, to ensure maximum flexibility.

### Proposed Enhancement
Restructure the repository to support multiple AI coding assistant platforms while maintaining the same AI-DLC methodology:

1. Mention specific instructions that support **Kiro CLI** - Using steering files (`.kiro/steering`)
2. Keep instructions that support **Amazon Q Developer IDE** - Existing support (`.amazonq/rules`)
3. Minor upgrades to some prompts to support **future platforms** - Extensible architecture

#### Structural Changes
- Rename `amazonq/` directory to `aidlc-rules/` (platform-agnostic)
- Users copy the same rule files to platform-specific locations:
  - Kiro CLI: `.kiro/steering/aidlc-rules/`
  - Amazon Q: `.amazonq/rules/aidlc-rules/`

#### Documentation Updates
- Add separate installation sections for each platform
- Include platform-specific verification steps
- Add screenshots showing rules loaded in both platforms
- Update prerequisites to list all supported platforms

#### Rule Content Updates
- Make tool references platform-agnostic (e.g., "use file writing tools" instead of "use fsWrite")
- Ensure all rules work across different AI assistants
- Maintain identical AI-DLC methodology across platforms

### Changes Summary
- Restructure directory from `amazonq/` to `aidlc-rules/`
- Update README.md with multi-platform installation instructions
- Add Kiro CLI steering file support documentation
- Update `core-workflow.md` to use platform-agnostic language
- Add platform-specific verification screenshots
- Update `content-validation.md` to remove tool-specific references

### Testing
- Verified rules load correctly in Kiro CLI
[...]

PR #616, opened 2026-07-20
feat: bounded Build & Test → Code Generation failure loop-back (2.6.20)

### LINKED ISSUE #611: [Feature]: bounded Build & Test → Code Generation failure loop-back — priced fixes, max-3 ledger, halt-and-ask
### Description

Build and Test's On-failure handling dead-ends when the root cause is upstream. Today (`core/aidlc-common/stages/construction/build-and-test.md`, Step 10): read the error, fix, re-run, and "if unable to fix after 2 attempts, log the failure in test-results.md and present the issue to the user at the approval gate". That is right for failures inside the stage's own remit (test config, build scripts, environment), but when the diagnosis points at the **generated code or an approach chosen at code-generation** — a library/version, container image, instance type, algorithm, flag — there is no sanctioned path back: the workflow either presents a failed gate or the agent writes the approach off, typically on an unpriced effort assumption.

Proposal: a bounded failure-escalation ladder with a sanctioned Build-and-Test → Code Generation loop-back.

1. **In-stage fix (max 2 attempts)** — existing behavior, for root causes inside this stage's remit.
2. **Classify and price** — when in-stage attempts are exhausted OR the diagnosis points upstream: decide whether the root cause lies in generated code / a code-generation approach choice; look for an identifiable fix in a swappable dimension (newer image, driver, wheel index, a CLI flag) and PRICE it — effort, cost, risk. Never declare a feasible path out of scope on an UNPRICED effort assumption.
3. **Autonomous bounded loop-back** — if `Construction Autonomy Mode: autonomous`, a priced fix exists, and fewer than 3 loop-backs have been used: record the diagnosis + priced fix plan, jump back to code-generation, apply the fix to the unit(s) the diagnosis names, and replay forward. The failed run's gate is NOT presented; its §13 learnings ritual defers to the eventual passing run.
4. **Halt-and-ask** — gated mode, bound exhausted, or no identifiable fix: log the failure and present a priced 3-option question (Retry with fix [price] / Accept failure and proceed to the gate / Abort). Giving up is the human's decision to make, never the agent's. A human-approved retry counts an entry but the human may override the bound.

**The loop-back counter is a crash-safe artifact ledger**, not session state: an append-only `## Loop-Back Log` in test-results.md, one `### Loop-back N — <timestamp>` entry (Diagnosis / Root-cause stage / Planned fix / Price) per attempt; the entry count IS the bound (max 3 per intent). Chosen over counting `STAGE_JUMPED` audit rows because it survives the backward jump (jumps reset checkboxes, never artifacts), is colocated with the diagnosis it must carry anyway, and is readable at the final gate; the `STAGE_JUMPED` rows remain the deterministic audit cross-check. Human-directed backward jumps do not count against the bound.

### Use Case

Trainable failure shape from the field (12-unit ML intent, Construction at 2.2.17): build-and-test failures whose fix is a swappable code-generation dimension — a training-container image lacking a driver, a wheel that needs a different index, an instance type without the required accelerator. Each is a priced, mechanical repair of an already-approved code-generation shape, not a new design decision — yet the protocol's only options were "agent silently exceeds its 2-attempt remit" or "present a failed gate and make the human re-drive the jump by hand". Under an autonomy grant the workflow should repair and replay within an explicit bound; when gated it should present the priced options rather than an unpriced shrug.

The machinery already exists upstream — this is protocol prose, not new tooling: `aidlc-jump.ts execute --target <slug> --direction backward` (backward jump resets the target + downstream stages to pending, leaves artifacts, emits the canonical `STAGE_JUMPED` event), and the Artifact Re-use ritual (Keep/Modify/Redo) already governs the forward replay over existing artifacts.

### Version

v2 (alpha)

### Area

Construction phase rules

### Additional Context

Implementation sketch (verified against v2 @ 2.5.0, commit 3f16023):

- `core/aidlc-common/stages/construction/build-and-test.md` — Step 10 On-failure becomes the 4-rung ladder; the artifact section gains the `## Loop-Back Log` shape (append-only; survives re-runs — Modify, never Redo, on loop-back re-entry since Redo would erase the ledger). Single-stage runs (`/aidlc --single build-and-test`) stop at rung 2 — there is no main-workflow position to move; log the priced options at that run's gate.
- `core/aidlc-common/protocols/stage-protocol.md` — new §1 subsection "Build-and-Test failure loop-back": the sanctioned exception to the NO EMERGENT BEHAVIOR RULE (like the existing revision escape hatch) and to checklist item 5's "complete the current stage before jumping" (a failed run is deliberately left in-flight; learnings defer to the passing run); the autonomous procedure (append ledger entry → execute the jump → replay); the halt-and-ask question template with priced option descriptions; the Artifact Re-use interaction (under an autonomous loop-back the 3-option re-use question is not presented — deterministically Modify for the unit(s) the fix names, Keep for the rest, Modify for build-and-test itself, every auto-decision still audited via `reuse-artifact`).
- `core/aidlc-common/protocols/stage-protocol-recovery.md` — crash-resume note: a latest `## Loop-Back Log` entry with a planned fix but no matching post-entry `STAGE_JUMPED` (Target: code-generation) means the session died between logging and jumping — re-execute the jump, don't re-diagnose; on any resume the loop-back count is the ledger's row count, not zero.
- `harness/{claude,kiro,kiro-ide,codex,opencode}/skills/aidlc/SKILL.md` — one-sentence exception on the "STAGE RITUAL IS ATOMIC" Key Principles bullet.
- Docs (`docs/reference/04-stages/construction.md` Build and Test section), a prose-pin test for the ladder + ledger, CHANGELOG/version, dist regen. Additive to the existing §1 halt-and-ask block (t76 pins its content — insert a sibling subsection, don't reword).

Design decisions to settle in review:
- **Jump routing**: the field version had the conductor call `aidlc-jump.ts execute` directly; upstream's cleaner path is via the engine (`next --stage code-generation` emits the validated jump print). Recommend the engine path.
- **Autonomy semantics**: rung 3 keys off the existing `Construction Autonomy Mode: autonomous` grant (the bolt ladder's answer), reading the loop-back as a repair of the already-approved code-generation shape — not a new autonomy inference (checklist item 6). The replayed code-generation gate is auto-approved under that grant with an explicit `--user-input "Autonomous loop-back N per stage-protocol §1"` marker. If maintainers prefer a fresh explicit grant, the ladder degrades gracefully (rung 3 collapses into rung 4's Retry option).
- **Swarm interaction**: on a loop-back replay where the engine emits `invoke-swarm`, the jump's fresh `STAGE_STARTED` floors the convergence ledger — all units re-dispatch by default. The protocol should instruct the cheap path: after `prepare`, run `check <unit> --check-cmd ...` on every unit first and claim already-green units at `finalize` without a worker turn; dispatch workers only for units the planned fix targets or that fail the check. (A unit-scoped redo is a possible follow-up; v1 keeps whole-stage replay.)

Field predecessor: a 2.2.17 installed-copy implementation of exactly this ladder; its ledger shipped in the field intent's test-results.md with zero entries — rungs 1–2 sufficed on that intent, and the bound exists for the intents where they don't.

jstrunk:
Closes #611.

## What

Gives Build & Test a sanctioned, bounded path back to Code Generation when the root cause of a failure lies in the generated code or an approach chosen there — replacing today's dead end ("2 attempts, then log and present the failed gate").

**The 4-rung failure-escalation ladder** (`build-and-test.md` Step 10):

1. **In-stage fix (max 2 attempts)** — existing behavior, for root causes in this stage's own remit.
2. **Classify and price** — root cause in generated code / a code-generation approach choice? Find a fix in a swappable dimension (newer image, driver, wheel index, a flag) and PRICE it (effort/cost/risk). Never declare a feasible path out of scope on an UNPRICED effort assumption.
3. **Autonomous bounded loop-back** — under `Construction Autonomy Mode: autonomous`, with a priced fix and fewer than 3 entries in the ledger: record the diagnosis + plan, jump back to code-generation via the engine, apply the fix to the unit(s) the diagnosis names, replay forward. The failed run's gate is NOT presented; its learnings ritual defers to the passing run.
4. **Halt-and-ask** — gated mode, bound exhausted, or no identifiable fix: a priced 3-option question (Retry with fix [price] / Accept failure / Abort). Giving up is the human's decision, never the agent's.

**The counter is a crash-safe artifact ledger**: an append-only `## Loop-Back Log` in test-results.md — the entry count IS the bound (max 3 per intent). It survives the backward jump (jumps reset checkboxes, never artifacts), is colocated with the diagnosis, and is readable at the final gate; `STAGE_JUMPED` audit rows remain the deterministic cross-check. Human-directed jumps don't count.

## Design decisions (vs. the field predecessor)

- **Engine-routed jump**: the procedure runs `aidlc-orchestrate.ts next --stage code-generation` and executes the exact `aidlc-jump.ts execute` command the engine prints — never a hand-composed jump call.
- **Autonomy semantics**: rung 3 keys off the existing `Construction Autonomy Mode: autonomous` grant; the replayed code-generation gate is auto-approved with an explicit `--user-input "Autonomous loop-back N per stage-protocol §1"` marker — a repair of the approved shape, not a new autonomy inference (checklist item 6 intact).
- **Swarm interaction** (rewritten for 2.5.0's run-floor-stamped convergence ledger): on a loop-back replay that re-emits `invoke-swarm`, run `check` on every unit first and claim already-green units at `finalize` without a worker turn; dispatch workers only for units the planned fix targets or that fail the check.
- **Artifact Re-use under autonomy**: the 3-option re-use question is not presented — deterministically Modify for targeted units, Keep for the rest, Modify for build-and-test itself (Redo forbidden — it would erase the ledger); every auto-decision still audited via `reuse-artifact`.
- **`--single` runs** stop at rung 2 (no main-workflow position to move); priced options are logged and presented at that run's gate.

## Changes

- `core/aidlc-common/stages/construction/build-and-test.md` — Step 10 ladder + `## Loop-Back Log` artifact shape.
- `core/aidlc-common/protocols/stage-protocol.md` — new §1 subsection "Build-and-Test failure loop-back (3.6 → 3.5)" (purely additive sibling of the pinned Halt-and-ask block; t34/t76 green), carve-out sentence on the NO EMERGENT BEHAVIOR RULE, EXCEPTION sentence on checklist item 5, Artifact Re-use auto-decision rule.
- `core/aidlc-common/protocols/stage-protocol-recovery.md` — crash-resume bullet (logged fix without a matching `STAGE_JUMPED` ⇒ re-execute the jump, don't re-diagnose; resume count = ledger rows).
- `harness/{claude,kiro,kiro-ide,codex,opencode}/skills/aidlc/SKILL.md` — exception on the STAGE RITUAL IS ATOMIC principle (codex had no such bullet; it gains the full bullet with the exception).
- New `tests/integration/t245-build-test-loopback-prose.test.ts` — 24 tests / 73 assertions pinning the ladder, §1 subsection, ledger paragraph, engine routing, swarm path, halt-and-ask template, recovery bullet, and the SKILL exception across authored + dist copies of all five harnesses.
- Docs: `docs/reference/04-stages/construction.md`, `04-stage-protocol.md`, `03-orchestrator.md`.
- Version 2.5.4 + CHANGELOG entry + README badge; all dists regenerated.

## Testing

- Smoke + unit tiers: **173 files, 4256 assertions, 0 failures**.
- New t245: 24 pass / 0 fail. Pinned-prose suites t34 / t76 / t44 / t87 / t68: 129 pass / 0 fail. Protocol readers t35–t37, t43, t47: 154 pass / 0 fail. Skill-parity/drift t181 / t55 / t239 / t129: 34 pass / 0 fail.
- `bun scripts/package.ts --check` and the coverage-registry check both clean.

## Notes for review

- Version claims 2.5.4 assuming companion PRs #612 (2.5.1) and #613 (2.5.2) land first; will rebase + re-bump per the CHANGELOG conflict policy as needed.
- No engine code changes — the jump tool, `STAGE_JUMPED` event, checkbox resets, and Artifact Re-use replay all exist today; this PR is protocol prose + stage definition + tests.
- A unit-scoped redo (avoiding whole-stage replay) is a possible follow-up noted in #611.

---

## Maintainer integration update (2026-08-19)

The maintainer-led integration discussed above was pushed to this branch as a fast-forward: the three original commits are untouched; three commits were appended (two v2 sync merges and the integration commit).

- **Both findings from the latest review are resolved.** Re-entry after the backward jump is now settlement-aware with two explicit routes (artifact-only workflows may take the all-covered `gate: true` fast path; sticky receipt-mode workflows re-emit per-Unit work), and on BOTH routes the planned fix and deterministic Modify/Keep re-use decisions land BEFORE the gate, with a fresh current-attempt `REVIEW_COMPLETED` required for every applicable Unit (`STAGE_JUMPED` invalidates prior receipts; the engine refuses approval without replacements). The swarm check-first cheap path still skips builder turns for already-green Units, but every claimed Unit gets a fresh terminal reviewer receipt in its fresh worktree before `finalize --claimed`; stale worktrees are discarded, never adopted.
- **Plan approval on replay**: the recorded Plan Approval answer stays authoritative (the loop-back repairs the approved plan; the delta goes in the ledger). In gated mode the human's "Retry with fix" answer IS the re-approval, carried via `--user-input` on the replayed approval report.
- **New behavioral test** `tests/integration/t304-loopback-review-receipt-replay.test.ts` drives the real engine tools end to end: approve code-generation, execute the engine-printed jump command, hit the settle gate, get refused ("2 of 2 applicable units have no fresh recorded review"), record fresh receipts, approve. The prose pins moved to `t305-build-test-loopback-prose.test.ts` (40 assertions, all seven harnesses, authored + dist).
- **v2 sync and re-slot**: merged v2 2.6.17; version is now **2.6.20** (supersedes the 2.5.x numbers in the description above); tests renumbered t304/t305 to clear slot collisions; t304's fixture adapted to State Version 8 and code-generation's current produces set. The SKILL exception now ships in all seven harnesses, including Cursor and GitHub Copilot.
- **Verification**: `package.ts --check`, coverage-registry `--check`, typecheck + biome all clean. Full sliced gate green: smoke+unit 234 files / 6754 assertions / 0 failures; integration 106 / 1464 / 0 (with `AIDLC_CLAUDE_SDK_LIVE=1`); deterministic e2e 31 / 144 / 0; live Claude TUI, Kiro ACP, Kiro TUI, and Codex exec slices green (three load-induced flakes in v2-side tests re-ran green alone).

### Fix-round update (2026-08-21)

The 2026-08-19 review findings are addressed at `c875078` (fast-forward; nothing rewritten):

- **P1 fixed in the engine** (this supersedes the "No engine code changes" note in the original description): `autonomousSwarmOwnsStage()` in `aidlc-state.ts` now mirrors the orchestrator's unit-major emission suppression as well as the candidate predicate, so the autonomous unit-major replay can mint interactive Unit receipts instead of `unit start` being refused as swarm-owned. New `t307-loopback-unitmajor-replay` drives the route end to end (start, fresh review, complete, approval) and pins the stage-major refusal boundary so the carve-out is not widened.
- **P2**: rung 1 of the failure ladder is limited to Build and Test's own test configuration, build scripts, and environment setup; generated source and test-code defects route through the ledgered loop-back regardless of size. The bounded unit-scoped repair path from the thread will be tracked as a separate issue.
- **P2**: crash recovery gains the third replay branch: a replay that re-emits `invoke-swarm` follows the construction module's Swarm interaction procedure.
- **v2 sync**: merged v2 2.6.18 (#767); the loop-back contract now lives in the new conditional construction protocol module (`aidlc-common/protocols/stage-protocol-construction.md`), in its pre-split position beside the Bolt halt-and-ask, with every cross-reference re-pointed. Version stays 2.6.20, and the CHANGELOG upgrade note now states the `tools/` update.
- **Verification**: package parity, coverage registry, typecheck, and lint clean; smoke+unit 236 files / 0 failures; integration 107 / 0; deterministic e2e 32 / 0 including `t301-express-scope-routing` on the merged tree; targeted t68/t304/t305/t307 slice 52/52.


apackeer:
## Review: bounded Build & Test -> Code Generation loop-back

Reviewed at head 579dd58. Since the PR is protocol prose + tests with no engine changes, every engine claim in the prose was verified against the tools it names, and they hold up well:

- The engine-routed jump is real: `next --stage code-generation` resolves direction via `aidlc-jump.ts resolve` and prints exactly the `execute --target ... --direction backward --scope ...` command the prose describes (`core/tools/aidlc-orchestrate.ts:2712-2714`).
- `STAGE_JUMPED` carries a literal `Target:` field (`core/tools/aidlc-jump.ts:445-451`), so the recovery bullet's "no matching STAGE_JUMPED (Target: code-generation)" check is greppable as written.
- Backward jumps touch only the state file: `handleExecute` flips checkboxes and state fields and writes audit rows, with no artifact deletion anywhere in the file, so the Loop-Back Log genuinely survives the jump (`core/tools/aidlc-jump.ts:313-332`, `:463`).
- The autonomous auto-approval clears the human-presence guard through the existing autonomy carve-out (`core/tools/aidlc-state.ts:1784-1786`), and `Construction Autonomy Mode` is a real, tool-read field (`core/tools/aidlc-lib.ts:2658-2662`), so rung 3 needs no new engine behavior.
- The swarm run-floor story matches the code: `SWARM_UNIT_CONVERGED` rows are stamped with the stage's latest main-workflow `STAGE_STARTED` (`core/tools/aidlc-swarm.ts:283-303`, `core/tools/aidlc-lib.ts:3451-3465`), the jump mints a fresh `STAGE_STARTED`, and `finalize` re-verifies every claimed unit so the no-worker-turn claim path is safe at the tool level (`core/tools/aidlc-swarm.ts:565-601`).

Deterministic tiers reproduce the PR's numbers in my checkout: smoke+unit 173 files / 4256 assertions green, t245 24/24, the pinned-prose suites (t34/t76/t44/t87/t68) and protocol readers green, `bun scripts/package.ts --check` and the coverage-registry check clean, typecheck + biome clean.

Findings, in severity order.

### Must-fix

**1. Gated "Retry with fix" collides with the engine's settle-to-gate routing, so the fix is never applied on the replay.** Because jumps preserve artifacts, every code-generation unit is still "covered" on disk after the backward jump. The off-swarm per-unit router therefore finds no uncovered unit and settles straight to a `gate: true` directive on the last unit (`core/tools/aidlc-orchestrate.ts:2343-2361`), and every harness SKILL instructs that `gate: true` on a per-unit stage means "every unit is already built, so run the learnings ritual and present the single approval gate" (for example `harness/codex/skills/aidlc/SKILL.md:80`) - no stage body, so the Artifact Re-use question never fires and the planned fix is never applied. The loop-back subsection's step 3 does say to apply the fix on re-entry, but the deterministic directive pushes the conductor the other way. Result if the conductor follows the directive: the gate is presented un-fixed, the human approves, build-and-test fails again, and each retry burns a ledger entry until the bound exhausts. The autonomous path has the same routing but its Artifact Re-use tail explicitly overrides it; the gated path has nothing. Needs either engine-side coverage invalidation on a loop-back jump, or explicit prose in the loop-back subsection stating that the re-entry directive arrives as `gate: true` and the planned fix must be applied (Modify on the targeted units) before that gate is presented - plus a test pinning whichever is chosen.

**2. Version slot is stale and the PR is CONFLICTING.** The body's assumption (companion PRs taking 2.5.1/2.5.2) is already broken: v2 is at 2.5.2 today (#595 merged as 2.5.1, #576 as 2.5.2). Per the CHANGELOG conflict policy this needs a rebase onto v2 plus a re-bump with the `## [2.5.4]` heading renamed to match; t68 will catch a miss.

**3. The PR falsifies a sentence it does not update.** `core/aidlc-common/protocols/stage-protocol.md:128` still says the Bolt halt-and-ask "is the one case where `autonomous` mode stops to consult the user." Rung 4 is now a second such case (autonomous run, bound exhausted or no identifiable fix). One-sentence fix; I checked that no test pins that sentence.

### Should-fix

**4. The no-fix branch cannot render the pinned template honestly.** Rung 4 explicitly fires when "no identifiable fix exists" (`core/aidlc-common/stages/construction/build-and-test.md:139-141`), but the only halt-and-ask template has a mandatory `Candidate fix: [fix] - estimated price: [...]` slot and a "Retry with fix" option (`core/aidlc-common/protocols/stage-protocol.md:214-224`), and t245 pins the exact prompt string. A no-fix variant (Accept failure / Abort, or Retry-as-is) is needed, with matching t245 assertions.

**5. `--single` promises a gate that isolated runs forbid.** `core/aidlc-common/stages/construction/build-and-test.md:149` says priced options are "presented at this run's gate", but the isolated stage-runner branch in every harness SKILL says "Do not ... present a workflow gate ... present the isolated-run summary and STOP" (`harness/claude/skills/aidlc/SKILL.md:50` and the same bullet in the other four). Reword to "in the isolated-run summary"; the same phrase also appears at `docs/reference/04-stages/construction.md:904` and in the CHANGELOG bullet. t245's single-run test pins "Stop at rung 2" but not the gate phrase, so the reword is test-safe.

**6. The crash-resume bullet is in the wrong recovery section.** It was appended under "Stage re-run" (`core/aidlc-common/protocols/stage-protocol-recovery.md:119-125`), which covers "user requested changes after approval". The scenario it describes (session died between logging and jumping) is detected on session resume, a different section (`stage-protocol-recovery.md:44`). A resuming conductor reading the resume path will not encounter it; move or cross-reference it under "Session resume". t245's recovery assertions are placement-agnostic.

**7. The swarm cheap path hides two unenforced premises.** (a) `prepare` hard-errors on replay if the prior attempt's worktrees or `bolt-<slug>` branches still exist (`core/tools/aidlc-worktree.ts:212-220`), and nothing in `aidlc-swarm.ts` ever discards them - a crash or halt-and-ask mid-swarm leaves them in place, so the first tool call of the prescribed replay can fail with an error the prose never mentions. (b) The premise "the prior run's code is in the base the worktrees forked from" is not guaranteed: swarm finalize merges only AIDLC metadata, never generated source (`core/tools/aidlc-orchestrate.ts:805-812`, `core/tools/aidlc-bolt.ts:412-462`); if the git code merge never ran on the failed attempt, every `check` is red and the cheap path silently degrades to full re-dispatch. A sentence handling the prepare collision (discard or adopt existing worktrees first) and softening the already-green premise would cover both.

### Nits

- The Artifact Re-use tail's abbreviated audit call (`aidlc-state.ts reuse-artifact --decision <keep|modify>`) omits the required positional slug and `--artifacts` flag, so run literally it errors (`core/tools/aidlc-state.ts:2765-2772`). Conductors execute this prose; worth spelling out the full form or pointing at the canonical block above it.
- The commit message says t243; the file is t245. Also t244 is unclaimed on v2 and in the open PRs I checked, so the gap will confuse the next slot hunt - consider t244 or a note.
- The max-3 bound, append-only rule, and Redo prohibition are all prose-only (nothing deterministic refuses a 4th backward jump or counts the ledger). Consistent with the PR's declared scope, but a cheap follow-up would be `aidlc-jump.ts execute` refusing a backward build-and-test -> code-generation jump when the ledger already holds 3 entries. Could ride the unit-scoped-redo follow-up already noted in the issue.

Happy to re-review after the rebase; the engine-claim verification above should carry over unchanged since no tool code moves.

core/aidlc-common/protocols/stage-protocol.md
```diff
@@ -148,6 +148,91 @@ options:
...
+(`finalize` re-verifies every claimed unit, so the claim is safe); dispatch
+workers only for the unit(s) the Loop-Back Log's planned fix targets or that
+fail the check. On the inline per-unit path all artifacts still exist, so the
+fix is applied through the Artifact Re-use ritual per unit.
+
+**Halt-and-ask (gated or unset mode, bound exhausted, or no identifiable
+fix):**
+
+```question
+prompt: "Build and Test failed: [short error]. Root cause: [diagnosis]. Candidate fix: [fix] — estimated price: [effort/cost/risk]. Loop-backs used: [N]/3. How would you like to proceed?"
```
apackeer: UX: “price” normally implies a monetary value, but this prompt defines it as a combined estimate of effort, cost, and risk. Users may reasonably look for a dollar amount or misread what is being compared. Please rename this to **estimated impact** and expose the dimensions separately, for example: `Estimated impact — effort: [...]; financial cost: [...]; risk: [...]`. The related `price` / `priced` / `unpriced` wording and the Loop-Back Log’s `Price` field should be changed consistently.

jstrunk:
Thanks for the thorough review — addressed all of it. Summary below, plus the rebase.

**Rebase + re-bump:** rebased onto current `v2` (9f91454). Bookkeeping conflicts resolved: version bumped 2.5.4 → **2.5.16** (pre-assigned to avoid colliding with sibling PRs rebasing in parallel), CHANGELOG heading/date and README badge updated to match. New commit: `fdf801a`.

**Test rename:** `tests/integration/t245-build-test-loopback-prose.test.ts` → `t250-build-test-loopback-prose.test.ts` (`t245` was claimed by unrelated merged work in the interim; `t250` is the pre-assigned collision-free slot). This also resolves the commit-message/file-number mismatch nit — the commit message and file now agree.

**Must-fix #1 (gated "Retry with fix" collides with settle-to-gate routing):** Confirmed the issue — a backward jump preserves artifacts, so every code-generation unit is still "covered" on disk, and the re-entry `next` call answers with `gate: true` straight to the approval gate, skipping the stage body (and the ordinary Artifact Re-use question) for the gated path exactly as you described. Went with the prose fix, mirroring the autonomous path's existing override: added a new **"Gated failure loop-back"** paragraph in `stage-protocol.md`'s Artifact Re-use section, sibling to "Autonomous failure loop-back", stating the same deterministic Modify/Keep decisions must be applied via `aidlc-state.ts reuse-artifact` **before** that gate is presented. Cross-referenced from the halt-and-ask's "Retry with fix" description. Pinned in `t250`.

**Must-fix #2 (version slot):** Done — 2.5.16, per above.

**Must-fix #3 (Bolt halt-and-ask "one case" sentence):** Updated `stage-protocol.md`'s Bolt halt-and-ask sentence to name rung 4 (bound exhausted / no identifiable fix, under `autonomous`) as a second case where `autonomous` mode stops to consult the user, alongside the Bolt halt-and-ask.

**Should-fix #4 (no-fix halt-and-ask variant):** Added a distinct no-fix template in `stage-protocol.md` — Accept failure / Abort only, no `Candidate fix` slot, no "Retry with fix" option (nothing to invent a fix to retry with). The stage body now picks the variant based on whether rung 2 actually produced a priced candidate fix. Pinned in `t250`.

**Should-fix #5 ("presented at this run's gate" vs. isolated runs):** Reworded to "in this/that run's isolated-run summary" in `build-and-test.md`, `docs/reference/04-stages/construction.md`, and the CHANGELOG bullet — matching the existing SKILL wording ("present the isolated-run summary and STOP"). Confirmed the phrase doesn't appear in any harness SKILL.md, so no changes needed there.

**Should-fix #6 (crash-resume bullet in the wrong section):** Moved the logged-but-not-jumped detection bullet from "Stage re-run" (wrong scenario — that's for user-requested changes after approval) to "Session resume" (the actual scenario — session died mid-jump). Left a short cross-reference in "Stage re-run" so a reader landing there isn't stranded.

**Should-fix #7 (swarm cheap-path unenforced premises):** Added both sentences to the swarm-interaction paragraph: (a) check for and discard/adopt a prior attempt's leftover worktrees/`bolt-<slug>` branches before calling `prepare` (which hard-errors on collision), and (b) softened the "already-green" claim — it holds only once the prior attempt's git code merge actually completed; otherwise every `check` comes back red and the path degrades gracefully to full re-dispatch rather than silently claiming unbuilt units.

**Nit (abbreviated audit call):** Spelled out `aidlc-state.ts reuse-artifact <slug> --decision <keep|modify> --artifacts "<...>"` in full in both the autonomous and new gated paragraphs.

**Nit (t243/t245 mismatch):** Resolved by the rename above.

**Nit (unenforced bound/append-only/no-Redo rules):** Left as-is per your note — consistent with the PR's declared scope.

**Verification:**
- `bun scripts/package.ts` / `--check`: clean, no drift
- `bun tests/gen-coverage-registry.ts --check`: OK
- `bun run check` (package --check + typecheck + lint): clean
- `bash tests/run-tests.sh --smoke --unit`: 178 files, 4396 assertions, 0 failures
- `t250-build-test-loopback-prose.test.ts` (renamed + expanded with new pinning assertions): 36 pass
- `t68-version-changelog-sync.test.ts`: 7 pass (2.5.16 verified in sync across `aidlc-version.ts`, CHANGELOG, README)

Force-pushed to `feat/611-build-test-loopback` at `fdf801a`. PR title updated to reflect 2.5.16.

apackeer:
## Re-review at fdf801a

All three must-fix and all four should-fix items from the round-1 comment body are verified fixed at this head, and the deterministic gates reproduce green in my checkout. One round-1 item is still open, though: the inline review comment on the "price" wording was not addressed, so I'm holding approval on that one item (details at the end). Everything else is done.

**Must-fix 1 (gated "Retry with fix" vs settle-to-gate routing)** - fixed as asked, prose route. The new "Gated failure loop-back" paragraph sits as a sibling of the autonomous override in stage-protocol.md's Artifact Re-use tail, states the `gate: true` re-entry mechanism explicitly ("Do not accept that directive at face value"), and orders the fix BEFORE the gate; the halt-and-ask's "Retry with fix" prose cross-references it. I re-confirmed the underlying routing on the rebased head (the covered-units settle path still emits `gate: true` on the last unit, `core/tools/aidlc-orchestrate.ts:2296-2312`), so the prose now matches what the engine actually does on both paths. Pinned by two dedicated t250 tests.

**Must-fix 2 (version slot)** - fixed. 2.5.16 across `aidlc-version.ts`, CHANGELOG heading, README badge; t68 green; `scripts/ci-changelog-guard.ts origin/v2` reports all 137 base entries preserved + 1 new. The pre-assignment also means this PR survives sibling merges below 2.5.16 without another re-bump, which is the right trade. One cosmetic consequence: 2.5.12-2.5.15 will read as gaps in the CHANGELOG until (unless) the sibling PRs claiming them land; nothing enforces against that.

**Must-fix 3 ("the one case" sentence)** - fixed; no "one case where" phrasing remains anywhere in core/docs/harness, and the replacement names rung 4 precisely (bound exhausted or no identifiable fix, under `autonomous`). Pinned.

**Should-fix 4 (no-fix variant)** - fixed. The no-fix template drops both the `Candidate fix:` slot and "Retry with fix", the chooser sentence keys on rung 2's actual output, and the t250 assertion isolates the question fence so the explanatory prose can still name the omitted option. Nice touch making the omission self-justifying ("a fabricated fix to retry with").

**Should-fix 5 (isolated-run summary)** - fixed; zero "at this/that run's gate" hits remain in core, docs, harness, or CHANGELOG, and the negative assertion in t250 guards regressions.

**Should-fix 6 (crash-resume placement)** - fixed; the detection bullet now lives under "Session resume" (where a resuming conductor actually reads) with a cross-reference left in "Stage re-run". t250 pins the ordering structurally.

**Should-fix 7 (swarm premises)** - fixed. The prepare-collision sentence is executable as written (`aidlc-worktree.ts discard` removes both the worktree and the `bolt-<slug>` branch, and `prepare` still hard-errors on either surviving), and the already-green premise is correctly scoped to "only once that attempt's git code merge actually completed" with the graceful-degradation outcome spelled out.

**Nits** - the audit call is spelled out in full in both override paragraphs (matches the tool's real usage line, `aidlc-state.ts:3010`); the t243/file mismatch is resolved by the t250 rename, and t250 is collision-free among the open PRs I checked (t248 is claimed twice, by #660 and #613 - not this PR's problem, but flagging it for whoever rebases second; t249 is #617's).

**My gate runs at fdf801a** (worktree checkout): smoke+unit 178 files / 4396 assertions green (matches your numbers exactly), t250 36/36, the pinned-prose suites (t34/t44/t76/t87) + t68 + t181 conductor-skill parity green, `bun run check` (package --check + both tsconfigs + biome) clean, coverage-registry `--check` clean.

Two small observations, neither blocking:

1. The "Gated failure loop-back" paragraph names only `Construction Autonomy Mode: gated`, while rung 4's trigger is "gated (or unset)". The mode-agnostic cross-reference from "Retry with fix" routes an unset-mode conductor there anyway, so behavior is covered; widening the paragraph's opening to "gated (or unset)" would make it exact.
2. The codex conductor SKILL gains the whole STAGE RITUAL IS ATOMIC bullet (it had none before this PR). That's the right call - t250's every-harness pin requires it and t181 parity stays green - just noting it's an alignment, not a carry-over.

### Outstanding: the "price" -> "estimated impact" rename (round-1 inline comment)

The inline review comment on stage-protocol.md from round 1 asked to rename "price" (which reads as a monetary value) to **estimated impact** with the dimensions exposed separately (`Estimated impact - effort: [...]; financial cost: [...]; risk: [...]`), changing the `price` / `priced` / `unpriced` wording and the Loop-Back Log's `Price` field consistently. That comment was not addressed at this head: the priced-variant prompt still says `estimated price: [effort/cost/risk]` (stage-protocol.md:227), the ledger shape is still `Diagnosis / Root-cause stage / Planned fix / Price` (stage-protocol.md:178, build-and-test.md:116), rung 2 still says "PRICE it" (build-and-test.md:131), and the wording has since spread to docs/reference/04-stages/construction.md and the CHANGELOG entry, and is now pinned by t250 in 16 assertions. So the rename now touches four surfaces: the two protocol/stage files, the reference doc, the CHANGELOG bullet, and the matching t250 strings.

Holding approval on that one item - everything else is resolved, so it should be a quick mechanical sweep. Happy to approve as soon as it lands.

One process item: CI shows `action_required` on this head - as a fork PR it needs a maintainer's "Approve and run" click before the v2 gate actually executes; worth doing before merge so the required checks land.

apackeer:
Thanks, Jeff. The previous re-review confirmed that you addressed all of the substantive findings from that pass. The one requested change still open is the terminology sweep from `price` to `estimated impact`, with effort, financial cost, and risk shown separately.

@leandrodamascena, could you take the second review pass on this PR? There is one interaction I would particularly like independently verified before final approval: a backward jump invalidates the existing per-unit Code Generation review receipts, while preserved artifacts make the replay route directly to the all-covered settle gate. Does the current replay protocol reliably produce fresh review receipts for every applicable unit before approval?

This is a verification question, not a new redesign request for Jeff. If the existing protocol already covers it, the terminology rename remains the only contributor change. If it exposes a real gap, we should identify the narrowest fix and handle the current-`v2` integration as maintainers. Jeff, please do not spend time rebasing this branch while we complete that check.

jstrunk:
Addressed the one outstanding item — the `price` → **estimated impact** terminology rename — at `9d532d8`, a commit on top of `fdf801a` (no rebase, per your request to leave the branch in place).

The halt-and-ask prompt now exposes the three dimensions separately, in the form you specified:

> Candidate fix: [fix] — estimated impact — effort: [effort]; financial cost: [cost]; risk: [risk].

and the "Retry with fix" option description carries the same expanded form.

The `price` / `priced` / `unpriced` wording and the Loop-Back Log's `Price` field were changed consistently across every surface the terminology reached:

- `core/aidlc-common/protocols/stage-protocol.md` — the halt-and-ask variants (now "impact-estimated variant" / no-fix variant), the ledger `Estimated impact` field, the autonomous procedure, and the give-up-violation sentence.
- `core/aidlc-common/stages/construction/build-and-test.md` — rung 2 is now **Classify and estimate impact** (`ESTIMATE ITS IMPACT — effort, financial cost, risk`), rungs 3/4, the Loop-Back Log field, and the single-stage carve-out.
- `docs/reference/03-orchestrator.md`, `docs/reference/04-stage-protocol.md`, `docs/reference/04-stages/construction.md`.
- The `CHANGELOG.md` 2.5.16 entry.
- `tests/integration/t250-build-test-loopback-prose.test.ts` — the 16 pinned assertions updated to the new wording.
- `dist/<harness>/` regenerated for all five harnesses.

Unrelated `aws-pricing` / pricing-model / price-performance text was left untouched. This is a prose-only refinement of the already-listed 2.5.16 entry, so no version bump (t68 stays green).

**Verification:**
- `bun run check` (package `--check` + typecheck + biome) — clean.
- `bun scripts/package.ts --check` — no drift across all five harness trees.
- `bun tests/gen-coverage-registry.ts --check` — fresh; ratchet held.
- `bash tests/run-tests.sh --smoke --unit` — 178 files, 4396 assertions, 0 failures.
- Targeted: `t250-build-test-loopback-prose.test.ts` 36/36, `t68-version-changelog-sync.test.ts` 7/7.

No functional change; only the terminology. Happy to fold this into the maintainer-led `v2` integration whenever that lands.

leandrodamascena (changes requested):
Re-review at `9d532d8`. The requested “price” to “estimated impact” terminology change is complete and consistent.

**[P1] Recreate every per-unit review receipt before the replayed settle gate**

A backward jump emits `STAGE_JUMPED`, invalidating all prior Code Generation review receipts. Because artifacts remain covered, replay routes directly to a settle directive naming only the final Unit. The loop-back override applies the fix and then presents or auto-approves the gate without requiring fresh §12a reviews for every applicable Unit. The engine consequently refuses completion. Require fresh reviews for every applicable Unit after all fixes/reuse decisions and before the settle gate, with a behavioral receipt test.

Relevant locations: `core/aidlc-common/protocols/stage-protocol.md:189-196`, `:260-267`, and `:1282-1295`.

**[P2] The swarm check-first path cannot finalize already-green Units without fresh reviews**

`prepare` creates a new `BOLT_STARTED` boundary, and `finalize` requires a terminal `REVIEW_COMPLETED` after that boundary. The proposed cheap path claims already-green Units without a worker turn but never instructs a fresh reviewer dispatch, so `finalize` rejects those claims. Require a reviewer pass in each claimed Unit’s worktree after `check`; this can remain “without a builder turn,” but not without review.

Relevant locations: `core/aidlc-common/protocols/stage-protocol.md:201-219` and `core/tools/aidlc-swarm.ts:273-330`.

Verification:

- `t250-build-test-loopback-prose`: 36/36 passed
- `t205-gate-revision-backstop`: 14/14 passed
- `git diff --check`: passed

The branch conflicts with current `v2`; these fixes should be included in the maintainer-led integration rather than requiring another contributor rebase.

apackeer:
Thanks Jeff, the estimated impact rename is confirmed complete. As discussed, maintainers are taking this forward from here: we'll integrate the branch with current v2 and fold in the remaining review items, keeping your commits credited. Nothing further needed on your end; we'll link the integration PR here when it opens.

apackeer:
Pushed the maintainer integration to this branch as a fast-forward; Jeff's commits are untouched, with three commits appended (two v2 sync merges and the integration commit). It resolves both findings from the last review (fresh per-unit reviews are now required on both replay routes and on the swarm check-first path, with a behavioral test, t304, proving the replayed gate refuses stale reviews and unblocks on fresh ones), and re-slots to current v2 as 2.6.20. Details in the description addendum below the original write-up.

@leandrodamascena ready for your re-review.

leandrodamascena (changes requested):
I re-reviewed `b9c654e0` against the prior findings and the current `v2.6.17` base. The previous fresh-review issues are substantially addressed, but one replay route remains blocked.

**P1: Autonomous unit-major replay cannot start Unit receipts**

`core/aidlc-common/protocols/stage-protocol.md:312-315` requires unit-major replay to use the ordinary per-unit walk and re-mint `unit start` / `unit complete` receipts.

The orchestrator correctly suppresses the swarm under unit-major at `core/tools/aidlc-orchestrate.ts:3284-3295`, but `core/tools/aidlc-state.ts:1406-1416` still treats any autonomous subagent per-unit Construction stage as swarm-owned. The prescribed `unit start` call is therefore refused at `core/tools/aidlc-state.ts:901-909`.

I reproduced this with an autonomous, unit-major Code Generation fixture:

```text
Refusing unit start: Construction Autonomy Mode is autonomous. The swarm referee owns per-unit bookkeeping...
```

Please make `autonomousSwarmOwnsStage()` return false when `Construction Iteration` is `unit-major`, matching the orchestrator's routing predicate, and add an executable replay test that drives Unit start through completion under this configuration.

**P2: Small generated-code defects bypass the Code Generation repair and review path**

`core/aidlc-common/stages/construction/build-and-test.md:128-131` includes “small code defects” in Build and Test's own in-stage remit. This conflicts with `core/aidlc-common/protocols/stage-protocol.md:231-234`, which routes failures rooted in generated code through Code Generation.

A direct Build and Test edit bypasses the loop-back ledger, Plan Approval enforcement, Unit receipts, and fresh Code Generation reviewer receipts. Please restrict rung 1 to Build-and-Test-owned configuration/scaffolding and route generated source or test-code defects through rung 2 regardless of size.

**P2: Crash recovery omits the autonomous swarm replay route**

`core/aidlc-common/protocols/stage-protocol-recovery.md:59-62` names only receipt-mode and artifact-only inline recovery after an existing jump. Autonomous stage-major replay instead returns `invoke-swarm`, whose stale-worktree cleanup and check/review/finalize procedure is defined at `core/aidlc-common/protocols/stage-protocol.md:317-341`.

Please add an explicit swarm recovery branch referencing that procedure so a resumed conductor does not treat an `invoke-swarm` replay as one of the two inline settlement routes.

The prior findings are otherwise resolved:

- Fresh per-Unit reviews after a backward jump are enforced by the engine and proven by `t304`.
- The swarm check-first path now requires a fresh reviewer receipt before `finalize --claimed`.
- The estimated-impact terminology and earlier protocol consistency issues are corrected.

Validation at this head:

- `t304` / `t305`: 41 tests passed
- Reviewer recovery and version-sync tests passed
- Package parity, typecheck, lint, and coverage registry passed
- GitHub currently reports no check runs for this head

I recommend keeping **Changes Requested** until the unit-major blocker is fixed.

jstrunk:
> **P2: Small generated-code defects bypass the Code Generation repair and review path**
> 
> `core/aidlc-common/stages/construction/build-and-test.md:128-131` includes “small code defects” in Build and Test's own in-stage remit. This conflicts with `core/aidlc-common/protocols/stage-protocol.md:231-234`, which routes failures rooted in generated code through Code Generation.
> 
> A direct Build and Test edit bypasses the loop-back ledger, Plan Approval enforcement, Unit receipts, and fresh Code Generation reviewer receipts. Please restrict rung 1 to Build-and-Test-owned configuration/scaffolding and route generated source or test-code defects through rung 2 regardless of size.

Maybe this should be its own issue/PR. Specifically for infrastructure development, I find many cases where generated code(cdk or terraform) looks right and may pass unit tests that get run in Code Generation but fail when actually deployed on AWS due to service details that CDK and terraform don't check. Reducing the round trip for the brute force testing and fixing is very important because it may take more than a few passes to actually get it right. It's not for lack of planning. It's just how developing against external resources works. It's a feature of AI assisted development that we shouldn't lose.

I'm okay with removing it from this PR. Where do you think that flexibility could fit?

leandrodamascena:
I agree this is a real and valuable AI-assisted development workflow, especially for CDK/Terraform and external AWS service behavior that static validation cannot reproduce.

My concern is not with preserving a fast repair path. It is that the current “small code defects” wording permits generated code to change outside Code Generation without any replacement review or lifecycle evidence.

For this PR, I recommend removing “small code defects” from rung 1 so the new loop-back contract remains internally consistent.

Then we should track a focused follow-up for a bounded, unit-scoped implementation repair path:

- only for localized implementation fixes discovered during real build/deploy validation;
- no change to the approved plan, dependency choice, architecture, or Testing Contract;
- identify the affected Unit(s);
- cap attempts;
- rerun the failing external validation;
- update relevant summary/traceability evidence;
- require a fresh reviewer result before Build and Test proceeds;
- escalate to the full Code Generation loop-back when the approach changes.

That preserves the short feedback loop without creating an unreviewed code-change bypass. It also aligns with the unit-scoped redo follow-up already noted in #611.

Separately, the P1 autonomous unit-major receipt failure remains reproducible and still needs the predicate fix before approval.

apackeer:
@leandrodamascena im going to bring this one home. we need this in my opinion. ill fix those P1s now

apackeer:
Done - pushed as another fast-forward (b9c654e0a..c875078). The unit-major P1 is fixed at the engine: autonomousSwarmOwnsStage() now carves out Construction Iteration: unit-major, matching the orchestrator's routing, and the new t307-loopback-unitmajor-replay test proves unit start / fresh review / unit complete / approval end to end while pinning the stage-major refusal so the carve-out is not widened. Both P2s are in too: rung 1 is restricted to Build and Test's own configuration and scaffolding (generated source and test-code defects route through the ledgered loop-back regardless of size; the bounded unit-scoped repair path from this thread gets its own issue), and crash recovery now has the invoke-swarm replay branch. The branch is also synced with v2 2.6.18, so the loop-back contract moved into the new construction protocol module in its pre-split position.

@leandrodamascena over to you for re-review.

leandrodamascena (approved):
Re-reviewed `c875078` against the previous findings. **Approved.**

All prior issues are resolved:

- Autonomous unit-major replay can mint fresh Unit lifecycle and review receipts; stage-major remains swarm-owned.
- Generated source and test-code defects now route through the ledgered Code Generation loop-back regardless of size.
- Crash recovery explicitly handles the `invoke-swarm` replay route.
- Both inline and swarm replay paths require fresh current-attempt reviews before settlement.
- The earlier terminology and protocol consistency findings remain resolved.

Validation:

- Focused slice: 52/52 passed (`t68`, `t304`, `t305`, `t307`)
- `bun run check`: passed
- Package parity: passed
- Coverage registry: passed
- `git diff --check`: passed

PR #626, opened 2026-07-22
fix: remove Codex PostCompact hook rejected by the hook output schema (2.5.9)

### LINKED ISSUE #593: [Bug]: Codex PostCompact hook rejected — `error: hook returned invalid PostCompact hook JSON output` (reproducible)
## Symptom (Codex CLI, reproducible)

After a context compaction, the Codex-only PostCompact hook fails every time:

```
• PostCompact hook (failed)
  error: hook returned invalid PostCompact hook JSON output
```

## Analysis

The PostCompact adapter case emits the same `hookSpecificOutput` envelope the SessionStart path uses:

- `harness/codex/hooks/aidlc-codex-adapter.ts:340-353` (`post-compact` case) runs the core `aidlc-session-start.ts` with `source: "compact"`, then calls `wrapContext(r.stdout, "PostCompact")` and writes the result to stdout.
- `wrapContext` (`aidlc-codex-adapter.ts:199-214`) wraps into:
  ```json
  {"hookSpecificOutput":{"hookEventName":"PostCompact","additionalContext":"..."}}
  ```
  The function's own comment says this shape is **"verified live for SessionStart"** — PostCompact was assumed to accept the identical envelope, but Codex rejects it as invalid PostCompact JSON. The two events evidently do **not** share an output contract (likely PostCompact expects a different key, a bare `additionalContext`, or no stdout payload at all).

### Why the test suite didn't catch it

`tests/unit/t149-codex-hook-adapter.test.ts:355-364` asserts the adapter *produces* `hookSpecificOutput.hookEventName === "PostCompact"` with the context — i.e. it pins the adapter to the **wrong** shape. The test validates the adapter against its own fixture (`tests/fixtures/codex-hook-payloads/payloads.json:119`), never against what real Codex accepts, so a green suite still ships a hook Codex refuses.

## Impact

- PostCompact is Codex-only and its job is to **re-inject the workflow mission after compaction** (`harness/codex/onboarding.fills.ts:54`, `SKILL.md:153`). When it fails, the deterministic mission reload silently doesn't happen: after a compaction the agent loses the re-injected workflow context and can drift off-protocol until the next event that reloads context.
- Every compaction in a Codex AIDLC session prints the error (noise) and skips the reload.

## Fix shape

1. Determine Codex's actual PostCompact stdout contract (early-access schema / probe) and adjust the `post-compact` case + `wrapContext` to emit it — PostCompact likely needs a distinct envelope from SessionStart's `hookSpecificOutput`.
2. If PostCompact accepts no structured context payload, fall back to injecting the mission via the mechanism the event *does* support (or accept that reload happens on the next SessionStart) rather than emitting a rejected envelope.
3. Correct `t149` to assert the real accepted shape, and add a contract note so the "verified live for SessionStart" assumption isn't silently reused for other events.

## Files

- `harness/codex/hooks/aidlc-codex-adapter.ts:340-353` (post-compact case), `:199-214` (`wrapContext`)
- `harness/codex/emit.ts:42` (hook wiring), `:172` (event-name map)
- `tests/unit/t149-codex-hook-adapter.test.ts:355-364`

Note: I could not confirm the *correct* PostCompact envelope from this repo (no Codex early-access here) — step 1 needs the live schema. Everything else (wrong-shape reuse, test blind spot, impact) is confirmed from source.

apackeer:
Fixes #593.

## Problem

Every compaction in a Codex AIDLC session printed:

```
• PostCompact hook (failed)
  error: hook returned invalid PostCompact hook JSON output
```

The shipped `post-compact` hook reused the SessionStart output envelope (`{"hookSpecificOutput":{"hookEventName":"PostCompact","additionalContext":...}}`). Codex's embedded `post-compact.command.output` schema (extracted from the codex-cli 0.145.0 binary) allows only `continue` / `stopReason` / `suppressOutput` / `systemMessage` with `additionalProperties: false` - there is no `hookSpecificOutput` and no context channel on this event at all (the binary literally carries the string "this event cannot emit additionalContext"). So the envelope was rejected on every compaction.

## Why removal (not a different envelope) is the fix

The mission reload never depended on PostCompact. After compaction, Codex queues a SessionStart with `source=compact` (`codex-rs/core/src/session/mod.rs` - `queue_pending_session_start_source(SessionStartSource::Compact)`, present at the 0.139.0 doctor floor and on current main) and delivers it at the next turn start. The already-registered session-start hook forwards `source` to the core hook, which renders the workflow-context block on `source=compact` with no duplicate `SESSION_STARTED` audit row (`PreCompact` owns `SESSION_COMPACTED`). The `session-start.command.output` schema does accept `hookSpecificOutput.additionalContext`, and its input schema enumerates `compact` as a source.

Pre-fix behavior was therefore: mission re-injected correctly via SessionStart(compact) PLUS a noisy, always-failing redundant PostCompact path. This PR removes the redundant path.

## Changes

- Remove the PostCompact wiring end to end: `hooks.json` entry, `SNAKE` event-name map entry, trust-seed hash, adapter `post-compact` case, and the live-captured fixture.
- Rewrite `t149` test 11 to drive `session-start` with `source: "compact"` and pin the surviving behavior (mission block injected, no new session audit row) - discharging the issue's "the test pinned the wrong shape" finding.
- Update `t150` wiring + trust-suffix expectations.
- Docs/onboarding (`docs/guide/11-session-management.md`, `docs/guide/harnesses/codex-cli.md`, Codex `AGENTS.md` fill, Codex `SKILL.md`): describe the compact-source SessionStart reload.
- Bump to 2.5.3 + CHANGELOG entry.

## Mixed-version safety

- An old installed `hooks.json` still invoking `post-compact` against the new adapter falls into the switch default: exit 0, empty stdout - safe degradation, no invalid JSON.
- Existing installs keep an orphaned `post_compact:0:0` trust entry in `$CODEX_HOME/config.toml` after re-copying `dist/codex/`. It is inert (trust keys are per-event; no other entry's index or hash shifts) and may be deleted.

## Verification

- `bun scripts/package.ts --check`: all harness trees in sync.
- Repo-wide sweep for `PostCompact|post-compact|post_compact`: hits only in CHANGELOG (new entry + one historical line); no stale references in code, tests, docs, or dist.
- Unit slice (t149 + t150 + t68): 3 files pass, 39/39 assertions, including the rewritten compact-source test.
- Contract evidence is source-level (codex binary schema + openai/codex Rust source at 0.139.0 and main); a live long-session compaction on Bedrock was not run.


raj-jain-aws:
Exploring why the approval button is not showing for my identity. Adding this dummy comment as a test

leandrodamascena (changes requested):
**Findings**

1. **[P1] Distinct compact events replay stale workflow context.** The surviving `session-start` path still enters the adapter-wide replay cache (`harness/codex/hooks/aidlc-codex-adapter.ts:105-120`), whose key is only `target + rawInput` and whose TTL is 30 minutes. Codex SessionStart input has no `turn_id`; for compact starts it contains the same session id, cwd, model, permission mode, and `source=compact`. Therefore two real compactions in one session produce the same key. I reproduced this against the PR by firing the rewritten test payload, changing `Current Stage` from `requirements-analysis` to `code-generation`, and firing the identical compact payload again. The second response still contained `Current Stage: requirements-analysis`. This makes the proposed replacement path return stale mission/routing state. Please exempt compact SessionStart from this replay strategy or add a reliable per-compaction discriminator, and add a regression test that performs two compact starts with a state change between them.

2. **[P1] The replacement is not immediate on the supported Codex 0.139-0.144 range.** The project still advertises and enforces a minimum of 0.139.0. In Codex 0.139, compaction queues `SessionStartSource::Compact`, but `run_pending_session_start_hooks` is only called at top-level `run_turn` entry; the mid-turn compaction branch continues directly to another model sample without draining the queued hook ([0.139 source](https://github.com/openai/codex/blob/rust-v0.139.0/codex-rs/core/src/session/turn.rs#L163), [mid-turn branch](https://github.com/openai/codex/blob/rust-v0.139.0/codex-rs/core/src/session/turn.rs#L273-L290)). Codex 0.145 added the missing drain immediately after mid-turn compaction ([0.145 source](https://github.com/openai/codex/blob/rust-v0.145.0/codex-rs/core/src/session/turn.rs#L354-L378)). Thus on every currently supported 0.139-0.144 install, removing PostCompact leaves the first post-compaction model continuation without the claimed deterministic mission reload. Since PostCompact cannot carry context, this likely requires raising and consistently enforcing/documenting the Codex floor to 0.145, or another supported immediate injection mechanism.

The implementation otherwise packages cleanly: `bun scripts/package.ts --check`, t149+t150 (32 tests), and t68 (7 tests) all pass. The PR is also based on 2.5.2/claims 2.5.3 while `v2` is now 2.5.5, so it must be rebased and regenerated after the behavioral fixes.

apackeer:
Addressed both P1 findings in `5075d684`:

- Compact-source `SessionStart` now bypasses the adapter replay cache. A regression test sends two byte-identical compact payloads around a workflow-state change and verifies the second response contains fresh context.
- The supported Codex CLI floor is now `0.145.0`, enforced by `/aidlc --doctor` and reflected across onboarding, docs, generated distributions, and live-test gates. A packaging test verifies `0.144.9` is rejected and `0.145.0` is accepted.
- Rebased onto current `v2` and regenerated every `dist/` tree.

Validation: focused tests passed (50 tests, 406 assertions), `bun scripts/package.ts --check`, typecheck, Biome, smoke, and unit tiers all passed. The default integration run reached the live Claude SDK tests, where 25 files were blocked before test logic by the sandbox rejecting writes to read-only `~/.claude.json`; deterministic integration tests remained green.

Ready for re-review.

leandrodamascena (changes requested):
The two previously reported P1 issues are fixed:

- Compact-source `SessionStart` now bypasses the replay cache, and the new two-compaction regression confirms fresh workflow context after a state change.
- Codex CLI `0.145.0` is now enforced by `/aidlc --doctor` and reflected in onboarding, documentation, generated distributions, and live-test gates.

Two merge blockers remain:

1. **[P1] The unit tier now fails its mechanism-honesty ratchet.** Adding `spawnSync` to `t150-codex-packaging.test.ts` correctly reclassifies that test from `none` to `cli`, but `tests/unit/gen-coverage-registry.test.ts:739` was not updated. Running smoke+unit fails with:

   ```
   Expected: EXPECTED_NONE_TO_CLI
   Received: + "unit/t150-codex-packaging.test.ts"
   ```

   Please add that path to `EXPECTED_NONE_TO_CLI`. The test itself documents this as the required maintenance action when a deterministic test begins spawning a CLI.

2. **[P1] The branch conflicts with current `v2`.** `git merge-tree` reports content conflicts in `CHANGELOG.md` and `docs/reference/05-agent-system.md`. Both this PR and `v2` currently claim version `2.5.6`, so the PR must rebase, preserve the existing `v2` changelog/docs changes, and re-bump to the next available version. Please coordinate the patch slot with PR #628, which also currently needs a re-bump.

Verification performed:

- `t149`: 21/21 passed
- `t150` + `t68`: 20/20 passed
- Packaging parity: passed
- Typecheck: passed
- Smoke+unit: 174 files passed, 1 file failed on the registry ratchet above
- Biome only reports the unrelated pre-existing warning in `core/tools/aidlc-doctor-bundle.ts:628`

apackeer:
Addressed both remaining blockers in `dfdb4aed`:

- Added `unit/t150-codex-packaging.test.ts` to `EXPECTED_NONE_TO_CLI`, preserving the mechanism-honesty ratchet now that the test deterministically spawns the doctor CLI.
- Rebased onto current `v2` (`9eb30dbd`), preserving the Kiro model-inheritance changelog/docs changes and regenerating every `dist/` tree. PR #628 has been asked to take `2.5.7`, so this PR now uses `2.5.8` to avoid another patch-slot collision.

Validation:

- Focused `gen-coverage-registry` + t149 + t150 + t68: 74 tests, 1,305 assertions, all passed
- Smoke + unit tiers: 175 files, 4,320 assertions, all passed
- Typecheck (production, tests, adapters): passed
- Biome: only the unrelated pre-existing `useOptionalChain` warning in `core/tools/aidlc-doctor-bundle.ts:628`

Ready for re-review.

leandrodamascena (approved):
The previously reported blockers are fixed:

- Compact-source `SessionStart` bypasses the replay cache and returns fresh workflow context.
- Codex CLI `0.145.0` is enforced and documented.
- `t150-codex-packaging.test.ts` is registered in the mechanism-honesty ratchet.
- The branch merges cleanly with `v2` and consistently uses version `2.5.8`.

Validation passed: 74 focused tests, packaging parity, typecheck, and the complete smoke/unit tiers with 175 files and 4,320 assertions.

No remaining findings. Approved.

PR #628, opened 2026-07-22
fix: approval-gate answer no-op ends the same-turn answer && report deadlock (2.5.3)

### LINKED ISSUE #592: [Bug]: Same-turn `answer && report --result approved` self-deadlocks the human-presence gate (QUESTION_ANSWERED consumes the HUMAN_TURN that approve needs)
## Symptom (observed on Codex CLI)

At the reverse-engineering approval gate, the agent chained the answer-log and the approve in one turn:

```
bun .codex/tools/aidlc-log.ts answer --stage reverse-engineering --details "Approve" \
  && bun .codex/tools/aidlc-orchestrate.ts report --stage reverse-engineering --result approved --user-input "Approve"
```

The user typed `approve` as a real chat message, yet `report` was refused:

```
{"emitted":"QUESTION_ANSWERED","stage":"reverse-engineering"}
{"kind":"error","message":"Transition rejected by aidlc-state.ts approve for \"reverse-engineering\": {\"error\":\"Refusing to approve \"reverse-engineering\": a real human has not acted at this gate since it opened. The approval gate requires a typed human turn before it can commit. ...\"}"}
```

An earlier attempt (before the user re-typed) was refused by the *answer* guard with the sibling message: `"Refusing to record this answer: a real human has not acted at this checkpoint this turn."` This is a **permanent deadlock** — every retry re-runs the same chain and can never advance.

## Root cause

Both guards share one predicate, `humanActedSinceGate` / `humanActedSinceLastAnswer` (`core/tools/aidlc-lib.ts:1422`): a human is present iff a `HUMAN_TURN` event appears *after the last gate resolution*, where a resolution is any of `GATE_APPROVED` / `GATE_REJECTED` / **`QUESTION_ANSWERED`** (`GATE_RESOLUTION_EVENTS`, `aidlc-lib.ts:1421`).

On the approve turn:

1. User types `approve` → `UserPromptSubmit→mint` hook emits **`HUMAN_TURN`**.
2. `aidlc-log.ts answer` runs, passes its own presence check, and emits **`QUESTION_ANSWERED`**.
3. `report → aidlc-state.ts approve` runs `humanActedSinceGate` → the newest resolution is now the `QUESTION_ANSWERED` from step 2, which sits **after** the `HUMAN_TURN` → predicate returns false → **refused**.

The `answer` call consumes the very human turn that the following `approve` requires. Because the freshness boundary for `approve` includes `QUESTION_ANSWERED`, a same-turn `answer → approve` for the *same* gate is self-defeating. (The `QUESTION_ANSWERED`-in-boundary rule is deliberate — it defeats fabricated cross-stage cascades, per the comment at `aidlc-lib.ts:1418-1420` — but it also kills a legitimate same-turn answer-then-approve of one gate.)

## Contributing cause — agent ran a command the protocol doesn't prescribe

The prescribed approve path is **only** `report --stage <slug> --result approved` (`core/aidlc-common/protocols/stage-protocol.md:162`, `:17`). It does **not** include `aidlc-log.ts answer`. The agent added the `answer` call on its own — understandable, since the approval gate is presented as a structured Approve/Request-Changes *question* (`harness/codex/skills/aidlc/SKILL.md:74`), which resembles the interview path where `answer` *is* correct (`stage-protocol.md:419`).

## Scope

The buggy logic is harness-neutral (`core/tools/aidlc-lib.ts`, `aidlc-state.ts`, `aidlc-log.ts`). It surfaced on Codex but any harness hits the deadlock if the conductor chains `answer && report`.

## Fix options (for discussion)

1. **Make the framework robust (recommended).** A `QUESTION_ANSWERED` should not be a freshness boundary for the *approve* path — the approve gate is resolved by `GATE_APPROVED`/`GATE_REJECTED`, not by an interview answer. Narrowing `approve`'s boundary to `{GATE_APPROVED, GATE_REJECTED}` (while the interview path keeps `QUESTION_ANSWERED`) makes a redundant same-turn `answer` harmless instead of fatal. Needs care against the cross-stage-cascade case the current rule guards (`aidlc-lib.ts:1418-1420`).
2. **Guardrail the agent.** Make the approval-gate instruction explicit that the approve path is `report` *alone* — never `aidlc-log.ts answer` — and/or have `report` tolerate/skip a self-emitted same-turn `QUESTION_ANSWERED` for the same stage.

Option 1 is the durable fix; option 2 reduces the trigger rate. Ideally both.

## Repro sketch

1. Reach any approval gate.
2. As the conductor, run `aidlc-log.ts answer --stage <slug> ... && aidlc-orchestrate.ts report --stage <slug> --result approved ...` in a single human turn.
3. `report` is refused with "a real human has not acted at this gate since it opened", and retries loop forever.

apackeer:
Fixes #592.

## Problem

At any approval gate, a conductor that chains the answer log and the approve in one turn:

```
bun .codex/tools/aidlc-log.ts answer --stage <slug> --details "Approve" \
  && bun .codex/tools/aidlc-orchestrate.ts report --stage <slug> --result approved --user-input "Approve"
```

deadlocks permanently. The `answer` call emits `QUESTION_ANSWERED`, which is a gate-resolution event in the human-presence predicate (`GATE_RESOLUTION_EVENTS`, aidlc-lib.ts), so it consumes the very `HUMAN_TURN` the following `report --result approved` requires. Every retry re-runs the same chain and can never advance. Observed on Codex CLI; the logic is harness-neutral, so every harness is exposed.

## Fix

Both halves of the issue's "ideally both" recommendation, with a safer variant of option 1:

1. **Framework robustness** (`core/tools/aidlc-log.ts`): `answer --stage <slug>` now recognizes when that same stage sits at `[?]` awaiting-approval and returns a successful skip - `{"skipped":"QUESTION_ANSWERED","stage":"<slug>","reason":"approval-gate-report-owned"}` - without emitting `QUESTION_ANSWERED`. The real `HUMAN_TURN` survives for `report --result approved`, so an existing `answer && report` chain recovers on the next typed human turn.

   Deliberately NOT the issue's option 1 (narrowing approve's freshness boundary to `{GATE_APPROVED, GATE_REJECTED}`): that would re-open the fabricated cross-stage cascade the boundary defends against (an interview answer consumes the turn, then a same-turn fabricated approve would pass). The skip keeps that property intact: ordinary interview answers remain workflow-global freshness boundaries, and the skip itself does not substitute for a `HUMAN_TURN` (both pinned by new tests).

2. **Guardrails** (prose): the stage protocol (checklist item 2, gate Part 3, section on question logging, event-emission paragraph), practices-discovery's affirmation gate, audit-format, `docs/reference/04-stage-protocol.md`, `docs/reference/12-state-machine.md`, and step 5 of all five harness SKILLs now state that approval gates are report-owned: never call `aidlc-log.ts decision` or `aidlc-log.ts answer` for an approval choice. The section 13 learnings ritual keeps its decision/answer pair (it is a non-gate question and its logged answer precedes the gate opening by contract).

## Tests

`tests/unit/t188-human-presence-gate.test.ts` gains three tests driving the public `aidlc-orchestrate.ts report` surface with the presence guard enabled:

- redundant approval answer is a no-op and report still approves (the deadlock repro, now green)
- an interview answer still cannot authorize a later same-turn approval (cascade property preserved)
- the approval-answer no-op does not substitute for a `HUMAN_TURN` (fabrication still refused)

One existing test was adjusted to activate presence tracking via a `decision` row instead of `gate-start`, keeping it an ordinary interview-answer scenario now that an open gate triggers the skip.

## Verification

- `bun scripts/package.ts --check`: all 5 harness trees in sync
- `bun run typecheck` (both tsconfigs) and biome on changed files: clean
- smoke+unit tier (`-P 8 --debug`): 174/174 files pass, 0 failed assertions
- integration tier (live Claude SDK, `-P 8 --debug`): 103/103 files pass, 0 failed assertions

## Version

2.5.3 + CHANGELOG entry + README badge. Note: 2.5.3 was also requested as the re-bump slot for #606; whichever merges second re-bumps and renames its CHANGELOG heading (t68 enforces).


leandrodamascena (changes requested):
The proposed fix resolves the happy-path deadlock, but the new early no-op introduces additional correctness gaps:

1. `core/tools/aidlc-log.ts:146-154` returns before checking human presence. Consequently, `answer "Request Changes" && report rejected` records a rejection with no `HUMAN_TURN`. I reproduced this against the PR head.
2. `core/tools/aidlc-log.ts:139-145` treats every answer targeting a `[?]` stage as an approval answer, regardless of `--details` or a preceding non-gate decision. Legitimate non-gate answers are silently acknowledged but omitted from the audit trail.
3. Gate classification occurs before acquiring the audit lock. A concurrent `gate-start` can move the stage to `[?]` after classification but before `QUESTION_ANSWERED` is appended, recreating the original deadlock.
4. Existing documented flows still log approval responses before reporting and, in some cases, omit `report awaiting-approval`, so the fallback does not activate:
   - `docs/reference/diagrams.md:488-528`
   - `docs/guide/07-interaction-modes.md:81-130`
   - `core/aidlc-common/protocols/stage-protocol.md:520,595`

Please make the skip conditional on a demonstrable approval response, preserve human-presence enforcement for rejection, perform classification and emission under one lock, and complete the documentation sweep.

The branch also currently conflicts with `v2`, including newer reviewer events and version files, so it needs rebasing and revalidation.

apackeer:
Thanks for the thorough round-1 review - all four findings were real, and all four are now addressed at 6be2d6b2 (rebased onto v2 @ 2.5.5, re-bumped to 2.5.6).

**1. Rejection path no longer bypasses presence.** The skip no longer returns before the human-presence check. A gate-choice answer with no fresh `HUMAN_TURN` refuses ("Refusing to acknowledge this approval choice..."), so the `answer "Request Changes" && report rejected` chain you reproduced now breaks at the answer instead of slipping through the successful no-op. Pinned by two new t188 tests (approve-shaped and reject-shaped, both with no human turn).

**2. Skip is conditional on a demonstrable approval response.** The no-op fires only when `--details` is gate-choice shaped: anchored, word-boundary matches for `Approve/Approved`, `Request Changes`, `Accept as-is`, `Reject/Rejected`, with optional trailing feedback after the label. Anything else at an open gate (e.g. a clarifying question's response) records `QUESTION_ANSWERED` normally and is never dropped from the audit trail. Pinned by a new t188 test; non-matches like "Approval workflow design looks good" and "Requesting more info" stay on the record path.

**3. Classification and emission under one lock.** `handleAnswer` now wraps the checkbox read, the presence check, and the append in a single `withAuditLock` section, so a concurrent `gate-start` (itself locked) cannot flip the stage to `[?]` between classification and the `QUESTION_ANSWERED` append. The tool's `emitAudit` routes held-lock emits through `appendAuditEntryUnlocked` - the same idiom `aidlc-state.ts` uses - to avoid self-deadlocking on the non-reentrant OS lock.

**4. Documentation sweep completed.** `docs/reference/diagrams.md` (section 11), `docs/guide/07-interaction-modes.md` (diagram + text fallback), and `docs/reference/04-stage-protocol.md` (diagram + Part 0) now show the report-owned gate flow: `report awaiting-approval` before the question, `report approved|rejected` after, no separate log-the-response step. The stage-protocol audit-log rules (the lines around your :520/:595 cites) now scope prompt/response logging to non-gate questions and state that the gate's audit rows are report-owned. A repo-wide grep for the stale patterns (`AUDIT_POST`, "append user's choice", "logs the user's response") comes back empty outside `dist/` regeneration.

**Rebase and revalidation.** Rebased onto v2 @ f122f193 (2.5.5, the reviewer-events work) and re-bumped to 2.5.6; the reviewer-event table rows in `audit-format.md` / `12-state-machine.md` are merged with the non-gate wording. Evidence: t188 17/17; smoke+unit 174/174, 0 failed assertions; live Claude SDK integration tier 103/103 (4 files were red in the main run - t72, t193, t238, t-journey-workspace - and all four pass on re-run; none of the 24 traces contains the new refusal or skip strings, so they are live-agent latency flakes, not regressions); `package.ts --check` clean; both tsconfigs clean.

One residual worth stating explicitly: an `answer` logged against a stage that is NOT at `[?]` (wrong slug) followed by a same-turn approve still deadlocks that turn - the skip keys on the target stage's checkbox state plus the details shape, deliberately, so it cannot become a general presence bypass. That failure now self-heals on the next typed human turn, and the prose sweep makes the wrong-slug call much less likely.

leandrodamascena (changes requested):
The lock/reentrancy changes and documentation sweep look sound, and the rejection-presence bypass is fixed. Two classification regressions and the version collision still block approval:

1. **Paraphrased approvals recreate #592.** At `[?]`, `answer --details "The user approved"` does not match `GATE_CHOICE_RE`, emits `QUESTION_ANSWERED`, consumes the `HUMAN_TURN`, and causes the following `report --result approved` to refuse. The same applies to realistic inputs such as `Yes, approve`, `LGTM`, and `Ship it`.

2. **Gate-prefixed clarifying answers are silently dropped.** After recording a non-gate `DECISION_RECORDED` at an open gate, `answer --details "Reject option B, use A"` returns `{"skipped":"QUESTION_ANSWERED",...}` and exits successfully without preserving the substantive answer. I reproduced the equivalent case with `Approve production deploys`. This violates the exact-user-input audit contract. The current test at `t188:520` only uses a non-triggering string and does not cover this path.

3. **The branch must rebase and re-bump.** Current `v2` already owns `2.5.6` for #606. `git merge-tree` reports a `CHANGELOG.md` conflict, and both branches contain different `## [2.5.6]` entries. Please preserve `v2`'s entry and bump this PR's version, badge, and heading to `2.5.7`.

The regex cannot reliably distinguish a gate response from a non-gate answer. A structural approach would avoid both runtime failures: if an unresolved non-gate `DECISION_RECORDED` exists after `STAGE_AWAITING_APPROVAL`, record its answer; otherwise treat `answer` targeting that `[?]` stage as the redundant gate call. Keep the new presence guard and single-lock transaction.

Please add regressions for a paraphrased approval and a gate-word-prefixed answer to an explicitly logged non-gate question.

apackeer:
Verified all three round-2 findings against `6be2d6b2` before changing the branch:

1. `answer --details \"The user approved\"` emitted `QUESTION_ANSWERED`; the following approved report returned an error directive and emitted no `GATE_APPROVED`.
2. After a gate-local `DECISION_RECORDED`, `answer --details \"Reject option B, use A\"` returned the approval-gate skip and preserved no `Details` row.
3. The base owned the PR's prior version and produced the expected changelog collision.

Fixed at `d0bbd00` after rebasing onto current `v2` @ `ccf284b`:

- Removed wording/regex classification. Under the existing single audit lock, `answer` now inspects the ordered audit stream: an unresolved `DECISION_RECORDED` after the current `STAGE_AWAITING_APPROVAL` records the exact answer; otherwise an answer targeting that `[?]` stage is the redundant report-owned gate call.
- Kept the human-presence guard on the redundant-answer path and the held-lock/unlocked-append transaction.
- Added regressions for `\"The user approved\"` and `\"Reject option B, use A\"` after an explicitly logged gate-local non-gate question.
- Preserved the base's 2.5.7 Kiro IDE entry and bumped this PR's version source, README badge, changelog heading, and generated trees to 2.5.8.

Validation:

- t188: 18/18
- focused gate + version + current-base adapter tests: 55/55
- smoke + unit before the final base advance: 175 files, 4,325 assertions, 0 failures
- all three TypeScript configs: clean
- Biome: 527 files, clean
- `bun scripts/package.ts --check`: all five harness trees clean

apackeer:
@leandrodamascena Ready for re-review at `d0bbd00`. The branch is rebased onto current `v2` (`ccf284b`), the release is re-bumped to `2.5.8`, both round-2 classification regressions are covered, and the validation evidence is in the updated comment above.

leandrodamascena (approved):
The previously reported blockers are fixed:

- Approval answers are classified from audit structure rather than wording.
- Paraphrased approvals no longer consume the human turn.
- Gate-prefixed answers to pending non-gate questions are recorded exactly.
- The branch merges cleanly with current `v2` and consistently uses version `2.5.8`.

Validation passed: 76 focused tests, packaging parity, typecheck, Biome, and the complete smoke/unit tiers with 175 files and 4,328 assertions.

No remaining findings. Approved.

Note: PR #626 must rebase and re-bump if this PR merges first.

PR #633, opened 2026-07-22
ci: bootstrap the docs deploy workflow on the default branch

apackeer:
## What

Prerequisite for #630 (the v2 Zensical documentation site): a byte-identical copy of that PR's `.github/workflows/docs.yml` added to `main`.

## Why

GitHub registers `workflow_dispatch` only for workflows that exist on the **default branch** (`main`). Without this file on main, the docs workflow cannot be manually dispatched at all (`gh workflow run docs.yml --ref v2` errors with "workflow not found"), which is needed to trigger the first deployment after the admin flips the Pages source to "GitHub Actions" - review feedback on #630.

## Behavior on main

Inert except for dispatch registration:

- `push`/`pull_request` triggers are scoped to `v2`, so nothing runs on main's own pushes or PRs
- a manual dispatch with `--ref v2` checks out and builds the v2 docs; a dispatch against main would fail fast at the build step (main has no `zensical.toml`/`pyproject.toml`), and deploys are additionally inert until the Pages source is switched

No other content on main is touched. Merge this before or together with #630; it has no effect on the currently served Jekyll site either way.

## Testing

actionlint clean; the workflow content itself is exercised end to end in #630 (strict build + deploy path verified on a fork preview).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,74 @@
+name: Deploy Documentation
+
+on:
+  push:
+    branches:
+      - v2
+  pull_request:
+    branches:
+      - v2
+  workflow_dispatch:
+
+# The build job validates every PR and push; only the deploy job needs
+# Pages/OIDC permissions, granted at job level below.
+permissions:
+  contents: read
+
+concurrency:
+  group: docs-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


leandrodamascena (approved):

raj-jain-aws (approved):
This is a low-risk, well-constructed change. Single file addition (.github/workflows/docs.yml), nothing else touched. The rationale is sound: planting the workflow on main is the only way GitHub will register workflow_dispatch so the v2 docs deploy can be manually triggered.

PR #637, opened 2026-07-22
fix: case-fold the swarm unit-name membership tests (2.5.12)

### LINKED ISSUE #621: Swarm batch never advances: bolt_dag unit-name casing vs forced-kebab SWARM_UNIT_CONVERGED slugs not case-folded
## Summary

Under autonomous Construction, the orchestration engine (`aidlc-orchestrate.ts`) re-emits `invoke-swarm` for the same Construction batch forever after that batch's units have genuinely converged and been finalized. The workflow cannot advance to the next batch or to the stage settle gate.

## Root cause

There is a unit-name **casing mismatch** between two sources that are compared without case-folding:

1. `bolt_dag.batches` in the compiled `runtime-graph.json` preserves unit names **exactly as authored** in `units-generation`'s `unit-of-work-dependency.md` — commonly uppercase, e.g. `[["U-16","U-AUTH","U-AZ"],["U-BENCH"]]`. (`computeBoltDag` / `parseBoltDag` in `aidlc-runtime.ts` do not canonicalize case.)

2. The swarm referee (`aidlc-swarm.ts`) **forces kebab-case lowercase** unit slugs — `prepare`/`check`/`finalize` reject non-kebab (e.g. `U-16` is rejected with `Invalid --slug: "U-16". Must be kebab-case`). So the `SWARM_UNIT_CONVERGED` audit rows `finalize` writes carry `Unit name: u-16` / `u-auth` / `u-az` (lowercase).

In `aidlc-orchestrate.ts`, `swarmConvergedUnits()` builds a `Set<string>` from the `Unit name` field of `SWARM_UNIT_CONVERGED` rows (lowercase), and `tryEmitSwarm()` computes owed units as `batch.filter((u) => !converged.has(u))` where `u` is the uppercase DAG name. `converged.has("U-16")` is always false against a set containing `"u-16"`, so every batch reads as fully unconverged and batch 1 is re-emitted indefinitely.

## Impact

Any intent whose `unit-of-work-dependency.md` authors unit names in uppercase (or any casing other than kebab-lower) will hang at the first autonomous swarm batch: `finalize` exits 0, metadata consolidates (Bolt Refs empties, SWARM_UNIT_CONVERGED rows land), yet `next` keeps returning `{"kind":"invoke-swarm", "units":[...same batch...]}`.

## Suggested fix

Case-fold both sides of the comparison in `aidlc-orchestrate.ts`:
- In `swarmConvergedUnits()`, store `unit.toLowerCase()` in the converged set.
- In `tryEmitSwarm()`, test `!converged.has(u.toLowerCase())` while still emitting the original-cased name (the conductor's `prepare` re-derives the kebab slug from it).

Alternatively, canonicalize unit names to kebab-lowercase at compile time in `computeBoltDag`, but that changes the stored graph shape other readers may depend on, so the comparison-site case-fold is the smaller blast radius.

## Environment

Observed on a downstream fork of the framework during an mvp-scope Construction stage (`code-generation`) with a 3-unit parallel batch under autonomous mode.

jstrunk:
Fixes #621.

## What

Case-folds the engine's swarm unit-name membership tests so an autonomous Construction run whose unit names are authored in uppercase (e.g. `U-16` in `unit-of-work-dependency.md`) can advance batches, settle the stage, and pass the approval gates. Before this fix, `next` re-emitted `invoke-swarm` for the first batch forever after that batch had genuinely converged — the silent hang in #621.

## Root cause (recap from the issue)

`bolt_dag.batches` preserves unit names **exactly as authored**, while the swarm referee (`aidlc-swarm.ts`) forces **kebab-lower** unit slugs — `prepare`/`finalize` reject non-kebab — so the `SWARM_UNIT_CONVERGED` rows carry `u-16`. `converged.has("U-16")` against a set holding `"u-16"` is always false, so every batch read as fully unconverged.

## The fix — all four comparison sites, not just the hang

Since the issue was filed (against 2.2.17-era code), the ledger reader moved to `aidlc-lib.ts` and grew more consumers (including 2.5.5's per-unit review enforcement). Fixing only `tryEmitSwarm` would have moved the failure downstream: the engine would advance the batches, but the settle report, the approve-time artifact guard, and the per-unit review-receipt check would each then refuse for the same casing reason. So:

- `aidlc-lib.ts` `swarmConvergedUnits()` — the shared reader stores **lowercased** `Unit name`s (documented in its header comment).
- `aidlc-orchestrate.ts` `tryEmitSwarm()` — folds the authored batch name at the membership test, and still emits the **original-cased** names in the `invoke-swarm` directive (the conductor's `prepare` re-derives the kebab slug from them).
- `aidlc-orchestrate.ts` `isSettledAutonomousSwarm()` — the report-side settle evidence gate folds the DAG name.
- `aidlc-state.ts` `isSettledSwarmForArtifactGuard()` — the approve-time artifact-guard exemption folds the DAG name.
- `aidlc-state.ts` per-unit review-receipt check (`reviewedUnits`) — a swarm's reviews run in kebab-lower-slugged Bolt worktrees, so their `REVIEW_COMPLETED` rows carry the lower slug too; the set and both membership ops fold (add, artifact-invalidation delete, and the missing-units filter).

Per the issue's suggested fix, this is the comparison-site case-fold — **not** canonicalizing in `computeBoltDag`/`parseBoltDag`, which would change the stored graph shape other readers depend on.

## Non-goal

`aidlc-swarm.ts prepare --units U-16` still rejects the raw uppercase name loudly (`Invalid --slug`) and the conductor adapts by deriving the kebab slug — that loud path is unchanged. #621 is specifically the silent hang in the deterministic engine.

## Changes

- `core/tools/aidlc-lib.ts`, `core/tools/aidlc-orchestrate.ts`, `core/tools/aidlc-state.ts` — the case-folds above (comment-documented at each site).
- `tests/unit/t201-swarm-batch-advance.test.ts` — new describe "authored-casing vs referee slug (issue #621)", 3 red-first cases: batch 1 converged under lowercase rows advances to batch 2 (`["U-BENCH"]` original-cased); a partial pass re-fans only the owed units original-cased; a fully-converged uppercase DAG settles and `report --approved` completes after per-unit reviews are recorded.
- `tests/integration/t185-stage-artifact-guard.test.ts` — new settled-swarm exemption case: an uppercase-authored DAG whose rows carry kebab-lower slugs passes `approve`.
- Version 2.5.12 + CHANGELOG entry + README badge; all dists regenerated. (Rebased once onto v2 after it advanced to 2.5.11 — the original 2.5.6 number this PR claimed was consumed upstream in the meantime; re-bumped past it.)

## Testing

- Red-first: all 4 new cases fail on the unfixed engine, pass with the fix.
- Full smoke + unit tiers via `tests/run-tests.sh`: **178 files, 4399 assertions, 0 failures**. Integration tier: pass.
- `bun scripts/package.ts --check`, `bun tests/gen-coverage-registry.ts --check`, `bun run typecheck`, and `bun run lint` all clean.
- `bun run check` (the exact command the repo's new v2 PR gate CI job runs) and `bun scripts/ci-changelog-guard.ts` (the new gate's changelog-deletion guard) both pass locally against this branch.

## Context

Found and mitigated during a field engagement (a 3-unit parallel batch under autonomous mode hung at `code-generation`); the fix here extends that mitigation to the consumer sites added since.

## Note on the Security Scanners signal

`Security Scanners` (`.github/workflows/security-scanners.yml`) triggers only on `push`/`pull_request` to `main`, plus a daily schedule — never on a PR against `v2`. So this PR cannot inherit that workflow's result regardless of #638's status; confirmed by reading the workflow's `on:` block rather than by absence of a run. It does fail on `main` itself today (e.g. run 29914525428): grype flags 17 pre-existing findings in `scripts/aidlc-evaluator/uv.lock` (`pillow` 12.2.0 — Dependabot #619 — plus `mcp` 1.23.3 and `pydantic-settings` 2.14.1), and gitleaks flags 7 deliberate fake credentials in test fixtures not yet in the `.gitleaks.toml` allowlist. Tracked in #638, orthogonal to this PR.

The checks that DO apply here are `v2`'s new PR gate (`ci.yml`, added since this PR opened): contract checks (`bun run check`), smoke+unit tests, and a changelog-deletion guard. All three verified green locally against this branch (see Testing above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.12] - 2026-07-24
+
+Fixes an autonomous-Construction hang: when unit names are authored in uppercase (e.g. `U-16`) in `unit-of-work-dependency.md`, the swarm's converged batches were compared against the referee's kebab-lower audit slugs without case-folding, so `next` re-emitted `invoke-swarm` for the first batch forever and the stage could never settle. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `next` now case-folds the unit-name membership tests (batch advance, the report-side settle evidence gate, the approve-time artifact-guard exemption, and the per-unit review-receipt check), while still emitting the original-cased unit names in `invoke-swarm` directives.
+* Fixes #621.
+
 ## [2.5.11] - 2026-07-24
 
 Intent Capture now keeps generated intent and stakeholder claims grounded in the user's description, confirmed answers, workflow-selected scope, or explicitly registered memory. Unsupported content is omitted, elicited, or surfaced as a human-owned assumption instead of being presented as fact. **Upgrade:** re-copy your `dist/<harness>/` shell into the project so the updated stage, Product Lead reviewer contract, and `claim-sources` sensor are installed.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.5.11-blue)
+![version](https://img.shields.io/badge/version-2.5.12-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -3513,6 +3513,13 @@ export function latestMainWorkflowStageStarted(
 // fields (pre-2.5.0 audit logs) fail closed: the affected units re-fan on the
 // next swarm pass, which finalize's re-verify makes safe. The timestamp check
 // stays as belt-and-braces.
+//
+// The set holds LOWERCASED names: rows carry the referee's forced kebab-lower
+// unit slug (prepare/finalize reject non-kebab), while bolt_dag.batches
+// preserve unit names as authored in unit-of-work-dependency.md (often upper,
+// e.g. "U-16"). Every consumer must lowercase the DAG-side name before its
+// membership test, or an uppercase-authored batch reads as unconverged
+// forever (#621).
 export function swarmConvergedUnits(
   projectDir: string,
   slug: string,
@@ -3526,7 +3533,7 @@ export function swarmConvergedUnits(
     if ((auditBlockField(block, "Run floor") ?? "") !== floor) continue;
     if (floor && timestamp < floor) continue;
     const unit = auditBlockField(block, "Unit name");
-    if (unit) converged.add(unit);
+    if (unit) converged.add(unit.toLowerCase());
   }
   return converged;
 }
--- a/core/tools/aidlc-orchestrate.ts
+++ b/core/tools/aidlc-orchestrate.ts
@@ -2012,8 +2012,10 @@ function isSettledAutonomousSwarm(
   if (r.state !== "ok") return false;
   const units = r.batches.flat();
   if (units.length === 0) return false;
+  // converged holds kebab-lower slugs (see swarmConvergedUnits); lowercase the
+  // authored DAG name for the membership test (#621).
   const converged = swarmConvergedUnits(projectDir, node.slug);
-  return units.every((unit) => converged.has(unit));
+  return units.every((unit) => converged.has(unit.toLowerCase()));
 }
 
 // Try to handle an eligible autonomous swarm stage, returning true (and emitting)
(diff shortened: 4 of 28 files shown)


apackeer:
Thanks, Jeff. #621 is a real current-`v2` bug, and this PR correctly identifies the downstream convergence, approval, and review comparisons that need to understand the worktree slug. I ran the focused `t185`/`t201` slice from this head: both files passed, with 37 assertions.

There is one earlier boundary we need to include before integrating it. The current conductor passes `directive.units` unchanged to `aidlc-swarm prepare`, and `prepare` forwards each value unchanged to `aidlc-worktree create --slug`. `U-16` is therefore rejected before any lowercase convergence row can be written. The new tests seed those rows directly, so they do not exercise that path; the current comment that the conductor re-derives the slug is not the shipped contract.

We also need to preserve authored unit identity independently from filesystem identity. The DAG currently accepts `U-AUTH` and `u-auth` as distinct names, while this patch folds both into one convergence/review key, so one row could satisfy both. The same global fold affects non-swarm per-unit reviews. The current-`v2` integration should preserve authored names in DAGs, directives, audit, and results; derive and validate a lowercase worktree slug inside deterministic swarm tooling; reject units that collide after that mapping; and use shared comparison logic for legacy lowercase rows. A real `prepare` -> review -> `finalize` -> `next`/approve uppercase test plus a collision case should cover the boundary.

This branch now conflicts and overlaps heavily with current `v2`, so please do not spend time rebasing it. Maintainers can port the useful downstream fixes and integrate the complete authored-name/worktree-slug split. I suggest keeping this PR open until that integration lands, then closing it as incorporated.

@leandrodamascena, could you take another review pass and confirm this identity boundary and current-`v2` integration direction before maintainers port the fix?

leandrodamascena (commented):
Review conclusion: this branch should not be rebased or merged; it can now be closed as superseded by `01893096`.

The identity boundary raised above is real. This patch lowercases both the authored DAG name and the audit/review key. Because the Unit-name contract is case-sensitive, distinct units such as `U-AUTH` and `u-auth` then share one convergence/review identity, so one receipt can satisfy both. Lowercasing also is not the actual filesystem-slug transform for names containing dots, underscores, or a leading digit.

Current `v2` implements the safer split: `boltSlugForUnit()` derives a deterministic collision-resistant internal Bolt/worktree slug, while directives, results, and `SWARM_UNIT_CONVERGED` retain the exact authored Unit name; convergence comparisons therefore remain exact. I verified the current end-to-end `1b legacy-safe Unit names complete the autonomous swarm lifecycle` case and the current `t201-swarm-batch-advance` suite: both pass.

On this PR head, its focused `t201` suite also passes (16/16). The seven `t185` failures reproduce unchanged on the parent commit, so they are baseline failures rather than regressions from this patch.

Recommendation: close #637 as superseded/incorporated and close #621 against the current `v2` implementation.

leandrodamascena:
Closing this PR as superseded by [#692](https://github.com/awslabs/aidlc-workflows/pull/692), merged in commit [`01893096`](https://github.com/awslabs/aidlc-workflows/commit/01893096).

#692 resolves the specific failure reported in #621 by preserving the authored Unit name as its identity while deriving a separate, collision-resistant slug for Bolt/worktree operations. I verified the legacy Unit-name swarm lifecycle and batch-advance tests against current `v2`.

Therefore, #621 can also be closed as fixed by #692. Thanks for identifying and testing the original issue.

PR #639, opened 2026-07-22
fix(kiro-ide): never read stdin in adapter entry paths (2.5.7)

apackeer:
## Problem

The `run()` export refactor in #560 introduced unconditional stdin acquisition
in both Kiro IDE adapter entry paths. Kiro IDE opens hook stdin but does not
write or close it, so agent-stop hooks hung whenever `USER_PROMPT` was empty.
The forwarding loop (`stop`) and `SESSION_ENDED` (`session-end`) silently never
ran.

Prompt-submit and post-tool-use hooks were unaffected because Kiro IDE
populates `USER_PROMPT` for those events.

## Fix

- The direct Kiro IDE adapter entry guard never reads stdin.
- The shared CLI dispatcher skips stdin acquisition only for the `kiro-ide`
  route; Codex and Kiro CLI retain their existing stdin behavior.
- Both paths pass an empty input to the adapter, which obtains its current
  event context from `USER_PROMPT`.

## Regression Coverage

- t218 test 12 holds stdin open and verifies the direct adapter exits.
- t218 test 12b exercises the actual agent-stop targets, `stop` and
  `session-end`, with empty and absent `USER_PROMPT`. Each invocation asserts
  its own effect: the forwarding-loop decision or a `SESSION_ENDED` audit
  delta.
- t218 test 12d exercises both agent-stop targets through the CLI dispatcher
  while stdin remains open.
- t218 test 12c verifies post-tool-use work still completes with open stdin.

## Validation

- `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts tests/unit/t68-version-changelog-sync.test.ts`
  (37 passing)
- Live Kiro IDE before/after verification of `HUMAN_TURN`,
  `SESSION_STARTED`, and `SESSION_ENDED`

## Release

Version `2.5.7`; `CHANGELOG.md`, the README badge, authored version source, and
all generated harness distributions are synchronized.

Related stdin-channel work in #614 and #615 remains open. Those changes compose
with this fix; if #615 lands first, this branch will need its context-channel
comments refreshed during rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.5.7] - 2026-07-23
+
+Fixes a Kiro IDE regression that silently disabled the agent-stop hooks: the forwarding-loop reminder and `SESSION_ENDED` never ran, because the IDE delivers an empty `USER_PROMPT` on agent stop and the adapter's entry guard then waited on stdin, which Kiro IDE opens but never writes. Hooks on prompt submit and after tool use were unaffected (the IDE populates `USER_PROMPT` for those events). Verified live on Kiro IDE against both the broken and fixed adapter. **Upgrade:** re-copy `dist/kiro-ide/` into your project (only `.kiro/hooks/aidlc-kiro-adapter.ts` changed).
+
+* Kiro IDE agent-stop hooks (forwarding loop, session end) work again: the adapter no longer reads stdin on any event (context arrives only via the `USER_PROMPT` environment variable; the stdin value was never consumed).
+
 ## [2.5.6] - 2026-07-22
 
 Unblocks agent delegation on the Kiro harnesses (#601): every shipped model pin is removed from the Kiro agent surfaces, so the conductor and all 14 personas inherit the session model. Previously every `invoke_sub_agent` spawn failed with `Invalid model ID` whenever the pinned IDs (`claude-opus-4.8` on the conductor JSON, `claude-sonnet-4.5` on the tier-projected persona JSONs and `.md` frontmatter) were not enabled on the user's Kiro install - blocking the subagent stages, every `reviewer:`-bound gate, the 2.5.0 ensemble dispatches, and any manual delegation. Verified fix shape (live on Kiro IDE): with both pin layers stripped, all 14 agents spawn and respond on a session model the pins never named. **Upgrade:** re-copy the entire `.kiro/` tree from `dist/kiro/` or `dist/kiro-ide/` into your project (agents, settings, tools, and knowledge all carry changes).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [2.5.7] - 2026-07-23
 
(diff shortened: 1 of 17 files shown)


leandrodamascena (changes requested):
The direct adapter fix works, but the same hang remains through the supported dispatcher path.

1. **`core/tools/aidlc.ts:965-982`**  
   `aidlc adapter kiro-ide <target>` still calls `await readStdin()` before invoking the adapter. With `USER_PROMPT=""` and stdin held open, the direct adapter exits normally, while the dispatcher hangs indefinitely. Please bypass stdin acquisition for `kiro-ide` in `runAdapter()` and add an open-stdin dispatcher regression test.

2. **`tests/unit/t218-kiro-ide-hook-adapter.test.ts:352-371`**  
   Test 12b claims to cover the affected `agentStop` hooks, but exercises `session-end` and `block`; `block` is a `preToolUse` hook. Replace it with `stop` and assert the forwarding-loop response.

3. **`tests/unit/t218-kiro-ide-hook-adapter.test.ts:363-371`**  
   Both `session-end` variants share one final `SESSION_ENDED` assertion, so one can silently fail while the other supplies the event. Assert the audit delta after each invocation.

Before merge, please also reconcile the `2.5.6` collision with PRs #606/#637 and the stdin-channel work in #614/#615.

apackeer:
Thanks @leandrodamascena - all three findings were correct, and all three are addressed in 2de50a99.

**1. Dispatcher path (`runAdapter` stdin acquisition)** - confirmed and reproduced before fixing: with `USER_PROMPT=""` and stdin held open, `aidlc adapter kiro-ide session-end` hung until killed while the direct adapter path exited cleanly. `runAdapter()` now skips stdin acquisition for the `kiro-ide` harness only (`core/tools/aidlc.ts`), mirroring the adapter's entry guard; codex and kiro CLI keep their reads. New t218 **12d** pins the dispatcher route with both agentStop targets, stdin held open, and the captured empty-`USER_PROMPT` shape - red against the unfixed dispatcher (timeout kill), green with the fix.

**2. 12b targets** - right on both counts: the live capture showed the hang biting agentStop, and those targets are `stop` + `session-end` (`block` is preToolUse). 12b now exercises exactly those two, and asserts the forwarding-loop response (`{"decision":"block"}` parsed from stdout) for `stop`.

**3. Shared assertion** - fixed. Each `session-end` invocation now asserts its own SESSION_ENDED audit-count delta (+1 per call), and each `stop` invocation asserts its own decision output, so neither the empty-string nor the absent-`USER_PROMPT` variant can silently no-op behind the other.

Gates on the new commit: `package.ts --check` clean, typecheck clean, full smoke+unit+integration tier 277/277 files / 5667 assertions, zero failures (t218 30/30, and the dispatcher suites t229 17/17 / t230 57/57 unaffected).

**Version collision**: acknowledged - #637 also claims 2.5.6 and #606 is re-slotting. Whichever of us merges second re-bumps and renames the CHANGELOG heading per the repo's conflict policy; I'll re-check at merge time.

**#614/#615 interplay**: checked for conflict - they compose rather than collide. #615 adds a per-target stdin read *inside* `run()` (gated to the two payload targets, raced against a 2s timeout) for the Kiro 1.x channel, while this PR removes the *unconditional* reads at the two entry points (adapter entry guard + dispatcher). Whichever lands second rebases trivially; the only follow-up is that once #615 merges, the entry-guard comment's "context arrives via USER_PROMPT only" should be softened to name both channels - happy to fold that in on whichever PR goes last.

leandrodamascena (approved):

PR #650, opened 2026-07-23
ci: add v2 PR gate (contract checks + smoke/unit tests + changelog guard)

leandrodamascena:
## Summary

The `v2` line has no test CI today — nothing runs on a pull request. This adds `.github/workflows/ci.yml` on `pull_request: [v2]` with three deterministic, credential-free jobs, plus a changelog-completeness guard.

## Jobs

| Job | Runs | Notes |
|---|---|---|
| `check` | `bun run check` | dist byte-parity (`package.ts --check`) + typecheck (all 3 tsconfigs) + biome lint |
| `test` | `bun tests/run-tests.ts --smoke --unit -P 8` | the deterministic tiers — no AWS / Bedrock / live Claude |
| `changelog-guard` | `bun scripts/ci-changelog-guard.ts origin/<base>` | fails if the PR deletes a `## [x.y.z]` heading present on the base branch |

## Why smoke + unit, not `--ci`

The `--ci` profile includes the **integration** tier, which drives the live Claude Agent SDK via AWS Bedrock. On a stock GitHub runner with no AWS credentials it pass-by-skips, so a green `--ci` run would silently mask the entire integration surface. Gating on the credential-free tiers keeps a green run meaningful. Integration can later be added as a non-blocking job guarded on repo secrets (OIDC → Bedrock). See `docs/reference/09-testing.md` for the tier map.

## Changelog guard

`tests/unit/t68` already guards version⇄changelog **sync** and heading **uniqueness**, but nothing guards against a PR **deleting** an existing entry (a bad rebase, a truncation, or a conflict resolution that clobbers someone else's changelog block). This CI-only guard diffs `## [x.y.z]` headings against the PR base ref and fails, naming any removed entry. It reuses t68's exact heading regex so the two never disagree.

## Verified locally

- `check`: typecheck + biome both exit 0; `package.ts --check` green on all 5 trees.
- `test`: **175 files, 4318 assertions, 0 failures** (`--smoke --unit -P 8`).
- `changelog-guard`: passes when entries are preserved (132 base entries, 0 removed); fails with exit 1 and names `## [2.5.5]` when a block is deleted.

## Not in this PR (repo config, apply manually)

Branch protection on `v2` — require the CI checks, require branches up to date before merge, required reviewers, and no self-review — is repository settings, not a workflow file. A maintainer with admin applies it. The required-status-check **contexts** GitHub exposes are the job **display names** (not the yaml job ids), so require exactly:

- `Contract checks (parity + typecheck + lint)`
- `Tests (smoke + unit)`
- `Changelog completeness`

(The `changelog-guard` job runs only on `pull_request` events, so it reports a context on every PR.)

## Versioning

No version bump: CI/test infra is not a user-visible surface, so per the AGENTS.md changelog policy it does not touch `aidlc-version.ts` / the README badge / `CHANGELOG.md` (test-only and internal-tooling changes live in commit messages).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,9 @@ jobs:
 
   changelog-guard:
     name: Changelog completeness
+    # pull_request only: the guard diffs against the PR event's base commit,
+    # which does not exist on workflow_dispatch (github.base_ref is empty there).
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -81,5 +84,9 @@ jobs:
       # Fail if this PR removes any `## [x.y.z]` heading that exists on the
       # base branch. Complements tests/unit/t68 (which guards version⇄changelog
       # sync and heading uniqueness, but not deletion of an existing entry).
+      # Diff against the FROZEN base commit the PR event fired from
+      # (pull_request.base.sha), NOT origin/<base_ref>: the branch ref moves as
+      # v2 accrues new entries while this PR is open, which would make the guard
+      # falsely report those newer entries as "removed" on any re-run.
       - name: Guard against deleted CHANGELOG entries
-        run: bun scripts/ci-changelog-guard.ts "origin/${{ github.base_ref }}"
+        run: bun scripts/ci-changelog-guard.ts "${{ github.event.pull_request.base.sha }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,78 @@
+name: CI
+
+# Deterministic PR gate for the v2 line. Runs the fast contract checks
+# (dist byte-parity, typecheck, lint) and the credential-free test tiers
+# (smoke + unit). The integration tier is deliberately excluded: it drives
+# the live Claude Agent SDK via AWS Bedrock and would either need repo
+# secrets or silently pass-by-skip here, so a green run stays meaningful
+# without it. See docs/reference/09-testing.md for the tier map.
+on:
+  pull_request:
+    branches:
+      - v2
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+# One in-flight run per PR ref; supersede older pushes to the same PR.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: Contract checks (parity + typecheck + lint)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: '1.3.14'
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      # `bun run check` = package.ts --check (dist drift guard) && typecheck
+      # (all three tsconfigs) && biome lint. Fails CI on a hand-edited or stale
+      # dist, a type error, or a lint warning.
+      - name: Run contract checks
+        run: bun run check
+
+  test:
+    name: Tests (smoke + unit)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: '1.3.14'
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      # smoke + unit are the deterministic, credential-free tiers (no AWS /
+      # Bedrock / live Claude). The runner exits non-zero on any failed file.
+      - name: Run smoke + unit tiers
+        run: bun tests/run-tests.ts --smoke --unit --parallel 8
+
+  changelog-guard:
+    name: Changelog completeness
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          # Need the PR base commit on disk to diff CHANGELOG headings against.
+          fetch-depth: 0
+
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: '1.3.14'
+
+      # Fail if this PR removes any `## [x.y.z]` heading that exists on the
+      # base branch. Complements tests/unit/t68 (which guards version⇄changelog
+      # sync and heading uniqueness, but not deletion of an existing entry).
+      - name: Guard against deleted CHANGELOG entries
+        run: bun scripts/ci-changelog-guard.ts "origin/${{ github.base_ref }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,13 @@ on:
   pull_request:
     branches:
       - v2
+    # Run on PR open, on every push to the PR branch (synchronize), and on
+    # reopen. This is GitHub's default set for `pull_request`; declared
+    # explicitly so the re-run-on-modify behavior is unambiguous.
+    types:
+      - opened
+      - synchronize
+      - reopened
   workflow_dispatch:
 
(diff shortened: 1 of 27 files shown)


apackeer (changes requested):
Thanks - this fills a real gap (the v2 line has had no PR test gate), and most of the design calls here are right: smoke+unit instead of `--ci` is the correct choice (`--ci` adds the integration tier at `run-tests.ts:131`, which would silently pass-by-skip without Bedrock creds), permissions are read-only, the concurrency-cancel matches the existing `docs.yml` convention, and the action SHA pins are honest (I verified `de0fac2e...` is `actions/checkout` v6.0.2 and `0c5077e5...` is `oven-sh/setup-bun` v2.2.0 via the GitHub API). The biome side-fix is legitimate and the dist copies regenerated cleanly, and the no-bump call is correct per the changelog policy.

Verified at the PR head (`be02c312`, and again with current v2 merged in): `bun run check` exits 0 (parity + all 3 tsconfigs + biome), smoke+unit 176 files / 4359 assertions / 0 failures, and the guard script's pass and fail paths both behave as documented when exercised directly.

Requesting changes for the first two items; the rest are smaller.

### P1: the guard compares against a moving branch, not the event's base SHA

`ci.yml:85` passes `origin/${{ github.base_ref }}`, which is wherever the base branch points at run time, not where it pointed when the PR event fired. Reproduced: this PR's head merges v2 at `ccf284b5` (CHANGELOG through 2.5.7); v2 has since advanced to `c38ba24a` (2.5.8-2.5.10). Running the guard at this PR's head against current `origin/v2` exits 1 claiming `## [2.5.8]`, `## [2.5.9]`, `## [2.5.10]` were removed - so re-running the currently-green check would fail, and every PR that sits open while v2 accrues changelog entries gets a false red on re-run. Use the frozen base instead:

```yaml
run: bun scripts/ci-changelog-guard.ts "${{ github.event.pull_request.base.sha }}"
```

(`fetch-depth: 0` already makes that SHA available.)

### P2: `workflow_dispatch` always fails the guard

`ci.yml:20` enables manual dispatch, but `github.base_ref` is empty outside `pull_request` events, so the guard receives `origin/` and exits 1 (`fatal: invalid object name 'origin/'` - reproduced). Note the P1 fix doesn't cover this: `github.event.pull_request.base.sha` is also empty on dispatch. Either gate the job:

```yaml
changelog-guard:
  if: github.event_name == 'pull_request'
```

or drop `workflow_dispatch` from the triggers.

### P2: branch-protection instructions name job ids, not check contexts

The PR body says "the job `name:`s match the required-check names" and lists `check` / `test` / `changelog-guard` - but those are the yaml job **ids**. The contexts GitHub exposes for required checks are the display names: `Contract checks (parity + typecheck + lint)`, `Tests (smoke + unit)`, `Changelog completeness` (confirmed via `gh pr checks` on this PR). An admin following the instructions verbatim would require checks that never report and block every merge. PR-body edit only.

### Should-fix: doc drift the PR body itself points at

The body says "See `docs/reference/09-testing.md` for the tier map", but that doc still describes CI as running `--ci` with the integration tier: `09-testing.md:21` ("When: CI push (--ci, every PR)") and `09-testing.md:203` ("CI pipeline ... `--ci`"), plus `docs/reference/11-contributing.md:76` and `:82` showing `--ci` as the CI command. Per the repo's Documentation Policy these should be updated in this PR, now that real CI runs smoke+unit only.

### Nits

- `ci.yml:66`: `--parallel 8` is inert here - `run-tests.ts:672` pins the smoke and unit tiers to serial (`effectiveParallel = 1`). Harmless, but worth dropping or commenting so nobody expects an 8-way speedup.
- `ci-changelog-guard.ts` copies t68's heading regex with a "kept in lock-step" comment, but nothing enforces that. A tiny pin test comparing the two patterns (or a shared exported constant) would make the claim real. Fine as a follow-up.

leandrodamascena:
Thanks for the thorough pass — the P1 catch is a real bug and I've fixed it along with P2a. Pushed in `e9b3214d`.

**P1 — moving base → frozen base.sha.** Confirmed and fixed. The guard now diffs against `github.event.pull_request.base.sha` instead of `origin/${{ github.base_ref }}`. I reproduced exactly what you described: at this PR's head, running the guard against current `origin/v2` reports `## [2.5.8]`/`## [2.5.9]`/`## [2.5.10]` as removed (they were never on this branch — v2 advanced after the PR opened), while guarding against the frozen merge-base (`ccf284b5`) passes clean with all 133 entries preserved.

One nuance worth noting for the record: on a normal `pull_request` run the default `actions/checkout` gives the *merge ref* (PR head merged onto current v2), whose working-tree CHANGELOG already carries 2.5.8–2.5.10 — so the false-red only bites because the guard reads `git show <base>:CHANGELOG.md` (the base tip) rather than the working tree. Either way, comparing against a moving ref is fragile by construction, so `base.sha` is the right fix regardless of which ref the runner checks out.

**P2a — workflow_dispatch.** Fixed by gating the job with `if: github.event_name == 'pull_request'`. Confirmed the failure first (`fatal: invalid object name 'origin/'`). `base.sha`/`base_ref` are both empty on dispatch, so the gate is the clean fix; kept `workflow_dispatch` on the other two jobs since they don't need a base.

**P2b — check contexts vs job ids.** Good catch — fixed the PR description to require the display names (`Contract checks (parity + typecheck + lint)`, `Tests (smoke + unit)`, `Changelog completeness`) instead of the yaml job ids, so an admin following it verbatim doesn't require checks that never report.

**Bonus (unrelated ask):** added a `paths:` filter to `docs.yml` so the site only rebuilds when something the build consumes changes (`docs/**`, `zensical.toml`, the link-rewrite script, `pyproject.toml`, `uv.lock`, the workflow itself) — not on every push to v2.

**Deferring to a follow-up** (all harmless, non-blocking):
- the `--parallel 8` nit on the `test` job — inert since smoke/unit pin serial; I'll drop it.
- the `--ci` doc drift in `09-testing.md`/`11-contributing.md` — real, but it's a test-strategy-doc edit separable from this CI PR; happy to sweep it here instead if you'd prefer it in-scope.
- the t68-regex "kept in lock-step" pin — agreed a shared exported constant would make the claim real.

Re-verified locally after the fix: `bun run check` exits 0, smoke+unit green, and the guard's pass/fail paths both behave against the frozen base.

apackeer (approved):
All three blocking items are fixed at `e9b3214d` - verified by running the guard at the new head, not just reading the diff. Approving. All three checks are green on the head commit.

**P1 (frozen base): confirmed fixed.** At this head, `bun scripts/ci-changelog-guard.ts ccf284b5...` exits 0 with all 133 base entries preserved, while the old form against current `origin/v2` still reproduces the false red (claims 2.5.8/2.5.9/2.5.10 removed) - so the switch to `github.event.pull_request.base.sha` is doing exactly the work it needs to. The API confirms this PR's event `base.sha` is `ccf284b5`. Your merge-ref nuance is right and worth having on the record: the false red only bites because the guard reads `git show <base>:CHANGELOG.md` rather than the working tree, but comparing against a moving ref is wrong by construction either way.

**P2a (workflow_dispatch): confirmed fixed.** The job-level `if: github.event_name == 'pull_request'` is the correct shape, and you're right it's the only viable one - `base.sha` is also empty on dispatch, so there was no ref to fall back to. Keeping dispatch on the other two jobs is fine since they need no base.

**P2b (check contexts): confirmed fixed.** The PR body now lists the three display names verbatim, plus the note that the guard reports a context on every PR. An admin can follow it as written. (I checked v2's protection via the API: the 1-approval review requirement and enforce-admins exist, but no required status checks are configured yet - so that section remains a real post-merge admin task, not already done.)

**On your doc-drift question: my preference is to fold it into this PR.** The PR body itself points readers at `docs/reference/09-testing.md` for the tier map, and the repo's Documentation Policy asks for same-commit updates. It's a small sweep: `09-testing.md` lines 21, 27, 68, 89, 203, 314 and `11-contributing.md` lines 76, 82 still describe CI as running `--ci` (integration tier included). Non-blocking on this approval - if you'd rather keep the PR frozen as-is, an immediate follow-up PR is acceptable, but in-PR is cleaner.

Two non-blocking notes for the record from this pass:

- The bonus `docs.yml` `paths:` filter is correct as far as it goes, but note the interaction with `docs-rewrite-links.ts` exiting 1 on dead out-of-tree link targets: before the filter, a PR that deleted a file linked from `docs/` failed the docs build on that PR; now the build skips, and the dead link surfaces later on whichever unrelated docs change builds next. An acceptable trade for not rebuilding the site on every push - just latent rather than immediate detection.
- `base.sha` freezes at PR-open time, so a CHANGELOG entry that lands on the base after this PR opened sits outside the guard's window (t68 still catches deletion of the newest entry via the version pin; only a mid-window middle-entry deletion escapes both). Strictly better than the moving ref and not worth more machinery - just noting the boundary.

Deferring the `--parallel` nit and the t68 regex pin (shared exported constant) to a follow-up: agreed on both.

PR #654, opened 2026-07-25
fix: inject resolved rule contents into the run-stage directive (#495)

### LINKED ISSUE #495: [Bug]: AI-DLC v2 `rules_in_context` is prose-only — the engine emits rule/memory **paths** in the run-stage directive but nothing forces the conductor to read them, so per-stage steering (org/phase memory) is silently skipped
## Description

The compiler resolves each stage's applicable rules into `rules_in_context` and the engine ships them in the `run-stage` directive, but it ships only the **file paths** — not the file **contents**. The conductor is then expected to open each path with a read tool before producing the stage's artifacts. Nothing in the engine enforces that read: it is a prose instruction in the stage protocol / persona, not a deterministic anchor. When the conductor skips the read (which it does, non-deterministically), the stage runs with **none of its steering applied**, and because the paths were "delivered" it looks like the rules were in context when they were not.

This is the same class of defect the framework already recognizes elsewhere — e.g. #485 (`recompose` "never under autonomous" is prose-only, no engine check) and #442 (conductor performs a gate action but skips the engine verb that records it). The stated design principle is **"determinism in tools, judgment in humans, knowledge in the LLM"**; `rules_in_context` delivery is currently a knowledge-in-the-LLM path with no deterministic anchor, so the knowledge silently fails to arrive.

## Observed (live run, Kiro IDE 1.0.89, greenfield MVP)

A `PostToolUse` payload probe captured every `read_*` and `execute_bash` on a real `/aidlc` run. Across the first three ideation stages:

- The engine emitted the correct `rules_in_context` on the `next` directive **every time** — for `intent-capture`, `feasibility`, and `scope-definition` the directive `tool_response` carried:
  ```json
  "rules_in_context": [
    "aidlc/spaces/default/memory/org.md",
    "aidlc/spaces/default/memory/team.md",
    "aidlc/spaces/default/memory/project.md",
    "aidlc/spaces/default/memory/phases/ideation.md"
  ]
  ```
- The conductor's actual reads in those three stages were: `intent-capture` → stage file + 2 persona files + memory-template + stage-protocol (5 files, **0 memory rules**); `feasibility` → stage file + `intent-statement.md` (**0 memory rules**); `scope-definition` → stage file only (**0 memory rules**).
- **Not one** of the four `rules_in_context` paths was read in any of the three stages.

When asked directly, the conductor confirmed it: *"I did not read those 4 files during intent-capture… rules_in_context listed them and I skipped the load."* It then read them (for the first time) only to answer the question, and its citations matched the on-disk files exactly — `org.md`'s `## Walking Skeleton` ("run the walking-skeleton Bolt first when greenfield") and `## Testing Posture` ("mvp → 80% line coverage") and `phases/ideation.md`'s `## Scope Discipline` ("No implementation details … in ideation artifacts") — rules that **should have shaped** the ideation artifacts but didn't. (The conductor even noted its own `feasibility` artifact listed a concrete tech stack, contradicting the `## Scope Discipline` rule it never loaded.)

This is **non-deterministic**, not a hard failure: a separate run on a different project (same engine, same graph wiring) *did* read the same four files and cite them. Same input, different behavior — the read is at the LLM's discretion.

## Root cause

`core/tools/aidlc-orchestrate.ts`, `buildRunStageDirective` (line numbers on 2.2.0):

```ts
const directive: RunStageDirective = {
  kind: "run-stage",
  stage: node.slug,
  ...
  rules_in_context: (node.rules_in_context ?? []).map((r) => r.path),  // :1072 — PATHS only
  ...
};
```

Only `r.path` is placed in the directive. The rule contents live on disk; whether they reach the model depends on the conductor choosing to read each path. There is no engine step, no gate, no hook that asserts the read happened.

Contrast with the persona delivery **twelve lines below**, which already does the deterministic thing — it reads the file and injects its **content**:

```ts
if (isFirstRunStageOfWorkflow(stateContent, node)) {
  const persona = readConductorPersona();               // :1087 — reads the FILE
  if (persona !== null) directive.conductor_persona = persona;  // :1088 — injects CONTENT
}
```

So the engine already has the pattern (`readConductorPersona()` → content in the directive); `rules_in_context` simply doesn't use it.

The directive contract only requires `rules_in_context` to be a `string[]` of paths (`checkStringArray(o, "rules_in_context", …)` in `aidlc-directive.ts`), and `resolveRulesForStage` (`aidlc-graph.ts:578`) is unit-tested for **which paths** resolve per scope (org/team/project/phase) — never for whether the conductor consumes them. Consumption is a live-run/LLM-behavior property, invisible to the directive-shape and graph-resolution tests. The gate/sensor layer doesn't check it either (steering application is advisory, not a sensor).

## Steps to Reproduce

1. Run any greenfield workflow with non-empty `org.md` / `phases/<phase>.md` memory (the shipped seed already has substantive `org.md`).
2. Reach `intent-capture` (or any early stage), let the conductor produce artifacts, and inspect what it actually read (payload probe, or just ask it to cite a line from each `rules_in_context` file it loaded *for that stage*).
3. Observe that the directive listed the memory paths but the artifacts were produced without reading them — and that a re-run may behave differently.

## Expected Behavior

The rules resolved into `rules_in_context` are reliably present in the model's context when it produces the stage artifacts, on every run.

## Actual Behavior

The engine delivers paths only; the conductor reads them or not at its discretion. Observed 0/4 memory rules read across three consecutive ideation stages in one live run; 4/4 in another — non-deterministic, silently skipped, no audit trace of the miss.

## Suggested Fix

Give `rules_in_context` the same deterministic anchor the persona already has: have the engine read each resolved rule file at directive-build time and inject its content, reusing the `readConductorPersona()` pattern. Keep the paths (backward-compatible) and add the content alongside:

```ts
// additive: paths stay; content is read at build time (skip empty/placeholder files)
rules_in_context: (node.rules_in_context ?? []).map((r) => r.path),
rules_content: (node.rules_in_context ?? [])
  .map((r) => ({ path: r.path, text: readRuleIfPresent(r.path) }))
  .filter((r) => r.text && r.text.trim().length > 0),   // team.md/project.md are comment-only placeholders → dropped
```

where `readRuleIfPresent` is the same best-effort `readFileSync`-guarded helper shape as `readConductorPersona()` (missing/empty file → omitted, directive stays well-formed). This removes the LLM's discretion from the steering-delivery path without touching rule resolution, the graph, or the gate model. Directive size grows by the resolved rule bytes per stage — the same order as the persona injection that already ships on the first stage.

## Additional Context

- **Related sibling defect (same root cause, reviewer surface):** the two reviewer agents (`aidlc-product-lead-agent`, `aidlc-architecture-reviewer-agent`) are expected to apply their `knowledge/<agent>/reviewing.md` checklist/output-format, but the IDE `.md` agent schema has **no resource/knowledge-preload field** (confirmed against `docs/custom-agents.md` and `docs/subagents.md`: frontmatter is `name/description/tools/model/mcpServers/permissions` for main agents and `name/description/tools/model/includeMcpJson/includePowers` for subagents — none preload a file). So `reviewing.md` reaches the reviewer only if the reviewer chooses to read it at runtime — the identical prose-only, non-deterministic path. A parallel fix is to inject `reviewing.md` content into the reviewer's delegation directive (engine-side), or absorb it into the reviewer `.md` body. `docs/subagents.md` also states steering files *do* apply to subagents deterministically, so a steering route exists but pollutes all agents.
- The design-principle framing ("determinism in tools … knowledge in the LLM", every "never" rule needs a deterministic anchor) is quoted verbatim from #485; this issue is a new surface of the same pattern, on the `rules_in_context` / steering-delivery path, which no existing issue covers (searched all 172 open+closed issues — no `rules_in_context` / knowledge-loading / memory-not-read report exists; #469 is the nearest but is about the reviewer `**Date:**` placeholder, a different surface).
- Found while running on Kiro IDE 1.0.89; the defective line is harness-independent shared core, so it projects into every harness `dist/` build.

wowzoo:
Fixes #495.

`buildRunStageDirective` ships `rules_in_context` as paths only, and nothing in
the engine forces the conductor to read them. Per-stage steering could therefore
be skipped silently — the paths listed, the files present with substantive
content, the artifacts written without them.

The engine already has the pattern this needs, twelve lines below the defect:
`conductor_persona` is delivered by reading `aidlc-common/conductor.md` at
directive-build time and baking in its contents, so no skill references it by
path. This applies the same treatment to the rule layers.

## The change

- **`aidlc-directive.ts`** — new `RuleContent` type; optional
  `rules_content?: RuleContent[]` on `run-stage`, registered in
  `RUN_STAGE_FIELDS`, validated by `checkOptionalRulesContent` (shape discipline
  mirrors `checkOptionalConsumesAbsent`).
- **`aidlc-orchestrate.ts`** — `readRuleContent` reads each resolved path
  best-effort; absent, unreadable, or placeholder-only yields no entry. Gated on
  `codekbCtx` so the ctx-less emit path stays byte-identical, mirroring
  `splitConsumesByPresence`'s own no-ctx skip.
- **`ruleTextIsSubstantive`** drops files whose every line is blank, a heading, a
  `>` blockquote, or an HTML-comment fence — the exact shape of the shipped
  `team.md` / `project.md` seeds, so an untouched install gains no directive
  noise. They appear as soon as they carry a real rule.

`rules_in_context` is unchanged and remains the authoritative roster;
`rules_content` is an additive subset over the files that carried content.

**Limitation — this reads the `default` space only.** `readRuleContent` resolves
each entry of `node.rules_in_context`, and those paths are baked by
`memoryDisplayPath()` whenever the graph is compiled — pinned to
`MEMORY_SPACE = "default"`, as is the `rulesDir()` the roster is walked from. Nothing
on this path consults the active-space cursor, so with a cursor on `teamB` the
injected content is still `aidlc/spaces/default/memory/`. That pin is the shipped
COMPILE/DISPLAY-family behaviour documented in `aidlc-graph.ts` and this PR does
not change it — but it is worth stating plainly, because injection makes the
default content deterministic rather than merely listed, and the conductor
demonstrably stops reading paths it has already been given (see the run below).
Making rule resolution cursor-aware is a separate, harness-wide question.

## No skill or protocol prose changes — and a live run explains why

I first rewrote the conductor's "read every file in `rules_in_context`" clause in
all five harness `SKILL.md` files, then measured whether it was needed. It is
not, so I reverted it and this PR touches no prose.

A live Kiro IDE 1.0.212 run, instrumented with a temporary `PreToolUse` probe on
`read_file|read_files` (the framework's own hooks cannot see reads — their
matchers are write/shell only), produced this in a single `read_files` call at
`intent-capture`:

| rule file | in `rules_content` | conductor read it |
|---|---|---|
| `org.md` (5439 chars) | yes | **no** |
| `phases/ideation.md` (1158) | yes | **no** |
| `team.md` (placeholder) | no | yes |
| `project.md` (placeholder) | no | yes |

The read boundary matched `rules_content` membership exactly: the conductor
skipped precisely the two files whose contents it had already received, and read
the two it had not — in the same call as the stage file and stage protocol, so
this was not laziness. Injection adds no duplicate read, and the existing prose
needs no change.

## Tests

New `t248-directive-rules-content.test.ts`, CLI-boundary (the builder has no
exports), 7 tests: content present and verbatim, subset-of-roster, placeholder
exclusion, a substantive `team.md` appearing, an emptied `org.md` dropping out,
the phase rule tracking the stage's phase, and the no-memory-tree case leaving
the field absent with routing unaffected. Registered in the coverage registry's
`none->cli` spawner list (the honesty ratchet requires the explicit edit).

## Verification

- `bun run check` — green (parity 5 harnesses, typecheck, lint 540 files)
- `bun tests/run-tests.ts --smoke --unit` — 178 files, 0 failed, 4395 assertions
- `bun scripts/ci-changelog-guard.ts <base>` — OK, 137 preserved, 1 new
- Fixture emit: `org.md` + `phases/ideation.md` injected, both placeholder seeds
  dropped; adding a rule to `team.md` makes it appear immediately

The integration tier's `t66` and `t89` also fail on base `257b43a3` (2.5.11 added
the `claim-sources` sensor without refreshing the designer-export golden fixture
or the sensor-import fixture dir). Both are outside the PR gate's tiers and
unrelated to this change; happy to file that separately.

Version bumped to 2.5.14 (2.5.12 is #615, 2.5.13 is #653) with a matching
CHANGELOG entry and README badge. `docs/roadmap.md` drops #495 from "Known gaps".

**Scope note:** the two edited files are shared core — byte-identical in all five
`dist/` trees — so the behaviour lands on every harness, not just one. Worth
naming the asymmetry: the harnesses with a native include (Kiro CLI's agent
`resources` glob, Claude's `@`-import stub, Codex's `AIDLC_RULES_DIR`,
opencode's `instructions`) already carry the memory tree in ambient context, so
for them this is a per-stage determinism guarantee over a roster the glob does
not distinguish rather than newly available content. On Kiro IDE, which has no
such native include, the injected content is the deterministic path. In both
cases the determinism is over the `default` space, per the limitation above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.14] - 2026-07-25
+
+Per-stage steering now reaches the workflow deterministically. The `run-stage` directive already listed the resolved rule paths in `rules_in_context`, but reading them was prose-only instruction, so an org or phase rule could be silently skipped and the stage's artifacts written without it. The engine now reads each resolved rule file at directive-build time and ships its contents in a new `rules_content` field, the same way `conductor_persona` is already injected. **Upgrade:** re-copy your `dist/<harness>/` shell into the project. No workflow, command, or file-layout change — rules you have already authored start arriving as content on the next stage.
+
+* The `run-stage` directive carries a new optional `rules_content` array of `{path, text}` entries — the on-disk contents of the rule files named in `rules_in_context`. `rules_in_context` is unchanged and remains the authoritative roster.
+* Rule files that hold only scaffolding (a heading, `>` guidance, or HTML-comment examples — the shape the shipped `team.md` and `project.md` seeds have before you affirm anything) are omitted, so an untouched install gains no directive noise. They appear as soon as they carry a real rule.
+* A missing or unreadable rule file is still not an error: the entry is dropped and the directive stays well-formed, with `rules_in_context` unaffected.
+
 ## [2.5.11] - 2026-07-24
 
 Intent Capture now keeps generated intent and stakeholder claims grounded in the user's description, confirmed answers, workflow-selected scope, or explicitly registered memory. Unsupported content is omitted, elicited, or surfaced as a human-owned assumption instead of being presented as fact. **Upgrade:** re-copy your `dist/<harness>/` shell into the project so the updated stage, Product Lead reviewer contract, and `claim-sources` sensor are installed.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.5.11-blue)
+![version](https://img.shields.io/badge/version-2.5.14-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-directive.ts
+++ b/core/tools/aidlc-directive.ts
@@ -55,6 +55,14 @@ export type DirectiveKind =
 // are read straight off the compiled stage-graph.json node; consumes/produces
 // carry RESOLVED aidlc-docs/... paths (the engine resolves vocabulary names →
 // paths at emit time; the conductor never re-derives them).
+
+// One rules_in_context file, with its on-disk content baked in. `path` is the
+// same projectDir-relative path that appears in rules_in_context; `text` is the
+// file's full contents at emit time. See RunStageDirective.rules_content.
+export interface RuleContent {
+  path: string;
+  text: string;
+}
 export interface RunStageDirective {
   kind: "run-stage";
   stage: string;
@@ -100,6 +108,18 @@ export interface RunStageDirective {
   // in-context, with no skill referencing that file by path. Absent on every
   // later directive (the persona persists in the session once delivered).
   conductor_persona?: string;
+  // rules_content — the CONTENT of each rules_in_context file that exists on
+  // disk with substantive (non-placeholder) text, read by the engine at
+  // directive-build time and injected here so the conductor receives its
+  // per-stage steering in-context WITHOUT having to read each path itself
+  // (the read was prose-only and skipped non-deterministically in live runs).
+  // Mirrors conductor_persona's deterministic-injection pattern. rules_in_context
+  // (the paths) is unchanged and authoritative; rules_content is an additive
+  // superset covering the files that carried real content. Absent on the
+  // ctx-less emit path (no projectDir to resolve/read against) and whenever no
+  // resolved rule file has substantive content — the directive stays
+  // well-formed without it.
+  rules_content?: RuleContent[];
   // next_stage: the display name of the in-scope stage that FOLLOWS this one,
   // resolved by the engine so the approval gate's Approve option can read
   // "Continue to <next_stage>" verbatim. null = this is the final in-scope
@@ -294,6 +314,7 @@ const RUN_STAGE_FIELDS = [
   "reviewer",
   "reviewer_max_iterations",
   "conductor_persona",
+  "rules_content",
   "next_stage",
   "unit",
   "consumes_absent",
@@ -456,6 +477,9 @@ function checkRunStageShared(
   checkStringArray(o, "sensors_applicable", kind, errors);
   checkString(o, "stage_file", kind, errors);
   checkOptionalString(o, "conductor_persona", kind, errors);
+  // rules_content: optional (present only on the live emit path when at least one
+  // rules_in_context file had substantive content). Each entry {path, text}.
+  checkOptionalRulesContent(o, "rules_content", kind, errors);
   // next_stage: optional-nullable on a run-stage directive. Present as a string
   // names the following in-scope stage; null means this is the final in-scope
   // stage; absent carries no name. So string OR null validates; any other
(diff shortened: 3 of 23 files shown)


apackeer:
Thanks, Keehyun. The diagnosis and initial implementation here were correct: required rule content must be delivered by deterministic tooling rather than left to discretionary path reads. I ran the focused test from this exact head; all 7 submitted cases passed.

Since this opened, #658 merged that same core direction into current `v2` and carried it through the additional failure paths found during review. The shipped implementation resolves the active space, fails closed on missing, unreadable, or invalid-UTF-8 required rules, preserves authored blockquoted policies, splits large bundles into integrity-bound `load-steering` parts, and enforces delivery at dispatched-worker boundaries. It also absorbs reviewer checklists into their generated agent bodies. The current focused suite passes all 32 cases.

The separate persona and supplemental-knowledge retrieval gap you called out is not lost; it is tracked in #694. Issue #495 itself is now resolved by the merged 2.5.33 implementation, although it remains open because changes merged to `v2` do not auto-close issues.

Please do not spend time rebasing or reworking this branch. I recommend closing #654 as incorporated by #658 and manually closing #495 as resolved. Thank you for the live evidence and initial patch that established the direction.

PR #658, opened 2026-07-25
fix: deliver stage steering deterministically - bounded rule chunks, dispatch-boundary enforcement, absorbed reviewer checklists (2.5.33)

### LINKED ISSUE #495: [Bug]: AI-DLC v2 `rules_in_context` is prose-only — the engine emits rule/memory **paths** in the run-stage directive but nothing forces the conductor to read them, so per-stage steering (org/phase memory) is silently skipped

apackeer:
Fixes #495.

## Problem

The engine resolved each stage's rules into `rules_in_context` but shipped only the file **paths** in the run-stage directive; whether the content ever reached the model depended on the conductor choosing to read each path. The issue documents live runs where 0/4 memory rules were read across three consecutive ideation stages - the directive "delivered" steering that silently never arrived, and the artifacts contradicted rules that should have shaped them. Two sibling surfaces had the same prose-only delivery: the reviewer agents' `reviewing.md` checklists, and the rules relayed into dispatched-subagent briefs.

## Fix - deterministic delivery on every hop

**Engine to conductor.** `next` now emits the active stage's substantive rules as bounded `load-steering` directives (ordered `{path, text}` content) before `run-stage`. Opaque `continue` tokens bind stage, workflow state, bundle hash, and next part; a fresh `next` restarts at part 1. Every directive stays below the common 28 KiB output floor - oversized rules split at Markdown headings, then UTF-8 boundaries, with **no size-based path fallback**. A missing, unreadable, or invalid-UTF-8 required rule stops before stage work with repair guidance (fail closed). Rules re-read every stage, so a §13 learning admitted mid-workflow reaches the very next stage. Comment-only template placeholders (`team.md`/`project.md` as shipped) are excluded until they carry an affirmed practice.

**Conductor to dispatched worker.** A new PreToolUse hook (`aidlc-dispatch-rules.ts`, backed by the shared resolver `aidlc-steering.ts` so the two hops cannot drift) appends the exact hash-marked rule bundle to every AI-DLC subagent brief that lacks it. Claude Code, Codex, and opencode rewrite the tool input (`updatedInput`); idempotent when the bundle is already present. Stage resolution precedence: explicit stage-file path in the brief, then the state file's Current Stage, then a unique slug mention.

**Kiro CLI degradation (deliberate).** Kiro cannot rewrite tool input, and a block-with-retry contract deadlocked in live testing (the conductor cannot reproduce a multi-KB bundle byte-exactly; every retry re-blocked). Kiro is also the one harness where the invariant already holds without the brief: every delegated agent preloads the full active memory tree natively via its `resources` glob. So on Kiro an incomplete brief proceeds with an advisory warning; an unloadable **required** rule (real missing steering, no preload fallback) still blocks with repair guidance. Kiro IDE keeps native memory-resource preload. This mirrors the existing harness-capability pattern (reviewer dispatch record, reviewer-scope enforcement).

**Reviewer checklists.** `knowledge/<agent>/reviewing.md` is absorbed into the two reviewer agents' generated bodies at build time on every harness (the reviewer set is derived from stage frontmatter, not hardcoded). The redundant per-agent knowledge globs and self-`.md` `resources` entries were removed from the Kiro reviewer JSONs - the checklist now arrives exactly once.

## Deliberate non-goal

Persona and supplemental **knowledge** files remain path-loaded (with actionable `context_warnings` for missing/unreadable optional files) pending a retrieval system - wholesale content injection of ~130KB knowledge trees is the wrong shape. This is the issue's "Additional Context" sibling surface, scoped out intentionally; follow-up issue to come.

## Verification

- Full tiers green on the branch: smoke+unit 178/178, integration green (t72 latency flake green solo; t145 codex-emitter case red only on a box with stray `/tmp` package roots - environmental).
- 9-slice live pre-merge gate run across Claude SDK/TUI, Kiro ACP/TUI, Codex exec, opencode; every red either fixed in-branch, flake-confirmed green solo, or **reproduced on clean v2 tip** (pre-existing: e2e `t113`, `t-tui-kiro-bugfix`, `t-tui-t51`/`t73`, `t-acp-kiro-journey`; these fail without this PR too). Kiro IDE slice skips clean on Linux (needs macOS + Kiro.app) - unverified there, low risk (no dispatch hook registers on the IDE).
- Live-verified behavior: the load-steering bundle delivered and applied on Claude SDK runs; the dispatch hook's exact-bundle rewrite observed on Codex; the earlier Kiro dispatch deadlock eliminated after the advisory change (zero blocked dispatches).
- New/updated tests: t248 (steering transport: substantive filter, chunking, budget, continue-token invalidation, dispatch rewrites per harness, stage-resolution precedence), t147 (Kiro adapter advisory contract), t238 (mob briefs carry rule content), t118/t53 (load-steering-aware test contracts), plus t89 fixtures + t66 golden realigned with the `claim-sources` sensor (pre-existing tip reds folded in per review).

## Notes for reviewers

- `BASH_MAX_OUTPUT_LENGTH=150000` + `AIDLC_DIRECTIVE_MAX_BYTES=120000` ship in the Claude settings so the directive line survives the tool-output cap; the engine default stays 28 KiB for the other harnesses.
- An adversarial two-reviewer pass ran mid-development; its P1 (a deliver-once/budget composition that silently downgraded overflowed files to path-reads) was reproduced, fixed, and superseded by the chunked design.
- Version 2.5.33 + CHANGELOG + README badge bump per policy (re-slotted from 2.5.12 at the merge-prep rebase onto v2 @ 57ae3d05; 2.5.32 was taken by #672).


leandrodamascena (changes requested):
I found six behavioral regressions that should be addressed before merge.

1. **[P1] Stop recovery skips steering content**

`core/hooks/aidlc-stop.ts:750-815` discards `rules_content` and forwards only the continuation token. The conductor is told to apply content "already returned," but that content was visible only to the hook subprocess. I reproduced this with a real active workflow: the block reason contained the token but none of `org.md`.

2. **[P1] Kiro IDE lacks the claimed preload fallback**

Kiro IDE delegates use Markdown frontmatter, while memory `resources` exist only in agent-v1 JSON documented as CLI-only (`harness/kiro-ide/manifest.ts:104-107`). The IDE also has no dispatch-rule hook registration. Consequently, its workers receive neither deterministic brief augmentation nor native memory preload.

3. **[P1] Large dispatch bundles produce truncated JSON**

`core/hooks/aidlc-dispatch-rules.ts:92-130,267-279` recombines all chunks into one unbounded hook response. With a 1.19 MB rule file, the hook exited successfully after emitting only 786,432 bytes of invalid JSON. This defeats the bounded engine transport precisely for rules large enough to require chunking.

4. **[P1] Blockquoted policies are silently dropped**

`core/tools/aidlc-steering.ts:23-33` classifies every `>` line as non-substantive, although `core/knowledge/aidlc-shared/rules-reading.md:24-33` defines only blank and HTML-comment bodies as empty. A `team.md` containing `> NEVER deploy without approval.` was absent from both `rules_content` and `rules_in_context` in my reproduction.

5. **[P2] Continuation tokens can be altered to skip chunks**

Tokens are unsigned base64url JSON (`core/tools/aidlc-orchestrate.ts:1668-1707`). Changing only `i` to `parts` passes the bundle, state, and route checks and immediately emits `run-stage`. I reproduced this with four parts by changing the first token index to `4`.

6. **[P2] Readable knowledge paths can overflow the global directive cap**

Only warnings are bounded. `inline_context_paths` remains unlimited (`core/tools/aidlc-orchestrate.ts:1399-1441`), while `emit()` rejects every directive above 28 KiB. Adding 260 valid knowledge files caused the final continuation to exit 1 with `refusing to emit a directive larger than 28672 bytes`.

Verification passed: 141 focused tests, package drift check, typecheck, lint, and `git diff --check`. The new tests do not cover these failure paths.

apackeer:
Addressed all six review findings in `5069d96`:

- Stop-hook recovery now carries the exact `rules_content`.
- Kiro IDE delegates receive active memory through always-included steering.
- Oversized dispatch responses fail before emitting partial JSON.
- Blockquoted policies are treated as substantive.
- Steering continuation tokens reject altered chunk indexes.
- `inline_context_paths` is bounded with an omission warning.

Also added regressions for these paths and corrected the related Kiro/plugin documentation and projections. Verification passed: generated-tree parity, typecheck, lint, `git diff --check`, and 238 focused tests across the affected files.

@leandrodamascena, please re-review when you have a chance.

leandrodamascena (changes requested):
I found five remaining issues that should be addressed before merge.

1. **[P1] Continuation tokens can still be forged to skip chunks**

The HMAC key is derived only from the public project path. I recomputed the MAC after changing the first token's index and skipped directly from part 1 of 4 to `run-stage`.

`core/tools/aidlc-orchestrate.ts:1693-1700,1831-1846`

2. **[P1] An invalid explicit stage path bypasses rule injection**

`promptStage()` returns `null` immediately when a path-shaped stage reference contains an unknown slug, instead of falling back to the workflow's `Current Stage`. I reproduced a dispatch completing with no rule augmentation by including `not-a-real-stage.md` in the brief.

`core/hooks/aidlc-dispatch-rules.ts:80-85`

3. **[P1] Plugin agents are excluded from deterministic delivery**

`isAidlcAgent()` requires the `aidlc-` prefix, excluding valid plugin agents such as `test-pro-metrics-agent`. Their dispatches therefore receive no active-stage rule bundle.

`core/hooks/aidlc-dispatch-rules.ts:47-53`

4. **[P2] Markerless or adversarially framed content is accepted as an exact bundle**

`hasExactBundle()` only checks whether each rule text appears somewhere in the prompt. It does not validate the generated digest or framing markers. I reproduced a prompt describing the exact rules as obsolete being accepted without appending the authoritative bundle.

`core/hooks/aidlc-dispatch-rules.ts:97-115,126-130`

5. **[P2] Kiro blocks valid oversized rules despite its advisory fallback**

The Kiro adapter treats every dispatch-hook exit 2 as an unloadable required rule. The hook also returns exit 2 when a valid bundle exceeds its output limit. I reproduced a 600 KB valid rule being blocked rather than proceeding through Kiro's native resource preload.

`harness/kiro/hooks/aidlc-kiro-adapter.ts:583-601`
`core/hooks/aidlc-dispatch-rules.ts:279-287`

The other previous findings appear resolved. Verification passed: package parity, typecheck, lint, smoke, all 167 unit files, and 225 focused tests. These adversarial paths are not covered by the current tests.

apackeer:
Addressed all five remaining review findings in `80cd9387`:

- Continuation HMACs now use a random 32-byte machine-local key (`0600` on POSIX), including an ignored pre-intent fallback; the public project path can no longer re-sign altered chunk indexes.
- Unknown explicit stage paths now fall back to the workflow's `Current Stage`.
- Deterministic dispatch delivery recognizes installed plugin agents through the agent roster.
- Idempotency now requires the complete digest-framed generated bundle; markerless or adversarially framed rule text cannot suppress injection.
- Kiro distinguishes valid oversized bundles from unloadable rules: oversized bundles use native active-memory preload, while missing/unreadable rules still block.

Added regressions for each path and updated the reference docs/changelog. Verification passed: package parity, TypeScript typecheck, Biome lint, `git diff --check`, and the full smoke + unit profile (179 files, 4,445 assertions). Focused steering tests passed 32/32. The integration profile passed aside from the documented live-LLM `t72-stage-reverse-engineering` timeout; the unrelated `t92` PATH issue passed all 47 tests when rerun with the repository toolchain on `PATH`.

leandrodamascena (approved):

PR #664, opened 2026-07-28
feat(plugins): graduate adds.scopes + edge-aware stage-number seeding (2.5.28)

apackeer:
## What

Graduates the plugin contribution seam's `adds.scopes` surface from deferred (declared -> drop-logged) to implemented, and makes new-stage number seeding edge-aware so multi-stage plugins compile regardless of filename order. Version 2.5.12.

Doc 18 has listed `adds.scopes` as designed-but-deferred with the note "when these graduate they set-union like the others" - this PR is that graduation. It is the mechanism half of porting an external methodology (aws-samples/sample-ai-plc) as an AIDLC plugin; the plugin itself is a separate follow-up PR. Everything here is generic mechanism work exercised by synthetic fixtures - no ai-plc content.

## Why

A plugin can ship its own stages and its own scopes, but until now a plugin scope could only contain plugin stages: the seam that would put an existing CORE stage under a plugin scope (`adds.scopes`) was drop-logged, never merged. That blocks the natural shape for a methodology plugin - e.g. a discovery flow that runs in place of core Ideation and then continues into core Inception onward under one scope.

Separately, compile seeded numbers for new slugs in file-walk (alphabetical) order. A multi-stage plugin whose flow order disagrees with its filename order (dependency `product-strategy` vs. dependent `discovery-document`) seeded numbers that contradicted its own `requires_stage` edges and hard-failed the lower-numbered-dependency compile invariant - the plugin could not load at all. test-pro never hit this because its two stages happen to sort correctly.

## How

**`adds.scopes` merge (compose hook).** Set-unions into the target stage's `scopes:` frontmatter, recorded in the per-plugin contribution sidecar, and the recompiled scope grid routes the core stage under the plugin scope. Two guard rails, each violation dropped-with-log (doctor-visible), never merged:
- ownership - the scope must belong to the contributing plugin (named exactly `<plugin>` or `<plugin>-` prefixed); a contribution cannot weld a core stage into a core or foreign-plugin scope;
- installed identity - the scope's `scopes/<name>.md` must already be installed, so a typo never becomes an all-SKIP phantom scope.

`select-plugins` strips merged scope memberships on disable (same sidecar mechanism as produces/sensors/consumes); re-enable + recompose restores them.

**Edge-aware number seeding (graph compile).** New slugs are seeded per phase by Kahn-ordering the batch over its own `requires_stage` edges, then assigning next-free contiguous indices. An authored `number:` acts only as a relative-ordering tiebreak among independent stages - its absolute value never lands in the graph. So the engine owns all number values, "number is display-only" and "a plugin claims no range" hold fully, and two uncoordinated plugins cannot collide on numbers. A `requires_stage` cycle among new stages is a loud compile error. Already-pinned rows are untouched; a bare-core install compiles byte-identically.

`adds.requires_stage` remains deferred (still drop-logged), unchanged.

## Testing

- t188 +3 tests: the merge/grid/strip round-trip (both owned-name shapes in one contribution); the foreign + phantom scope guards; an adversarial seeding case where filename order is opposite to edge order (63 assertions total).
- Full smoke+unit tier: 178 files PASS. Integration slice (t188, t224, t222, t65, t184, t-custom-harness-compile, plugin content suites): PASS.
- `bun scripts/package.ts --check` clean; `bun run typecheck` clean.
- Validated end-to-end by composing a 9-stage plugin (with internal chain edges and filename order opposite to flow order) plus test-pro into the same fresh install: both compose with zero drops, 43 stages, numbers seeded in flow order, disable/re-enable round-trips.

## Notes for reviewers

- Docs updated: reference 18 (merge semantics, surface table, scopes section, status) and the authoring guide (the number-semantics paragraph now explains hint-not-claim; the `adds.scopes` bullet moves to implemented).
- Known pre-existing red, unrelated: t66's designer-export golden is stale on the v2 base (missing 2.5.10's claim-sources fields) - reproduced on unmodified origin/v2.
- CHANGELOG + version + README badge bumped to 2.5.12 per policy.


scripts/plugin-hooks-template/compose.ts
```diff
@@ -1365,12 +1398,13 @@ try {
       recordContrib(target, "produces", addedProduces);
       recordContrib(target, "sensors", addedSensors);
       recordContrib(target, "consumes", addedConsumes);
+      recordContrib(target, "scopes", addedScopes);
```
leandrodamascena, line 1544: **P1 — Serialize contribution merges across plugin hooks**

_(Anchoring here on the new `scopes` sidecar record; the mutation this concerns is the `writeFileSync(stageFile)` at ~1480 and the sidecar write at ~1492.)_

I saw that each plugin's compose hook reads and rewrites the same stage file with no lock: the `writeFileSync(stageFile)` and the sidecar write both run unprotected, while the recompile at `aidlc-graph.ts:2160` already runs inside `withAuditLock`. So the transaction is only half-locked — two concurrent composes can both succeed and both write sidecars claiming their scope merged, but the last writer's stage file wins and silently drops the other plugin's `adds.scopes` merge.

I reproduced this with two simultaneous compose runs (the bun-fallback path a plugin's SessionStart hook takes when `aidlc` isn't on PATH): both exited 0, both sidecars were present, and the final stage contained only one scope. Please extend the existing `withAuditLock` to cover the whole stage + sidecar + compile transaction, not just the compile.

core/tools/aidlc-utility.ts
```diff
@@ -636,6 +637,7 @@ function stripDisabledPluginContributions(
         if (record) {
           if (record.produces?.length) content = removeListValues(content, "produces", new Set(record.produces), false);
           if (record.sensors?.length) content = removeListValues(content, "sensors", new Set(record.sensors), false);
+          if (record.scopes?.length) content = removeListValues(content, "scopes", new Set(record.scopes), false);
```
leandrodamascena, line 661: **P1 — Use the same lock for selection changes and compose**

I saw that `select-plugins` strips scopes and deletes the sidecar without coordinating with compose, so a concurrent compose can reintroduce a scope after its plugin is disabled.

My probe finished with `plugins: ["aidlc"]`, yet the disabled plugin's scope remained welded into the core stage and its sidecar was gone. Because re-enable + recompose never recreates the sidecar, no later disable can clean up that orphaned membership. Selection changes and compose need to share one lock.

leandrodamascena (changes requested):
Thanks for this work, Arden — the scope-ownership checks and the topological seeding are well done, and all the focused tests pass. I saw, though, that the stage mutations run unlocked while only the compile is serialized — that gap allows silent lost updates and lets disabled-plugin contributions stay active. Two P1s below worth a second look before merge.

**Verification:**
- t188-plugin-compose: 65 passed
- t224-plugin-selection: 12 passed
- `bun scripts/package.ts --check`: clean

apackeer:
Pushed 3125341e on top of the review fixes - it hardens the compose/selection seam this PR graduates, plus closes the remaining review nits.

**Concurrency: one workspace lock across compose, select-plugins, and graph compile.** The adds.scopes merge made contribution state multi-writer in practice (two plugin compose hooks at SessionStart, or a select-plugins racing a compose), so:

- A compose hook now acquires the workspace audit lock for its whole run (stage mutations, contribution sidecars, compile, runner regeneration) and hands ownership to the tools it spawns via `AIDLC_WORKSPACE_LOCK_OWNER_PID`. `graph compile` accepts the inherited lock only after verifying the declared owner is its live parent AND the stamped lock holder (new `auditLockOwnedByProcess`, unit-tested in t161).
- `select-plugins` wraps its whole transaction (validation, strip, selection write, recompile, rollback) in `withAuditLock`, with a ~60s acquire budget since it legitimately queues behind a full compose. `withAuditLock` gained optional `maxRetries`/`retryMs` params; every other caller keeps the old default.
- New t188 coverage: two concurrent composes both land their scope merges and sidecars; select-plugins blocks behind a live compose and cannot leave an orphaned scope membership; the lock dir is always released.

**Fixes folded in:**

- Scope ownership and name reads go through `frontmatterScalar`, so a quoted `plugin:`/`name:` scalar (as YAML formatters emit) no longer fails the exact-match guard closed on the plugin's own scope. Test added.
- `buildGraphStage` dedupes `requires_stage`, so runtime topo sees the same edge cardinality as the compile-time number seeder; the flow-order test now also asserts the compiled edge list and runs `graph topo`.
- Compose pins the resolved absolute `AIDLC_PROJECT_DIR` into its spawned-tool env. Previously a relative `CLAUDE_PROJECT_DIR` re-resolved against the child's cwd, sending the compile to `<proj>/<proj>`: the frontmatter merge landed but the scope never reached the grid (reproduced end-to-end on a fresh dist install). The relative-project test now asserts zero drops and the scope EXECUTE in the recompiled grid, so it fails without the pin.
- Coverage registry regenerated for the new tests; stale comment enumerations updated (structural adds now list `scopes`; removed the stage-schema claim of a phase-prefix compile cross-check that the topo-seed rework deleted).

No new version bump - this squashes into 2.5.12's story; the CHANGELOG gained one bullet under the existing heading.

Verified from a clean worktree at the tip: `bun run check` (typecheck x3 + biome + `package.ts --check` parity) clean; t188 69/69 + t161 green; full smoke+unit tier 178 files / 4397 assertions green.

leandrodamascena (changes requested):
Thanks for the follow-up. The new shared lock closes the two previously reported races when every caller uses the same project-path spelling. I found three remaining P1 paths that bypass or corrupt the graduated mechanism, plus three P2 parsing/recovery issues.

1. **[P1] A symlink alias bypasses the workspace mutex**

Lock identity is derived from the lexically resolved project path, not its real path. Two processes addressing the same workspace through its physical path and a symlink therefore hash to different lock directories and both acquire successfully while mutating the same stages, sidecars, graph, and grid. I reproduced this directly: `auditLockDir(real) !== auditLockDir(alias)`, and both `acquireAuditLock(..., 0, 1)` calls returned true. This keeps both previously reported races reachable through path aliases.

`scripts/plugin-hooks-template/compose.ts:36-44`
`core/tools/aidlc-lib.ts:2963-2974,3196-3228`

2. **[P1] Plugin content can spoof the manifest identity and claim another plugin's scope**

`pluginNameFromRoot()` prefers the first `plugin:` field found anywhere under stages/scopes/contributions over the host plugin manifest. The ownership guard then compares the target scope owner against that spoofed name, and the sidecar is also keyed by it. I reproduced a plugin whose manifest name was `attacker` but whose contribution declared `plugin: victim`: it merged `victim-scope` with no drop and created victim-attributed contribution state. Disabling plugins subsequently treated the merge as belonging to `victim`, not the installed attacker package.

`scripts/plugin-hooks-template/compose.ts:101-146,1255-1276,1406-1437`

3. **[P1] Duplicate scope identities inside one incoming plugin are installed and poison the runtime roster**

The installed-name collision precheck snapshots the destination roster once, but `copyTreeNoClobber()` does not add names copied earlier in the same source tree to that roster. Two incoming files declaring the same `name:` both pass. I reproduced `alpha.md` and `beta.md` both declaring `duplicate-scope`: compose exited successfully and installed both files; scope-table and runner generation then failed with `Duplicate scope name`, leaving the install malformed.

`scripts/plugin-hooks-template/compose.ts:465-493,934-971,1221`

4. **[P2] Quoted `adds.scopes` entries are rejected even though quoted scope identity is supported**

The new scalar helper correctly handles quoted `name:` and `plugin:` values, but the `adds.scopes` list parser still accepts only an immediately unquoted kebab token. I reproduced `- "quoted-scope"` being parsed as zero entries and drop-logged. The quoted-scalar test covers the scope file, not the contribution value, so it does not exercise this mismatch.

`scripts/plugin-hooks-template/compose.ts:1336-1344`

5. **[P2] Disable can remove a scope membership that predated the plugin**

Compose compares list entries using their raw spelling, so an existing `- "quoted-existing-scope"` is considered different from the contributed unquoted value and receives a second entry that is recorded in the sidecar. Disable-time stripping canonicalizes quotes and removes both entries. I reproduced the authored quoted membership disappearing after `select-plugins aidlc`, violating the sidecar contract that only values actually added by compose are stripped.

`scripts/plugin-hooks-template/compose.ts:1012-1015,1254-1276`
`core/tools/aidlc-utility.ts:560-565,636-646`

6. **[P2] A compile failure can permanently wedge newly copied plugin stages**

Compose copies new stages before compilation. If compilation fails, such as on the newly detected stage cycle, the files remain installed and only a retry marker is written. After correcting the plugin source, no-clobber treats the corrected stage as a collision and retains the invalid installed copy, so every retry compiles the same broken stage until the user manually removes it.

`scripts/plugin-hooks-template/compose.ts:934-971,1607-1618`

Verification on head `403f6490`:

- Package parity passed for all five harnesses and `git diff --check` was clean.
- The symlink-lock, identity-spoofing, duplicate-scope, quoted-entry, and disable-strip cases were reproduced locally against generated installs.
- The focused lock tests passed. The integration test file could not start in this isolated worktree because its `smol-toml` dependency was not installed; required CI is green at this head.

The canonical-path compose/compose and compose/select races from the previous review appear fixed. The findings above concern alternate lock identity and plugin-input cases not covered by the new concurrency tests.

apackeer:
@leandrodamascena Addressed all six findings from the latest review in `0d95f26e`:

- Workspace lock identity now canonicalizes the project realpath, so symlink aliases share the same mutex.
- Plugin identity comes from the host manifest (including the OpenCode manifest shape); owned stages, scopes, agents, and contributions with a mismatched `plugin:` are dropped.
- Incoming scope/agent names are reserved as files are accepted, so duplicate identities in one plugin tree are rejected before compile.
- Quoted `adds.scopes` entries are parsed and merged canonically.
- Structural list union and disable-time stripping preserve equivalent pre-existing quoted entries and remove only the recorded plugin addition.
- Compose snapshots copied files and contribution writes; graph compile failure rolls them back before writing the retry marker, allowing corrected source to compose on retry.

Regression coverage was added for each path. Verification at the pushed head:

- `t188-plugin-compose`: 73/73
- `t161-per-intent-lock-reaper` + `t224-plugin-selection`: 25/25
- smoke + unit: 178 files, 4,398 assertions
- TypeScript typecheck, Biome lint, and `bun scripts/package.ts --check`: clean

Ready for re-review.

apackeer:
Rebased onto the current `v2` branch and force-pushed with lease. I also re-bumped this PR from `2.5.12` to `2.5.28` (`v2` is at `2.5.26`, and `2.5.27` is already claimed by #646), regenerated all distributions, and refreshed the coverage registry.

Validation completed locally:
- `bash tests/run-tests.sh --smoke --unit`: 180 files, 4,452 assertions, 0 failures
- `bun run check`: package parity, all TypeScript configs, and Biome passed
- Focused plugin tests: t161 13/13, t188 73/73, t224 12/12
- `git diff --check` passed

All substantive review findings are addressed. @leandrodamascena, rereview requested.

leandrodamascena (changes requested):
Round 3 looks good. I re-ran the two prior P1 repros against the reworked code and both stay **closed**: the concurrent-compose lost-update still serializes (waiter cleanly acquired after the holder released, no throw), and select-plugins-vs-compose still runs under the same 600-retry `__workspace__` lock — no orphan. The `rollbackComposeWrites` transaction is a nice add, and I tried to break the new recovery path (re-stamp double-owner, deadlock, double-apply on retry) — all refuted at the code level: the compile child only *reads* the owner stamp (never re-stamps), rollback is idempotent behind `composeTransactionOpen`, and merges dedup via Set. Solid work.

One thing still open, and the `realpathSync` change actually made it more reachable:

## [P2] intent-birth and recompose still lock on the default ~5s budget while compose/select-plugins hold ~60s

Both `handleIntentBirth` (`aidlc-utility.ts:3543`) and the recompose/scope-change path (`aidlc-utility.ts:4573`) call `withAuditLock(projectDir, () => {…})` with **no trailing budget arg** → the default 50×100ms (~5s), whereas `COMPOSE_LOCK_RETRIES` and `SELECT_PLUGINS_LOCK_RETRIES` are both 600 (~60s). I reproduced it cross-process: with a live 600-budget compose holding the `__workspace__` sentinel, an intent-birth **threw at 5269ms** — "Failed to acquire audit lock … after retries."

And the new `realpathSync` canonicalization (`aidlc-lib.ts:2964`) — which is itself a correct fix, it closes a real two-bucket lost-update hole — now collapses a symlink-path birth and a physical-path compose onto the *same* lock. They used to key different buckets and run in parallel, so the budget mismatch bites **more** often now, not less.

The fix is a one-liner at each site: give intent-birth and recompose the same 600 budget the other `__workspace__` holders use, so a legitimate birth/recompose waits behind a long compose instead of dying.

## Note (not a blocker): SIGKILL-mid-compose recovery gap still open

This commit only added `realpathSync` to the lib — and no `process.on("exit")` net would help anyway, since it doesn't fire on SIGKILL. It stays P3: `rollbackComposeWrites` is parent-in-memory + best-effort, so a SIGKILL mid-rollback can leave a partial restore. But the compile child refuses an inherited lock when its ppid doesn't match, the reaper frees a dead-PID owner immediately, and the affected derived targets (graph/grid) regenerate next session. Fine for a follow-up.

**Verification:** t188 73 pass, t161 20 pass, t68 green (2.5.28, single heading), `bun scripts/package.ts --check` clean across all 5 trees.

apackeer:
Addressed the final P2 in `9b16a3bc`.

- `intent-birth`, `recompose`, and `select-plugins` now share the 600-retry (~60s) workspace-lock budget, so workflow mutations wait behind a legitimate compose instead of failing around five seconds.
- Added a cross-process regression that holds the live workspace lock past the old budget and queues both `intent-birth` and `recompose`. It reproduced the old behavior with both commands exiting 1, then passed after regeneration.
- Regenerated all five harness distributions and updated the existing `2.5.28` changelog entry; no additional version bump.

Verification at the pushed head:
- `t188-plugin-compose`: 74/74
- `t161-per-intent-lock-reaper`: 13/13
- `t194-recompose`: 17/17
- `t20` + `t68`: 24/24
- smoke + unit: 180 files, 4,452 assertions, 0 failures
- `bun run check`: package parity, all TypeScript configs, and Biome clean
- `git diff --check`: clean

Ready for re-review.

leandrodamascena (approved):

PR #666, opened 2026-07-28
fix: route plugin slash commands through terminal dispatch (v2.5.45)

leandrodamascena:
## Summary

Fix `/aidlc plugin ...` slash commands being interpreted as freeform workflow requests.

Before this change, commands such as `/aidlc plugin list` reached `aidlc-orchestrate.ts next plugin list`. Because the orchestrator did not classify the leading `plugin` noun as a terminal command, it could ask the user to compose a scope or advance an active workflow instead of invoking the plugin utility.

This change introduces one shared plugin-command grammar used by the public dispatcher, orchestration engine, and Kiro pre-LLM interceptor.

## Behavior

The following commands now dispatch directly to their existing utility handlers:

| Slash command | Utility command |
| --- | --- |
| `/aidlc plugin list` | `aidlc-utility.ts plugin-list` |
| `/aidlc plugin list --json` | `aidlc-utility.ts plugin-list --json` |
| `/aidlc plugin sync` | `aidlc-utility.ts plugin-sync` |
| `/aidlc plugin select [names]` | `aidlc-utility.ts select-plugins [names]` |
| `/aidlc plugin help` | Global AI-DLC help |

Plugin commands are terminal utilities. They do not enter the scope-selection funnel, resume a workflow, or advance the active stage.

Missing and unknown plugin verbs now produce deterministic errors rather than being interpreted as intent descriptions.

## Implementation

- Add `parsePluginCommand()` as the shared parser for the `plugin` noun.
- Extend `classifyTerminalCommand()` with the `plugin-verb` command family.
- Use the shared parser from the public `aidlc` dispatcher.
- Add an orchestrator terminal branch before workflow state inspection.
- Preserve trailing arguments such as `--json` and plugin selection names.
- Extend Kiro turn-scoped terminal latching to cover plugin commands.
- Prevent a Kiro conductor retry from issuing a bare `next` after a plugin command.
- Add `plugin select [names]` to user-facing help.
- Correct the CLI guide stale direct-only plugin-selection wording.
- Regenerate Claude, Codex, Kiro CLI, Kiro IDE, and OpenCode distributions.

## Regression Coverage

- Plugin list routing with `--json`.
- Plugin sync routing.
- Plugin selection argument preservation.
- Plugin help routing.
- Missing and unknown plugin verb errors.
- Shared terminal-command classification.
- Public dispatcher parity.
- Kiro off-band dispatch and `plugin-verb` latch creation.
- Active-workflow protection preventing plugin commands from emitting `run-stage`.
- Coverage registry freshness.

## Version and Upgrade

Bumps AI-DLC from `2.5.11` to `2.5.12` across all generated harness distributions.

**Upgrade:** re-copy the appropriate `dist/<harness>/` shell into the project.

## Validation

- `bash tests/run-tests.sh --unit`: 166 test files, 3,287 assertions, zero failures.
- `bun run check`: package parity, TypeScript, and Biome passed.
- `bun scripts/build-binaries.ts`: native binary build passed.

The native artifact was produced under `build/binaries/native/aidlc`. Release binary artifacts are ignored and are not committed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.45] - 2026-08-06
+
+Plugin management commands now route deterministically through every `/aidlc` slash-command harness instead of being misread as freeform workflow requests. **Upgrade:** re-copy your `dist/<harness>/` shell into the project so the shared plugin parser and harness routing updates are installed.
+
+* `/aidlc plugin list [--json]`, `/aidlc plugin sync`, and `/aidlc plugin select [names]` now dispatch to their existing utility handlers without asking to compose a scope or advancing an active workflow.
+* `/aidlc plugin help`, a missing plugin verb, and unknown plugin verbs now produce deterministic help or errors across the dispatcher, orchestrator, and Kiro pre-LLM interceptor.
+* Kiro terminal-command interception now latches plugin utilities for the current turn, preventing a conductor retry from rolling the workflow forward after the plugin command completes.
+
 ## [2.5.44] - 2026-08-06
 
 Human-readable AI-DLC artifacts now follow the conversation language instead of defaulting to English. The framework rule layer gains four `## Mandated` rules in `org.md` that define how the conversation language is resolved, why it stays stable across short turns like `Approve`, which artifacts are localized, and which protocol tokens stay English. The rules travel to delegated agents and reviewers inside the active-stage rule bundle the dispatch-rules hook already delivers, so a subagent that never sees the conversation can still resolve the language from the rule layer, `aidlc-state.md`, or the upstream artifacts it consumes. **Upgrade:** fresh installs receive the rules with their seeded `default` space. An existing workspace keeps its own memory tree — runtime self-healing intentionally does not overwrite one — so merge the four rules into each `aidlc/spaces/<space>/memory/org.md` by hand and start a fresh session for them to load.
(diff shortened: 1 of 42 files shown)


apackeer:
Thanks, Leandro. This bug still exists on current `v2`, and the shared parser and terminal-routing direction are right.

One path needs correction before merge: Kiro’s compiled-binary route sends the internal names `plugin-list`, `plugin-sync`, and `select-plugins` to the public binary, which expects `plugin list`, `plugin sync`, and `plugin select`. It currently relays the resulting unknown-command error as though the utility completed. Could you update that translation and add coverage with `AIDLC_COMPILED_EXECUTABLE` set?

The same-turn latch should also render `plugin …` rather than `--plugin …`. There is no need to rebase solely for the branch drift; we can handle current-`v2` integration and release metadata afterward.

leandrodamascena:
Thanks for catching this. I updated Kiro’s compiled-executable path to translate the internal utility names back to the public CLI grammar: `plugin-list` → `plugin list`, `plugin-sync` → `plugin sync`, and `select-plugins` → `plugin select`. The TypeScript utility path remains unchanged.

I also fixed the same-turn latch label so plugin commands render as `plugin …`, not `--plugin …`, and added regression coverage with `AIDLC_COMPILED_EXECUTABLE` set for `list --json`, `sync`, and `select` with multiple names.

Validation: 100 focused tests passed, package parity passed, typecheck passed, coverage registry passed, and Biome passed on all changed authored/test files.

Pushed in `44a056ec`.

apackeer (approved):
Approved at 44a056ec. The August 4 follow-up addresses the compiled Kiro argv translation and same-turn latch labeling, with regression coverage. I reviewed the full authored diff; package parity, typecheck, Biome, and the affected unit tests pass locally after timeout-only reruns.

Before merge, resolve the current-v2 conflicts, refresh the version/changelog/badge and generated distribution/coverage metadata, and rerun checks on the integrated head.

PR #668, opened 2026-07-29
test: --no-llm CI profile + hermetic git in run-tests

apackeer:
## What

Two test-runner hardening changes, ported from the SEEK AI-DLC pilot fork (SEEK-Jobs/aide-ai-dlc-pilot-external, drift commit by Helen Giapitzakis), where both have been running as the fork's CI profile. Test infrastructure only: no core/, no harness/, dist/ untouched, no version bump per the changelog policy.

### 1. --no-llm flag (or AIDLC_NO_LLM=1)

Forces the Claude gate closed even when the claude CLI is on PATH, so real-LLM integration/e2e files take their per-file SKIP path while deterministic tests in those same tiers still run. Today the only way to avoid live-LLM cost/flakiness in --integration is to skip the tier entirely, losing its deterministic files too.

This gives CI a full-tier deterministic profile. It would pair well with the new v2 PR gate: the ci.yml workflow currently runs smoke/unit, and --no-llm makes adding the deterministic integration tier affordable.

Verified: with claude on PATH, `--integration --no-llm` skipped 24 Claude-required files and ran 79 deterministic files; a banner names the forced-closed gate so a skipped-green run cannot be mistaken for live coverage.

### 2. Hermetic git for spawned test files

GIT_CONFIG_GLOBAL/GIT_CONFIG_SYSTEM point at the platform null device in every spawned test file's env, so test git reads no developer/system gitconfig. The worktree/Bolt/swarm/multi-repo tiers drive real git against throwaway temp repos; on a machine with commit.gpgsign=true and an SSH signer (e.g. 1Password's op-ssh-sign), a fixture seed commit tries to sign and dies "Could not connect to socket" whenever the signing agent is down - a suite failure unrelated to the code under test. Fixtures already pass -c user.email/-c user.name per commit, so identity is unaffected.

## Changes

- tests/run-tests.ts: noLlm flag + env default, usage help, gate-closed branch ahead of the missing-CLI branch, NULL_DEVICE git config env
- tests/smoke/t05-run-tests-parallel.test.ts: two self-tests driving the real runner (flag and env forms), each proving a claude-required file SKIPs while a deterministic sibling PASSes
- docs/reference/09-testing.md: flag documented

## Testing

- smoke+unit: 178 files, 0 failed
- --integration --no-llm: 24 skipped / 79 ran, deterministic files all green (3 pre-existing base reds on v2 - t66 designer-export golden, t89 claim-sources sensor fixture, t163 reaper steal race - reproduce identically on 9f914544 without this change; happy to file them as a separate issue)
- bun run typecheck clean

### DIFF
--- a/docs/reference/09-testing.md
+++ b/docs/reference/09-testing.md
@@ -317,9 +317,8 @@ bash tests/run-tests.sh       # POSIX compatibility wrapper
 
 # Output modifiers
 --verbose       # Write per-test logs to tests/logs/
---no-llm        # Force the Claude gate closed: real-LLM integration/e2e files
-                # SKIP while deterministic tests in those tiers still run.
-                # Also via AIDLC_NO_LLM=1.
+--no-llm        # Force all live-model gates closed while deterministic
+                # integration/e2e tests still run. Also via AIDLC_NO_LLM=1.
 --debug         # Implies --verbose; streams per-test output and writes SDK/TUI
                 # driver traces to tests/logs/
 --filter PAT    # Only run tests whose filename matches extended regex PAT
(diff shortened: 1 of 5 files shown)


leandrodamascena (changes requested):
1. **[P2] `--no-llm` still permits live Codex tests**

   `tests/run-tests.ts:287` closes only the Claude gate. With `AIDLC_CODEX_EXEC_LIVE=1`, Codex e2e tests can still invoke Bedrock, so the profile is not fully deterministic or cost-free. Please disable all live-model test gates when `--no-llm` is active.

2. **[P2] Deterministic TUI preflight is skipped**

   `t-tui-preflight.serial.test.ts` uses `tui-drive.ts` but never invokes Claude or spends tokens. Nevertheless, `--no-llm` classifies and skips it. Please refine the dependency classification so deterministic TUI tests continue running.

3. **[P2] Git isolation removes required `safe.directory` configuration**

   Setting `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` to the null device can break Git-based tests in CI containers or mounted workspaces with different ownership. Please preserve required Git safety configuration while disabling nondeterministic settings such as commit signing.

**Validation**

- Runner smoke test passed: 22 tests.
- Package parity passed.
- `git diff --check` passed.
- Typecheck unavailable because `bun-types` is missing locally.
- Lint unavailable because `biome` is missing locally.

apackeer:
Addressed the latest review feedback in `5740269`:

- `--no-llm` now forces all current live-model gates closed, including Codex exec, Kiro ACP/TUI/IDE, Claude TUI, and opencode run. `--all --debug` cannot reopen the TUI gate.
- TUI dependency classification now requires executable code to target `claude`, so the deterministic TUI substrate preflight and Kiro TUI tests are not incorrectly skipped.
- Git isolation now uses a process-unique generated config, preserves protected `safe.directory` entries from system, global, and command scopes, disables commit/tag signing, and removes higher-precedence command-scope overrides before spawning tests.

Validation:
- `bun run check` passed (package parity, all TypeScript configs, and Biome across 540 files).
- Runner smoke `t05`: 22/22 passed, 63 assertions.
- Classifier unit `t134`: 3/3 passed, 17 assertions.
- `git diff --check` passed.

leandrodamascena (approved):
LGTM

PR #669, opened 2026-07-29
fix(question-rendering): forbid echoing the ```question fence as literal chat text (2.5.15)

apackeer:
## Problem

The orchestrator sometimes prints a fenced ` ```question ` block as literal chat text instead of rendering it through the harness's native question mechanism (AskUserQuestion on Claude Code; structured/numbered prompts on the other harnesses). The user gets a non-interactive wall of spec text at exactly the moments that need a real answer - approval gates, the interaction-mode choice, the ladder prompt, halt-and-ask, the learnings gate - and loses the native "Other" escape hatch.

## What changes

Contract prose only - no commands, flags, or state formats:

- `core/aidlc-common/protocols/stage-protocol.md`: a harness-neutral paragraph stating a ` ```question ` fence is a spec rendered THROUGH the annex, never printed verbatim ("spec in, native prompt out; never echo the fence"), with a carve-out for the protocol's own illustrative fences.
- Every harness question-rendering annex (claude, codex, kiro, kiro-ide, opencode) gains a "Never echo the spec (non-negotiable)" section phrased for that harness's mechanism: names the echo a protocol violation, enumerates the five structured-question sites, and instructs the self-catch (about to write a triple-backtick question block into a reply = STOP, that belongs in the tool call).
- New smoke test `tests/smoke/t250-question-fence-never-echo.test.ts` pins the load-bearing phrases across all six files so the rule cannot silently rot. Zero LLM, zero subprocess.

Version 2.5.18 + CHANGELOG entry + README badge (prose ships in dist/, so this bumps; upgrade = re-copy the dist shell).

## Scope note

This is the knowledge-layer fix: it makes the rule present and emphatic where the conductor reads. A deterministic Stop-hook backstop (scan the ending turn for an echoed fence while a question is pending, block with a corrective reason) is a natural follow-up on Claude/Codex where the Stop payload carries transcript_path; kept separate deliberately - the Stop hook is a delicate, flow-altering surface that deserves its own review.

## Testing

- smoke+unit: 179 files, 0 failed (includes the new t250, 25 assertions)
- integration: 100/103 green; the 3 reds (t66, t89, t163) are pre-existing on base 9f914544 - verified by stashing this change back to the pristine base and reproducing identical signatures
- `bun scripts/package.ts --check` clean across all 5 harnesses; `bun run typecheck` clean


leandrodamascena (changes requested):
The overall direction is good, and strengthening the rendering contract is a reasonable knowledge-layer fix. However, I found three issues that should be addressed before merging.

1. The “every structured-question site” list is incomplete (`harness/*/skills/aidlc/question-rendering.md:30-36`). The section says the rule applies to every structured-question site “with no exceptions,” but it omits the consolidated-summary confirmation in `stage-protocol.md:297-307`, which also uses a fenced ` ```question ` specification. Because this prose steers the orchestrator, the incomplete exhaustive list can leave a known rendering path unreinforced. Please include every current site or introduce the list with “including, but not limited to.” The new test also codifies exactly these five sites.

2. The illustrative-fence regression check is vacuous (`tests/smoke/t250-question-fence-never-echo.test.ts:104-112,137-142`). It counts every occurrence of the string ` ```question `, including several inline references introduced by this PR. Consequently, it remains green even if every actual illustrative fenced block is removed. Please count opening fence lines instead, for example: `t.match(/^```question\s*$/gm)?.length ?? 0`.

3. The harness-neutral contract incorrectly implies native prompt UI everywhere (`core/aidlc-common/protocols/stage-protocol.md:19-23`). Kiro CLI, Kiro IDE, and opencode explicitly have no structured-question tool and render numbered prose. Please describe this as a native tool or the annex-defined fallback rendering, with the escape supplied either by the tool or by the numbered-prose format.

Verification: the new t250 test passes all 25 tests; `bun scripts/package.ts --check` passes across all five harnesses; PR CI is green. Local typecheck and lint could not run because the isolated worktree lacked `bun-types` and `biome`, while the corresponding CI job passed. The branch also currently conflicts with `v2`, now at 2.5.33, so the release files require rebasing and a new version bump.

apackeer:
Addressed the latest review in `ed59f8b7` and rebased onto current `v2` (release metadata is now `2.5.37`).

- Made the structured-question site list explicitly non-exhaustive (`including but not limited to`) and added consolidated-summary confirmation.
- Fixed `t250` to count actual opening `question` fence lines, including indented fences, rather than inline string mentions.
- Updated the harness-neutral protocol to cover both native question tools and annex-defined numbered-prose fallbacks.

Verification:
- `t250-question-fence-never-echo`: 25 pass, 0 fail
- `t68-version-changelog-sync`: 7 pass, 0 fail
- `bun run check`: package parity, typecheck, and Biome all pass

The branch conflict is resolved and GitHub now reports the PR as mergeable; refreshed CI is running.

leandrodamascena (approved):

Issue #632, opened 2026-07-22
[Bug]: reverse-engineering is not stored in the current intent


jeromevdl:
### Description

When working on a brownfield project, with `bugfix` and `refactor` scopes (maybe others but it didn't happen with `feature`), the reverse engineering outputs are stored in `aidlc/spaces/default/codekb/<repo>` instead of the dedicated intent folder (`aidlc/spaces/default/intents/<intent>/inception/reverse-engineering`). As a result, when chaining 2 intents (bug/refactor), the reverse engineering stage overrides the previously generated assets (`architecture.md` and so on).

### Steps to Reproduce

Start an intent in bugfix/refactor scope (at least) and progress to the reverse engineering phase (generally the 1st one)

### Expected Behavior

Assets are generated in `aidlc/spaces/default/intents/<intent>/inception/reverse-engineering` (per intent folder)

### Actual Behavior

Assets are generated in `aidlc/spaces/default/codekb/<repo>` and override previously generated assets.

### AI-DLC Version

v2 (alpha)

### Release / Commit

2.2.0

### AI-DLC Phase

Inception (requirements, user stories, application design)

### Platform / IDE

Kiro IDE

iuryeng:
Hi @jeromevdl — thanks for the detailed report. I dug into where this behavior stands at the current v2 HEAD (2.5.x) while contributing elsewhere in the repo, and I'm sharing what I found in case it helps with triage.

**What I found: the space-level placement appears to be the intended design in current releases, rather than a storage bug.**

- The stage declares it directly: the `core/aidlc-common/stages/inception/reverse-engineering.md` frontmatter ships `outputs: "aidlc/spaces/<active-space>/codekb/<repo>/ (9 artifacts: …)"`, with `condition: "… Always rerun for freshness"` and a `reverse-engineering-timestamp` artifact acting as the staleness marker for the store.
- The engine encodes it in two places: `KNOWN_CODEKB_STAGES` in `core/tools/aidlc-orchestrate.ts` and `core/tools/aidlc-state.ts` both treat reverse-engineering as a codekb stage.
- The workspace release notes (CHANGELOG, the spaces/intents release) describe the per-repo code-KB as a deliberately **space-level** store — a sibling of `memory/` and `intents/`, shared by every intent in the space — and 2.5.x added audit events for codekb writes (`codekb > <repo> > <name>` breadcrumbs) plus gate-revision counting for the RE gate, so the per-intent trail still records what was (re)generated and when.

[...]

PR #670, opened 2026-07-29
fix: guard the shared codekb against silent narrower-rerun overwrites (2.5.35)

apackeer:
## Summary

Reverse Engineering no longer silently discards a prior intent's code knowledge when a later intent performs a narrower scan. The space-level per-repo CodeKB remains shared across intents, but its timestamp now records structured scan coverage and a content fingerprint so the stage can verify freshness before deciding whether to rescan.

Addresses #632. This PR deliberately does not move Reverse Engineering outputs into per-intent storage; it preserves the shared CodeKB design while preventing unexamined reuse and silent narrower overwrites.

## Changes

- Add a structured `Scope of Analysis` block to `reverse-engineering-timestamp.md`, recording full/partial scope, source intent, analyzed paths/components, shallow paths, and a content fingerprint.
- Add the read-only `aidlc-utility codekb-scope-diff` verb:
  - status mode reports `NO_STORE`, `CURRENT`, `STALE`, `UNVERIFIED`, or `UNKNOWN_SCOPE`;
  - `--compare <timestamp.md>` reports `COVERS` or `NARROWER` and lists coverage that would be discarded;
  - `--mint --paths <csv>` computes the fingerprint written into the timestamp;
  - all modes support `--json`.
- Compute fingerprints with `git write-tree` over a temporary index restricted to analyzed paths. The result covers working-tree content and respects gitignore rules without mutating the real index.
- Add a Reverse Engineering rerun guard. A verified-current store with relevant coverage can be reused by explicit human choice, which skips the scan; stale, unverifiable, legacy, or irrelevant stores require a full or focused rescan choice.
- Compare delivered scan coverage before replacing an existing store. A narrower result is surfaced at the approval gate with the exact paths/components being discarded.
- Add deterministic unit coverage for scope parsing, fingerprint stability, all status verdicts, comparison behavior, minting, and CLI routing.
- Bump the shipped version, README badge, and changelog to 2.5.12, then regenerate every harness distribution.

## User experience

For a current CodeKB that covers the new intent, the user can reuse it and avoid running Reverse Engineering again. When a rescan is needed, the workflow distinguishes full and focused scans. If the delivered result is narrower than the shared store it replaces, the approval gate makes that loss explicit instead of silently accepting it.

Existing CodeKB stores predate scope tracking and report `UNKNOWN_SCOPE` until one post-upgrade scan records the structured block.

## Test plan

- `bun scripts/package.ts --check`: all five harness trees in sync.
- `bun run typecheck`: passed across all three TypeScript configurations.
- `bun run lint`: 541 files checked with no warnings or errors.
- Focused unit suite: 80 passed, 0 failed across `t248-codekb-scope-diff`, `t68-version-changelog-sync`, `t239-documentation-parity`, `t244-scope-matrix-doc-sync`, and `gen-coverage-registry`.
- Rebased cleanly onto current `v2` at `207db2ea` before push.

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the project license.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.35] - 2026-08-03
+
+Reverse Engineering no longer silently discards a prior intent's code knowledge on rerun. The shared per-repository codekb now records and compares actual scan coverage before replacement, keeps full-root freshness checks stable, and resolves every repository before advancing a multi-repo stage. **Upgrade:** re-copy your `dist/<harness>/` shell into the project; existing stores predate scope tracking and report `UNKNOWN_SCOPE` until their first post-upgrade scan writes the new scope block.
+
+* `reverse-engineering-timestamp.md` now ends with a structured `## Scope of Analysis` block recording full/partial coverage, analyzed and shallow paths, components, intent, and a content fingerprint.
+* The direct `aidlc-utility codekb-scope-diff` verb reports `NO_STORE`/`CURRENT`/`STALE`/`UNVERIFIED`/`UNKNOWN_SCOPE`, compares incoming coverage as `COVERS` or `NARROWER`, and mints working-tree fingerprints for recorded paths.
+* Full-root fingerprints exclude the framework-owned `aidlc/` tree so writing codekb, scope-draft, state, and audit files does not immediately make a store stale. Invalid or unmatched pathspecs now produce `unknown`/`UNVERIFIED` instead of Git's stable empty-tree hash.
+* `kind: full` now requires explicit repository-root (`./`) coverage, and a full store can only be replaced without warning by another valid full scan.
+* Multi-repo reruns resolve all reuse/rescan decisions before one lifecycle report, scan only repositories that need replacement, and use repository-specific scope drafts so parallel scans cannot overwrite one another.
+
 ## [2.5.34] - 2026-08-03
 
 The plugin contribution seam now merges `adds.scopes`: a plugin can put an existing core stage under one of its OWN scopes, so a single plugin scope can carry a run through both the plugin's stages and the core stages it routes onto. Previously the surface was declared-but-deferred (drop-logged, never merged). **Upgrade:** re-copy your `dist/<harness>/` shell and re-install/re-sync your plugins so the updated compose hook and `select-plugins` strip are in place; already-installed plugins recompose automatically on the next session start.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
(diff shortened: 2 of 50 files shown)


leandrodamascena (changes requested):
Thanks for tackling the silent CodeKB overwrite problem. I found several paths where the new guard can still silently lose coverage.

1. **[P1] Full-root fingerprints immediately invalidate themselves**

   `aidlc-lib.ts:1288` fingerprints `./`, which includes the committed `aidlc/.../codekb` tree. Writing the new artifacts and minted fingerprint changes that tree, so the next status check reports `STALE` without any source change. I reproduced different before/after hashes.

2. **[P1] Failed pathspecs collapse to a stable empty-tree fingerprint**

   The status from `git add` at `aidlc-lib.ts:1288` is ignored. A typo or mixed valid/invalid path list can therefore mint Git’s empty-tree hash. Repeating the same invalid pathspec after source changes produces the same hash and incorrectly reports `CURRENT`. Treat a failed `git add` as uncomputable instead of hashing the empty index.

3. **[P1] Full-scope invariants are not enforced**

   `aidlc-utility.ts:4271` treats incoming `kind: full` as unconditional `COVERS`, regardless of its listed paths. Conversely, `parseReScope()` permits a full store with no analyzed paths, which any later partial scan also covers. Require a full scope to identify repository-root coverage and require a full incoming scan when replacing a full store.

4. **[P1] Reusing one repository skips the entire multi-repo stage**

   `reverse-engineering.md:110` runs the stage-level `report --result skipped` for each reused repository. In an intent where repo A is reused and repo B needs scanning, this advances the stage before B runs. Resolve all repository decisions first and report skipped only when every repo is reused.

5. **[P2] Parallel repositories share one scope draft**

   The stage permits parallel scans but every repo writes `<record>/inception/reverse-engineering/scope-draft.md`. One repo can overwrite another’s draft before comparison. Use a repository-specific draft path.

**Validation**

- New scope-diff tests: 23 passed.
- Version, documentation and coverage tests: 57 passed.
- Package parity, typecheck, lint and `git diff --check`: passed.
- The live RE integration test did not complete within 120 seconds.
- The branch requires rebasing and re-versioning against current `v2`.

apackeer:
Addressed the latest review and rebased this PR onto current `v2` as `2.5.34`.

- Full-root fingerprints now exclude the framework-owned `aidlc/` tree relative to the project root, including when the project is nested inside a parent Git repository.
- Failed or mixed-invalid pathspecs now produce `unknown` / `UNVERIFIED` instead of Git's stable empty-tree hash.
- Full/partial scope invariants are enforced: `kind: full` requires `./`, partial scopes cannot claim `./`, and only another valid full scan can replace a full store without `NARROWER`.
- Multi-repo reuse decisions are resolved before one lifecycle report; isolated all-reuse runs use the `--single` completion path.
- Parallel comparisons use repository-specific temporary scope drafts, preserve results by repository, and remove drafts after comparison.
- Documentation, changelog/version, coverage metadata, and all generated harness trees were updated.

Validation:

- Focused unit slice: 5 files, 97 assertions, 0 failures (`t68`, `t154`, `t239`, `t248`, coverage registry).
- `bun run check`: package parity, all TypeScript configs, and Biome passed.
- `git diff --check` passed.

The live Reverse Engineering model journey was not rerun; the deterministic utility and shipped-stage contracts cover these review fixes.

leandrodamascena (approved):

PR #671, opened 2026-07-29
fix: make sensor output parsing robust to leading stdout noise (2.5.31)

apackeer:
## Problem

Both sensor JSON parse sites do a bare JSON.parse of a wrapped tool's exit-0 stdout: the dispatcher (aidlc-sensor.ts) on a per-sensor script's result object, and the linter sensor (aidlc-sensor-linter.ts) on eslint's --format json array. When a sensor fires against a repo whose package manager prints a run banner or lockfile warning before the JSON (pnpm is the canonical case, common when sensors run against sibling repos in a multi-repo workspace), the parse throws and the sensor's real verdict is silently discarded as an advisory PASS (script-error: bad-output / eslint-bad-output). The user-visible symptom is an absence: findings that should have surfaced at the gate never appear, and the run reads clean.

## What changes

- New exported stripStdoutNoise(stdout, startChar): slice leading noise down to the first structural JSON character. Applied at both sites: the dispatcher slices to the first `{`, the linter sensor to the first `[` (self-contained script keeps its own copy - no cross-import).
- Graceful degradation preserved: pure-noise input returns unchanged, so JSON.parse still throws and the existing bad-output advisory paths fire exactly as before. ESLINT_SPEC pin, base-path handling, and all other sensor semantics untouched.
- New unit test t251-sensor-stdout-noise: JSON preceded by noise parses to the real verdict; pure noise still degrades; clean JSON unchanged.

Version 2.5.31 + CHANGELOG + README badge (re-bumped past 2.5.30 on rebase onto v2 tip d0cd10a6).

## Testing

- smoke+unit: 179 files, 4402 assertions, 0 failed (t251 6/6)
- integration sensor slice (t95 sensor-fire, t49 bolt-sensor-failures): 2 files, 0 failed - live dispatch semantics unchanged
- bun scripts/package.ts --check clean across all 5 harnesses; bun run typecheck clean; coverage registry check fresh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.31] - 2026-07-31
+
+Sensor output parsing now tolerates leading stdout noise. When a sensor runs against a sibling repository, that repo's package manager (pnpm and similar) can print a run banner or lockfile warning before the wrapped tool's JSON; previously the dispatcher and the linter sensor did a bare parse of that stdout, so any preamble made the parse throw and the sensor's real verdict was silently discarded as an advisory PASS. Both parse sites now slice stdout to the first structural JSON character before parsing, and still degrade gracefully when even the sliced output is not JSON. **Upgrade:** re-copy your `dist/<harness>/` shell into the project so the updated `aidlc-sensor` dispatcher and `linter` sensor are installed. No configuration or command changes.
+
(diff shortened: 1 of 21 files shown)


leandrodamascena (approved):
Thanks for addressing the silent verdict loss. I verified that the canonical pnpm-style leading banner is stripped correctly and the failing sensor verdict survives.

One question: was the observed package-manager noise confirmed on `stdout` rather than `stderr`? The parser only reads `stdout`, and this distinction is particularly relevant to the linter’s direct `bunx` invocation.

A few non-blocking hardening notes:

- The linter’s first-`[` strategy can select bracketed or ANSI-colored log text instead of the ESLint array.
- Trailing wrapper output still causes `JSON.parse` to fail.
- Parsing one balanced JSON value would handle both cases more reliably.
- The branch needs rebasing and re-versioning against the current `v2`.

The implemented fix is correct for its documented canonical input, so these do not block approval.

PR #674, opened 2026-07-29
feat: optional declared workspace manifest (repos.json) + sync tool + doctor rows (2.5.36)

apackeer:
## What

Multi-repo workspaces gain an optional declared manifest and a sync tool, layered on the existing auto-discovery model without changing runtime discovery. Teams can declare their expected child repositories in `repos.json` and reconcile a fresh workspace with one command.

## User-visible changes

- Optional workspace-root `repos.json`: `{ "org": "...", "repos": [{ "name": "...", "branch": "...", "url": "..." }] }`. Names must be safe immediate children; `branch` and `url` are optional. `branch` selects the branch for a fresh clone and is advisory for an existing checkout; `url` overrides the derived GitHub SSH URL. The legacy `defaultBranch` key remains tolerated.
- New tool: `bun <harness>/tools/aidlc-workspace-sync.ts [--force]`. It clones declared-but-missing repos, updates only a managed `.gitignore` block, and generates `aidlc.code-workspace`. Existing repos are never re-cloned or branch-switched. Exit codes are `0` in sync, `1` blocked/error with live paths unchanged, and `2` synced with advisory warnings.
- Orphan removal requires `--force` plus conservative proof that every local object graph is recoverable from a live remote. Dirty, ignored, hidden-index, stashed, unreachable-object, linked-worktree, submodule, LFS, or locally dependent remote state blocks removal.
- Sync is serialized and transactional. Clones and generated files are staged; concurrent manifest/output/orphan changes abort; final installs are no-replace; rollback restores prior state. Removed checkouts and replaced generated files are retained in ignored recovery directories for manual inspection and deletion.
- `/aidlc --doctor` adds advisory rows for uncommitted workspace records, manifest/on-disk drift, and a stale managed `.gitignore` block. Manifest-specific rows appear only when `repos.json` exists and never alter the doctor exit code.
- User Guide chapters document the manifest contract and command. The manifest drives sync and doctor only; repositories present on disk remain the runtime source of truth.

## Testing

- `t255-workspace-sync`: 48 assertions covering schema/path validation, declared-branch clones, orphan safety and remote recovery proof, races, locking, rollback, dangling symlinks, and retained recovery data.
- `t256-workspace-doctor`: 9 assertions covering clean, drifted, missing, malformed, and schema-invalid manifest states.
- Focused workspace/doctor/lock suite: 3 files, 71 assertions, all passing.
- Final smoke+unit: 187 files, 4,618 assertions, all passing.
- Lock-steal race integration check: 2 assertions, passing.
- Focused Claude TUI rerun: 2 files, 6 test assertions, all passing.
- `bun scripts/package.ts --check`, `bun run typecheck`, Biome, and `git diff --check`: passing.
- Live pre-merge slices exercised Claude SDK/TUI and Codex exec. Kiro ACP/TUI slices skipped because the installed CLI is unauthenticated; Kiro IDE requires macOS. The live composer `t193` remains nondeterministic because the model may create a custom scope instead of selecting stock `bugfix`. Deterministic E2E still has the two standing `v2` failures `t113`/`t122`, addressed separately by `fix/t113-t122-base-reds`.


leandrodamascena (changes requested):
Thanks for bringing this upstream. The optional-manifest model is a useful addition, but the destructive sync path currently cannot uphold its stated data-safety and atomicity contracts. I found four P1 blockers and two P2 consistency issues.

1. **[P1] `--force` can delete ignored and otherwise hidden local-only files**

`isClean()` relies on configurable `git status --porcelain`. That omits ignored files, and a repository can also configure `status.showUntrackedFiles=no`. Such a checkout is classified as recoverable and recursively deleted. I reproduced both an ignored secret and a hidden untracked file disappearing with exit 0. The safety contract needs to account for all filesystem content that removal would destroy, not only the default porcelain view.

`core/tools/aidlc-workspace-sync.ts:136-140,308-327,371-374`

2. **[P1] Local remote-tracking refs do not prove that commits are recoverable remotely**

`hasUnpushed()` compares against local `refs/remotes/*` without fetching or querying the remote. A stale or fabricated tracking ref therefore makes a unique local commit appear pushed. Local-only tags, secondary-worktree HEADs, submodule repositories, and LFS objects are also outside the check. I reproduced deletion after the real remote branch had been removed while its cached tracking ref remained. The command exited 0 and deleted the only remaining copy.

`core/tools/aidlc-workspace-sync.ts:142-150,308-327,371-374`

3. **[P1] Repository names can escape the workspace root**

Manifest validation requires only a non-empty string. A name such as `../escaped` is passed through `join(ROOT, name)` and then to `git clone`. I reproduced a repository being cloned successfully outside the workspace with exit 0. Nested names also disagree with the immediate-child runtime discovery contract. Names need the same safe single-segment validation used by the runtime, plus a resolved-path containment check.

`core/tools/aidlc-workspace-sync.ts:111-130,278-283,352-357`

4. **[P1] The apply phase is not atomic and exit 1 can follow destructive partial changes**

Clones, orphan removals, `.gitignore`, and the workspace file are mutated sequentially with no staging or rollback. I reproduced a valid first clone remaining after the second clone failed. I also reproduced an orphan being deleted before a later `.gitignore` write failure. Both runs exited 1 despite the documented `error (nothing changed)` contract.

`core/tools/aidlc-workspace-sync.ts:17-18,35-43,350-381`

5. **[P2] Preflight approval can become stale before orphan deletion**

The tool records an orphan as removable, performs potentially slow clones, and then deletes the orphan without reacquiring a lock or rechecking its Git/filesystem state. I reproduced creating local work while a slow clone was in progress; sync still deleted the checkout and exited 0. Concurrent syncs also race the same destinations. At minimum, destructive predicates need revalidation immediately before removal under a workspace lock.

`core/tools/aidlc-workspace-sync.ts:268-327,350-374`

6. **[P2] A fresh clone can be on the wrong declared branch while sync exits 0**

Branch checks run only for repositories already present during preflight. `git clone` does not pass the declared branch and newly cloned repositories are not checked afterward. With a remote default of `main` and manifest branch `develop`, the first sync cloned `main` and reported fully in sync with exit 0; only the second invocation detected the mismatch and returned 2.

`core/tools/aidlc-workspace-sync.ts:278-305,352-364,383-395`

Additional diagnostic gap: workspace doctor validates only the `repos` array and string names, so it can report a schema-invalid manifest as synchronized even though sync rejects it (`core/tools/aidlc-workspace-doctor.ts:89-106` versus `aidlc-workspace-sync.ts:111-130`).

Verification on head `2a613133`:

- `t255-workspace-sync` and `t256-workspace-doctor`: 30 passed, 124 assertions.
- Package parity passed for all five harnesses.
- Independent reviewers reproduced ignored-file deletion, stale-remote data loss, path escape, partial apply, TOCTOU deletion, concurrent invocation, and wrong-branch first-clone behavior; a defense pass failed to invalidate any of them.

The focused tests cover ordinary dirty/unpushed/stashed cases, but not the filesystem and remote-recoverability conditions that must hold before recursive deletion.

apackeer:
Addressed all review blockers in `98d3cab`:

1. Orphan checks now override configurable status defaults and detect ignored/untracked content, empty directories, assume-unchanged/skip-worktree files, and stashes.
2. Recoverability is verified against live remotes rather than cached tracking refs, including local refs/tags, reflog-only commits, linked worktrees, submodules, LFS storage, and remotes stored inside the checkout.
3. Sync and doctor share a strict manifest parser with safe single-segment names, duplicate rejection, and resolved immediate-child containment.
4. Clones and generated files are staged under the workspace root and installed through reversible same-filesystem renames; apply failures roll back, retaining recovery data if rollback itself cannot complete.
5. The full reconcile is serialized under a workspace lock, and destructive predicates are recomputed immediately before each orphan move.
6. Fresh clones pass the declared branch to `git clone --branch` and verify the resulting checkout.

The doctor schema mismatch is also fixed by using the same parser as sync.

Additional adversarial coverage protects empty local directories and self-contained local remotes, two further ways the previous proof could lose the only copy.

- focused `t255`/`t256`: 43 tests, 213 assertions, 0 failures
- full unit tier: 168 files, 3,321 assertions, 0 failures
- typecheck and lint: pass
- package parity: all five harnesses pass
- `git diff --check`: pass

leandrodamascena (approved):

Issue #684, opened 2026-07-30
[Feature]: Benchmarks evaluating AI-DLC performance


KBB99:
### Description

A group of benchmark tasks that evaluate the efficacy of different version of AI-DLC. For example passing a series of notes describing an HTML page hosted on CloudFront and evaluating three main metrics (for now): cost, quality, and human approvals.

### Use Case

As new AI-DLC workflows spend more tokens leading to higher costs I would like to ensure that the generated artifacts are of higher quality. Benchmarks should also allow make it possible to optimize the flow automatically.

### Version

v2 (alpha)

### Area

Other

KBB99:
## First v2 evaluation run — v2 (2.5.25, `d2ae3cca`) vs the v1.0.1 baseline on sci-calc-v2

We restored the evaluator from the `evaluator-keep` branch and ran the full pipeline against the current `v2` dist via the **claude-cli adapter** (real `claude` TUI driving `/aidlc --scope mvp`, golden test case `sci-calc-v2`). As far as we know this is the first scored v2 datapoint; v2 had not been evaluated since the evaluator was removed in #433. Results, next to the recorded v1.0.1 release numbers from the trend history:

| Metric | Golden | v1.0.1 | **v2 (this run)** |
|---|---|---|---|
| Unit tests | 180 pass | 121 pass | **293 pass, 0 fail (100% line coverage, 626/626)** |
| Contract tests (OpenAPI) | 88/88 | 77/88 | **88/88** |
| Security findings | 0 | 1 | **0** |
| Lint findings | 0 | 0 | 660 (265 errors) — see caveat 4 |
| Qualitative doc score | 0.854 | 0.754 | **0.749** |
| Wall-clock | 23.8m | 14.9m | **~7.5h** (see caveat 2) |

### Headline findings

**Functional correctness is a decisive win.** All 11 of v1.0.1's contract-test failures pass in v2 — every domain-error case (divide-by-zero→400, sqrt(-1), ln(0), log base 1, asin(2), …). The generated test suite is 2.4× larger than v1.0.1's and exceeds the golden target (293 vs 180) at 100% line coverage. The per-unit functional-design stages + adversarial reviewer loop appear to genuinely carry the error-envelope spec through to code.

[...]

PR #448, opened 2026-06-29
fix: remove --test-run / Test Run Mode (#369) (2.1.4)

apackeer:
## What

Removes the `--test-run` flag and Test Run Mode mechanism entirely (closes #369).

`--test-run` was designed for CI/automated runs: it auto-approved every approval gate and skipped the structured-question and learnings rituals. But it worked unconditionally in any session, so an agent in an interactive human session could append `--test-run` to `report`/`approve` and rubber-stamp its own incomplete work, bypassing the gate that is supposed to keep a human in the loop (the bug reported in #369).

The issue proposed gating `--test-run` behind `AIDLC_CI=true`. We are doing the cleaner thing instead: removing it. With the current SDK and TUI test drivers, the test harness no longer needs an auto-approve bypass. The SDK driver auto-answers each `AskUserQuestion` gate via its declarative answer policy, and the TUI driver drives the real gate, so tests reach `WORKFLOW_COMPLETED` through the normal human-in-the-loop path. There is nothing left for `--test-run` to do that a real driver does not already do.

## Breaking changes (for CI/scripts)

- `--test-run` is removed from the parsers everywhere (`aidlc-orchestrate.ts report`, `aidlc-state.ts approve`, `aidlc-log.ts decision`/`answer`, the `/aidlc <scope> --test-run` slash form, and `aidlc-runtime.ts compile`) and no longer does anything. (Mechanically: `report`/`approve` silently ignore the now-unknown flag and run the gate normally; `decision`/`answer` error on it as an unknown flag; only `enable-test-run` hard-errors.)
- The `aidlc-utility.ts enable-test-run` subcommand is removed, along with the `Test Run Mode` state-file field.
- The `TEST_RUN_MODE_ENABLED` audit event is removed. Together with the park-umbrella additions this PR rebases on top of (`WORKFLOW_PARKED` / `WORKFLOW_UNPARKED`, +2), the audit taxonomy is now 68 events. `GATE_APPROVED`, `QUESTION_ANSWERED`, and `DECISION_RECORDED` no longer carry the optional `Test-Run: true` field, and `MEMORY_EMPTY` rows are never stamped `Test-Run`.
[...]

PR #691, opened 2026-07-31
fix(evaluator): PTY driver deadlock + multi-select gate handling

KBB99:
## Summary

Two fixes to the cli-harness PTY driver (`_pty_terminal.py`), both hit while producing the first scored v2 evaluation run (#684). Root cause for both: the evaluator was written against v2 ~2.1.x when `--test-run` auto-approved gates; v2 removed that bypass in 2.1.4 (#448), so every gate — including multi-select widgets — now surfaces to the driver.

### 1. `drive_until()` pipe deadlock

When a menu stayed painted (any widget Enter doesn't dismiss), the loop re-sent Enter every iteration **without draining PTY output**. The TUI's output filled the kernel pipe buffer, the TUI blocked writing, the driver blocked on `write(2)` sending the next keystroke, and the pair froze permanently — confirmed twice via process sampling (`sample` showed both sides parked in `write`), costing multi-hour runs each time.

Fix: answer a given menu screen once — re-answer only when the rendered screen changes — and always `_drain()` after answering.

### 2. Multi-select gates spun forever

`answer_gate_default()` only knew single-select menus (bare Enter accepts the highlighted option). On a multi-select widget — e.g. the v2 Learnings ritual — **Enter toggles the checkbox instead of selecting**, so the driver toggled `[ ]`↔`[✔]` indefinitely (observed for 19h). This is exactly the failure the repo's own e2e driver documents as the t73 hang ("the loop toggled `[ ]`↔`[✔]` forever", `tests/harness/tui-drive.ts`).

Fix: port tui-drive.ts's per-widget key model to the Python driver —
- checkbox option lines (`1. [ ]`) → **Space** toggles the Recommended option, then **Right** advances a multi-tab form / **Enter** commits a lone question
- `Submit answers` review screen → plain **Enter**
- anything else → single-select, bare Enter (unchanged)

## Validation

- 51/51 cli-harness tests pass; `ruff check` + `ruff format --check` clean.
- Proven in anger: with these two fixes the claude-cli adapter drove a complete 21-stage v2/mvp workflow end-to-end (three `/aidlc --resume` sessions), producing the scored results on #684 — 88/88 contract tests, 293 generated tests at 100% coverage.

## Known gaps NOT addressed here (follow-ups)

Found in the same exercise, needing owner decisions rather than mechanical fixes:
1. **2h adapter timeout** (`timeout_seconds: 7200`, `adapter.py`) is far too small for v2/mvp (~7.5h observed). Suggest raising it and/or teaching the adapter to chain `/aidlc --resume` sessions (which worked flawlessly when driven manually).
2. **`find_aidlc_docs()` doesn't know v2's layout** (`aidlc/spaces/<space>/intents/<intent>/`), so a completed run still reports "no aidlc-docs" — we extracted docs manually for #684.
3. **Stage-5 scorer needs an explicit `--region`** (NoRegionError otherwise); should default from config or the dist's pinned region.
4. **No token-usage capture** in the CLI adapters — `run-metrics.yaml` reports zero. Stopgap documented on #684: sum `AWS/Bedrock` CloudWatch metrics over the run window from `run-meta.yaml`.

Refs #684.

### DIFF
--- a/evaluator/packages/cli-harness/src/cli_harness/adapters/_pty_terminal.py
+++ b/evaluator/packages/cli-harness/src/cli_harness/adapters/_pty_terminal.py
@@ -203,11 +203,27 @@ class PtyTerminal:
         return bool(_MENU_CARET_RE.search(screen) and _SELECT_FOOTER_RE.search(screen))
 
     def answer_gate_default(self) -> None:
-        """Accept the highlighted (default/Recommended) option on a single-select gate.
+        """Answer the painted gate, mirroring tui-drive.ts's per-widget key model.
 
-        Mirrors tui-drive's single-select behavior: the engine highlights the
-        Recommended option by default, so a bare Enter selects it.
+        - Submit review screen: Enter commits the whole form.
+        - Multi-select option (numbered line with a checkbox): Enter TOGGLES
+          instead of selecting (the t73 toggle-forever hang), so Space toggles
+          the Recommended option ON, then Right advances a multi-tab form or
+          Enter commits a lone multi-select question.
+        - Single-select: bare Enter accepts the highlighted Recommended option.
         """
+        screen = self.screen_text()
+        if "Submit answers" in screen:
+            self.send_key("Enter")
+            return
+        if re.search(r"\d+\.\s*\[[ xX✔]\]", screen):
+            self.send_key("Space")
+            time.sleep(0.15)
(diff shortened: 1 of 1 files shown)


leandrodamascena:
Hey there! I'm closing this because we are working in a new evaluator for v2.

Issue #681, opened 2026-07-29
[Bug]: Promote Codex harness lifecycle and human-authorization reliability fixes to v2


snkinard:
## Description

A sustained brownfield evaluation of the AI-DLC v2 Codex harness exposed
several interacting reliability and authorization defects. A local harness
snapshot was repaired and adversarially validated, but those fixes need to be
ported into the canonical `v2` source, tested there, released, and consumed
without project-local patches.

The defects are related and should be handled as one promotion effort because
they share routing, audit, and gate-integrity boundaries:

1. Per-unit artifact existence could be mistaken for completion, allowing a
   paused or partial unit to be skipped.
2. Resume state retained the stage but not the exact unit lifecycle checkpoint,
   reason, and next action.
3. Ambiguous or automatic structured-question responses could mint
   human-presence evidence.
4. Human-presence freshness used timestamp/file ordering that was unsafe across
   audit shards in the same second.
5. Reusable decision receipts could be rebound from an old answer to a
   different key, unit, source digest, or typed value.
6. Decision-ledger writes were vulnerable to lost updates.
7. Autonomous execution and rejection paths could weaken mandatory human
   gates.
8. The public audit CLI could fabricate authority-bearing audit rows, including
   `HUMAN_TURN`.
9. A paused unit could route back as ordinary stage work without an explicit
   conductor hard-stop signal.

## Environment

- AI-DLC harness snapshot: `2.5.13`
- Canonical target: `v2` branch (observed at
  `9f9145442429440ccd93f22e27186aa428007e5a`)
- Platform: OpenAI Codex on macOS
- Codex version: `0.146.0`
- Workflow phase where first observed: Construction / Code Generation

## Proposed canonical changes

- Add explicit, attempt-scoped `UNIT_STARTED`, `UNIT_PAUSED`, `UNIT_RESUMED`,
  and `UNIT_COMPLETED` receipts.
- Treat artifacts only as evidence checked by `UNIT_COMPLETED`, never as the
  transition itself.
- Persist and report active unit, unit state, pause reason, and exact next
  action.
- Route a paused unit ahead of new work while emitting `unit_state: paused`;
  require the conductor to stop until an explicit resume.
- Require structured-question completion to be an explicit completed object
  with a genuinely nonempty, non-cancellation answer and no auto-resolution
  metadata.
- Bind human presence to the current Session/Turn and fail closed on
  cross-shard same-second ambiguity.
[...]

PR #692, opened 2026-07-31
fix: harden human-authorization boundaries + unit-of-work lifecycle receipts (2.5.49)

apackeer:
Closes the verified authorization and lifecycle gaps from #681 (the brownfield Codex evaluation). Seven of the nine reported claims are addressed; the remaining two (decision-ID-bound reusable receipts, decision-ledger locking) describe a receipt subsystem v2 does not have and are deliberately left to a follow-up design discussion rather than bolted on here.

## What each claim gets

**Claim 8 - the public audit CLI could forge authority rows.** `aidlc-audit.ts append` / `append-batch` now refuse the receipts the engine's guards read as authorization evidence (`HUMAN_TURN`, `GATE_APPROVED`, `GATE_REJECTED`, `QUESTION_ANSWERED`, `REVIEW_REQUESTED`, `REVIEW_COMPLETED`, `SWARM_UNIT_CONVERGED`, `AUTONOMY_MODE_SET`, and the four new `UNIT_*` receipts) with an error naming the owning tool. `append-raw` refuses a body carrying a taxonomy `**Event**:` line, and a caller-supplied `Event` field key is refused on every emit path (a second `**Event**:` line would spoof the multiline event queries; `Timestamp` stays legal - park/unpark rows carry it and the emitter's own line is written first). Owning tools and hooks emit through the library import, untouched. Test fixtures that simulate owning emitters set `AIDLC_ALLOW_DIRECT_AUDIT_EVENTS=1`; the suite runner sets it globally.

**Claim 7 - autonomous execution could weaken mandatory gates.** `aidlc-bolt.ts set-autonomy --mode autonomous` requires a fresh human turn (the ladder-prompt answer), and the grant consumes that turn exactly like a gate resolution - one human turn can no longer both grant autonomy and approve something else. De-escalating to `gated` never needs presence.

**Claim 4 - same-second cross-shard ordering was unsafe.** `humanActedSinceGate` now reads shards individually: buffer position remains the tiebreak only *within* a shard (where it reflects append order), and when the deciding human-turn/resolution pair shares one second-precision timestamp across *different* shards - where execution order is unknowable - the gate fails closed and requires a fresh turn instead of trusting shard filename order.

**Claim 3 - ambiguous/auto-resolved question responses.** Cancellation boilerplate ("Cancelled", "dismissed", "timed out", empty text - the exact `status=completed` / answer `Cancelled` repro from the issue) is refused as an interview answer (`aidlc-log.ts answer`), an approval choice (`approve --user-input`), and rejection feedback (`reject --feedback`). A substantive answer that merely contains a cancellation word ("cancel the standing order via cron") still passes. The presence mint itself stays content-blind by design.

**Claims 1, 2, 9 - unit lifecycle.** New `UNIT_STARTED` / `UNIT_PAUSED` / `UNIT_RESUMED` / `UNIT_COMPLETED` receipts (taxonomy 74 → 78), emitted by the new `aidlc-state.ts unit start|pause|resume|complete` verb on inline per-unit Construction stages:

- **Artifacts are evidence, receipts are the transition.** `unit complete` verifies the unit's required artifacts on disk (kind-filtered, same rules as the engine's coverage walk) before committing. Once any receipt exists for a stage, the engine requires a receipt per unit - a scaffolded or partially-written unit can no longer read as done by file existence alone. Receipt-free workflows keep today's artifact-driven coverage, so nothing re-opens on upgrade.
- **Pause carries the checkpoint.** `unit pause` requires `--reason` and `--next-action`; the state file mirrors them under `## Runtime State`, `resume` returns them as JSON, and the session-start hook injects an `Active Unit:` context line so a fresh session lands on the exact stopping point.
- **A paused unit hard-stops the engine.** `next` emits an ask naming the unit, its reason, and next action (`unit_state: paused`) until an explicit `unit resume`; `report --result awaiting-approval` is refused while a unit is paused; `unit start` refuses while another unit is open (single-active-unit invariant, so resume/restart races cannot create two active units).
- **No swarm interference.** The verb refuses under `Construction Autonomy Mode: autonomous` (same rule as `park`); the referee's `SWARM_UNIT_*` ledger remains the sole per-unit authority on the autonomous path.

## Tests

- `t261-audit-authority-floor` - CLI forge refusals across all three append paths, the `Event` field-key floor, set-autonomy presence + consume-once, cross-shard fail-closed vs same-shard append order, and the non-answer floors including the issue's named repro.
- `t260-unit-lifecycle-receipts` - receipts-not-artifacts completion, single-active-unit, pause checkpoint mirroring, the paused-unit engine ask, and the approval refusal.
- Taxonomy pins moved in lock-step (t28, t81, t239, registry docs, user guide); coverage registry regenerated.

Gates run: `package.ts --check`, typecheck, lint, smoke + unit (183 files / 4481 assertions) all green on the rebased head. E2E ran sliced across the full 67-file tier: all deterministic journeys green including the construction-worktree scopes, swarm referee, and stop hook; the two reds (t113, t138) fail identically on the pristine base commit (t113's helper predates reviewer enforcement; t138 is the known live-SDK timeout) - both exonerated by baseline runs.

## Notes for reviewers

- **Breaking (CI only):** scripts that drove gates against bare fixtures via `aidlc-audit.ts append HUMAN_TURN`-style calls need `AIDLC_ALLOW_DIRECT_AUDIT_EVENTS=1`. The CHANGELOG entry carries the upgrade note.
- Claims 5/6 from the issue (decision-ID binding, ledger locking) are intentionally out of scope - proposed follow-up: a design issue for typed decision receipts, referencing the reporter's repaired-snapshot evidence if they can share it.
- Version 2.5.34: 2.5.31 is claimed by #686, 2.5.32 by #687, 2.5.33 by #689. Test slots: #687 took t259, so the authority-floor test lands as t261.


core/tools/aidlc-audit.ts
```diff
@@ -244,12 +259,73 @@ export interface AuditEntryInput {
...
+// parser takes the first match.
+const RESERVED_FIELD_KEYS = new Set(["Event"]);
+
 function validateAuditEntry(entry: AuditEntryInput): void {
   if (!VALID_EVENT_TYPES.has(entry.eventType)) {
     throw new Error(
       `Invalid event type: ${entry.eventType}. Must be one of: ${[...VALID_EVENT_TYPES].join(", ")}`
     );
   }
+  for (const key of Object.keys(entry.fields)) {
```
leandrodamascena, line 378: [P1] Reject control characters in audit field names

Only the exact key `Event` is rejected, while field names are rendered verbatim. I reproduced an allowed `ERROR_LOGGED` append using a field name containing `\n**Event`, which `findAllEvents()` then recognized as a forged `HUMAN_TURN`. The same path can forge `UNIT_COMPLETED` or any protected receipt. Validate field names against a strict single-line grammar before rendering.

core/tools/aidlc-state.ts
```diff
@@ -868,6 +873,199 @@ function handleUnpark(_args: string[]): void {
...
+  const action = args[0];
+  const VALID_UNIT_ACTIONS = new Set(["start", "pause", "resume", "complete"]);
+  if (!action || !VALID_UNIT_ACTIONS.has(action)) {
+    error(
+      `Usage: aidlc-state.ts unit <start|pause|resume|complete> --stage <slug> --unit <name> [--reason <text>] [--next-action <text>]`,
+    );
+  }
+  const rest = args.slice(1);
+  const slug = getFlagValue(rest, "--stage");
+  const unit = getFlagValue(rest, "--unit");
```
leandrodamascena, line 862: [P1] Validate units against the authoritative DAG

`unit start` accepts any nonempty unit name without checking the resolved Bolt DAG. I started `rogue-unit`; `next` then routed legitimate `unit-a`, but starting it failed because the rogue checkpoint remained active. This leaves the workflow deadlocked. Require a safe unit identifier that exists in the authoritative unit set, and reject newlines/path traversal in all state-mirrored values.

core/tools/aidlc-state.ts
```diff
@@ -868,6 +873,199 @@ function handleUnpark(_args: string[]): void {
...
+      error(
+        `Refusing unit ${action}: Construction Autonomy Mode is autonomous. The swarm referee ` +
+          "owns per-unit bookkeeping (SWARM_UNIT_* receipts); interactive unit receipts apply " +
+          "only to gated Construction.",
+      );
+    }
+
+    const checkpoint = activeUnitCheckpoint(pd, slug);
+
+    if (action === "start") {
```
leandrodamascena, line 916: [P1] Do not allow completed units to reopen while remaining settled

After completing every unit, `unit start` accepts a previously completed unit. Its old `UNIT_COMPLETED` receipt remains in the settled set, so `next` returns the final `gate:true` directive even though that unit now has an active `UNIT_STARTED` checkpoint. Reject restarting a completed unit or make any nonterminal checkpoint override its earlier completion receipt and block approval.

core/tools/aidlc-lib.ts
```diff
@@ -3531,6 +3590,87 @@ export function swarmConvergedUnits(
...
+  projectDir: string,
+  slug: string,
+): Set<string> {
+  const audit = readAllAuditShards(projectDir);
+  if (!audit) return new Set();
+  const floor = latestMainWorkflowStageStarted(audit, slug);
+  const done = new Set<string>();
+  for (const { timestamp, block } of findAllEvents(audit, "UNIT_COMPLETED")) {
+    if (auditBlockField(block, "Stage") !== slug) continue;
+    if (floor && timestamp < floor) continue;
```
leandrodamascena: [P1] Bind unit receipts to an exact stage attempt

The attempt floor compares second-precision timestamps using `<`. When a prior `UNIT_COMPLETED` and a new attempt’s `STAGE_STARTED` occur in the same second, the old receipt is accepted for the new attempt. I reproduced this directly. Use an exact run-floor/attempt identifier, as the swarm receipt path already does, rather than timestamps alone.

core/tools/aidlc-bolt.ts
```diff
@@ -802,6 +804,27 @@ function handleSetAutonomy(args: string[]): void {
...
+  // (`isAutonomousMode` in approve / answer) keys off the field this writes, so
+  // an unattended model that could flip it would unlock every gate at once.
+  // Require a HUMAN_TURN since the last gate resolution, exactly the approve
+  // guard's semantics: the human's ladder answer mints the fresh turn, and
+  // humanActedSinceGate treats the autonomous-mode AUTONOMY_MODE_SET row as a
+  // resolution so the grant consumes the turn (one authority action per human
+  // turn). The ladder choice is set-autonomy-owned — logging it via `aidlc-log
+  // answer` first would consume the turn as QUESTION_ANSWERED and refuse here.
+  // De-escalation (gated) restores gates; it never requires presence.
+  if (flags.mode === "autonomous" && !humanPresenceGuardDisabled() && !humanActedSinceGate(pd)) {
```
leandrodamascena: [P1] Make autonomy authorization check-and-consume atomic

`set-autonomy` calls `humanActedSinceGate()` before acquiring the audit lock, then appends `AUTONOMY_MODE_SET` and writes state in separate operations. Concurrent autonomy grants or an approval racing the grant can both observe the same fresh `HUMAN_TURN` and commit, violating the stated consume-once rule. Hold one lock across the presence check, audit append, and state update.

core/tools/aidlc-state.ts
```diff
@@ -2191,6 +2404,16 @@ function handleReject(args: string[]): void {
       `Refusing to reject "${slug}": --feedback must contain the human's requested changes.`,
     );
   }
+  // Same non-answer floor as approve: a dismissed gate question is neither an
+  // approval nor a change request — it must not commit GATE_REJECTED and spin
+  // the revision loop on cancellation boilerplate.
+  if (isNonAnswer(feedback)) {
```
leandrodamascena, line 2469: [P2] Require human presence before recording rejection

The rejection path now rejects cancellation boilerplate, but it still never calls `humanActedSinceGate()`. A conductor can submit any substantive generated feedback and commit `GATE_REJECTED` without a human turn, even though this receipt is treated as authority-bearing and consumes presence for later decisions. Apply the same human-presence guard used by approval.

leandrodamascena (changes requested):
The authorization and unit-lifecycle additions have several fail-open paths that defeat the new authority floor or allow the workflow to reach a gate with an active/incomplete unit. The inline findings include direct reproductions for audit receipt forgery, rogue-unit deadlock, completed-unit restart, and same-second attempt reuse.

core/tools/aidlc-audit.ts
apackeer: Addressed in `8249d4b8`. Audit field keys now use a strict single-line label grammar while `Event` remains reserved; values escape CR/LF and U+2028/U+2029, and `append-raw` also rejects line-breaking headings and canonical event injection. `t261-audit-authority-floor` covers newline, CRLF, Unicode separator, NUL, and append-raw injection cases.

core/tools/aidlc-state.ts
apackeer: Addressed in `8249d4b8`. Every lifecycle verb now validates the unit as a safe lowercase path segment, resolves the authoritative Bolt DAG, and requires membership in that resolved unit set. State-mirrored reason and next-action values reject line-breaking control characters. `t260-unit-lifecycle-receipts` covers rogue, traversal, multiline, stale-cache, and malformed-DAG cases.

core/tools/aidlc-state.ts
apackeer: Addressed in `8249d4b8`. Settlement is now reduced from the complete lifecycle stream in chronological order: `UNIT_COMPLETED` adds the unit, while any later nonterminal lifecycle row removes it. Restarting a completed unit therefore reopens it and blocks the gate until it completes again. The regression is in `t260-unit-lifecycle-receipts`.

core/tools/aidlc-lib.ts
apackeer: Addressed in `8249d4b8`. Lifecycle receipts now carry an exact `Run floor` token (`<boundary-event>:<timestamp>#<ordinal>`), and readers require an exact field match to the current attempt. This distinguishes same-second re-entry; receipt mode also remains sticky across attempts. `t260-unit-lifecycle-receipts` reproduces and rejects the same-second prior-attempt receipt.

core/tools/aidlc-bolt.ts
apackeer: Addressed in `8249d4b8`. `set-autonomy` now holds one audit lock across human-presence validation, `AUTONOMY_MODE_SET` emission, and the state write, using the unlocked emitter inside the critical section. The concurrent-grant regression in `t261-audit-authority-floor` verifies that one fresh turn permits exactly one successful grant.

core/tools/aidlc-state.ts
apackeer: Addressed in `8249d4b8`. The rejection transition now applies the same fresh-human-turn guard as approval before mutating state or emitting `GATE_REJECTED`; the emitted rejection consumes that turn. `t188-human-presence-gate` covers refusal without presence and refusal to reuse the same turn for a second rejection.

leandrodamascena (changes requested):
The corrective commit resolves several earlier findings, but lifecycle integrity still has fail-open and upgrade-breaking paths.

1. **[P1] Bind `unit start` to the engine-routed stage/unit pair**
`unit start` checks DAG membership but not dependency readiness. Starting `unit-b` before `unit-a` succeeds, and `next` then routes `unit-b`, overriding topological order.

2. **[P1] Do not order same-second attempt boundaries by shard filename**
A new `GATE_REJECTED` in another shard can lose the tie to an older `STAGE_STARTED`, allowing the prior attempt’s `UNIT_COMPLETED` receipt to remain settled.

3. **[P1] Require artifacts to be regular files**
`existsSync()` accepts directories named `*.md`. I created directories for every required artifact and `unit complete` emitted `UNIT_COMPLETED`.

4. **[P2] Preserve in-flight swarms across upgrades**
`finalize` now requires the new stamped `SWARM_STARTED`. Worktrees prepared before upgrading cannot finalize or be re-stamped without being discarded and rebuilt.

5. **[P2] Preserve legacy safe unit names**
The new lowercase-only validator rejects previously accepted names such as `api_v2` and `WebUI`, breaking existing receipt-free workflows despite the compatibility claim.

Focused validation passed: 111 tests, 476 assertions, package parity, and remote CI.

apackeer:
Addressed @leandrodamascena's follow-up review in `a7596c67`.

1. `unit start` now consumes steering continuations and requires the exact engine-routed `directive.stage` / `directive.unit` pair, so dependencies and earlier same-batch units cannot be bypassed. The compiled dispatcher path is covered too.
2. Attempt-floor readers preserve shard identity. Same-second boundaries in different shards produce a deterministic fail-closed ambiguity floor instead of being ordered by shard filename.
3. Unit completion and engine coverage now require regular files; directories named `*.md` no longer count as artifacts.
4. Pre-upgrade in-flight swarms can finalize only when the frozen, hash-verified audit proves the ordered legacy `SWARM_STARTED -> BOLT_STARTED -> STATE_FORKED -> AUDIT_FORKED` preparation and its derived attempt still matches current state. Stale/incomplete preparations remain refused.
5. Safe legacy unit names such as `api_v2` and `WebUI` are accepted while traversal, separators, whitespace, and control characters remain rejected.

Verification:
- focused unit routing/lifecycle/attempt slice: 46/46 green
- compiled-dispatcher lifecycle slice: 21/21 green
- swarm referee e2e: 16/16 green
- coverage registry + version/changelog checks: 41/41 green
- `bun run typecheck`, `bun run lint`, `bun scripts/package.ts --check`, `bun tests/gen-coverage-registry.ts --check`, and `git diff --check` all pass

leandrodamascena (changes requested):
The corrective commit resolves several earlier findings, but authority and lifecycle integrity still fail open under same-second cross-shard races.

[P1] Fail closed when a resolution competes with human presence across shards
When `GATE_APPROVED` exists in one shard and `GATE_REJECTED` plus `HUMAN_TURN` share its timestamp in another, `humanActedSinceGate()` returns true. The global order is unknowable, so the human turn may predate the latest resolution.

[P1] Do not order same-second unit lifecycle events by shard filename
With `UNIT_STARTED` and `UNIT_PAUSED` in different shards at the same timestamp, `activeUnitCheckpoint()` can report `in-progress` based on lexical shard order. This ignores a potentially effective pause and permits operations that should remain blocked.

Focused validation reproduced both issues. `git diff --check` and remote CI passed; full local guards were not rerun.

apackeer:
Addressed the latest cross-shard race findings in `5bac4587`.

1. `humanActedSinceGate()` now requires a candidate human turn to be provably newer than every latest resolution. If any same-second resolution is in another shard, authority fails closed regardless of shard filename order.
2. Unit lifecycle reduction now preserves shard identity, keeps each shard's known append order, and resolves same-second cross-shard ties conservatively: paused, then another non-terminal state, then completed. This shared reduction protects both active checkpoints and completion settlement.

Regression coverage includes both filename orientations, the exact three-event authority repro, same-shard reverse ordering, pause/start and pause/resume ambiguity, completion/non-terminal ambiguity, and strictly later lifecycle events.

Verification:
- focused authority/lifecycle slice: 47/47 assertions green
- adjacent presence + shard-ordering slice: 25/25 assertions green
- version/changelog sync: 7/7 assertions green
- `bun run typecheck`
- focused Biome check
- `git diff --check`

leandrodamascena (approved):
Latest updates address the blocking authority and lifecycle issues from the previous reviews. Focused lifecycle, authority, swarm, parity, typecheck, lint, and diff checks pass.

Approving. Two non-blocking compatibility edge cases would still be worth fixing before merge:

- Backward jumps to inline Construction stages while autonomy remains enabled cannot emit Unit lifecycle receipts because `unit start/complete` are refused.
- Safe legacy Unit names beginning with a digit, such as `2fa`, were accepted by `v2` but are now reported as malformed.

Neither warrants another blocking review round, but addressing them would make the upgrade path more robust.

PR #701, opened 2026-08-02
docs: install bun and uv from a package manager, not a piped shell script

comdaze:
## Problem

Installing the two prerequisites this project asks the reader to install — `bun` and `uv` — required piping a network script into a shell, and in most places that was the only method documented. Plenty of organisations prohibit that pattern outright, so those users currently have no documented path to set the framework up.

## Change

Every such step now uses a package manager, with a link to the vendor's own installation guide for the other methods:

| Tool | Documented now | Vendor guide |
| --- | --- | --- |
| `bun` | `brew install bun` / `npm install -g bun` | https://bun.com/docs/installation |
| `uv` / `uvx` | `brew install uv` / `pipx install uv` | https://docs.astral.sh/uv/getting-started/installation/ |

Commands were verified against the Homebrew API rather than assumed: formula `bun` is in homebrew-core at **1.3.14**, which is the version this repo already pins for CI, and formula `uv` resolves. `brew install bun` is the homebrew-core spelling, not the older `oven-sh/bun/bun` tap form. Both documentation URLs return 200.

On Windows, `npm install -g bun` replaces the `irm … | iex` PowerShell form, which also let the README's three bun blocks (bash / PowerShell / CMD) collapse to two — `npm install -g bun` covers PowerShell and CMD alike.

## Deliberately not changed: Claude Code's installer

Anthropic documents the native installer as the recommended path and has **deprecated npm installation** of Claude Code. Replacing it would push readers onto a deprecated method for a tool this repository does not own, so that guidance is left to its vendor. This PR is limited to the two prerequisites the project itself asks for.

## Files

- `core/tools/aidlc-utility.ts` — the `--doctor` bun fix hint. This is the only user-visible string change; no behaviour, flag, or exit code moves.
- `harness/{claude,codex,kiro,kiro-ide,opencode}/onboarding.fills.ts` — all five harnesses, so the generated `CLAUDE.md` / `AGENTS.md` carry it too. `opencode` is the one earlier passes over this text had missed.
- `README.md`, `docs/guide/01-getting-started.md` (prerequisite table, uv block, troubleshooting row), `docs/guide/15-troubleshooting.md`, `docs/guide/harnesses/{kiro-cli,kiro-ide}.md`, `docs/reference/{06-hooks-and-tools,11-contributing,14-claude-features}.md`.
- `dist/` regenerated with `bun scripts/package.ts`.

Checked exhaustively rather than by eye: after this change no file under `core/`, `harness/`, `docs/` or `README.md` contains `bun.sh/install`, `astral.sh/uv/install`, or an `irm … | iex` form for either tool.

## Verification

- `bun scripts/package.ts --check` — all five harness trees in sync
- `bun run check` — clean (parity + typecheck + biome `--error-on-warnings`)
- `bun tests/run-tests.ts --smoke --unit --no-llm --parallel 8` — **184 files / 4522 assertions / 0 failures**
- `tests/unit/t68-version-changelog-sync.test.ts` — green
- `scripts/ci-changelog-guard.ts` against `v2` — clean

Version bumped to **2.5.34** with the README badge and a matching `CHANGELOG.md` entry, per the PR checklist in `CONTRIBUTING.md`. Happy to re-bump if `v2` ships 2.5.34 before this merges.

The integration tier is not green in my environment for reasons unrelated to this change: Anthropic blocks the live Claude substrate from my region, and `t66` and `t92` reproduce identically on an unmodified `v2` checkout. The runs above used the new `--no-llm` profile from #668, which was very handy for exactly this. Happy to re-run anything specific.


apackeer:
Thanks, Sean. The package-manager alternatives are useful, but we don't want to make them the only documented path because that assumes users already have Homebrew, Node/npm, pipx, or another package manager installed. Both Bun and uv's official guides lead with standalone installers for macOS/Linux and Windows, then list package-manager alternatives. We'd like our guidance to follow the same structure: retain or update the vendor standalone commands as the default, then add clearly labeled package-manager alternatives and state their prerequisites.

One scope correction as well: Bun is the runtime prerequisite shared by every harness; uv/uvx is not. In the shipped framework, uvx is used by the four optional AWS MCP servers in the Claude Code distribution. The repository also uses uv and Python for its maintainer documentation build. Please keep uv installation guidance in those optional/contributor contexts rather than presenting it as a general framework prerequisite.

Current `v2` has moved in the surrounding documentation, generated, and release files, so please don't spend time on a mechanical rebase. Maintainers can port the agreed wording onto current `v2`.

apackeer:
Closing this version rather than asking for a rebase. Thanks again for surfacing the installation-policy concern and for the thoughtful patch.

PR #705, opened 2026-08-03
feat: code-generation joins the unit-major construction walk (2.5.56)

apackeer:
Completes the unit-major increment from #512 - the deferred half its description called out: "code-generation still runs last; that is the follow-up, by widening the walk's block filter." Addresses the second consequence in the #503 RFC (time-to-first-code scales with total design surface, not with the skeleton).

## What

Under `Construction Iteration: unit-major`, the walk's block now covers EVERY per-unit Construction stage in graph order - the four inline design stages plus code-generation. For each Unit of Work in Bolt build order, the engine emits design 3.1-3.4 then code-generation for THAT unit before the next unit begins, so the first working code lands after one unit's design documents instead of after `4 x N` design passes, and a walking-skeleton-first bolt plan is honored end to end.

- **The widening:** `constructionUnitMajorBlock` (was `constructionDesignBlock`) drops the `mode: inline` filter; graph order already puts code-generation last per unit via `requires_stage`, so no ordering logic changes. `emitForSlug`'s unit-major branch drops its inline test to match.
- **Swarm interlock (the one real design decision):** `eligibleAutonomousSwarmBatches` refuses emission while the knob is set. The walk's coverage signal is disk artifacts (`unitCovered`); the swarm's is `SWARM_UNIT_CONVERGED` audit rows, which walk-built units never write - an autonomous swarm firing mid-walk would re-fan units the walk already built. One owner, one coverage signal. The refusal deliberately lives in the emission helper, NOT in `isAutonomousSwarmCandidate`: `isSettledAutonomousSwarm` must keep granting the report-side approve exemption for units a prior stage-major swarm legitimately built in worktrees, even if the knob flips afterwards.
- **Zero gate-machinery changes:** the per-stage gates still cascade at block end via the existing fully-covered `pick === null` delegation - now five gates (4 design + code-generation) once the whole (stage x unit) grid including code-generation is covered. Code Generation's per-unit Step 3 Plan Approval still hard-stops before generation, so a human sees each unit's design -> plan -> code in sequence even though the stage-level gates come later.
- **Stage-major default untouched** - byte-identical, proven by t210's deep-equal cases which are unchanged and green.

Version 2.5.39, CHANGELOG entry (states the behaviour change for existing unit-major users), README badge, docs (`stage-protocol.md`, `delivery-planning.md` Step 7, construction stage reference) and all five harness SKILL.md files updated.

## Why this shape (and not per-unit gates)

The state model has one checkbox per stage, so a stage's gate can only fire when it is complete for ALL units; firing functional-design's gate after only the first unit would mark the stage complete and orphan the rest. Per-unit stage checkboxes are the RFC's Option C (pipelined fan-out) - a bigger state-model change, out of scope here and better sequenced after #610's wave proposal. The five-gate late cascade keeps every existing guard (early-approve refusal, revision re-entry, presence guard) on stock machinery.

## Behaviour change for existing unit-major users

A workflow with the knob already set resumes onto the widened walk (code-generation joins the remaining grid), and the autonomous swarm no longer fires while the knob is set - the walk owns the build, serially in Bolt build order. Unset the knob (or use the stage-major default) to keep parallel batch swarms. Called out in the CHANGELOG.

## Tests

- **t267 (new):** the five-gate cascade ends on code-generation's real gate on the last unit and hands over to build-and-test; a degenerate block (all design stages `[x]`) walks code-generation-only correctly; the early-approve guard refuses while a unit's code-generation is uncovered; revision re-entry re-emits exactly the uncovered (code-generation, unit) pair.
- **t209 (updated):** case 3 is now the time-to-first-code assertion (design-covered alpha -> `code-generation/alpha`, NOT `functional-design/beta`); new case 3b pins the unit advance after the full five-stage column.
- **t210 (updated):** case 4 flips to pin the swarm suppression (knob + autonomous -> per-unit run-stage, not invoke-swarm); new case 4b is the knob-off negative control proving the same fixture still swarms without the knob. The knob-absent/junk deep-equal byte-identity cases are unchanged.
- **gen-coverage-registry:** t267 added to the none->cli reclassification pin.

## Known non-blocking note

Carried forward from #512 and now slightly wider: the sensor-fire hook keys on `Current Stage`, so mid-walk code-generation artifact writes fire under an earlier design stage's sensor list (advisory-only; the design stages share identical sensors, code-generation's differ). Same family as #512's shipped note; not fixed here.

## Verification

- Rebased onto v2 at 2.5.35 (#670); version re-bumped per the CHANGELOG conflict-trap policy.
- Deterministic default tier (smoke + unit + integration, 288 files) green. Two reds triaged: gen-coverage-registry (real - the t267 pin, fixed in this branch) and t163-reaper-steal-race (known contention flake, green solo, structurally untouched by this diff).
- `package.ts --check` clean on all five harnesses, typecheck clean, t68 version sync green post-rebase.


leandrodamascena (changes requested):
Thanks for completing the unit-major walk and for covering the ordering, gate cascade, revision re-entry, and swarm negative control. The focused suites and package parity pass, but the widened walk creates one blocking interaction with the forwarding-loop hook and one sensor-attribution defect.

### 1. P1: mandatory code-generation Plan Approval cannot cleanly stop the loop

This PR supports `Construction Iteration: unit-major` together with `Construction Autonomy Mode: autonomous`: `eligibleAutonomousSwarmBatches` suppresses swarm emission, and `emitUnitMajorRunStage` emits ordinary `code-generation/<unit>` directives from the walk (`core/tools/aidlc-orchestrate.ts:2620`, `:3098-3107`).

Code Generation nevertheless requires an explicit human answer before generation and explicitly forbids inferring approval from a forwarding-loop continuation (`core/aidlc-common/stages/construction/code-generation.md:130-145`). The stop hook cannot represent that hard stop in this path:

- `isPendingQuestionStop` rejects every pending Construction question when autonomy is `autonomous` (`core/hooks/aidlc-stop.ts:431-445`), so the blank Plan Approval question causes the hook to block and feed the conductor back into the forwarding loop for the autonomous eight-block ceiling instead of releasing to the human.
- Even without autonomy, unit-major normally emits `directive.stage: code-generation` while the state still says `Current Stage: functional-design` (or another earlier design stage). `isPendingQuestionStop` derives the slug from `Current Stage`, receives only `directive.unit`, and consequently searches `construction/<unit>/functional-design/*-questions.md` rather than the actual `code-generation-questions.md`.

The existing stop-hook test pins the first half directly: a blank code-generation question under autonomous Construction is expected to BLOCK (`tests/integration/t121-stop-hook-enforce.test.ts:930-943`). Before this PR that behavior was compatible with the autonomous swarm owner; after this PR the ordinary unit-major code-generation body must present the mandatory human Plan Approval while the state still carries the autonomous grant.

This contradicts the PR statement that Step 3 “still hard-stops before generation.” At best the user is trapped behind repeated stop-hook continuations until the cap releases; at worst a continuation is treated as permission to proceed despite the stage prohibition. Please make the pending-question carve-out resolve the active `directive.stage` rather than only Current Stage, and exempt this mandatory Plan Approval from the autonomous keep-running rule when unit-major routes code-generation through the interactive walk. Add an integration case that drives the actual unit-major directive into the stop hook with a blank per-unit code-generation Plan Approval.

### 2. P2: code-generation sensor results are attributed to the earlier design stage

The walk emits a later stage without moving Current Stage, but `aidlc-sensor-fire.ts` selects `sensors_applicable`, passes `--stage`, and therefore chooses diagnostic/audit placement exclusively from the state pointer (`core/hooks/aidlc-sensor-fire.ts:154-177`, `:207-216`).

Reproduction shape: cover alpha’s four design stages in a feature/unit-major workflow while Current Stage remains functional-design; `next` emits `code-generation/alpha`. TS/JS writes run the linter/type-check entries through `--stage functional-design`, so any `SENSOR_FAILED` row and `.aidlc-sensors/<stage>/...` detail are recorded under functional-design while code-generation appears clean.

I saw the PR body’s known-note disclosure. This was mostly harmless while the interleaved stages shared document-oriented work; widening the interleave to source generation makes stage attribution materially wrong for code diagnostics. The hook needs the active directive stage (or an equivalent runtime marker) rather than the state cursor during an interleaved unit-major body.

### Merge state

The PR is currently `CONFLICTING` / `DIRTY` against `v2` and needs a rebase. I did not find a current version collision for `2.5.39`; the conflict itself still needs resolution and regenerated parity.

### Verification

- t209, t210, and t267 through the official unit runner: 24/24 passed.
- t121 stop-hook integration suite: 54/54 passed; its autonomous pending-question case confirms finding 1’s incompatible invariant.
- Coverage registry plus version/changelog sync: 41/41 passed.
- `bun scripts/package.ts --check`: all five harnesses in sync.
- `git diff --check`: clean.
- Remote CI is green, including contract/typecheck checks.

I recommend keeping this at `CHANGES_REQUESTED` until the Plan Approval/stop-hook interaction is fixed and covered. The sensor attribution issue should also be corrected or explicitly split into a tracked follow-up with the resulting diagnostic limitations documented.

apackeer:
Addressed both findings in `3cfcf4bf`, rebased onto current `v2` / 2.5.42.

- Plan Approval now resolves the active directive stage and unit through load steering and task-status synchronization, including autonomous unit-major execution.
- Sensor dispatch now uses a state-bound active-directive marker, with stale-marker fallback and isolated-run cleanup.

Verification: 11 focused files, 267 assertions passed; package parity, typecheck, and lint pass. Please re-review when CI completes.

leandrodamascena (changes requested):
The corrective commit fixes sensor attribution and allows mandatory Plan Approval to reach the human, but two blocking unit-major paths remain.

[P1] Preserve the first stage when task synchronization updates status
The sync-statusline hook moves `Current Stage` to each interleaved directive. After the final `code-generation` unit, the cursor remains on `code-generation`, so the completed-grid path presents that gate directly and skips the four earlier design gates. My probe expected `functional-design` but received `code-generation`.

[P1] Restrict the autonomous stop exception to Plan Approval
The exception currently accepts any blank `[Answer]:` in a code-generation questions file. An unrelated clarification question therefore stops an autonomous unit-major run, defeating the autonomy guard. My probe used a generic “Which edge case?” question and the hook allowed the stop.

The official focused suites passed: 115 tests and 324 assertions. Both missing negative probes failed as described. Remote CI and `git diff --check` are green.

apackeer:
Addressed both follow-up findings in `383831f9`.

- Unit-major task/status synchronization now preserves the durable first-stage cursor, `In Progress`, and checkbox for a valid interleaved directive, so the completed-grid gate cascade starts at `functional-design` instead of skipping directly to `code-generation`. The transient active-directive marker is still refreshed for sensor attribution.
- The autonomous stop exception now accepts only the active unit's exact pending Plan Approval section in `code-generation-questions.md`; unrelated blank clarification questions keep the autonomous loop running.
- The plan-approval dispatch guard now resolves the active directive before falling back to `Current Stage`, preserving deterministic enforcement while the durable cursor remains on the earlier design stage.

Verification: focused unit/integration slice passed 5 files / 158 assertions; package parity, all TypeScript configs, lint/check, coverage-registry freshness, and `git diff --check` pass. Please re-review when CI completes.

leandrodamascena (approved):
The latest commit resolves the previously blocking gate-cascade and Plan Approval stop issues. Focused unit-major, hook, sensor, parity, typecheck, lint, and CI checks pass.

Approving with one recommended follow-up: switching a fully converged stage-major swarm to `unit-major` causes `next` to re-emit `code-generation/alpha` instead of presenting the final gate because swarm convergence is no longer considered on the unit-major path. This is a supported but uncommon mid-stage configuration change, so it does not warrant another blocking review round.

The branch still needs rebasing onto current `v2` and the resulting version/changelog conflicts resolved before merge.

PR #709, opened 2026-08-04
fix: materialize active-space cursor after clone (2.5.42)

### LINKED ISSUE #706: [Bug]: `aidlc/active-space` is gitignored but nothing ever writes it — only `space switch` calls setActiveSpaceCursor, so a cloned workspace permanently lacks the shipped cursor
### Version

2.5.36 (behaviour present well before this release; verified against 2.5.35 and 2.5.36)

### Summary

`dist/<harness>/.gitignore` ignores `aidlc/active-space` alongside `…/intents/active-intent`, treating both as "per-user session cursors" that the user's own first `/aidlc` will write. But the engine only ever writes the **space** cursor from one place: the `/aidlc space <name>` switch handler. `birthIntent` writes `active-intent`; nothing writes `active-space`.

The result: a user who clones a shared workspace has **no** `aidlc/active-space` and nothing in the normal workflow creates one. `activeSpace()` silently falls back to `"default"`, so behaviour is correct — but the shipped cursor is destroyed on the first commit and never comes back, and the packager's own stated reconciliation ("their first `/aidlc` cursor-write stays untracked") does not hold.

### Evidence

**Only one caller writes the space cursor** (`core/tools/aidlc-lib.ts:1714` defines it):

```
$ grep -rn 'setActiveSpaceCursor' core/ harness/
core/tools/aidlc-lib.ts:1714:export function setActiveSpaceCursor(projectDir: string, name: string): void {
core/tools/aidlc-utility.ts:86:  setActiveSpaceCursor,
core/tools/aidlc-utility.ts:4194:  setActiveSpaceCursor(projectDir, target);
```

`core/tools/aidlc-utility.ts:4194` is inside `handleSpace`, reached only after the existing-space guard:

```ts
const spaces = listSpaces(projectDir);
if (!spaces.some((s) => s.name === target)) {
  die(`Unknown space "${target}". ...`);
}
setActiveSpaceCursor(projectDir, target);
```

**The intent cursor, by contrast, is written at birth** — `core/tools/aidlc-lib.ts:1904`, at the end of `birthIntent`:

```ts
setActiveIntentCursor(projectDir, dirName, space);
```

So the two cursors the `.gitignore` groups together are not symmetric:

| Ignored path | Auto-created by normal use? | Where |
|---|---|---|
| `aidlc/spaces/*/intents/active-intent` | Yes — every intent birth | `aidlc-lib.ts:1904` |
| `aidlc/.aidlc-clone-id` | Yes — minted when absent | `aidlc-lib.ts:2115`+ |
| `aidlc/active-space` | **No** — only `/aidlc space <name>` | `aidlc-utility.ts:4194` |

**Confirmed on a real run.** A workspace that has completed intent birth and audit writes:

```
$ ls -a aidlc/
.aidlc-clone-id   spaces
$ cat aidlc/spaces/default/intents/active-intent
260803-word-book
$ cat aidlc/.aidlc-clone-id
cd425b11e555
$ cat aidlc/active-space
cat: aidlc/active-space: No such file or directory
```

The clone-id and intent cursor both materialised. The space cursor never did.

### Why this contradicts the packager's stated design

`scripts/package.ts:370-380` ships the cursor deliberately and explains how shipping-it and ignoring-it are supposed to reconcile:

```
// NOTE: it is GITIGNORED in the user's workspace (a per-user session cursor,
// vision 5.1 - teammates legitimately point at different spaces at once), yet
// dist must SHIP it as part of the shell. The two reconcile: the dist
// .gitignore ignores aidlc/active-space for the END USER (their first /aidlc
// cursor-write stays untracked), while OUR repo commits the shipped pointer
// once (git add -f on the seed commit) ...
```

The reconciliation rests on "their first `/aidlc` cursor-write". That write does not exist for the space cursor. So for any workspace committed to git, the shipped pointer is silently dropped and never regenerated.

`docs/guide/03-spaces-and-intents.md:82` makes the same symmetry claim to users:

> **Two cursors** — `active-space` and `active-intent` — record *where you are right now*. They are per-user (gitignored) …

and `docs/guide/03-spaces-and-intents.md:40` annotates the tree with `active-space ← cursor: which space you're in (gitignored, per-user)`, which reads as "this file exists in your workspace" — it does not, for anyone who cloned.

### Impact

Low severity (no incorrect behaviour — the fallback is correct and documented at `aidlc-lib.ts:988-1000`), but it has real costs:

1. **The shipped shell is not reproducible from a commit.** `dist/` ships `aidlc/active-space`; a clone of a committed workspace lacks it. The workspace shell is otherwise fully reconstructible.
2. **Teaching/demo friction.** Walking someone through the space model, the file the docs and the architecture reference (`docs/reference/01-architecture.md:404`) point at is absent, so the explanation does not close. This is what surfaced it for us — we run AI-DLC workshops from a git-hosted workspace and had to un-ignore the path in our own harness trees to keep the space walkthrough coherent.
3. **The `.gitignore` comment is misleading** as written, since it justifies ignoring a file on the basis of a per-`/aidlc` write that never happens for that path.

### Suggested fix

Any one of these resolves the inconsistency; (a) seems closest to the documented intent:

**(a) Make the cursor auto-create, matching `active-intent`.** Have `birthIntent` (or the same seam that ensures workspace dirs) write the space cursor when absent, e.g. `if (!existsSync(cursor)) setActiveSpaceCursor(projectDir, activeSpace(projectDir))`. `setActiveSpaceCursor` is already best-effort and idempotent, and writing the resolved value is a no-op semantically. The packager comment then becomes true as written.

**(b) Stop ignoring `aidlc/active-space` in the dist `.gitignore`.** The shipped `default` pointer travels with the repo like the rest of the shell; a user who never switches spaces never dirties it, and a user who does switch is making a deliberate, arguably shareable choice. (`active-intent` still warrants ignoring — it genuinely changes on every birth.)

**(c) Keep both as-is and fix the docs + comment** to say the cursor is absent until the first explicit `space switch`, and that absence means `default`. Cheapest, but leaves `dist/` shipping a file that cannot survive a commit.

### Notes

- `handleSpaceCreate` (`core/tools/aidlc-utility.ts:4438`+) intentionally does not move the cursor ("Switch to it with `/aidlc space <name>`"), so creating a space does not create the cursor either. That separation seems deliberate and is not what this issue is about.
- `/aidlc --doctor` is unaffected: `workspace shell ready` checks the engine dir and `aidlc/spaces/default/memory/` only (`core/tools/aidlc-utility.ts:1545`), not the cursor.

apackeer:
## Summary

- recreate a missing gitignored `aidlc/active-space` cursor during SessionStart and active-intent writes
- publish the cursor atomically without replacing a concurrent explicit space switch
- ignore transient staging files, update the workspace documentation, regenerate every harness distribution, and bump the release metadata to 2.5.42

## Verification

- targeted unit + integration slice: 4 files, 83 assertions, PASS (`2026-08-06T00-58-08Z`)
- `bun run check` (package parity, all TypeScript configs, Biome over 570 files)

Fixes #706

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.42] - 2026-08-06
+
+Restores the gitignored `aidlc/active-space` cursor after cloning a committed workspace. The cursor still defaults safely to `default` when absent, but SessionStart and active-intent writes now atomically materialize it without overwriting a concurrent explicit space switch. **Upgrade:** re-copy your `dist/<harness>/` shell so the updated library and SessionStart hook are installed; no repository migration or configuration change is required.
+
+* A fresh clone now recreates `aidlc/active-space` on SessionStart, even before an intent exists, so the shipped workspace model remains visible without committing per-user navigation state.
+* Intent birth, intent switching, and flat-layout migration recreate a missing space cursor through the same best-effort cursor primitive.
+* Cursor creation stages complete bytes and publishes them through an atomic no-replace link, so it cannot overwrite a concurrent `/aidlc space <name>` switch.
+
 ## [2.5.41] - 2026-08-05
 
 Code Generation's plan-before-generation ordering is now enforced deterministically. A field report showed the conductor generating code first and backfilling `code-generation-plan.md` beside `code-summary.md`, turning the plan into a retroactive summary; the new plan-approval PreToolUse guard refuses the developer-agent dispatch until the target unit's plan is on disk and the human has answered "Approve Plan". **Upgrade:** re-copy your `dist/<harness>/` shell into the project so the new hook and its registrations are installed.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.5.41-blue)
+![version](https://img.shields.io/badge/version-2.5.42-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/hooks/aidlc-session-start.ts
+++ b/core/hooks/aidlc-session-start.ts
@@ -18,7 +18,7 @@
 //   clear   → SESSION_STARTED
 //   compact → no emission (PreCompact already fired)
 //
-// The hook is a no-op if aidlc-state.md is absent in cwd (no active workflow).
+// With no aidlc-state.md, only cursor/include bootstrap runs; workflow work is a no-op.
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { appendAuditEntry } from "../tools/aidlc-audit.ts";
@@ -28,6 +28,7 @@ import {
   activeIntentUuid,
   activeSpace,
   clearSessionIntentUuid,
+  ensureActiveSpaceCursor,
   errorMessage,
   findIntentByUuid,
   getField,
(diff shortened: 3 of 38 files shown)


leandrodamascena (approved):
The implementation correctly restores the missing cursor and protects concurrent explicit space switches. Focused tests, package parity, typechecks, lint, and CI are green.

Approved for the behavioral change. Before merge, please rebase and re-bump because `2.5.37` is already occupied on `v2`.

Suggested non-blocking follow-ups:

- Stop shipping `aidlc/active-space=default`, or exclude it from upgrade-copy instructions, so upgrades cannot reset a user’s non-default space.
- Add a fallback for filesystems without hard-link support, or surface the degraded materialization through `doctor`.

Verification: 58 unit assertions, 32 integration assertions, `bun run check`, package parity, and `git diff --check` all passed.

PR #712, opened 2026-08-04
feat: add a tutorial scope for a taught single-sitting run (2.5.38)

wowzoo:
## What

Adds a tenth core scope, `tutorial`, for a taught single-sitting run that reaches working code without dropping the design stages.

Neither existing teaching-adjacent scope covers this. `poc` skips `application-design`, `units-generation`, and `delivery-planning` on the premise that a spike is throwaway — but those design stages are the part of the method worth showing. `workshop` skips all of Ideation and opens at `reverse-engineering`, so it assumes a facilitator-framed project and an existing codebase to analyze; that does not fit a from-scratch exercise, and its 25 executing stages are more than one sitting holds.

`tutorial` keeps the design spine and instead drops the three stages that cost the most while producing artifacts that are optional downstream:

- `user-stories` — `mode=mob`, parallel support-agent dispatch plus integration and disagreement triage. The most expensive planning stage.
- `refined-mockups` — re-refines the same screens as `rough-mockups`, which also produces `mockup-visual-ref`, so the UI reference `code-generation` consumes stays intact.
- `functional-design` — `for_each=unit-of-work`, and its artifacts are optional inputs to `code-generation`.

Result: 12 EXECUTE of 32 behind 9 approval gates. NFR, infrastructure, CI, and the operation phase are out of scope for a local exercise and are skipped.

Authored as data only, per [Contributing § Adding a Scope](docs/reference/11-contributing.md#adding-a-scope): one scope file plus a `scopes:` tag on the 12 member stages. No TypeScript edit registers it — `validScopes()` derives from the scope files.

## User-visible changes

- New scope `tutorial`, selected with `--scope tutorial`. `depth: Minimal` with `testStrategy: Minimal` pinned in frontmatter, so no `--depth`/`--test-strategy` flag is needed at launch.
- `keywords` is empty by design, so the scope is never inferred from freeform text and a taught session always starts from the same grid. A freeform intent mentioning "tutorial" still routes to `feature`; the scope must be passed explicitly.
- `skeleton: on`, so Construction opens with the walking-skeleton ceremony when practices resolve to scope-dependent — the first-Bolt gate and the ladder prompt are shown to the learner rather than described.
- EXECUTE (12): the 3 initialization stages, `intent-capture`, `scope-definition`, `rough-mockups`, `requirements-analysis`, `application-design`, `units-generation`, `delivery-planning`, `code-generation`, `build-and-test`. `operation` is the only phase skipped in full, so `init` emits one `PHASE_SKIPPED`.
- `/aidlc --doctor` now reports `10 scopes valid`.
- Nothing changes for the nine existing scopes. **Upgrade:** re-copy your `dist/<harness>/` shell to pick up the new scope file and the recompiled grid.

## Testing

Verified on `dist/claude` unless noted.

| Check | Result |
|---|---|
| `compile --check`, `scope-table --check` | exit 0 on all five harness trees (one caveat below) |
| `resolve tutorial` | EXECUTE set == the 12 tagged stages, SKIP 20 |
| `next --scope tutorial` | `12 of 32 stages, 9 approval gates` |
| `intent-birth --scope tutorial` | writes `- **Scope**: tutorial`; first post-init stage is `intent-capture` (not `reverse-engineering`, as `workshop` would give) |
| `AWS_AIDLC_DEFAULT_SCOPE=tutorial` + `doctor` | reported valid; `10 scopes valid` |
| `detect-scope --from-text --input "tutorial for teaching the method"` | returns `feature`, not `tutorial` — confirms `keywords: []` opts out of inference |
| `designer export` | `scopes` 9 → 10 with `stages=32`, `artifacts=122`, `agents=14` unchanged, so the blast radius stays on the scope axis |
| `typecheck`, `lint`, `package.ts --check`, `gen-coverage-registry --check` | all exit 0 |
| `tests/unit` | 85 failures outside `t255`, byte-identical to the set this branch's base (`c73ee984`) produces on this machine |
| scope-axis integration (`t30`, `t39`, `t39-count`, `t66`, `t128`, `t130`, `t141`, `t31`, `t145`, `t188`, `t224`, `t40`, `t41`, `t92`) | 360 pass, 0 fail |

On the unit suite: I ran the base in a clean worktree for comparison. Raw totals are 92 failures on this branch against 91 on base, but every one of the differing names is inside `t255` (`workspace-sync`), which never mentions `scope` and whose failing subset varies per run (7 here, 6 on base; earlier runs gave 5→6→6 and 5→5→5). Excluding `t255`, both sides fail the same 85 tests — so nothing outside that flaky suite moved.

Tests and docs that hardcoded the nine-scope set were extended, as Contributing warns they must be: `t39` (per-scope phase sequence, now 30 observables), `t39-count`, `t30`, `t60`, `t67`, `t125`, `t225`, and the `t66` designer-export golden. Doc updates land in the same change: the scopes guide (section, routing row, use-case row, and the `t244`-guarded Stage-by-Scope Matrix column), the customization valid-values list, and the orchestrator scope-to-stage table and breakdown.

## Notes on two things I found but did not fix

- **`opencode`'s `scope-table --check` fails, on `v2` too.** Its `skills/aidlc/SKILL.md` writes the marker comment with an em dash (`— do NOT hand-edit`) while the checker expects a hyphen, so it reports "missing scope-table markers" regardless of this change. I refreshed that file's table while preserving its own marker lines, and left the mismatch alone as pre-existing. Happy to fix it here or in a separate PR if you prefer.
- **The doctor sample output in the getting-started and CLI guides reads `29 advisories`, but `v2` emits 27.** I updated only the scope count in those samples (`9` → `10`) and left the advisory number untouched, since it was already stale before this change.

## Two decisions worth a second opinion

- **Version.** I bumped to `2.5.38`. The first pass targeted `2.5.37`, but the question-rendering fix took that number while I was preparing this, so I rebased onto `c73ee984` and re-bumped per the Changelog Policy conflict-trap. If it happens again before merge I will do the same.
- **No runner.** I did not set `runner: true`, so there is no `/aidlc-tutorial` command. The scope is fully usable via `/aidlc --scope tutorial`, and only four high-traffic scopes ship runners today. Say the word if you would rather it have one.


leandrodamascena (changes requested):
The new scope is well integrated into packaging, generated distributions, documentation, and the static scope grid. However, the runtime behavior does not currently match the tutorial experience described by the PR.

### 1. [P1] Walking-skeleton gate occurs after all units

`core/scopes/aidlc-tutorial.md:23-25` says that `skeleton: on` places the first-Bolt gate and autonomy ladder in front of the learner.

For this scope, `code-generation` is the first Construction stage. It is also a per-unit stage. The orchestration logic in `core/tools/aidlc-orchestrate.ts:2926-2959` suppresses the gate for every uncovered unit and restores the real gate only after all units are covered.

For a plan containing units A and B, the resulting sequence is:

```text
code-generation(A, gate:false)
code-generation(B, gate:false)
code-generation(B, gate:true)
```

This contradicts `core/aidlc-common/protocols/stage-protocol.md:112-137`, which requires the walking-skeleton gate after the first Bolt and the ladder before subsequent Bolts. By the time the ladder appears, every `code-generation` Bolt has already run, so selecting autonomous or gated execution has no practical effect.

The scope needs either a Construction route compatible with Bolt-level gating or orchestration support that settles and gates the first unit separately before continuing the per-unit stage.

### 2. [P1] Rough mockups do not reach code generation

`core/scopes/aidlc-tutorial.md:43-45` says that `rough-mockups` produces `mockup-visual-ref`, preserving the UI reference consumed by `code-generation`.

That artifact does not exist. `rough-mockups.md:12-15` produces `wireframes`, `user-flow`, and questions. Neither `application-design.md:19-31,60-64` nor `code-generation.md:16-32,68-80` consumes or reads those artifacts. The actual consumer is `refined-mockups`, which this scope skips.

Consequently, the tutorial asks the learner to create mockups but then generates code without supplying those mockups to the developer agent. This breaks the advertised design-to-code teaching arc.

Please either retain `refined-mockups`, teach `application-design` or `code-generation` to consume the rough artifacts, or remove `rough-mockups` from this scope and revise the claim.

### 3. [P1] Skipping user stories conflicts with mandatory downstream work

The scope skips `user-stories`, but several retained stages still require story-based outputs:

- `units-generation.md:129-134` requires every story to be mapped and every unit to have stories.
- `delivery-planning.md:125-131` requires Requirements → Stories → Architecture verification and story-to-requirement coverage.
- `code-generation.md:84` mandates story-to-code-step traceability.
- `code-generation.md:157-160` requires a stories summary in the developer delegation.

These are not described as optional when the stories artifact is absent. A faithful tutorial run must therefore invent stories that were never produced or violate the retained stage protocols.

If omitting `user-stories` is important for session length, the downstream stages need an explicit no-stories fallback based on requirements. Otherwise, `user-stories` should remain in the route.

### 4. [P2] No behavioral coverage for the new scope

`tests/.coverage-registry.json:3870-3874` explicitly records `tutorial` with an empty `coveredBy` list and `status: UNCOVERED`.

The added tests validate discovery, counts, generated grids, and initialization metadata, but none advances a tutorial workflow through delivery planning with multiple units, skeleton-stance classification, first-unit code generation, first-Bolt approval, autonomy selection, remaining units, and build/test completion.

A focused orchestration test would have exposed the delayed walking-skeleton gate described above. Please add at least one deterministic route test for the intended multi-unit tutorial behavior.

### Verification

`bun run check` passes, including package drift, typecheck, and lint. All 203 tests directly related to the changed files pass. The broader CI profile produced unrelated or timing-sensitive failures; deterministic failures checked against `origin/v2` reproduce there as well and were not attributed to this PR.

The static integration is sound, but the behavioral issues above prevent the scope from delivering the tutorial flow it documents.

apackeer (changes requested):
Directionally, this is sound: `origin/v2` has no from-scratch, single-sitting teaching scope that preserves the design spine, and a declarative `core/scopes/` definition plus stage membership is the correct extension layer. The generated distributions are also in sync.

I'm requesting changes for the following:

1. **[P1] Walking-skeleton timing:** `core/scopes/aidlc-tutorial.md:23` promises the first-Bolt gate and autonomy ladder, but multi-unit code generation emits every uncovered unit with `gate:false` and presents the real gate only after all units are covered. A tutorial replay produced alpha/false, beta/false, then beta/true. The learner therefore chooses autonomy after all Bolts have already run.
2. **[P1] Missing stories fallback:** the scope skips User Stories, while Units Generation still requires every story assigned and every unit to have stories, Delivery Planning verifies Requirements -> Stories -> Architecture, and Code Generation mandates story traceability plus a stories summary. Retain User Stories or add an explicit requirements-based fallback to every retained stage.
3. **[P1] Mockup chain is broken:** Rough Mockups does not produce `mockup-visual-ref`; it produces `wireframes` and `user-flow`. Only Refined Mockups consumes them, while Application Design and Code Generation do not. Retain Refined Mockups, wire the rough artifacts into downstream stages, or remove the claim/stage.
4. **[P1] Rebase and re-bump:** the branch uses `2.5.38`, while current `v2` is `2.5.59` and already contains an unrelated `2.5.38` entry. Please rebase and bump the version source, README badge, and changelog together.
5. **[P2] Add routed behavior coverage:** the coverage registry explicitly marks `scope:tutorial` as `UNCOVERED`, and several broad scope tests still iterate only the old nine scopes. Please add a deterministic multi-unit tutorial journey covering delivery planning, first-unit generation, first-Bolt approval, autonomy selection, remaining units, and build/test.
6. **[P2] Complete the documentation sweep:** stale nine-scope references remain in README, the glossary, CLI valid values, architecture/skill references, and harness-engineering docs. The orchestrator table also labels Functional Design as `2.9` instead of `3.1`. The help description's "full arc" wording conflicts with skipping Operation entirely.

UX-wise, the explicit-only `tutorial` name, Minimal defaults, absence of a dedicated runner, and doctor count change fit the existing product model. The first-Bolt promise and "full arc" wording do not yet match the behavior.

wowzoo:
Closing this for now — I'll revisit it against a newer AI-DLC version.

The base has moved from 2.5.38 to 2.6.2, and 2.6.1 reshaped the graph this PR indexes: the framework now ships 33 stages, `application-design` became `domain-design`, a new `contract-design` was inserted at 2.8, and Delivery Planning moved to 2.9. This branch adds a scope row to twelve stage files, one of which — `core/aidlc-common/stages/inception/application-design.md` — no longer exists. So this is not a conflict to resolve but a grid to re-derive against the new graph.

Better to re-derive the tutorial scope once the design-stage rework has settled than to land a grid keyed to the old one.

PR #404, opened 2026-06-18
feat: first-class observability artifacts across NFR stages (2.5.75)

jeromevdl:
Fixes #398 - observability was missing from the NFR chain: `nfr-requirements` listed it as a category while `requirements-analysis`, `nfr-design`, and `infrastructure-design` ignored it, and no observability artifact flowed between stages.

Maintainer-side integration of @jeromevdl's design onto current `v2` (see the integration comment for the full walkthrough):

- NFR Requirements produces `observability-requirements.md` and NFR Design produces `observability-design.md`, both mapped to `[service]` unit kinds via `produces_kinds`.
- NFR Design consumes `observability-requirements` and Infrastructure Design consumes `observability-design` as required inputs; the NFR Design question focus list covers observability decisions.
- The architecture/platform boundary is preserved: Infrastructure Design continues producing the platform-specific `monitoring-design.md` (now framed as implementing the observability-design strategy), and Observability Setup continues consuming it.
- Designer export grows to 126 artifacts / 14 agents; docs, fixtures, test pins, and all seven harness dists regenerated.

Verified with the full smoke+unit+integration tier green on this head (305 files unit+integration locally, plus CI).

PR #716, opened 2026-08-05
feat: project and propagate stale stage results

djoo-lgcns:
## Contribution context

Head branch: `feat/stage-validity-projection`
Version metadata: `2.6.16`

## Local validation

- Focused validity/directive/single-stage suites: **passed** (199 assertions)
- Authority and lifecycle suites: **passed** (148 assertions)
- Deterministic integration profile: **passed** (103 files, 1,414 assertions; live-model gates disabled)
- Generated harness packaging and drift check: **passed**
- Typecheck, lint, and coverage-registry freshness: **passed**
- Live Bedrock preflight: **not run** (local AWS credentials are expired)

## Summary

Separate historical stage completion from current AI-DLC-owned artifact validity.

This revision keeps runtime resolution instance-aware but persists a compact stage-level receipt:

- one dependency-free shared artifact vocabulary plus runtime resolver;
- active Bolt DAG unit/kind expansion;
- `produces_kinds` filtering;
- aggregate structure/content fingerprints per canonical artifact;
- receipt-recorded input-edge propagation;
- fail-open migration from receipt-less, schema-1, and superseded Draft receipts.

## Problem

`completed` is historical execution state. It does not prove that a result still matches the artifact basis against which it completed.

A filename-only or static-edge implementation is insufficient for v2 because:

- `build-test-results` and `load-test-results` use `test-results.md` physically;
- per-unit outputs depend on the active Bolt DAG;
- `produces_kinds` prunes outputs by unit kind;
- an optional consume declared in frontmatter may not have existed in the completed attempt;
- `requires_stage` can represent ordering as well as semantic dependency.

### Shared runtime artifact resolution

`aidlc-artifact-vocabulary.ts` owns canonical-to-physical filename aliases and CodeKB stage ownership; `aidlc-artifact-resolution.ts` uses that leaf vocabulary to resolve concrete runtime artifact instances.

For per-unit producers it uses the active Bolt DAG and unit kinds. A malformed DAG fails closed. A no-DAG legacy or isolated single-stage run may fall back to existing unit directories. `produces_kinds` filters the expected output matrix.

The resolver is also used by the existing state artifact guard and orchestrator artifact paths, preventing canonical IDs such as `build-test-results` from being checked at nonexistent filenames.

### Compact schema-2 completion receipt

Each main-workflow `STAGE_COMPLETED` event may record a `Validation Basis` with:

- graph-contract fingerprint;
- project type;
- canonical input and output artifact IDs;
- unique producer and required semantics;
- resolved instance and present counts;
- an aggregate structure fingerprint per canonical artifact;
- an aggregate content fingerprint per canonical artifact;
- receipt-recorded input edges (`from_stage` → `to_stage_artifact`) captured from the artifact instances present when completion is reported. This is report-time evidence, not runtime read instrumentation.

Structure fingerprints aggregate `structureHash` sums across all present unit instances of a canonical artifact; content fingerprints aggregate `contentHash`. A single canonical row can therefore back multiple unit instances, and the receipt's storage cost stays constant regardless of unit count.

The state machine treats a stage's receipt row as authoritative for its own outputs; the resolver decides which files count. Absent optional inputs remain absent in the aggregate. Structure-only changes and content-only changes are distinguishable through the two fingerprints.

### Fail-open migration

Legacy audit ledgers, schema-1 receipts, and superseded Draft receipts are not rejected. Migration triggers on the next `STAGE_COMPLETED` at approve time. Recent iteration receipts also fall through fail-open until the stage completes again.

### Downstream propagation

When a stage's own row becomes stale under the current graph contract, its receipt-recorded `observed_input_edges` are unwound to mark any downstream stage that consumed the invalidated aggregate. Cycles terminate without looping.

## Scope

Validity here is dependency validity between AI-DLC-owned artifacts represented by the current graph contract and report-time completion receipts.

Explicitly out of scope for this PR:

- correctness of source trees, generated code, or Git working state as a whole;
- Git commit provenance or repository snapshot equivalence;
- CI or CD pipeline execution validity;
- deployment target validity;
- external-system validity.

The feature complements historical stage completion; it does not replace it. `completed` remains the answer to *"did this stage produce output once?"* Validity answers *"does that output still describe the current state under the current graph contract?"*

## Tests

The test commit covers:

- live v2 contracts for `code-generation` and `build-and-test`;
- canonical filename aliases;
- active Bolt DAG expansion and `produces_kinds` filtering;
- malformed-DAG fail-closed and no-DAG fallback behavior;
- compact aggregate receipt parsing;
- structure versus content drift;
- per-unit fan-in aggregation without exposing unit rows in the receipt;
- absent versus present optional inputs;
- receipt-recorded transitive propagation;
- reopened-stage propagation;
- cycle-safe propagation;
- filesystem/audit integration.

## Historical independent A/B evaluation

Compared the pre-review prototype against the frozen upstream baseline on the same Task CRUD goal, evaluator revision, Kiro CLI environment, and post-completion mutation payload.

> **Review follow-up:** the experiment below established the detector and propagation signal, but its hard routing barrier was intentionally removed in response to review. The current implementation keeps the normal directive kind and attaches a machine-readable `stage_validity` advisory; enforcement and recovery policy are deferred.

### Why a forked evaluator

`awslabs/aidlc-workflows` `v2-evaluator` (`scripts/aidlc-evaluator/`) runs the paired A/B pipeline end-to-end but couples three assumptions that block reproducibility outside AWS-issued environments and outside the evaluator repository:

- Bedrock is the LLM backend for both the human-analog and the qualitative scorer, hard-coded through `boto3` — an environment without Bedrock credentials silently falls back to canned strings, which reads as success in the transcript;
- Kiro dispatch uses the v1 `/skill aidlc-orchestrator` slash — Kiro CLI 2.16.2 (v3 agent engine) rejects the v1 slash and requires ACP (Agent Client Protocol) for headless multi-turn runs;
- The Kiro distribution (`dist/kiro/.kiro`) is auto-detected from the evaluator's own checkout, so both variants read the *evaluator's* skills/tools/agents rather than each ref's own dist — the A/B result is invalid because the candidate feature never actually runs.

The [`feat/codex-cli-evaluator-backend`](https://github.com/djoo-lgcns/aidlc-workflows/tree/feat/codex-cli-evaluator-backend/scripts/aidlc-evaluator) fork addresses these three, strictly opt-in and backward compatible:

- LLM backend shim (`packages/shared/src/shared/llm.py`) with a `codex-cli` backend (Codex CLI subprocess, `--ephemeral`, `--sandbox read-only`) alongside the original `bedrock` path — resolved per-component through `AIDLC_EVAL_{HUMAN,SCORER}_BACKEND` / `AIDLC_EVAL_LLM_BACKEND` env vars; unset env keeps Bedrock as before.
- v2 workspace layout detection and dispatch through `kiro-cli acp` — a bidirectional JSON-RPC 2.0 adapter that answers agent-initiated `session/request_permission` / `terminal/*` / `fs/*` requests, streams `session/prompt` responses, and applies a two-part timeout (per-turn maximum plus a 120-second notification-idle watchdog). The v1 chat path is preserved as a fallback via `AIDLC_EVAL_KIRO_MODE=chat_watchdog`.
- Per-ref `dist/kiro/.kiro` sparse-checkout (`_setup_dist_from_rules`) that installs each ref's own dist into that ref's workspace and records a sha256 `kiro-dist-manifest.txt`. The legacy auto-detect remains available with a WARN, so any user relying on the previous behavior sees no functional change.

None of these changes affect what the evaluator measures; they change only where and how it runs. Scripts using the shipped defaults (Bedrock, evaluator-repo dist) continue to work.

### Frozen revisions

- Baseline: `upstream/v2` @ `d7807fe1230defdbf81e544ac5d044e012bb08d7`
- Evaluated candidate: pre-review prototype @ `dbb8986af71a6842a89214a7604a9dfdb35ff3db` (5 commits: 3 feature + 2 chore)
- Evidence was captured against pre-rebase-2 tips (baseline `upstream/v2` @ `18bcc4688e20258e713de750d725f83981f88bde`, candidate `feat/stage-validity-projection-rebased` @ `a4bca7009699f1afa323884b7f2069d8783f1a5b`). Current reviewed head: `1d331c046dac350a514ba6ef6f44d0c3541d82df`. The detector/resolver lineage remains, while review follow-ups changed surfacing to advisory, sealed completion authority, and made capture failures untracked and non-blocking.
- Evaluator (fork used to run the paired evaluation): `djoo-lgcns/aidlc-workflows` `feat/codex-cli-evaluator-backend` @ `d068cad8033c002aeab8ddb28f6384ddf7e2ba12`
- Kiro CLI: `kiro-cli 2.16.2`
- Model: `claude-opus-4.8`
- Mutation payload SHA256: `29e28a39cb8d0c2b224bdce9eb5c60abc82ace4b39452d6a65e2a279ec367af1` (appended byte-for-byte to each variant's `unit-of-work.md`)

### Historical deterministic detection probe

Model-independent evidence from the pre-review enforcement prototype. `bun .kiro/tools/aidlc-orchestrate.ts next` in each post-mutation workspace produced disjoint results. The candidate JSON below is historical; the current implementation emits the normal directive plus `stage_validity`.

**Baseline:**

```json
{
  "kind": "done",
  "reason": "Workflow complete — no in-scope stage remains after build-and-test",
  "narration": "That is everything on the plan. Your work is finished and written up."
}
```

**Candidate:**

```json
{
  "kind": "error",
  "message": "Completed stage result(s) no longer match their validation basis. Directly stale: units-generation, code-generation. Downstream revalidation required: build-and-test. Re-enter the earliest affected stage with /aidlc --stage units-generation."
}
```

This evidence isolates the detector: the candidate's `aidlc-validity.ts` + `aidlc-artifact-resolution.ts` project stage-level validity from the mutation and propagate it along the graph's consume/produce edges without any model in the loop. The current PR retains that detection but reports it as an advisory instead of returning the historical error directive.

### Kiro transcript verdicts

Both variants received identical ACP prompts (`/aidlc --status`, `/aidlc --stage code-generation`, `/aidlc`). Human-reviewed verdicts on the captured transcripts:

| Signal | Baseline | Candidate |
| --- | --- | --- |
| Historical deterministic stale signal | 0 / 1 | 1 / 1 |
| Completion Validation Basis receipt | — | 1 / 1 |
| Direct stale detection | — | 1 / 1 |
| Downstream revalidation propagation | — | 1 / 1 |
| Earliest affected stage indication | — | `units-generation` |

Candidate second turn (`/aidlc --stage code-generation`, 12 tool calls, tool-driven receipt):

> "Completed stage result(s) no longer match their validation basis. Directly stale: units-generation. Downstream revalidation required: none. Re-enter the earliest affected stage with /aidlc --stage units-generation."

Candidate third turn (a follow-up `/aidlc` retry, 31 tool calls) — the barrier is persistent:

> "A plain `/aidlc` won't clear this — the fix is specific: re-enter at Units Generation so its output can be reconciled before Code Generation runs again. Please invoke `/aidlc --stage units-generation`."

Baseline second turn (`/aidlc --stage code-generation`, 47 tool calls, model-driven interpretation only):

> "The new requirement was added as a 'post-completion evaluation change,' so it reads as a deliberate scope extension past v1 — but I won't override an affirmed rule on my own."

The baseline transcript never emits `validation basis`, `directly stale`, or a specific re-entry command from an engine tool; the model reasons about a diff it happened to notice. The candidate transcript surfaces `aidlc-validity.ts` output verbatim in both turns.

### Post-hoc scope registration (transparency)

The workflow composed a custom `task-crud-api` scope at runtime but did not persist its definition. `/aidlc` therefore failed with `Unknown scope "task-crud-api"` in both workspaces before any stage-validity logic could run. To unblock resume without altering the workflow's semantics, both workspaces were augmented with:

- `.kiro/scopes/aidlc-task-crud-api.md` — frontmatter + description only
- a `task-crud-api` entry in `.kiro/tools/data/scope-grid.json` whose EXECUTE/SKIP stage grid reflects each workspace's `aidlc-state.md` exactly (baseline 9 EXECUTE / 23 SKIP; candidate 7 EXECUTE / 25 SKIP)

The stage grid was copied from state, not authored freshly, so it is a re-registration of the scope the workflow already ran under — it exposes the completed workflow to the orchestrator's re-entry path; it does not modify what "completion" means or which stages were considered part of the workflow.

### Reproduction artifacts

The rest of the comparison report is duplicated by the sections above (frozen SHAs, transcript quotes, human-reviewed verdicts). The unique details are:

| Item | Baseline | Candidate |
| --- | --- | --- |
| Initial regression test outcome | `install_failed` | `completed` |
| Initial validation receipts | 0 | 58 |
| `.kiro/tools/aidlc-validity.ts` in workspace | absent | present |
| `.kiro/tools/aidlc-artifact-resolution.ts` in workspace | absent | present |
| Mutation `before` sha256 (`unit-of-work.md`) | `ecce128c0e31946f…` | `08866f157b40d189…` |
| Mutation `after` sha256 (`unit-of-work.md`)  | `311659d203704fad…` | `402b41ded0b1fc6d…` |
| Mutation `payload` sha256 (byte-for-byte identical across variants) | `29e28a39cb8d0c2b224bdce9eb5c60abc82ace4b39452d6a65e2a279ec367af1` | (same) |

The candidate's initial CRUD produced a runnable test suite; the baseline's dependency install failed. This is a side effect of the candidate's shared runtime resolver and unit-aware dist installation surfacing during the initial regression check, not the stage-validity feature itself. The `.kiro/tools/*` presence check confirms each workspace ran under its ref's own `dist/kiro/.kiro/` — the sparse-checkout provenance held.

### Scope of this evidence

Covers AI-DLC-owned artifact dependency validity — the feature this PR proposes. It does not claim source-tree, Git snapshot, CI-run, deployment, or external-system validity.

## Follow-up work

This PR intentionally limits validity to AI-DLC artifact dependencies represented by the current graph contract and report-time completion receipts. Follow-up work may include:

- distinguishing semantic dependencies from ordering-only `requires_stage` edges;
- validating that stage frontmatter declares every semantic artifact read by the stage body;
- adding source-tree and Git snapshot provenance for `workspace_requires` stages;
- associating Build and Test validity with commands, repository snapshots, and CI results;
- projecting validity per unit rather than only at stage granularity;
- handling cross-intent and post-merge invalidation.

These are graph-contract and observation extensions; they do not replace the runtime artifact-resolution correctness required by this PR.


djoo-lgcns:
I flagged default test suite: **passed** on the code paths this PR changes.
On this loaded environment, 4 files surface known env-dependent flakes: two timing-sensitive integration tests (t276/24 ~5493 ms, t249/19 ~5005 ms) that pass on upstream/v2 at 4062 ms / 4398 ms respectively but brush the 5 s limit here (~500 ms initialization overhead from the validity module loading); one (t240/9) is a pre-existing upstream/v2 timeout for the same test; two t230 dispatcher tests show test-isolation flake and pass when run in isolation.
None exercise this PR's stage-validity code.

apackeer (changes requested):
Victor, thanks for this PR - it is substantial, carefully argued, and the A/B
evidence in the body is validation work we rarely see from contributors. I
reviewed the full diff against `v2`: core source, the seven generated harness
trees, docs, and tests, with targeted test runs and reproduction probes against
the shipped dist, first at `dbb8986a` and re-verified against the current head
`4fbea2ce`.

### Part A - Direction

**Sound, and worth landing.** The baseline gap is real: the completion guard
checks output existence only at transition time, and `next` then trusts `[x]`
forever, so `done` can be returned over drifted artifacts. Receipts on existing
`STAGE_COMPLETED` rows rather than a second mutable state file, fail-open
migration, and propagation that follows dependencies recorded in completion
receipts rather than every static `consumes` declaration are the right calls.
The shared filename resolver also fixes a real pre-existing bug (consumed
`build-test-results`/`load-test-results` paths were built as `<artifact>.md`
instead of `test-results.md`), and you have kept it current with the new
upstream `traceability` artifact. One credit up front: the artifact-guard
weakness present on the earlier head (an `existsSync` fallback that let a
directory satisfy the guard, and acceptance of the non-physical
`load-test-results.md`) is resolved on the current head -
`producesArtifactsExist` now checks
`isRegularFile(join(dir, artifactFileName(name)))` only, verified by probe.

### Part B - Recommendation: split detection from enforcement

This PR contains two features of different maturity, and I would like to see
them land separately:

- **PR 1, detection (close to ready):** the resolver, schema-2 receipts, and
  the validity inspection, shipped advisory end to end. Advisory has to hold on
  both paths: the `next` barrier becomes a warning, AND receipt-capture
  failures degrade to an untracked completion with a visible warning instead of
  aborting approve/advance (see findings 2 and 3). Because directive
  `narration` is presentation-only by contract
  (`core/tools/aidlc-directive.ts:40-43`), drift should surface as a
  machine-readable field on the relevant directives and in `--status` output,
  not as prose alone - otherwise conductors and scripts can silently ignore
  it. If the statusline consumes validity, it should read a cached or bounded
  summary rather than re-running the full inspection per render: inspection
  re-fingerprints every tracked artifact
  (`core/tools/aidlc-validity.ts:701-708`), and you have already measured
  ~500ms of module-load overhead on a loaded machine. Detection-only
  deliberately keeps routing over known drift; that is acceptable as a staged
  rollout precisely because it is explicit and machine-visible. It also starts
  accumulating receipts in real ledgers immediately, shrinking the fail-open
  migration window and producing field data on drift frequency before anything
  blocks.
- **PR 2, enforcement (own PR, own design discussion):** the hard barrier plus
  its recovery semantics. As shipped here, any post-completion byte change to a
  schema-2-tracked artifact - a typo fix as much as a scope rewrite - blocks
  all routing. The jump branch precedes the barrier, so nobody is trapped, and
  the supported forward recovery is deliberate: a backward jump re-executes the
  target and everything downstream (the only other exit today is reverting the
  artifact to its receipt bytes). Whether that should stay the ONLY response is
  the open design question: a waiver/acknowledge path would need old/new
  hashes, a recorded reason, fresh human authority, and a defined effect on
  every affected downstream receipt (re-capturing just the producer would leave
  completed consumers stale, and a redo-jump resets only the target, so
  downstream receipts recompute stale on the next inspection). That is real
  design work, and a follow-up PR is the right home for it - your
  deterministic-barrier A/B probe carries over intact as its demonstration.

To be fair about the cost: turning the barrier into a warning is subtractive,
but the machine-visible reporting and `--status`/statusline surfacing is new
plumbing. I think it is the right trade, and I am happy to help with the
mechanics. If you would rather keep one PR, blocking gated behind an env flag
with advisory as the default achieves the same staging, though the separate PR
serves the enforcement design conversation better.

### Part C - Findings

1. **P1 - Protect `STAGE_COMPLETED` now that it carries routing authority**
   (`core/tools/aidlc-validity.ts:505-516`, `core/tools/aidlc-audit.ts:328-346`).
   The public `aidlc-audit.ts append` CLI accepts `STAGE_COMPLETED`, and a
   receipt-less completion row deliberately deletes both tracked maps (the
   schema-1 supersede). One diagnostic append therefore converts any stale
   tracked stage to untracked/fail-open and silently disables validity.
   Reproduced against the shipped dist on both heads. `CLI_PROTECTED_EVENT_TYPES`
   exists exactly for rows that guards read as authority; `STAGE_COMPLETED` now
   qualifies (the synthetic single-stage emission path will need the internal
   route).

2. **P2 - Compute the receipt before appending `GATE_APPROVED`**
   (`core/tools/aidlc-state.ts:2570-2576`). The `stageValidationAuditFields`
   spread is evaluated after `GATE_APPROVED` is already in the append-only
   ledger. A capture failure - reproduced with a malformed Bolt DAG while
   resolving per-unit consumed artifacts for `build-and-test` - strands an
   approval with no `STAGE_COMPLETED` and no state write, contrary to the
   atomic-emission contract documented above the try block. Hoist the capture
   into a local before the first append; `handleAdvance` and
   `handleCompleteWorkflow` are already safe because `STAGE_COMPLETED` is their
   first emission.

3. **P2 - A legal optional dangling consume fails closed**
   (`core/tools/aidlc-validity.ts:268-280`, call site `:309`). `uniqueProducer`
   throws on zero producers even for `required: false` consumes, but
   `aidlc-graph.ts:1037-1038` defines that as a first-class valid graph state,
   and plugin contributions make it reachable. The throw blocks approve/advance
   during capture and every `next` during inspection. Skip the absent optional
   edge; more generally, since the receipt is documented as optional, capture
   errors should degrade to an untracked completion plus a warning rather than
   block the completion path.

4. **P2 - The receipt fingerprints report-time bytes, not the basis the stage
   read** (`core/tools/aidlc-validity.ts:306-321`). Inputs are enumerated from
   static `consumes` and hashed when completion is reported. If an input
   changes between the stage reading it and `report`, the receipt binds the
   output to bytes it never consumed and then reports valid; an optional input
   appearing in that window is recorded as consumed when it was not. The PR
   body's "observed input edges ... captured while the stage was reading"
   overstates what ships. Either bind completion to a stage-start/read-time
   snapshot, or narrow the guarantee in the body, docs, and `audit-format.md`.

5. **P2 - The latest merge left two divergent filename maps, and the current
   head fails lint.** Upstream added `artifactFilename` to
   `core/tools/aidlc-lib.ts:3253-3255` (traceability only) while this PR ships
   `artifactFileName` in the resolver (test-result aliases + traceability).
   Consequences on `4fbea2ce`: (a) `bun run check` fails - the merge kept
   upstream's `const filename = artifactFilename(name)` in
   `resolveArtifactPath` (`core/tools/aidlc-orchestrate.ts:1417`) while the
   return arms use the resolver's function, leaving the local unused; (b)
   `producesArtifactFile` (`aidlc-lib.ts:3271-3299`), which feeds the
   gate-revision backstop (`aidlc-state.ts:432,446`) and review-artifact
   matching, resolves `build-test-results` to the non-physical
   `build-test-results.md` - it rejects a real write to `test-results.md` and
   would match the filename that never exists (probe-verified). Please
   consolidate the filename mapping into one leaf module imported by both
   `aidlc-lib` and the resolver; the same consolidation is a good home for
   `KNOWN_CODEKB_STAGES`, which the resolver currently re-declares
   (`aidlc-artifact-resolution.ts:26`) although `aidlc-lib.ts` exports the
   canonical set.

6. **P2 - Rename the new test**
   (`tests/unit/t263-stage-validity-propagation.test.ts`). `v2` already has
   `t263-reviewer-terminal-ordering.test.ts`. t277-t282 have since been taken
   or claimed by other in-flight work; t283 is unclaimed as of this review.

7. **P2 - Version slot collision after the latest merge.** The branch's 2.5.75
   bump collides with the 2.5.75 that #404 has since merged to `v2` (base is
   now 2.6.2). Please re-merge upstream and take the next free patch slot above
   the then-current base, renaming the CHANGELOG heading (t68 catches the
   duplicate otherwise). I would not hard-code a number here: today's next slot
   is already claimed by other in-flight work. The PR body also still says
   2.5.63.

Non-blocking note: `README.md:365` says 38 `aidlc-*.ts` engine tools; the
current head's tree has 39 after the upstream merge. CI has not yet run on the
fork branch and will need a maintainer approval.

### Part D - Validation

On `4fbea2ce`: `bun scripts/package.ts --check` passes (all seven harness trees
in sync) and `bun run typecheck` passes; `bun run check` fails per finding 5.
On `dbb8986a`: targeted runs were green (the new validity tests, 21 assertions;
artifact guard; audit authority floor; version-changelog sync). Findings 1-3
were each reproduced against the shipped dist and re-verified on `4fbea2ce`;
each fix should land with a regression test.

djoo-lgcns:
Thanks a lot for the detailed review — I really appreciate the time you put into going through the implementation, tests, and the A/B validation.
Your comments were especially helpful in clarifying the boundary between validity detection and enforcement, and they also surfaced several edge cases that were worth fixing properly.


Addressed the requested changes in `63825674` and refreshed the branch onto current `v2` at `1d331c04`.
- Split detection from enforcement: `next` keeps its normal directive kind and carries an optional machine-readable `stage_validity` advisory. `--status` performs the detailed inspection; the continuously rendered statusline remains cheap.
- Protected `STAGE_COMPLETED` from the public audit CLI. The synthetic `report --single` lifecycle pair now uses the internal owning-emitter path and remains atomic.
- Compute validation capture before `GATE_APPROVED`. Capture/inspection failures fail open as receipt-less, untracked completions with a persistent `Validation Warning`; routing continues.
- Skip legal optional consumes with zero producers while retaining strict errors for required-missing and ambiguous producers.
- Narrowed the docs and PR description to report-time receipt bytes rather than runtime read observation.
- Consolidated filename aliases and CodeKB stage ownership in the dependency-free `aidlc-artifact-vocabulary.ts` leaf. Build/load test results now resolve consistently to `test-results.md`.
- Renamed the test to `t299-stage-validity-propagation.test.ts`, merged current upstream `2.6.15`, and moved this PR to `2.6.16`. README now reflects the actual 42 `aidlc-*.ts` modules after the upstream knowledge additions plus this PR.

Verification on the final merged head:

- focused validity/directive/single-stage suites: 199 assertions, pass
- authority/lifecycle suites: 148 assertions, pass
- deterministic integration: 103 files / 1,414 assertions, pass
- typecheck, lint, package parity, coverage registry, and version/changelog sync: pass

The live Bedrock preflight was not rerun because I currently lack the local AWS credentials; the deterministic integration profile was run with live-model gates disabled.


I also agree that a separate PR is the better place for the enforcement design conversation. Once this detection/advisory piece is settled, I’d be glad to follow up there and work through the remaining enforcement and recovery semantics. I also appreciate your offer to help with the mechanics as we move that discussion forward.

apackeer:
Thanks @djoo-lgcns for sticking with this through the whole review cycle. The rework landed exactly where the discussion pointed: detection stays deterministic in the engine, the advisory rides the normal directive without touching routing, and legacy workflows fail open with no migration step. The A/B evidence you put together made the case for the detector better than any design doc would have.

From here the maintainers will take over the merge mechanics so you do not have to keep chasing the release train: the rebase onto current v2, the version and test-number re-slot (2.6.16 and t299 were both claimed by PRs that merged while this was in review), and a couple of small follow-ups from our second pass, the main one being that untracked-only histories will report through `/aidlc --status` rather than attaching a warning to every `next`, so pre-upgrade workflows are not nagged each turn. Maintainer edits are enabled on the PR, so these land as commits on your branch and the feature merges under your authorship.

The enforcement and recovery semantics conversation stays open as the follow-up you proposed, and we would be glad to have you drive it; your deterministic barrier probe carries over intact as its demonstration. Thanks again for a genuinely well-built contribution.

djoo-lgcns:
Thanks @apackeer — I have opened #878 for the enforcement and recovery design discussion. It keeps #716 as the fixed detection substrate, proposes a minimal known-drift barrier with explicit backward revalidation, leaves waiver out of the first implementation, and calls out five policy decisions for alignment before code begins. I will leave this PR branch to the maintainer-owned rebase, release/test re-slot, and untracked-only status follow-ups you described.

apackeer (approved):
Approving. Every finding from the 2026-08-17 review has been addressed and verified on the current head:

- Detection is split from enforcement: `next` keeps its directive kind and carries an optional machine-readable `stage_validity` advisory (also across steering continuations); untracked-only histories report via `/aidlc --status` only, so upgraded mid-flight workflows are not nagged per turn.
- `STAGE_COMPLETED` is CLI-protected; receipt capture is fail-open and ordered before `GATE_APPROVED`; the filename vocabulary is unified in one leaf module.
- The remediation and merge mechanics landed as maintainer commits on this branch per the takeover note, with the feature merging under the author's credit.

Verification: full sliced pre-merge gate on this head (smoke+unit 243 files, integration 107 files, deterministic e2e, plus live Claude TUI, Kiro ACP, Kiro TUI, and Codex exec slices, each red retested green-alone or attributed to pre-existing base issues), and the PR CI is green 4/4. One known limitation is documented in the state-machine reference (scope-blind resolution for express stage-level artifacts) with a follow-up planned, and the enforcement/recovery design conversation continues in the follow-up PR the author proposed.

PR #721, opened 2026-08-06
feat: add interactive harness installer (scripts/install.ts)

osataken:
Interactive CLI to install a harness into a project. Supports harness picker, version detection, --all for bulk update, kiro/kiro-ide migration, aidlc/ no-clobber merge, .gitignore backup, and input validation. 15 integration tests.

# Summary

Adds `scripts/install.ts` — an interactive harness installer that replaces the manual `cp -r dist/<harness>/ your-project/` workflow with a guided CLI experience. Users can now install, update, or migrate harnesses with a single command.

## Changes

- **New file: `scripts/install.ts`** (823 lines) — the installer with:
  - Interactive harness picker (numbered options, accepts name or number)
  - Target directory prompt with tilde expansion (`~/my-project`)
  - `.aidlc-install.json` metadata file written on install (tracks harness + version)
  - Version detection: reads metadata file, falls back to `aidlc-version.ts` regex for legacy installs
  - Kiro/kiro-ide disambiguation via metadata or `.kiro.hook` heuristic
  - Kiro ↔ kiro-ide migration (replaces `.kiro/`, preserves `aidlc/`)
  - `--all` flag: updates every outdated harness in one command, exits clean when all current
  - `aidlc/` workspace shell: always uses no-clobber merge (user state, intents, audit, memory preserved)
  - `.gitignore`: backs up existing as `.gitignore.bak`, installs new, shows merge note
  - Unified file plan showing `new`/`overwrite`/`merge`/`backup + replace` per entry
  - Input validation: harness names (alphanumeric + hyphens), target paths (no null bytes, control chars, length cap)
  - `--doctor` post-install validation
  - `NO_COLOR` support, ANSI styling matching project conventions (✓/✗, `─` separators, bold/dim/cyan)
  - "What's next" guidance with harness-specific invoke command
  - Codex trust-seeding note
  - `--all` + `--harness` mutual exclusion with clear error
- **New file: `tests/integration/t239-install.test.ts`** — 15 integration tests covering all harnesses, skip/force behavior, version detection, `--all`, error cases
- **Modified: `AGENTS.md`** — added `scripts/install.ts` bullet to project structure
- **Modified: `README.md`** — added installer to Quick Start section + repo layout diagram
- **Modified: `docs/guide/01-getting-started.md`** — added "Option A — Interactive installer" before manual copy steps

## Design Decisions

1. **`scripts/install.ts` (separate file) over extending `package.ts`** — The installer targets end-users; `package.ts` targets maintainers. Different audiences, different concerns. The `codex trust` subcommand was precedent for CLI extensions in package.ts, but install logic is complex enough to warrant separation.

2. **No plugin installation** — The plugin mechanism exists and works, but has no user-facing documentation yet. Offering plugin selection in the installer without a guide on what plugins are would confuse users. Plugin support can be re-added when `docs/guide/XX-plugins.md` ships.

3. **`aidlc/` always uses no-clobber merge** — Even with `--force`. This directory contains user-generated state (intents, audit shards, workflow state) and user-edited memory (`team.md`, `project.md`). Overwriting it destroys work. New framework files from newer versions still get added; existing files are never touched.

4. **`.gitignore` backup + replace (not append)** — The shipped `.gitignore` contains both generic patterns and AI-DLC-specific ones. Appending just the AI-DLC block was considered but the generic section (node_modules, editor files) might conflict with the user's existing structure. Backup + replace with a merge note keeps it simple and reversible.

5. **`.aidlc-install.json` metadata file** — Solves kiro/kiro-ide disambiguation (both use `.kiro/`), enables version detection without regex-parsing source files, and future-proofs for upgrade logic. Falls back to `aidlc-version.ts` regex + `.kiro.hook` heuristic for legacy (pre-installer) copies.

6. **`--all` only updates outdated harnesses** — Not a "reinstall everything" nuke. If all are current, exits with "Nothing to update." This prevents unnecessary file churn in CI.

7. **`--force` does not imply `--all`** — `--force` requires `--harness` (or interactive selection). This keeps intent explicit: "overwrite this one harness" vs "update everything that's stale."

8. **Kiro ↔ kiro-ide migration deletes `.kiro/`** — Both harnesses own the same directory. You can't merge one into the other meaningfully (hook formats differ, agent configs differ). Clean replacement is the only safe path. `aidlc/` is preserved since it's a sibling.

9. **`gen-coverage-registry.test.ts` addition** — The test registry auto-classifies test files by mechanism (body-derived). Our `t239-install.test.ts` spawns `bun` (classified as "cli" mechanism) while declaring a `covers:` target — placing it in the `EXPECTED_NONE_TO_CLI` allowlist. This is the registry acknowledging the classification, not a placement decision about where the test file belongs.

10. **No `--update` flag** — Today it would be identical to `--force` (blind overwrite). The flag name implies smart behavior (preserve user files, show changelog, run migrations) that doesn't exist yet. `--force` is honest about what it does. `--update` can arrive when the logic justifies it.

## User experience

**Before:** Users manually copy 3–6 items from `dist/<harness>/` into their project using `cp -r` commands from the README. No version awareness, no conflict detection, no guidance on what to do next.

**After:**
```bash
$ bun scripts/install.ts

AI-DLC Installer v2.5.41
─────────────────────────────────────
Set up the AI-DLC workflow engine in your project.

  Press Enter to use the current directory.
  Target project directory? (.) ~/my-project

  Existing: kiro (v2.4.0)

What would you like to do?

  1. Update kiro to v2.5.41
  2. Install a different harness
  3. Both — update existing + install new

  Reply with a number or name: 1

Installing kiro → ~/my-project
  ↑ kiro: v2.4.0 → v2.5.41
  ✓ aidlc (merged, user data preserved)
  ✓ .gitignore (old saved as .gitignore.bak)
  ✓ .kiro
  ✓ AGENTS.md
─────────────────────────────────────
✓ Installation complete — 4 item(s) copied.

Note: Your original .gitignore was saved as .gitignore.bak.
  Merge your custom entries back into the new .gitignore.

What's next:
  /aidlc --doctor     verify the install
  /aidlc <description> start a workflow
```

Non-interactive mode for CI:
```bash
bun scripts/install.ts --harness kiro --target ./project --force
bun scripts/install.ts --all --target ./project
```

## Checklist

* [x] Changes are documented

## Test Plan

1. Run `bun test tests/integration/t239-install.test.ts` — all 15 tests should pass
2. Fresh install: `bun scripts/install.ts --harness kiro --target /tmp/test --force --doctor` — should copy files and doctor should pass (39/39)
3. Re-install same version: run the above again — should show "already at v2.5.41" and still copy
4. Version upgrade: fake an older version in `.kiro/.aidlc-install.json`, re-run — should show the upgrade arrow
5. `--all` with outdated: install two harnesses, fake old versions, run `--all` — both should update
6. `--all` when current: run `--all` again — should say "Nothing to update"
7. Kiro → kiro-ide migration: install kiro, then install kiro-ide with `--force` — should show "Migrating" and replace `.kiro/`
8. `.gitignore` backup: create a `.gitignore` in target, install — should create `.gitignore.bak`
9. `aidlc/` preservation: create files in `aidlc/spaces/default/state/`, install with `--force` — user files should survive
10. Invalid input: `--harness "../evil"` should error, `--all --harness kiro` should error

## Acknowledgment


apackeer:
osataken, thanks for putting this together and for documenting the implementation lessons on #722. This PR explores a repository-local Bun installer for copying and refreshing the generated harness trees, including version detection and Kiro/Kiro IDE migration.

We are moving the installation and upgrade experience in the direction described in Distribution RFC #722: a released, checksum-verified `aidlc` binary, transactional `aidlc init` refreshes, ownership-aware merges, version diagnostics, and upgrade/rollback support.

Would you mind taking another look at #722 and aligning this work with that delivery path? Your feedback there is useful, but we would prefer not to introduce a separate interim installer contract that users and CI would later need to migrate away from.

PR #726, opened 2026-08-08
fix(engine): a completed workflow named no path to new work

comdaze:
## The problem

A completed intent is a finished record, never a container to re-enter, so the only move onward is to birth a new intent. That move — `next --new-intent` → `intent-birth` — is named in **no directive, no hook message, and no help text**.

Meanwhile `aidlc-session-start.ts` asks the conductor to offer *"the standard resume options (Resume / Redo / Jump / Start Fresh)"*, and that menu is defined nowhere:

- **`Start Fresh`** occurs exactly once in the whole tree — in the line asking for it.
- **`Redo`** *is* defined elsewhere, as the intra-stage Keep/Modify/Redo gate loop in `stage-protocol.md`. That is a different mechanism, so a conductor that goes looking finds a definition that actively misleads it rather than none.

Every scope runner forwards freeform text into `next` and ends its loop on `{kind:"done"}` (`aidlc-runner-gen.ts`), so the observable failure is:

```
$ /aidlc --scope feature "add the thing"        # after a previous workflow closed out
Workflow complete — no in-scope stage remains after feedback-optimization (scope: feature).
```

…and the conductor stops. The user asked to start; the engine answered with a fact and no next step. `intent-birth` has supported this all along — its own source comment says *"Births the FIRST intent into the active space on a fresh workspace, OR a new intent for new work alongside an active one"* — only the user-visible surfaces were silent.

Found while driving the `vibe` scope, where one-session-per-container is by design and so the closed case is the *normal* one from the second session onward. But the defect is generic; `/feature <request>` dead-ends identically.

## The change

Three surfaces, one property: **a completed workflow must name the move.**

| File | Change |
| --- | --- |
| `core/tools/aidlc-orchestrate.ts` | the `done` branch names `--new-intent` in `reason` and says so in `narration`; a stronger form fires when freeform text was supplied, because that caller is demonstrably not just checking status |
| `core/hooks/aidlc-session-start.ts` | the resume line is keyed on the `Status` it already read — nothing to resume on a completed workflow, so it names the new-intent command instead |
| `core/tools/aidlc-utility.ts` | `--help` lists `intent birth` |

For the hook I **removed the dangling half rather than defining it in a second place**. Adding a "standard resume options" section elsewhere would leave two files describing one thing, and the next drift would be between them.

No command, flag, or output-format changes. Behaviour is unchanged beyond the message text.

## Tests

`tests/unit/t275-completed-workflow-new-intent-path.test.ts` — five assertions across the three surfaces plus the two dangling names.

It is deliberately explicit about **which assertion catches the class of bug and which catches only the literal string**: guard 4 (every `--flag` the guidance names must exist in the engine's parser) catches "the prose names a command that does not exist", which is the general shape; guard 3 catches the two literal names. Guard 4 would *not* have caught `Start Fresh`, because that was never a flag. Stated in the file, because a guard whose real coverage is unstated gets read as broader than it is.

**Verified non-vacuous:** restoring only the generated `dist/` copies to the pre-fix bytes and re-running reds all five; repackaging greens them again.

## Verification

- `bun scripts/package.ts --check` — clean
- `bun run typecheck` — clean
- `bunx biome check --error-on-warnings core harness scripts plugins tests` — three findings, all pre-existing in `tests/unit/t267-usage.test.ts`; **zero** in any file this PR touches (left alone deliberately — unrelated scope)
- `bun tests/gen-coverage-registry.ts --check` — clean (`t275` added to `EXPECTED_NONE_TO_CLI`, it spawns)
- `bash tests/run-tests.sh --smoke --unit` — 206 files, 3 failing, **none of them this change**:
  - `t248-codekb-scope-diff` + `t255-workspace-sync` — reproduce identically on an unmodified `v2` worktree (22 assertions)
  - `t150-codex-packaging` — `this test timed out after 5000ms`, no assertion diff, passes standalone

## Version bump

Bumps to **2.5.60** with the README badge and a matching `CHANGELOG.md` entry, per the `CONTRIBUTING.md` checklist and `t68`.

At ~1.6 releases/day this patch number will very likely be taken before review completes — **I will rebase and re-bump on request.**


apackeer (changes requested):
Thanks @comdaze. For context: this PR tackles the completed-workflow dead end - once an intent finishes, nothing in the directives, hook messages, or help text names the path to starting new work, and session-start still offers resume options that do not fit a completed intent. The fix adds a continuation path in shared core (engine directives + session-start wording + help text), with regenerated dists and a new test (t275).

Directionally, this is sound: the completed-workflow dead end exists on v2, and shared core is the correct layer. The generated dist changes are also in sync.

I found four blocking implementation issues:

1. The new continuation is carried by a `done` directive, but every harness and generated scope runner treats `done` as terminal: present the summary and stop. The original dead end therefore remains. The supplied-new-work case needs an actionable `ask` or `print` path.

2. Help advertises `intent birth` and `/aidlc next --new-intent`, but `intent birth` is a retired reserved verb that the parser rejects, and `next` is not a public `/aidlc` verb. Use `intent create` for the public form and describe the internal result accurately as a `print` naming `intent-create`.

3. The session-start wording removes Redo and says completed intents cannot be re-entered. Existing resume, redo, and backward-jump paths remain implemented, and the docs explicitly say the prior intent remains resumable.

4. The completed-workflow command interpolates freeform text without shell quoting. Requests containing quotes or shell syntax produce malformed/actionable command text; use the established command renderer.

The targeted suite and packaging checks pass, but t275 currently pins the retired vocabulary and terminal behavior rather than catching these conflicts.

UX-wise, this currently exposes an unusable command, removes a valid resume option, and offers a continuation that the directive protocol requires the conductor to stop before acting on.

apackeer:
Directionally, this is sound: the completed-workflow dead end exists on v2, and shared core is the correct layer. The generated dist changes are also in sync.

I found four blocking implementation issues:

1. The new continuation is carried by a `done` directive, but every harness and generated scope runner treats `done` as terminal: present the summary and stop. The original dead end therefore remains. The supplied-new-work case needs an actionable `ask` or `print` path.

2. Help advertises `intent birth` and `/aidlc next --new-intent`, but `intent birth` is a retired reserved verb that the parser rejects, and `next` is not a public `/aidlc` verb. Use `intent create` for the public form and describe the internal result accurately as a `print` naming `intent-create`.

3. The session-start wording removes Redo and says completed intents cannot be re-entered. Existing resume, redo, and backward-jump paths remain implemented, and the docs explicitly say the prior intent remains resumable.

4. The completed-workflow command interpolates freeform text without shell quoting. Requests containing quotes or shell syntax produce malformed/actionable command text; use the established command renderer.

The targeted suite and packaging checks pass, but t275 currently pins the retired vocabulary and terminal behavior rather than catching these conflicts.

UX-wise, this currently exposes an unusable command, removes a valid resume option, and offers a continuation that the directive protocol requires the conductor to stop before acting on.

comdaze:
Closing this — all four findings are valid, and they share one root cause that makes a rework the wrong response.

**My premise was false.** I claimed the path to new work after a close-out "appears in no directive, no hook message, and no help text." It appears in the engine, and has all along:

- `aidlc-orchestrate.ts` — `--resume` emits `askDirective("An existing workflow was found… Resume from last checkpoint, redo the current stage, jump to a stage, or start fresh.")`
- the `choice.includes("fresh")` branch emits `printDirective("Start-fresh accepted. … run \`next --new-intent --scope <scope> \"<description>\"\` — the existing workflow stays in place and the new intent starts alongside it.")`

That is exactly the mechanism this PR was written to add. The session-start hook's line naming "Resume / Redo / Jump / Start Fresh" is not a dangling reference — those four labels are the four implemented branches, and the conductor's job is to run `next --resume` so the engine renders the menu.

**How I got it wrong:** I grepped for the display label `"Start Fresh"`, found it in exactly one place (the hook line asking for it), and concluded the referent did not exist. The code spells the choices lowercase inside `choice.includes("fresh")`. I searched for the name and declared the thing absent — which is the same class of mistake as the one I thought I was fixing, and my own new test's fourth assertion ("every `--flag` the guidance names must exist in the engine") was supposed to catch precisely this shape.

So each of your findings traces back to that:

1. **`done` is terminal, so the fix fixes nothing** — correct, and the new-work path does not need to live there at all, because `--resume` → `resumed` → per-choice `print` already routes it.
2. **`intent birth` is retired; `next` is not a public verb** — correct. I carried vocabulary from a fork at 2.5.33 into a PR against 2.5.59 without checking the verb still existed. `intent create` is the right form.
3. **Redo and backward jump remain implemented** — correct, and worse than a wording problem: I removed a working option from a menu whose branches are all live.
4. **Unquoted freeform text in the command** — correct. `createPrintDirective` in the same file already uses `JSON.stringify()` for exactly this, with a comment saying why. No excuse.

The real failure in the session that prompted this was mine as a conductor, not the framework's: on a completed workflow I ran a bare `next`, got `done`, and stopped — instead of offering the resume menu the session-start hook had just told me to offer.

**One residual gap looks real and I would rather raise it separately than reshape this PR:** `--help` lists `intent list` and `intent switch` but not the creation verb, so `intent create` is undiscoverable from the help text even though it is the public form. Happy to send that as a one-line PR if you think it is worth having; say the word and I will, otherwise I will leave it.

Thank you for the review — it was more useful than a merge would have been.

PR #729, opened 2026-08-09
fix(continue-workflow): lead the load-steering block with the token, not the payload

comdaze:
## The defect

`continuationReason()` in the continue-workflow hook emits the `rules_content` payload **before** the `continue "<token>"` command. Measured on `v2` at 2.5.59, with the existing t121 fixture:

```
token at offset 332, payload at offset 160
```

A rule bundle runs to many KB — 16,583 chars across 37 entries on a stock install — and a harness may truncate hook output. The token is therefore the part that gets cut.

## Why that is a stuck loop rather than a lost message

Without the token there is nothing to act on. The reader sees a wall of rules, notes correctly that they are already on disk and ambient, and does nothing. The engine re-emits `load-steering` on the next turn, and again on the one after.

Observed live as **seven identical deliveries with no progress**, each one paying the full payload in context.

And re-running `next` is not a recovery: the chain's position lives in the token, so `next` returns the head of the chain with the same token every time. Draining a stuck chain meant reading the token out of the directive JSON by hand.

## The fix is ordering, not size

Truncated rule text is recoverable — the method files are on disk and reach the model through each harness's always-on include regardless. **A truncated token is not.** So the actionable instruction leads and the bulk context follows.

Every clause keeps its existing wording; only the order changes. A comment records why, so a future tidy-up does not quietly restore the old order.

## Test

`t121`'s existing `(a) load-steering reason carries exact content and continues the exact token` case gains an ordering assertion:

```ts
expect(tokenAt).toBeLessThan(payloadAt);
```

**Verified failing on unmodified `v2` before the source change** (the offsets above), passing after. Its existing `toContain` assertions on both clauses are untouched and still pass, which is what pins that this is a reorder and not a rewrite.

## Verification

- `bun scripts/package.ts --check` — clean
- `bun run typecheck` — clean
- `bun test tests/unit/t68-version-changelog-sync.test.ts` — 7/7 (version, badge, CHANGELOG agree at 2.5.60)
- `bunx biome check --error-on-warnings core harness scripts plugins tests` — three pre-existing infos in `tests/unit/t267-usage.test.ts`, none in the files touched here (left alone deliberately — unrelated scope)
- `bash tests/run-tests.sh --smoke --unit` — 205 files, two failing: `t248-codekb-scope-diff` and `t255-workspace-sync`, both reproducing identically on an unmodified `v2` worktree

## Version

Bumps to 2.5.60 with the README badge and a matching `CHANGELOG.md` entry, per the `CONTRIBUTING.md` checklist and `t68`. At the current release cadence this patch number will probably be taken before review completes — **I will rebase and re-bump on request.**

---

Context, since you reviewed #726: that one I closed rather than reworked — its premise was false, the resume menu it claimed was missing is fully implemented in the engine, and all four of your findings traced to that. This PR is a different shape on purpose: one behaviour, one file, an assertion verified red before the fix, and no claim that anything is absent.


apackeer (changes requested):
Sean, I reviewed this change to the continue-workflow Stop-hook recovery for `load-steering`. It moves the opaque `continue` command ahead of the serialized rule payload so host-side truncation is less likely to strand the chain. The change is implemented in the shared core hook, projected across every harness distribution, and covered by a new `t121` ordering assertion plus the 2.5.60 release metadata.

Directionally, this is sound. On `origin/v2`, the hook puts the large rule payload before the only actionable continuation token, and a fresh `next` restarts delivery at part one rather than recovering the lost position. The shared core hook is the right layer, the scope is proportionate, and `bun scripts/package.ts --check` confirms the `dist/` edits are generated rather than hand-maintained.

I am requesting changes for one blocking semantic regression:

1. `core/hooks/aidlc-continue-workflow.ts:943` now tells the conductor to `Run ... continue` before it tells it to apply the current `rules_content`, changing the old `apply ... before continuing` / `Then run` sequence to `Run` / `as you go`. The steering contract requires the current chunk to be applied and retained before its opaque token is invoked (`core/aidlc-common/protocols/stage-protocol.md:743`, `core/tools/aidlc-directive.ts:75`). If the new imperative order is followed literally, the chain advances past the current chunk without retaining it. Please keep the literal command/token ahead of the payload, but phrase it as the command to run only after applying the payload below. The `t121` assertion at `tests/integration/t121-stop-hook-enforce.test.ts:789` should also pin that semantic order instead of checking only character offsets.
2. `CHANGELOG.md:6` says the wording is unchanged, but this patch changes `before continuing` / `Then run` to `Run` / `as you go`. Please make the release note match the corrected message.

Focused verification passed at `tests/logs/2026-08-09T04-03-55Z` with `RESULT: PASS`. The version sync test passed, all five generated harness trees are in sync, typecheck and the focused Biome check passed, and no new test number was introduced. The passing test does not clear the finding because it checks token position, not the required apply-before-continue instruction.

UX impact is otherwise narrow: there are no new or renamed commands, flags, stages, defaults, or structured output fields. The only runtime surface is the agent-visible Stop-hook nudge; putting the token earlier is consistent with the intended recovery, while telling the agent to run it before applying the current payload is not. The documented upgrade remains to re-copy the relevant `dist/<harness>/` shell, with no CI or script break.

comdaze:
Thanks for the review — the blocking finding is correct, and it is the more
important of the two defects in play here. Moving the token ahead of the payload
also inverted the imperative, and `apply-then-continue` is the contract
(`stage-protocol.md`'s inline sequence, and `aidlc-directive.ts`'s own comment on
`LoadSteeringDirective`: "applies rules_content in order and immediately invokes
continue"). Read literally, my version told the conductor to advance the chain
past a chunk it had not retained — a quieter failure than the truncation I set out
to fix, because it looks like progress.

## The original defect is still present on current `v2`

Re-verified today against `cc08c161` (2.6.65), unmodified, by planting the
ordering assertion:

```
expect(tokenAt).toBeLessThan(payloadAt)
Expected: < 160
Received: 332
```

Byte-identical offsets to the 2.5.59 measurement in the PR body — 332 and 160.
`core/hooks/aidlc-continue-workflow.ts` has been touched 5 times across the 56
commits since (Copilot harness, Resume waits, Stop progress, `--resume`), but the
`load-steering` branch itself has not moved.

I also measured the payload it has to survive, which the original PR body did not
quantify per-scope. In a clean project whose only intent is `feature`, the first
`next` emits `load-steering` with a continuation token, and with a 267-line memory
layer (`org.md` + `team.md` + `project.md`) the bundle is **21 heading-sized
segments — a 17,145-byte payload inside an 18,017-byte directive, split across 2
parts** — with the **683-byte token after it**. Because the bundle is the
substantive text of the method files, it **grows as a team sediments practices**:
the token gets pushed further out the longer a project is used, so this is not a
fixed-size risk.

Two details worth having in the record:

- **Recovery is not available to the reader.** The chain position lives only in the
  token, so a truncated delivery cannot be repaired with `next` — it returns part 1
  with the same token. The observable behaviour is the same payload re-delivered
  every turn, which is how it was first noticed (seven identical deliveries).
- **A stage that resolves an empty bundle is exempt**, which may be a useful probe
  for you: with `chunks.length === 0` the engine returns `run-stage` directly and
  never enters this branch. I hit that with a plugin stage of my own that declares
  no rules — not one of the built-in scopes — and the memory layer still reached the
  model through the harness's always-on include. So the exposure is every stage that
  resolves a non-empty bundle, which is all the core ones.

## Proposed wording

The two orders are deliberately opposite, and the third sentence is what keeps
them from being read as one:

```ts
return (
  `The AIDLC workflow still has rules to load${where}. Two steps, in this order: ` +
  "**first** apply every path/text entry in the exact `rules_content` payload printed " +
  "below, **then** run " +
  `\`bun ${harnessDir()}/tools/aidlc-orchestrate.ts continue "${continueToken}"\` ` +
  "and keep following each load-steering step it returns until it answers `run-stage`. " +
  "The command is printed above the payload only so it survives output truncation; " +
  "it is still the second thing to do, and applying the payload first is what makes " +
  "the chain advance over rules you have actually taken in. " +
  "Do not summarise or narrate these rule chunks to the user.\n\n" +
  `${exactContent}`
);
```

Without that "printed above … only so it survives truncation" clause, a reader
resolves the print-order/execution-order conflict by assuming they are the same —
which is exactly how the regression got in. The `**` emphasis is not load-bearing;
drop it if it does not match house style, as long as the first/second relationship
stays explicit.

## On pinning the semantic order in `t121`

Agreed, and I would keep it as **two** assertions rather than one, because they
fail independently:

```ts
// (i) TEXTUAL — the token survives truncation.
const tokenAt   = reasonText.indexOf('continue "steering-token-495"');
const payloadAt = reasonText.indexOf('"path":"aidlc/spaces/default/memory/org.md"');
expect(tokenAt).toBeLessThan(payloadAt);

// (ii) IMPERATIVE — apply-then-continue, the contract order.
const firstAt = reasonText.indexOf("**first** apply");
const thenAt  = reasonText.indexOf("**then** run");
expect(firstAt).toBeGreaterThanOrEqual(0);
expect(thenAt).toBeGreaterThan(firstAt);
expect(reasonText).toContain("only so it survives output truncation");
expect(reasonText).not.toContain("as you go");   // the phrase that inverted it
```

Your point about the old assertion generalises, and it is the part worth keeping in
the comment: **an offset comparison cannot see an inverted imperative.** Check (i)
passed on my regressed build, which is why the suite read as clean. I verified the
new assertion has teeth the other way too — reverting the wording turns it red and
names the missing clause.

One caveat if you port this as-is: (ii) matches the literal `**first**` / `**then**`
markers, so it pins the *wording*, not the semantics in the abstract — including the
emphasis I just said was optional. Those two things have to be decided together: if
you drop or reword the markers, the assertion moves with them. I could not find a
check for "the imperative is apply-first" that does not bottom out in matching some
phrase, so the honest framing is that (ii) guards this specific sentence against
being re-inverted, rather than proving the property in general.

## Second finding (`CHANGELOG.md`)

Correct — the note claimed the wording was unchanged when it had changed. A
re-submission should carry a release note describing the corrected message, not
just the reordering.

## How would you like this landed?

You asked me not to spend time on a mechanical rebase, and after 56 commits the
original diff is not portable anyway. So this comment carries the agreed wording
for a maintainer to port. If you would rather have a fresh PR against current `v2`
— hook + `t121` + release note, three files — say so and I will open one.

apackeer:
Thank you, Sean. The maintainers will take it over from here and carry the fix through to merge.

apackeer (approved):
Thanks, Sean, for isolating the truncation failure and working through the action-order issue.

I reviewed the current head `7bec324e709d7075d560fcb317cfe8794d3019a4` against `4f944c00721907b186f2e5b1c8a95ae448214a52`.

The direction was established by the earlier maintainer review. The maintainer takeover preserves the continuation token before the bulk payload while explicitly requiring each rule chunk to be applied before invoking `continue`. It adds independent regression coverage for transport and execution order, updates release metadata to `2.6.79`, and regenerates all seven harnesses.

Focused `t121` and version-sync tests passed. Package parity, all TypeScript configurations, and Biome also passed. GitHub CI, deterministic integration/e2e, smoke/unit, Markdownlint, Gitleaks, Semgrep, and dependency audit are all green on this exact head. No live harness gates were run.

APPROVE.

PR #744, opened 2026-08-12
feat: add automatic Plannotator workflow gates

cfsegoviaz:
# Summary

This PR adds automatic, digest-bound Plannotator gates to AI-DLC question and approval boundaries in Kiro CLI workflows.

Previously, human decisions depended on chat/manual coordination or newest-file heuristics. That made it possible to lose reviewer feedback, bind a decision to the wrong artifact, or continue without independently checking every pending lifecycle artifact. The new flow opens the correct Plannotator interaction for an explicit `aidlc-docs` path, validates the typed result, and resumes only when the current artifact digest satisfies the gate contract.

## Changes

- Adds the workspace-scoped `review_aidlc_gate` MCP server and Kiro configuration, plus installation and provenance-verification guidance.
- Introduces typed interaction, completion, question parsing, provenance, MCP gate, and MCP server modules shared by normal workspace mode and the evaluator supervisor.
- Restricts review targets to canonical Markdown under `aidlc-docs`, with workspace/symlink confinement, bounded inputs and outputs, nonce/digest binding, concurrency protection, and single-consumption ledger state.
- Maps AI-DLC questions to Plannotator interview controls, applies answers atomically, uses a single custom `Other` path, and rejects unknown or malformed nested response fields.
- Returns requested-change feedback through a bounded Base64 envelope with size and SHA-256 metadata while excluding raw feedback and absolute paths from persisted audit/run metadata.
- Reworks the Kiro supervisor to drain every pending question document and gate every unapproved non-question artifact digest instead of selecting only the newest file.
- Propagates completed/incomplete/failed run status through the CLI evaluator and documents normal MCP mode versus strict external supervision.
- Adds focused unit, property, security, orchestration, and server tests plus representative screenshots.

## User experience

Before this change, users had to coordinate question completion and approval through chat or manual file editing, and the automation could not prove that the decision matched the exact artifact version.

After this change:

1. AI-DLC writes and validates a canonical question or review artifact.
2. Kiro calls `review_aidlc_gate` with the exact workspace-relative path and interaction type.
3. Plannotator opens interview controls for questions or document annotation mode for approvals.
4. The MCP validates provenance, path, digest, typed outcome, and postconditions.
5. AI-DLC resumes only for `answers_submitted` or `approved`; requested changes keep the current stage open and all failure/unknown outcomes fail closed.

**Automatic question interview**

Automatic AI-DLC question interview

**Document review and annotation**

AI-DLC document review and annotation

## Checklist

* [x] Changes are documented

## Test Plan

- [x] Run `uv run pytest packages/cli-harness/tests` from `scripts/aidlc-evaluator` — 69 passed.
- [x] Run Ruff check and format verification against all 22 changed/new Python files — passed.
- [x] Run `npx markdownlint-cli2` against all seven changed/new Markdown documents — 0 issues.
- [x] Parse `.kiro/settings/mcp.json` with `python3 -m json.tool` — passed.
- [x] Run `git diff --check` — passed.
- [x] Perform three semantic pre-PR reviews; final review reported no blocking findings.
- [ ] Repository-wide monolithic `uv run pytest` collection is not a valid verification command today because pre-existing packages contain duplicate test module names; the complete affected `cli-harness` suite is covered above.

Reviewer smoke test:

```bash
uv sync --project scripts/aidlc-evaluator --package aidlc-cli-harness
uv run --project scripts/aidlc-evaluator --package aidlc-cli-harness aidlc-plannotator-mcp --help
kiro-cli chat --v3 --require-mcp-startup
```

## Acknowledgment


leandrodamascena:
Thank you for the substantial work on this contribution. The digest-bound human gate is a valuable capability, and the attention to provenance, validation, and fail-closed behavior is appreciated.

I believe this would fit better as an optional plugin rather than being embedded directly into the AI-DLC workflow. The current implementation couples common workflow rules, question formatting, approval outcomes, audit metadata, and persisted configuration to Plannotator. Although an abstraction exists internally, the workspace MCP accepts only Plannotator and the Kiro configuration enables it by default.

There is also a portability concern: the implementation is available only for Kiro, while the workflow rules it changes are shared across all supported environments. This makes a common AI-DLC lifecycle contract depend on both one external product and one specific harness.

I recommend exploring this as an optional plugin while keeping the core workflow independent of Plannotator. Since we are targeting `v2` as the GA direction, the proposal would benefit from further discussion through a dedicated issue or RFC against that branch. The RFC can evaluate the cross-harness design, discuss tradeoffs, and determine whether the capability should be accepted before implementation proceeds.

PR #747, opened 2026-08-12
fix: make learning persistence content-addressed and provenance-bound (2.6.36)

### LINKED ISSUE #735: [Bug]: aidlc-learnings.ts persist silently drops a learning when its cid marker collides with an unrelated intent's — RULE_LEARNED still emitted, audit and file state diverge
### Description

`aidlc-learnings.ts persist` can silently drop a learning when its idempotency marker collides with an unrelated intent's. `cidMarker(slug, candidateId)` (`core/tools/aidlc-learnings.ts:408`) keys the marker only on `(stage slug, candidate id)`, with no intent-scoping — but candidate IDs restart at `c1` on every stage run, while the destination files (`project.md`/`team.md`) are workspace-level, accumulating learnings from every intent that's ever run (by design — see `stage-protocol.md` §13). So once a stage type runs more than once in a workspace, a later intent's low-numbered candidate can collide with an earlier, unrelated intent's marker. `handlePersist` then skips the write, but still emits `RULE_LEARNED` and still reports success.

### Steps to Reproduce

1. In an AI-DLC v2 workspace, run any gated stage in one intent (Intent A) through its §13 Learnings Ritual, and persist a learning candidate (e.g. `c1`) for stage `<slug>` via `aidlc-learnings.ts persist --slug <slug> --selections-json <path>`.
2. Switch to (or create) an unrelated Intent B in the same workspace, run the same stage `<slug>`, and persist a different learning that happens to also be numbered `c1` (candidate ids restart at `c1` on every stage run, so this isn't contrived — it's the common case for any stage run more than once in a workspace).
3. Observe the CLI output of the second `persist` call.
4. Inspect `project.md`/`team.md` and compare against the CLI's own `rule_learned` count.

### Expected Behavior

Both intents' candidates should be persisted as separate lines, and the reported `rule_learned` count should match the number of lines actually written.

### Actual Behavior

The second `persist` call reports `rule_learned: 1` (success) and emits a `RULE_LEARNED` audit row, but `project.md`/`team.md` never receives Intent B's line — the dedup check sees Intent A's marker already present and silently skips the write. Audit and file state diverge with no error surfaced anywhere. Confirmed 9 times in real usage, from 1-of-N candidates dropped to a full 4-of-4 batch on `code-generation`.

### AI-DLC Version

v2 (alpha)

### Release / Commit

v2.3.0 (commit `29a31f7`). Re-verified unchanged on the unreleased `v2` branch HEAD `e3bf9be1` (CHANGELOG-stamped `2.5.64` as of 2026-08-11) — not a stale-version issue.

### AI-DLC Phase

Not phase-specific

### Platform / IDE

Claude Code

### AI Model

Claude (Anthropic), via Claude Code — the defect itself is in deterministic TypeScript with no LLM in the loop, so reproduction and verification were done via direct `bun` invocation of the tool, independent of model.

### Environment

N/A — reproducible via direct `bun core/tools/aidlc-learnings.ts persist ...` invocation against a workspace fixture; no OS/IDE-specific dependency observed.

### Additional Context

The marker's own doc comment states its actual purpose as single-run crash recovery ("a same-day re-run... is a no-op") — it was never meant to be unique across intents, but it's the only key checked against a file that spans every intent in the workspace. Once we knew this could happen, we could no longer trust `persist`'s own success report — every gated stage needed an extra independent verification pass, and recovering from one collision meant a manual detect → diagnose → hand-write → re-verify cycle. That's a standing per-stage tax, not a one-off.

*Suggested fix:* include the intent slug (or record-dir name) as a third component of `cidMarker`'s key. `aidlc-lib.ts` already exports `activeIntent(projectDir, space?, explicit?)`, which resolves the current intent's record-dir name deterministically — `handlePersist` already has `projectDir` in scope, so this is a signature change plus a call-site update, no new plumbing required. New markers can't collide with old-format markers already in the file, so this is forward-compatible without any migration.

We ran into this repeatedly enough that we had to patch our own vendored `v2.3.0` copy just to keep the ritual reliable — verified against a live two-intent repro matching the steps above. Happy to open a PR for it if that works for you.

vdevanathan-olo:
Fixes #735

# Summary

Both review rounds on this PR found that the original fix's marker — scoped to `(intent, stage, candidate-id)` — did not yet provide a *stable* idempotency identity. This update addresses all four findings (3×P1, 1×P2) plus the doc-drift and mechanical notes from the second review, rebases onto current `v2`, and re-verifies everything end-to-end.

### [P1] Candidate IDs still collide across stage attempts within one intent
Candidate ids restart at `c1` on every `surface()` call, so a later run of the same stage in the same intent reused `c1` for a genuinely different learning — the old marker treated that as the same key and silently dropped the second write.

**Fix:** `cidMarker`'s third component is now a content hash of the learning's own text (`contentHash()`, sha256 truncated to 8 hex chars — matching this repo's existing `shortHash` convention in `aidlc-doctor-bundle.ts`), not the positional candidate id. An exact retry (crash recovery) still hashes identically and dedups; two different learnings landing on the same positional candidate id never collide. `Candidate-ID` is kept in the audit row for human traceability; `Content-Hash` is the actual dedup identity going forward.

### [P1] Selections not bound to their originating intent
`persist` re-resolved `activeIntent()` live at execution time, so a surface-under-A / switch-to-B / persist sequence wrote under B's marker instead of A's; the audit read/write also resolved the live cursor independently inside the lock.

**Fix:** `surface()`'s output and the selections-json schema now carry `space`/`intent`, bound once at surface time. `persist` reads and uses these directly — it never calls `activeIntent()`/`activeSpace()` itself. `withAuditLock`, `readAllAuditShards`, and `appendAuditEntryUnlocked` (both the learning and the sensor write paths) are all threaded the same pinned `(intent, space)`, so the audit row and the practice line can never land under different intents (this is the construction-order point from the second review's design note).

### [P1] Existing markers duplicated after upgrade
The new marker format didn't recognize either prior format, so retrying an already-persisted learning post-upgrade could duplicate it.

**Fix:** Both the audit-row check and the line-marker check now also recognize the two prior formats — the original pre-#735 `<!-- cid:<stage>:<candidate> -->` and #735's own first fix's `<!-- cid:<intent>:<stage>:<candidate> -->`. The audit-side legacy match (`priorLegacyAuditRow`) requires the row to *lack* a `Content-Hash` field, which is load-bearing, not decorative: without that guard, a row the new code itself wrote for a genuinely different learning that happens to reuse the same candidate id would be mistaken for a legacy retry, silently reintroducing the first finding. Caught this exact regression in my own first attempt via the dedicated test for finding #1, fixed before it reached this diff.

### [P2] Ambiguous intent resolution must not fall back to `unscoped`
**Fix:** `resolveSurfaceIntent()` distinguishes a genuinely empty workspace (zero intent records — a legitimate flat/legacy shape, `intent: null` is correct) from real ambiguity (multiple records, no valid cursor — now fails closed with an explicit error naming the fix). This check runs *before* `surface()` touches any per-intent path — `readStateFile`/`readRuntimeStageRow` are now passed the resolved intent explicitly, so an ambiguous workspace fails with this specific message rather than an unrelated-looking "state file not found" (which is what happens if the ambiguity check runs after the state read, since that read's own internal resolution hits the identical ambiguity first).

### [P2] Marker format documentation + doc drift
Updated `core/aidlc-common/protocols/stage-protocol.md` (§13, both the `surface` and `persist` steps) and `docs/reference/07-sensor-system.md` to describe the new marker shape and the surface-time provenance binding. Regenerated `dist/` via `bun scripts/package.ts` (`--check` clean across all 7 trees).

### Mechanical: rebase, version, test-slot rename

Rebased twice as `v2` kept moving while this PR was open — see the timeline comment below for what changed between pushes.

- Rebased onto current `upstream/v2` (`3d02c500`, "ci: gate PRs on the deterministic integration + e2e tiers").
- Version bumped to **2.6.17** — `2.6.15` and `2.6.16` both landed on unrelated merged PRs (#731, #772) after this branch first claimed `2.6.15`.
- Regression test renamed to **`t300`** — `t278` (the original slot) was claimed by merged PR #617, and this branch's own follow-up re-slot, `t284`, was in turn claimed by merged PR #731's `t284-knowledge-sync-rebind`. All touch-points updated: header comment, `describe()` title, `gen-coverage-registry.test.ts`'s registry entry, and both `tests/.coverage-registry.json` entries.
- Found and fixed a pre-existing defect while doing the first rename: the old `t278` file's `covers:` header (`cli:aidlc-learnings(persist)`) didn't match this project's actual claim syntax (`subcommand:<tool>:<sub>`) and silently registered zero coverage claims — traced through `parseCoversHeader` to confirm, not guessed. The corrected header now credits both `persist` and `surface`.

## User experience

**Before this update:** the fix from the first push still collided within a single intent across separate stage runs, could misattribute a write after an intent switch, could duplicate a learning on upgrade, and could silently unscope a genuinely ambiguous workspace.

**After:** all four are closed, each with a dedicated regression test that reproduces the reviewer's exact scenario before asserting the fix.

**Upgrade:** re-copy your `dist/<harness>/` shell into the project. The selections-json schema is now stricter — `space`/`intent` are required fields (bound from `surface()`'s own output), not inferred; a hand-built selections file missing them now fails loudly rather than being silently accepted.

## Checklist

* [x] Changes are documented

## Test Plan

- [x] `tests/unit/t300-learnings-cid-collision-followup.test.ts` (renamed/superseding `t278`, then `t284`) — 11 tests, one dedicated to each of the four findings plus the original #735 cross-intent regression and the new schema validation:
  - two intents' identically-numbered candidates both persist (original #735 regression, re-verified under the new marker)
  - a same-intent same-day re-run remains a no-op
  - two *different* learnings on the same positional candidate id, same intent, separate persist calls, both land (finding #1)
  - surface-under-A / switch-to-B / persist still writes under A's marker (finding #2)
  - the original pre-#735 2-part marker is recognized on retry (finding #3)
  - #735's own first-fix 3-part marker is recognized on retry (finding #3)
  - missing/malformed `space`/`intent` in selections-json fails loudly
  - zero intent records resolves as legitimately unscoped (not ambiguous)
  - multiple intent records with no valid cursor fails closed
  - multiple intent records *with* a valid cursor still resolves normally
- [x] `bun run check` (`package.ts --check` + `typecheck` + `lint`) → exit 0.
- [x] `bun tests/run-tests.ts --smoke --unit --parallel 8` (CI's first job) → 235 files, 6765 assertions. 3 failing files (`t248-codekb-scope-diff`, `t248-copilot-packaging`, `t298-knowledge-transaction`) — all confirmed pre-existing and unrelated to this change; `t298` specifically re-verified against a pristine, unmodified `v2` checkout in a throwaway worktree (identical failure, identical stale UUID).
- [x] `bun tests/run-tests.ts --integration --e2e --no-llm --parallel 8` (CI's second job) → 173 files, 7 failing assertions across 2 files (`t224-plugin-selection`, `t92`) — both match this repo's previously-documented environment-specific failures (tool-version/git-fingerprint differences), unrelated to this change.
- [x] This branch's own new/renamed test, `t300-learnings-cid-collision-followup.test.ts`, passes 11/11 in both runs above.

## Acknowledgment


leandrodamascena (changes requested):
Thank you for identifying the cross-intent collision. The underlying bug is real, but the proposed marker does not yet provide a stable idempotency identity.

### [P1] Candidate IDs still collide across stage attempts within one intent

Candidate IDs restart at `c1` whenever candidates are surfaced. Adding the intent directory prevents A/c1 from colliding with B/c1, but a later execution of the same stage in A still reuses A/stage/c1. I reproduced two differently worded `c1` selections in one intent: the second persist reported `rule_learned: 0` and its text never reached `project.md`.

Location: `core/tools/aidlc-learnings.ts:208-222,406-415,485-489`.

### [P1] Selections are not bound to their originating intent

Neither the surfaced output nor selections JSON carries space or intent identity. `persist` resolves whichever intent is active at execution time. I reproduced surfacing a selection under intent A, switching to B, then persisting it; the learning was written with B's marker. The intent is also resolved before the audit lock, while audit and destination paths resolve the live cursors again inside it.

Bind stable space/intent/candidate provenance into the surfaced selection and validate/pin it throughout the persist transaction.

Location: `core/tools/aidlc-learnings.ts:225-232,268-271,428-442,461-478`.

### [P1] Existing markers are duplicated after upgrade

The new implementation searches only for the new marker format. On retrying a learning previously persisted as `<!-- cid:<stage>:<candidate> -->`, the old audit row is found but the old line marker is not. The tool appends the same learning again under the new marker. I reproduced one audit row with two practice lines after upgrade.

Add explicit legacy-marker compatibility or an in-place migration and pin exactly one audit row and one practice line.

Location: `core/tools/aidlc-learnings.ts:413-415,485-499`.

### [P2] Ambiguous intent resolution must not fall back to `unscoped`

With multiple intents and no valid active cursor, `activeIntent()` returns null. Persist currently converts that ambiguity into the shared `unscoped` identity, recreating the original collision class. Fail closed unless this is an explicitly recognized legacy flat workspace.

Location: `core/tools/aidlc-learnings.ts:430-434`.

The focused suites pass (`114/114`) and package parity is clean, but they do not cover the scenarios above. After addressing them, the branch also needs a rebase onto current `v2`, a rebump to `2.5.70`, and its new test renamed from `t278` to `t279` because #617 now owns `t278`.

apackeer:
Independent second review at head 349f05a2, run alongside @leandrodamascena's - I concur with CHANGES_REQUESTED and won't repeat his findings. Adding what my round surfaced that isn't covered yet:

### [P2] The marker format is documented in two places this PR does not update

- `core/aidlc-common/protocols/stage-protocol.md:1264` still reads `<!-- cid:<slug>:<id> -->` (2-part). This file ships into every `dist/<harness>/` tree, so the fix needs a `bun scripts/package.ts` regen as well. The new code comment at `core/tools/aidlc-learnings.ts:412` cites stage-protocol.md section 13 while the protocol still shows the old key.
- `docs/reference/07-sensor-system.md:317` carries the same stale 2-part format.

Repo policy (AGENTS.md, Documentation Policy) asks for these to land in the same commit as the rename/format change.

### [NOTE] t278 rename mechanics

When moving the test off the taken t278 slot, three more spots need the same rename: the in-file header comment (line 3), the `describe()` title (line 124), and the `EXPECTED_NONE_TO_CLI` entry at `tests/unit/gen-coverage-registry.test.ts:772`; then re-run `bun tests/gen-coverage-registry.ts --check`.

### [NOTE] Slot targets have moved again

As of 2026-08-17, v2 is at 2.6.2, so the re-bump target is past the 2.5.70 mentioned above, and open PRs claim nearby version and t-number slots. Re-check what's free at rebase time - t68 catches a version/heading mismatch, and the registry check catches the test-list side.

### Confirmations from my round (these still stand)

- `bun scripts/package.ts --check` at head passes all 7 trees - dist genuinely regenerated, not hand-edited.
- Targeted slice green: 6 files / 84 assertions (t278-learnings, t97, t158, t99-learnings-gate-flow, gen-coverage-registry, t68).
- One design fact worth keeping in view for the provenance binding asked for above: the audit-side idempotency check (`priorAuditRow`, `core/tools/aidlc-learnings.ts:485`) staying keyed on (stage, candidate) without intent is correct today only because audit shards live inside the intent record dir (`core/tools/aidlc-lib.ts:3137-3142`). If `persist` starts accepting selections that carry their own intent provenance, the audit read/write must be scoped to that same pinned intent rather than the live cursor, or the audit row and the practice line can land under different intents.

vdevanathan-olo:
Thanks both — all four findings addressed, plus the doc-drift and mechanical notes. Rebased onto current `v2` (`074cf4c7`) and re-bumped to `2.6.15` (`2.5.70`/`2.6.2` are both well behind current `v2` now). PR description is fully rewritten to match; summary here for the diff itself:

@leandrodamascena

- **P1 same-intent collision**: `cidMarker`'s third component is now a content hash of the learning's text, not the candidate id — candidate ids restart at `c1` every run, a content hash is what's actually stable. Regression test reproduces your exact repro (two differently-worded `c1` selections, same intent).
- **P1 provenance binding**: `surface()` now binds `space`/`intent` once; `persist` reads them from the selections-json and never re-resolves the live cursor. Audit read, audit write, and lock identity are all pinned to that same value — not just the practice-line write.
- **P1 legacy duplication**: both the audit-row check and the marker-line check now recognize the two prior formats (pre-#735's 2-part, and this PR's own original 3-part). The audit-side check requires the row to *lack* a `Content-Hash` field to count as legacy — otherwise it would reintroduce the first finding by treating a fresh, differently-worded row as a legacy retry. (Caught myself doing exactly that in an earlier iteration — a dedicated test for finding #1 failed and surfaced it before it reached this diff.)
- **P2 fail-closed ambiguity**: multiple intent records with no valid cursor now fails closed with an explicit error. Zero records is still legitimately unscoped — distinguished from real ambiguity, both paths have dedicated tests.

@apackeer

- Marker format docs in `stage-protocol.md` and `07-sensor-system.md` updated, `dist/` regenerated in the same commit.
- `t278` → `t284` (all three rename touch-points + registry check). Slot re-checked fresh against live `v2` right before this push, not reused from either of our earlier numbers.
- Your design note on `priorAuditRow`'s scoping was exactly right and is what the provenance-binding fix above implements — the audit read/write are now pinned to the same intent as the practice-line write, not the live cursor.
- One thing I found independently while doing the rename: the old `t278` file's `covers:` header (`cli:aidlc-learnings(persist)`) doesn't match this repo's actual claim syntax and silently registered zero coverage claims — traced through `parseCoversHeader` to confirm. Fixed in `t284`'s header and re-minted `tests/.coverage-registry.json`.

Full test plan and file-by-file breakdown is in the PR description. `bun run check` clean; unit+integration sweep is 311 files / 5953 assertions with only the same 4 pre-existing environment-specific failures, reconfirmed against a pristine `v2` checkout via `git stash` this session.

vdevanathan-olo:
Update: rebased again — `v2` moved and reopened conflicts, now resolved.

Since yesterday's push, `v2` advanced three more commits (`074cf4c7` → `3d02c500`, via #731/#772/#791), which reintroduced conflicts here in two ways:

- **Version-slot collision**: this branch had claimed `2.6.15`; #731 landed as `2.6.15` and #772 right after as `2.6.16`. Rebumped to **2.6.17**.
- **Test-slot collision**: this branch's regression test had been renamed to `t284`; #731 also added its own `t284-knowledge-sync-rebind`. Renamed ours to **`t300`**.

No changes to the substantive fix itself (still the same content-hash marker / surface-time provenance binding / legacy-marker compat / fail-closed ambiguity resolution described above) — just the rebase, the slot renumbering, and regenerating `dist/` to match.

Re-verified end-to-end after resolving: `bun run check` clean; `bun tests/run-tests.ts --smoke --unit` (235 files) and `--integration --e2e --no-llm` (173 files) both run — the only failures are 5 pre-existing/environment-specific ones unrelated to this change (3 previously known, plus `t298-knowledge-transaction`, which I confirmed fails identically on a clean, unmodified `v2` checkout). This branch's own test (`t300`) passes 11/11 in both runs.

PR now shows `mergeable: MERGEABLE` — no outstanding conflicts.

apackeer (changes requested):
Independent second review, rounds 2 and 3, at heads 264705653 and cf958147a. First, credit where due: all four findings from the earlier rounds are genuinely fixed - I verified each against the code, and the regression suite reproduces the exact reviewer scenarios. The 2026-08-18 re-push is also a verifiably pure rebase/re-slot: a diff-of-diffs of the two PR deltas differs only in the version strings and the t284 -> t300 rename, with `aidlc-learnings.ts` byte-identical, and the targeted slice is green at head (7 files / 96 assertions; `package.ts --check`, registry `--check`, and `bun run check` all clean).

Three findings from my rounds that have not reached this thread yet - all reproduced deterministically at cf958147a:

### [P1] The practice line is not pinned to the surface-time space - only the lock and audit are

`practiceFilePath` (`core/tools/aidlc-learnings.ts:107-108`) calls `memoryDirFor(projectDir)` with no space argument, and that helper defaults to the live `aidlc/active-space` cursor (`core/tools/aidlc-graph.ts:332-334`). `persist` threads `pinnedSpace` into `withAuditLock`, `readAllAuditShards`, and both `appendAuditEntryUnlocked` calls, but the marker dedup read and the line write both hit whichever space is active at persist time.

Repro: seed an intent record in space `default`; write a selections file pinned `{"space": "default", "intent": "<record>"}`; write `other` to `aidlc/active-space`; run `persist`. Result: exit 0, `rule_learned: 1`, the practice line plus the pinned intent's marker land in `aidlc/spaces/other/memory/project.md` (default's project.md is never created), while the RULE_LEARNED row lands under `aidlc/spaces/default/intents/<record>/audit/` with `Destination:` pointing at the other space's file. The default space's rules resolver never loads the practice, and a later replay under default appends the line there too (hasRow true, hasLine false), leaving a stray copy in `other`.

This is the same defect class as the intent-axis finding this update fixes, one level up, and the crash-replay story the docs advertise ("the selections-file is the replay artefact") makes the switch-before-replay window real. It also contradicts the PR body's "persist ... never calls `activeIntent()`/`activeSpace()` itself" - it does, transitively, on the write path. The fix is one line - `memoryDirFor(projectDir, pinnedSpace)` threaded through `practiceFilePath`/`ensureFile` (`selFile.space` is required and always a string) - plus a space-switch variant in t300's finding-#2 group.

### [P2] The sensor branch still dedups on the unstable (stage, candidate_id) key

`priorAuditRow(auditContent, "SENSOR_PROPOSED", stageSlug, sel.candidate_id)` at `core/tools/aidlc-learnings.ts:698-703`, while manifest existence is keyed on the sensor id. Reproduced both ways: a DIFFERENT sensor reusing `c1` (candidate ids restart per surface batch) gets its manifest written and bound but its SENSOR_PROPOSED row silently skipped (`sensor_proposed: 0` for a run that wrote a manifest); the SAME sensor retried under `c2` emits a duplicate row. No data loss - manifests and binds are id-keyed - but the audit trail under- and over-counts, which is the integrity property this PR exists to fix. This pre-exists the PR, so a follow-up issue is fine by me; the fix is shim-free though: match on the `Sensor ID` field, which every legacy SENSOR_PROPOSED row already carries.

### [P2] The legacy-compat window still silently drops one new learning per pre-upgrade (stage, candidate_id) pair

Both legacy checks are candidate-id-keyed (`priorLegacyAuditRow` at :636, the two legacy marker probes at :639-640). A NEW, differently worded learning whose positional candidate_id collides with a pair persisted BEFORE the upgrade hits hasRow and hasLine together and is dropped. Repro: seed the pre-#735-format line plus a hash-less audit row for an OLD text under c1, then persist a NEW different text as c1 -> exit 0, `rule_learned: 0`, new text absent. So "two different learnings landing on the same positional candidate id never collide" does not hold across the upgrade boundary. The audit rows cannot disambiguate (no hash), but the legacy-marked LINE's text is sitting in `content`: only count a legacy marker/row match as a retry when the marked line's text equals `sel.text`. Alternatively, keep the retry-wins trade and state the residual window in the CHANGELOG bullet - as written it overclaims.

### Notes (non-blocking)

- The re-slot has gone stale again: #792 merged taking `2.6.17` AND the t300 number (`tests/integration/t300-plugin-kit.test.ts` is on v2 now), so this branch needs a third re-slot. t68 catches the duplicate CHANGELOG heading but NOT a cross-suite t-number duplicate, and nearby slots are moving fast - re-verify both against live v2 at push time.
- A selections file written before this change (the docs' "replay artefact") now fails post-upgrade with "selections-json is malformed: missing or non-string space". Loud, and the remedy is a re-surface - just worth one clause in the CHANGELOG bullet, since the policy asks for the errors users see.
- Two stale comment fragments in `tests/unit/t97.test.ts:36`: the header still writes the marker's third component as `<id>` (it is now the content hash) and says the fixtures "resolve no active intent", though persist no longer resolves anything - the fixtures pass `intent: null`.

Requesting changes on the P1; the two P2s are fine as follow-ups (or, for the legacy window, one honest CHANGELOG sentence).

vdevanathan-olo:
Thanks for the thorough second-round review! All three findings are fixed, pushed at `b3e19f0d`.

**[P1] `practiceFilePath` not pinned to the surfaced space** — fixed. `practiceFilePath` now takes the surface-time-pinned `space` and threads it through to `memoryDirFor`, instead of falling back to whatever space happens to be live-active when `persist` runs.

**[P2] Sensor dedup keyed on the unstable `(stage, candidate_id)`** — fixed, with one correction to the suggested approach: matching on Sensor ID *alone* (dropping stage entirely) turns out to introduce a new bug. Binding a sensor to a stage's frontmatter is legitimately per-stage even though the manifest itself is a per-project singleton — so dropping stage from the key means once *any* stage proposes a sensor, no *other* stage could ever bind that same sensor to its own frontmatter again. I caught this with a dedicated regression test before it shipped (two unrelated stages independently proposing the same sensor id — the second stage never got the frontmatter bind). The fix now keys on `(stage, sensor id)` instead — same shape as before, just with `candidate_id` replaced by the sensor's own stable id, which closes exactly the instability you found without losing cross-stage binding.

**[P2] Legacy-marker match not gated on text** — fixed. A legacy-marker match now requires the marked line's own text to equal the current selection's text (`legacyLineMatchesText`), so a different learning reusing an old candidate_id post-upgrade is no longer mistaken for a retry of the original and silently dropped.

Also fixed the two stale comment fragments in `t97.test.ts:36`, and re-slotted to `2.6.21`/`t306` (`v2` had claimed `2.6.17`/`t300` again via #792 since the last push — checked live against every currently-open PR's claimed version/test-slot to avoid a fourth collision, though given how fast `v2` is moving that may still need a re-check at actual merge time).

**Current state:**
- 7 new regression test groups (one per finding across all three review rounds, 15 tests total in `t306`) — each confirmed red against the pre-fix code and green after.
- Rebased cleanly onto `v2` @ `fbb1460c` — conflicts were purely mechanical (CHANGELOG/README/version-bump), no code conflicts. `mergeable: MERGEABLE`.
- Full local suite (`bun tests/run-tests.ts` smoke + unit + integration + e2e `--no-llm`) passes for everything touching this change; `typecheck`, `lint`, `scripts/package.ts --check`, and `tests/gen-coverage-registry.ts --check` all clean.

Ready for another look whenever you get a chance.

apackeer:
@vdevanathan-olo Vinod, this looks good to me. i will take over and prepare this for merge. Thankyou for your contribution. it is sincerely appreciated.

@leandrodamascena can you take a second look after i preppare for merge?

apackeer (approved):
Prepared for merge as promised above. `92bbe5fc2` (pushed to Vinod's branch under maintainer edits) closes the three findings my round-4 pass surfaced on top of `b3e19f0d1`, so nothing is left outstanding on my side and I am flipping my earlier changes-requested to approve.

What the takeover commit adds:

- **Fail-closed unscoped replay** (the round-4 P1): `persist` now refuses an `intent: null` selections replay once intent records exist in the pinned space (first statement inside the audit lock, `core/tools/aidlc-learnings.ts`). Previously `intent: null` degraded to `undefined`, which the audit layer resolves to the live/lone intent (`core/tools/aidlc-audit.ts:572-579`), so an unscoped learning surfaced before an intent existed could be audited under an intent born later. The error directs the user to re-run the stage's `surface` step.
- **Provenance validation**: `space` is validated with the existing `validSpaceFlag()` and a non-null `intent` must be a path-safe record-directory name, both before any path is built. A malformed selections file (for example `space: "../../x"`) can no longer write outside `aidlc/spaces/` or redirect audit/lock paths.
- **Docs drift**: the `persist` row in `docs/reference/06-hooks-and-tools.md` now states the surface-time-bound destination and the `(Stage, Content-Hash)` / `(Stage, Sensor ID)` dedup keys, and `core/knowledge/aidlc-shared/audit-format.md` adds `Content-Hash` to `RULE_LEARNED`'s required fields.
- The `2.6.21` CHANGELOG entry is amended in place (no version change), including the exact error a pre-upgrade selections file receives and the re-surface remedy. `t306` gains 4 regression tests (19 assertions total); dist regenerated via `scripts/package.ts`.

Verification at `92bbe5fc2`: targeted slice green, 10 files / 130 assertions / 0 failed (t306, t97, t99, t112, t158, t68, gen-coverage-registry, plus the doc-pin suites t239, t28-audit-event-sync, t86); `package.ts --check`, `gen-coverage-registry.ts --check`, typecheck, and lint all clean. Both round-4 repro probes now exit non-zero with nothing written.

@leandrodamascena over to you for the second look. Your 2026-08-13 findings were addressed back at the round-2 push (content-addressed marker, surface-time space/intent binding, fail-closed ambiguity) and everything found since is closed at this head, so the whole diff is ready for a fresh pass. Slot note: `2.6.21`/`t306` are free on `v2` @ `fbb1460c` today, but with 2.6.19/2.6.20 claimed by open PRs this needs the usual re-check at actual merge time.

leandrodamascena:
Final maintainer pass completed at `44419dd8f`.

All findings from the latest review are resolved:

- Learning identity now uses the full SHA-256 digest, preventing silent data loss from short-hash collisions.
- Older candidate-ID and 8-character hash markers remain safely supported.
- Duplicate selections in the same batch now produce exactly one practice line and one audit row.
- Persist now rejects missing, invalid, or symlinked space/intent provenance.
- Persist rejects a stage slug that differs from the stage recorded at surface time.
- Documentation, generated distributions, changelog, and version were updated to `2.6.36`.

Validation completed:

- 140 focused tests passed with 1,701 assertions.
- The dedicated regression suite passed all 27 tests.
- Integration tests that timed out during the parallel full-suite run passed when rerun individually.
- Remaining full-suite failures are unrelated and reproducible on pristine `v2`.
- An independent final code review found no remaining issues.

The PR is ready for final approval and merge.

leandrodamascena (approved):
All blocking findings are resolved at `44419dd8f`. The focused regressions, generated parity, typecheck, lint, and isolated integration reruns are green. Approved for merge.

PR #748, opened 2026-08-12
fix: make consumes_absent respect produces_kinds when flagging a real gap (2.5.68)

### LINKED ISSUE #736: [Bug]: consumes_absent's expected flag doesn't check produces_kinds — a kind-gated unit gets a false "real gap" flag for an artifact it was never supposed to produce
### Description

`splitConsumesByPresence()` (`core/tools/aidlc-orchestrate.ts:1113`) computes each `consumes_absent` entry's `expected` flag purely from whether a producing stage is on the active scope's path — it never checks the current unit's `kind` against that producer's own `produces_kinds` gate. Its sibling function two lines below, `resolveProduces()`, already does exactly this kind-filtering on the produces side via `filterProducesByKind(node.produces_kinds, ..., unitKind)`. So a unit whose `kind` correctly excludes it from producing a given artifact still gets flagged `expected: false` ("a real gap") on the next stage's directive, because the producing stage itself ran.

### Steps to Reproduce

1. In an AI-DLC v2 Construction workflow, run a per-unit stage that has a `produces_kinds` gate (e.g. `functional-design`, which gates `business-rules` to `[service, spec, library]`, excluding `ui`) for a unit tagged `kind: ui`.
2. Confirm the unit correctly does not produce the gated artifact (`business-rules.md` correctly absent).
3. Advance to the next stage that consumes that artifact as `required: true` (e.g. `nfr-requirements` consuming `business-rules`) and inspect its `run-stage` directive's `consumes_absent` entries.

### Expected Behavior

The `consumes_absent` entry for the kind-excluded artifact should report `expected: true` — its absence is by design, not a gap.

### Actual Behavior

The entry reports `expected: false` — flagged as "a real gap — surface it per the recovery protocol" — even though the artifact's absence is correct and intentional for this unit's kind. Confirmed live at a `nfr-requirements` directive for a `ui`-kind unit in our own workflow.

### AI-DLC Version

v2 (alpha)

### Release / Commit

v2.3.0 (commit `29a31f7`). Re-verified unchanged on the unreleased `v2` branch HEAD `e3bf9be1` (CHANGELOG-stamped `2.5.64` as of 2026-08-11) — not a stale-version issue.

### AI-DLC Phase

Construction

### Platform / IDE

Claude Code

### AI Model

Claude (Anthropic), via Claude Code — the defect itself is in deterministic TypeScript with no LLM in the loop, so reproduction and verification were done directly against the compiled stage graph, independent of model.

### Environment

N/A — reproducible via direct inspection of the compiled `stage-graph.json` and the `run-stage` directive it produces; no OS/IDE-specific dependency observed.

### Additional Context

This is a dated integration gap, not a misuse of either feature: `consumes_absent`/`expected` shipped in PR #482 (v2.1.7, closing #461) before `produces_kinds` existed at all; `produces_kinds` shipped later in PR #511 (v2.2.18, closing #502), wired only into the produces side. Nothing since has touched `splitConsumesByPresence`. Any project tagging unit kinds should eventually hit this too, since `produces_kinds` is used across 4 construction stages.

*Suggested fix:* thread `unitKind` into `splitConsumesByPresence` (mirroring `resolveProduces`'s existing pattern) and check the producing stage's `produces_kinds` gate — via the already-exported `filterProducesByKind` helper — before computing `expected`. `buildRunStageDirective` already receives `unitKind` and already passes it to `resolveProduces`; it just needs to pass it through here too.

We ran into this in our own workflow and had to patch our own vendored `v2.3.0` copy to unblock it — verified against the real `produces_kinds` data in our own compiled graph. Happy to open a PR for it if that's welcome.

vdevanathan-olo:
Fixes #736

# Summary

`consumes_absent`'s `expected` flag now checks a producer's own `produces_kinds` gate before flagging an absent artifact as a real gap, so a kind-tagged unit that correctly never produces a given artifact no longer gets a false "real gap" alarm on the next stage's directive.

## Changes

- `splitConsumesByPresence(consumes, scope, codekbCtx)` → `splitConsumesByPresence(consumes, scope, codekbCtx, unitKind)`, mirroring `resolveProduces`'s existing `unitKind` parameter.
- `producerOnPath`'s computation now also checks `filterProducesByKind(producer.produces_kinds, [artifact], unitKind).length > 0` — a producing stage being on the active scope's path is necessary but no longer sufficient; it must also actually be scoped to produce this specific artifact for this unit's kind.
- `buildRunStageDirective` (the one call site) already received `unitKind` as a parameter and already passed it to `resolveProduces` — it now passes it through to `splitConsumesByPresence` too.
- `filterProducesByKind` returns its input unchanged when `unitKind` is `null`, so an untagged unit or a stage with no `produces_kinds` map sees byte-for-byte identical behaviour — this is a strict narrowing of the true case, not a behaviour change for anything that isn't kind-gated.
- Regenerated `dist/` via `bun scripts/package.ts` (not hand-edited).
- Version `2.5.68` (README badge + CHANGELOG).

## User experience

**Before:** a `ui`-kind unit correctly never produces `business-rules.md` (`functional-design`'s own `produces_kinds` map scopes it to `[service, spec, library]`, excluding `ui`). The next stage that consumes `business-rules` as `required: true` (e.g. `nfr-requirements`) still flags it `expected: false` — "a real gap, surface it per the recovery protocol" — on its own `run-stage` directive, because `functional-design` itself did run. The conductor is pointed at a recovery protocol for an artifact that was never supposed to exist.

**After:** the same `ui`-kind unit's `consumes_absent` entry for `business-rules` reports `expected: true` — its absence is by design, not a gap. A `service`-kind (or untagged) unit's identical absence at the same stage is unaffected and still correctly reports `expected: false`, since `business-rules` genuinely is in scope for that kind.

**Upgrade:** re-copy your `dist/<harness>/` shell into the project.

## Checklist

* [x] Changes are documented

## Test Plan

- [x] New regression test `tests/unit/t279-consumes-absent-kind-gate.test.ts`: (1) a `ui`-kind unit's absent `business-rules` is `expected: true`, not a false alarm — confirmed RED against the pre-fix `dist/`, then GREEN after regenerating; (2) a `service`-kind unit's identical absence stays `expected: false` (real gap, regression anchor); (3) an untagged unit's identical absence also stays `expected: false` (unchanged, regression anchor).
- [x] `bun test tests/unit/t113.test.ts tests/unit/t116-directive-path-resolution.test.ts tests/unit/t207-unit-kind-schema.test.ts tests/unit/t279-consumes-absent-kind-gate.test.ts tests/unit/gen-coverage-registry.test.ts tests/unit/t68-version-changelog-sync.test.ts` → 137 pass, 0 fail.
- [x] `bun run check` (`package.ts --check` + `typecheck` + `lint`) → clean (the only lint output is 3 pre-existing infos in an unrelated file, `t267-usage.test.ts`, unaffected by this branch).
- [x] `bun test tests/unit/t208-unit-kind-pruning.test.ts tests/unit/t236-ensemble-evidence-gate.test.ts` → confirmed the 21 failures in these two files reproduce identically on a pristine, unmodified `v2` checkout (unrelated environment-specific failures — missing question-flow/contribution-file fixtures, not this change) — not a regression from this branch.

## Acknowledgment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.5.68] - 2026-08-12
+
+`consumes_absent`'s `expected` flag now respects a producer's own `produces_kinds` gate, so a kind-tagged unit that correctly never produces a given artifact no longer gets a false "real gap" alarm on the next stage's directive. Fixes #736. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
+
+* `splitConsumesByPresence` now takes `unitKind` (mirroring `resolveProduces`'s existing parameter) and checks the producing stage's `produces_kinds` gate before computing `expected`.
+* An untagged unit, or a stage with no `produces_kinds` map, sees byte-for-byte identical behaviour — this is a strict narrowing of the true case, not a behaviour change for anything that isn't kind-gated.
+* No command or flag changes; no breaking change for CI or scripts.
+
 ## [2.5.64] - 2026-08-11
 
 Composed-workflow routing is now deterministic from proposal through execution, and the affected test-runner and live-journey paths fail closed. Stock matches present exactly the grid that will run, incomplete model-produced grids cannot match a stock scope, in-flight recomposition preserves the running plan, isolated stages no longer wait on nonexistent questions, and mid-flow new work carries one typed continuation contract with a named scope. **Upgrade:** re-copy your `dist/<harness>/` shell into the project.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.5.64-blue)
+![version](https://img.shields.io/badge/version-2.5.68-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-orchestrate.ts
+++ b/core/tools/aidlc-orchestrate.ts
@@ -1506,10 +1506,22 @@ function resolveConsumes(
 //     check against; everything stays in `consumes`, exactly as before.
 //   - a path still carrying the {unit-name} placeholder → existence is
 //     unknowable pre-Bolt; it stays in `consumes`.
+//
+// "A producer is on the path" is necessary but not sufficient — a producer
+// on path can still legitimately not produce THIS artifact for THIS unit's
+// kind, via that producer's own produces_kinds gate (mirrors resolveProduces's
+// existing unitKind filtering below). Without this, a kind-excluded artifact
+// (e.g. a `ui`-kind unit correctly never producing an artifact a stage's own
+// produces_kinds map scopes to `[service, spec, library]`) is flagged
+// `expected: false` — a false "real gap" alarm, since the producing stage
+// itself did run. `unitKind` defaults to null (an untagged unit, or a
+// non-per-unit stage), which keeps prior behaviour exactly:
+// filterProducesByKind is a no-op when unitKind is null.
 function splitConsumesByPresence(
   consumes: ResolvedConsume[],
   scope: string,
   codekbCtx?: CodekbCtx,
+  unitKind: string | null = null,
 ): { present: string[]; absent: Array<{ path: string; expected: boolean }> } {
   if (!codekbCtx) return { present: consumes.map((c) => c.path), absent: [] };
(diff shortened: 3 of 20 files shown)


leandrodamascena:
Thank you for reporting this issue and for taking the time to isolate it with a focused implementation and regression test. The bug was valid against this branch's original base, [`e3bf9be1`](https://github.com/awslabs/aidlc-workflows/commit/e3bf9be14265e3e1621bfbd34ca04c70b124c250).

Since then, [#617](https://github.com/awslabs/aidlc-workflows/pull/617), merged as [`327d3626`](https://github.com/awslabs/aidlc-workflows/commit/327d3626), introduced an earlier kind-aware filter in `resolveConsumes()`. When a per-unit producer's `produces_kinds` excludes the current Unit kind, the artifact is omitted before presence splitting, so it appears in neither `consumes` nor `consumes_absent`.

This preserves `expected:true` for its original meaning: an artifact is absent because its producer is outside the active scope. A kind-inapplicable artifact is not an input for that Unit and is therefore omitted entirely.

I reproduced #736's exact `ui` Unit / `nfr-requirements` / `business-rules.md` scenario on current `v2`:

```json
{"stage":"nfr-requirements","unit":"web","businessRulesAbsent":[]}
```

The merged `t278-per-unit-wave` suite also pins this omission behavior. Verification completed:

- PR head focused suites: 137 passed
- Current `v2` affected suites: 108 passed
- Package parity passed on both
- Exact #736 reproduction passed on current `v2`

Rebasing this PR would make its implementation redundant and its `expected:true` assertion conflict with the current contract. For that reason, I'm going to close this PR as superseded by #617 and treat #736 as fixed by that merged implementation.

Thank you again for identifying the original integration gap and contributing the analysis and test coverage.

Issue #742, opened 2026-08-12
[Bug]: Approval recovery deadlocks after confirmed artifacts invalidate the reviewer receipt


akash-doshi:
## Summary

Approval recovery deadlocks when artifacts change after summary confirmation. The post-confirmation write invalidates the existing reviewer receipt. Approval then requires a fresh receipt, but the advisory review budget may already be exhausted. The only recovery is to choose Request Changes, even when the user wants no content changes, because that starts a new gate attempt and resets the review budget.

Enterprise customer feedback showed this sequence:

1. Approval was rejected because an artifact had no native-tool write after confirmation.
2. The artifact was written again to satisfy the confirmation ordering requirement.
3. That write invalidated the existing reviewer receipt.
4. Approval then failed because a fresh receipt was required but the review budget was exhausted.
5. The user had to select Request Changes and answer a follow-up asking what should change, despite requesting no content change.

This exposes internal receipt, budget, and transition mechanics without offering a direct recovery action.

## Reproduction

1. Seed an approval stage with a one-pass advisory reviewer.
2. Confirm the summary and open the approval gate.
3. Record advisory review iteration 1 against the current artifact fingerprint.
4. Write a declared artifact after that receipt.
5. Attempt approval.

The artifact write makes the receipt stale. A second advisory review is then refused because the one-pass budget is exhausted. The only current reset is a human `Request Changes`, even when no content change is required.

## Expected behaviour

- The system explains that the current review is stale and offers to refresh it.
- One action starts a fresh advisory review against the confirmed artifacts and returns to approval.
- Users do not need to reject content they already approved.
- Receipt and review-budget details remain diagnostic rather than primary user-facing instructions.

## Actual behaviour

- Approval stops with a stale-receipt or exhausted-budget error.
- The UI offers no recovery action.
- Request Changes is the only way to reset the review attempt.
- The next prompt asks for content changes even when none are required.

## Related issue

This is separate from #719. That issue covers the Claude conductor recording summary confirmation after the first artifact write. This one covers the recovery deadlock after artifact and reviewer receipts fall out of order, including the Copilot user experience.

## Proposed remediation

The ask is one refresh action that reruns the advisory review and returns to the same gate:

- Add a `Refresh review and retry approval` action when an artifact fingerprint makes the reviewer receipt stale.
- Internally start a new review attempt, run one advisory pass against the current confirmed artifacts, and return to the same approval gate.
- Preserve the prior review findings in history.
[...]

Issue #741, opened 2026-08-12
[Bug]: Pipeline stages can skip links and accept artifacts written by the conductor


akash-doshi:
## Summary

Pipeline stages can repeat an earlier link, skip the final link, and still produce artifacts that look complete to the deterministic guards. The write audit records the tool and file but not the required pipeline actor, so conductor-written files can satisfy the same evidence as files written by the expected agent.

Enterprise customer feedback from a brownfield multi-repo microservices system exposed this sequence:

- The lead pipeline link ran four times.
- The required final link did not run.
- The conductor wrote the full declared artifact set.
- The stage guard could not distinguish those writes from valid final-link output.

The lifecycle state remained safe because the gate was never reached. The artifact store still accepted output from the wrong actor.

## Reproduction matrix

Use a two-link pipeline and a two-repository intent. Exercise these cases:

| Case | Expected result |
|---|---|
| Interrupt after link 1 | Resume at link 2 without repeating link 1 |
| Final link never runs | Gate remains closed |
| Conductor writes a declared final-link artifact | Write cannot satisfy final-link provenance |
| One repository has all artifacts, another is missing one | Completion guard refuses the stage |
| A repository has approved reuse evidence | Reused repository is exempt, remaining repositories are still complete |

Current behavior has no durable link receipt, accepts actorless writes, and treats one matching CodeKB artifact as enough for Reverse Engineering.

## Expected behaviour

- Pipeline links run once and in the declared order.
- Recovery resumes from the first missing link.
- The final link owns the declared stage artifacts.
- Completion requires every declared artifact for every selected repository.

## Actual behaviour

- Pipeline ordering exists only in conductor guidance.
- There is no durable receipt for link completion or final-writer identity.
- Any Write or Edit can create the same native-write evidence.
- Reverse Engineering can pass its artifact precondition when only one matching file exists in one repository.
- Multi-repository `produces` paths can point at the workspace root instead of the registered repository set.

## Root cause

- `harness/copilot/skills/aidlc/SKILL.md:86,92,113` leaves pipeline dispatch and ordering to conductor prose.
- `core/aidlc-common/stages/inception/reverse-engineering.md:43-48,150-194` requires developer then architect execution, but no link receipt enforces it.
- `core/hooks/aidlc-write-audit-log.ts:133-168` records the tool, file, and context without the expected pipeline actor.
- `core/aidlc-common/protocols/stage-protocol.md:817-819` does not require contribution evidence for pipeline stages.
[...]

Issue #740, opened 2026-08-12
[Bug]: Copilot Stop hook races with Resume and conversational turns


akash-doshi:
## Summary

The Copilot Stop hook continues workflow execution while a human answer is still pending. It also resumes straight after normal conversational replies because the Copilot adapter does not write the human-turn marker the core Stop logic depends on.

Enterprise customer feedback surfaced three failures:

1. A Resume menu was displayed repeatedly while Stop continued the stage before the user selected an option.
2. Normal questions such as asking why the workflow was stuck were followed by synthetic continuation turns.
3. The interactive no-progress cap did not stop repeated nudges because non-advancing audit rows changed its progress signature.

### Pending Resume

1. Seed an active stage and start a new Copilot session.
2. Run `next --resume` and leave the Resume question unanswered.
3. Invoke Stop.

Today Stop can probe `next` and continue the stage before `report --result resumed --user-input "1"` records the answer.

### Conversational turn

1. Record an engine touch for an active interactive stage.
2. Submit a text-only human question through the Copilot UserPromptSubmit adapter.
3. Invoke Stop after the model answers without running an orchestration command.

Today the adapter writes the audit row but not the human-turn marker, so Stop can inject another workflow turn.

### No-progress cap

1. Invoke Stop repeatedly without changing stage or directive state.
2. Append a human or subagent audit event between calls.

Today the audit line-count change resets the streak even though the workflow did not advance.

## Expected behaviour

- A pending Resume question blocks workflow continuation until the human answers.
- A normal human message gives the model one clean response turn without a synthetic continuation.
- The no-progress cap measures workflow progress rather than audit activity.
- Stop does not run inside delegated subagent turns.

## Actual behaviour

- Session start presents Resume, Redo, Jump, or Start Fresh without recording a pending-answer state.
- Stop independently probes `next` and can continue the active stage before the answer is reported.
- Copilot's UserPromptSubmit adapter records `HUMAN_TURN` but does not create the marker used by transcript-free Stop handling.
- Human and subagent audit events can reset the no-progress streak even when the stage has not advanced.

## Root cause

- `core/hooks/aidlc-session-start.ts:272-289` presents resume context without a durable pending-answer latch.
- `core/tools/aidlc-orchestrate.ts:4758-4838` handles the resume report without persisting pending or consumed answer state.
[...]

PR #749, opened 2026-08-12
fix: stabilize Copilot Stop continuation and Resume waits (2.6.12)

### LINKED ISSUE #739: [Bug]: Copilot Stop hook restarts completed steering and replays continuation tokens
## Summary

The Copilot Stop hook restarts steering after `continue` has already returned `run-stage`. It regenerates the first continuation token and hands it back to the model. Because `continue` accepts a used token again, the session keeps repeating the same stage directive.

This came from enterprise customer feedback on AI-DLC v2 running against a brownfield multi-repo microservices system. In one session, a single continuation token was executed 31 times and reinjected through Stop 25 times. The repeats consumed context, duplicated stage work, and forced compaction.

## Reproduction

You can reproduce the transport failure without a Copilot binary or model provider:

1. Use a rule bundle large enough to produce more than one steering part.
2. Run `next` and retain the returned token `T1`.
3. Run `continue T1` and retain the new token `T2`.
4. Run `continue T1` again instead of using `T2`.

The second use of `T1` returns the same part again instead of being rejected.

A Stop call after steering completes also runs a fresh `next`, which starts delivery from part 1 again and supplies a new copy of `T1` to the model.

The expected trace is:

```text
next          -> load-steering part 1, T1
continue T1   -> load-steering part 2, T2
continue T1   -> replay of part 2 today, denied after the fix
continue T2   -> part 3 or run-stage
Stop after run-stage -> preserve run-stage, never restart part 1
```

## Expected behaviour

- Each steering part is consumed once per Copilot session.
- After `run-stage`, Stop preserves or restates that final directive.
- Stop does not restart steering from part 1.

## Actual behaviour

- Copilot PostToolUse receives the successful shell result but AI-DLC does not retain it.
- Stop runs a fresh bare `next` and regenerates part 1.
- Previously used tokens remain valid and replayable.

## Root cause

- `harness/copilot/hooks/aidlc-copilot-adapter.ts:73-94,796-823` declares `tool_result` and `toolResult` but does not read the result text or persist the latest directive.
- `core/hooks/aidlc-continue-workflow.ts:856-965,1245-1267` runs a fresh `next` and embeds the returned steering token in the Stop reason.
- `core/tools/aidlc-orchestrate.ts:2266-2339,5323-5381` validates tokens but does not record consumption.
- `core/tools/aidlc-orchestrate.ts:244-265` records the final stage marker without steering-token progress.

## Proposed remediation

The ask is a per-session directive receipt plus continuation replay denial:

- Persist the latest successful `next`, `continue`, or `report` directive per Copilot session from PostToolUse.
- Store only bounded transport metadata: session ID, directive kind, stage, part, parts, token, and state hash.
- Make Stop reuse the pending final directive after `run-stage` instead of calling a fresh `next`.
- Reject a `continue` call that reuses a consumed token in the same session.
- Allow a fresh `next` to reset the receipt. Missing or invalid receipts should fail open.

## Acceptance criteria

- [ ] Replaying a consumed token is denied with a clear recovery instruction.
- [ ] Stop after `run-stage` does not emit `load-steering` for the same stage and state.
- [ ] A multi-part steering bundle reaches `run-stage` exactly once.
- [ ] Existing non-Copilot harnesses keep their current stateless transport behaviour.
- [ ] Tests cover PostToolUse capture, replay denial, fresh-`next` recovery, and Stop after `run-stage`.

## Verification plan

Extend these tests:

- `tests/unit/t248-steering-content-delivery.test.ts` for deterministic multi-part token delivery.
- `tests/unit/t249-copilot-adapter.test.ts` for PostToolUse receipt capture and PreToolUse replay denial.
- `tests/integration/t121-stop-hook-enforce.test.ts` for Stop after a final directive.
- `tests/e2e/t122-stop-hook-e2e.test.ts` for the live hook journey when the optional SDK environment is available.

Run:

```bash
bun test \
  tests/unit/t248-steering-content-delivery.test.ts \
  tests/unit/t249-copilot-adapter.test.ts \
  tests/integration/t121-stop-hook-enforce.test.ts \
  tests/e2e/t122-stop-hook-e2e.test.ts
```

The deterministic unit and integration cases must pass without Copilot or a model provider. The optional live journey must show one ordered token chain and one final `run-stage`.

## Related issues

- #740 covers Stop behavior while human or subagent interaction is pending.
- #741 covers pipeline corruption after the replayed stage directive.

## Environment

- AI-DLC v2.5.60 at `e3bf9be`
- Copilot CLI 1.0.x
- Interactive Copilot session

akash-doshi:
## Problem

Enterprise customer feedback exposed a Copilot continuation loop:

1. Copilot received a successful `next` or `continue` result.
2. AI-DLC did not retain that delivered directive.
3. Stop ran a fresh `next`, restarted steering at part 1 and returned a reusable token.
4. Resume and conversational waits could be followed by another synthetic workflow turn.

One captured session executed the same continuation token 31 times and received 25 Stop reinjections.

## Fix

- Preserve the latest delivered `load-steering` or `run-stage` directive and pass it through every existing Stop safeguard instead of probing a fresh `next`.
- Replace the fixed `.transaction` protocol with the repository's owner-stamped lock convention: dead-owner and post-grace recovery, CAS revalidation, canonical marker visibility and doctor reporting.
- Move token replay denial into an atomic engine cursor for exact-context Copilot-owned markers. The engine compares the complete token digest and publishes the exact successor token or tokenless `run-stage` before stdout.
- Preserve sessionless and non-Copilot stateless continuation deliberately. Follow-up #762 owns the all-harness invariant.
- Narrow Copilot command detection to execution-shaped direct, source-dispatcher and real compiled forms. Inspection commands pass, one terminal `2>&1` is supported, unsafe direct compounds are refused and ambiguous wrappers run untracked.
- Reject a foreign explicit `--project-dir` before any current-project coordination write.
- Track exact host/carried IDs when available. Missing correlation runs untracked and cannot permanently deny execution or guess a Post mutation.
- Keep Resume questions and conversational turns session-scoped without adding Cancel or broader #740 lifecycle behavior.

The marker stores routing metadata and hashes only. It does not persist rules content, prompts, audit prose, plans or artifacts.

## Scope

- Based on post-#751 `v2` at `4569754`
- Version `2.6.6`
- 6 authored behavior files plus focused tests/docs
- Generated projections across all seven harnesses
- 65 files total: `+13,608/-932`

Generated core projections account for most of the visible diff. No workflow schema, stage graph, plan approval, review recovery, NFR behavior or code-generation policy changes are included.

## Validation

- Final generated affected matrix: 239 tests passed, 2,687 assertions, zero failures
- Authored focused matrix before generation: 212 passed, 1,498 assertions, zero failures/skips
- Real native compiled dispatcher: `t249` 43/43, 629 assertions
- Coverage mechanism registry: 34/34, 1,202 assertions
- Candidate t92 standalone: 47/47
- Full no-LLM candidate profile at P4: 391 files, five failed files, 21 failed assertions
- All five full-profile failed files are also present in the pristine post-#751 control profile
- t145 x4 and t272 x3 reproduce identically on pristine post-#751 `v2`
- Typecheck passed across all three TypeScript projects
- Biome passed with three pre-existing informational notices in untouched `t267`
- Package parity passed for Claude, Codex, Copilot, Cursor, Kiro, Kiro IDE and opencode
- Coverage registry and ratchet passed: 327/564 units, function coverage 156/317
- `git diff --check` passed
- Native binary was produced and used by t249. The `pathless-single-audit` release gate fails identically on candidate and pristine control.

### Live Copilot CLI canary

Validated locally on Copilot CLI 1.0.79 with GPT-5.6 Terra through Bedrock BYOK:

- `next 2>&1` returned `load-steering`
- exact `continue` reached `run-stage`
- repeated continuation was safely denied
- a deterministic installed Copilot-shell control proved the engine sequence `load-steering -> run-stage -> error` for exact token replay
- final marker retained `run-stage` with one settled continuation result

## Arden Review Follow-Up

- Leaked marker transactions replaced by owner-stamped lock recovery and doctor integration
- Contention distinguished from foreign ownership
- Inspection commands and terminal `2>&1` supported without substring blocking
- Copilot CLI correlation remains tracked; unavailable correlation safely degrades to untracked
- Direct/source/compiled foreign-project regressions added
- Shared marker reference and changelog corrected
- Atomic engine compare-and-advance added with exact same-token concurrency coverage
- Sessionless behavior pinned and all-harness replay filed as #762

## Boundaries

- Coordination state is not cryptographic protection against an arbitrary process running as the same OS user.
- Atomic rename assumes one local filesystem.
- PostToolUse is delivery evidence, not proof of final model consumption.
- Copilot CLI 1.0.79 on macOS/noninteractive is live-verified. VS Code remains documented and fixture-covered but not live-verified.
- Untracked wrappers lose hook-only delivery and Resume conveniences, while the engine cursor still owns replay for Copilot-installed projects.
- Cancel, report/park exactly-once and the remaining #740 prompt/compaction/TUI matrix stay out of scope.

## Issues

- Closes #739.
- Refs #740. Existing partial Resume/conversation/no-progress fixes remain, but #740 stays open.
- Follow-up #762 owns the all-harness replay invariant.
- #741 and #742 remain separate.

## Reviewer Focus

1. Owner-stamped marker transaction and doctor recovery in `aidlc-lib.ts` / `aidlc-utility.ts`.
2. Atomic Copilot-owned token cursor in `aidlc-orchestrate.ts`.
3. Claim/settlement, structural classifier and project isolation in the Copilot adapter.
4. Sessionless compatibility and duplicate-claim concurrency in t248/t249/t283.


apackeer:
Akash, this PR adds a Copilot-specific receipt for successful orchestration results and uses it to preserve the latest continuation or `run-stage` guidance when the shared Stop hook blocks. It also brings Copilot's `UserPromptSubmit` path in line with the shared human-turn marker contract.

**Part A - Direction: sound.** I verified the problem against `origin/v2`: Copilot's `post-tool` path ignores the successful tool result, and the shared Stop hook probes a fresh `next`, which can restart steering from part 1. The adapter is the right layer for a Copilot-session receipt, the authored/generated split is correct, and package parity passes.

**Part B - Review: request changes.**

1. `harness/copilot/hooks/aidlc-copilot-adapter.ts:232` accepts a direct orchestration command with `--project-dir <other-project>`, but `:265` stores that foreign result under the current hook project's session and state digest. I reproduced a current-project Stop prompt replaying the other project's `intent-capture` rules and token. Please parse the effective target and either require it to canonicalize to the hook project or bind the receipt to that target, with a regression test.

2. The PR says `Closes #739`, but the consumed-token half of that issue is unchanged. `core/tools/aidlc-orchestrate.ts:5327` still accepts the same valid continuation token repeatedly; on this head, two `continue T1` calls returned the same `run-stage`. Please implement the session-scoped replay denial from #739, or keep #739 open and scope this PR explicitly to the Stop-handoff fix.

3. This branch is three `v2` commits behind. Its version, README badge, and changelog are still `2.5.64`, while current `origin/v2` is `2.5.69`. Please rebase, claim `2.5.70` in all three authored surfaces, add the Copilot upgrade note, regenerate `dist/`, and rerun parity. The sync test currently passes only because the stale values agree.

4. The receipt-specific reasons at `harness/copilot/hooks/aidlc-copilot-adapter.ts:1038-1047` drop load-steering's "keep following until `run-stage`" and anti-narration instructions, plus run-stage's report/park/never-rubber-stamp guidance from `core/hooks/aidlc-continue-workflow.ts:956-975`. Please retain those safeguards in the replacement text.

5. A `run-stage` receipt remains valid across `PreCompact` because compaction does not change the state digest, but the full directive it references may no longer be in context. The resulting Stop prompt says to continue from current context and forbids `next`/`continue`, leaving no recovery path. Please invalidate that receipt on compaction, persist enough directive data, or add a safe one-time `next` fallback, with coverage.

Non-blocking notes: expired receipt files are never garbage-collected, and the PR's "9/9" receipt-test count is not reflected by the seven added test blocks; the reproducible full totals are `t249` 32/32 and `t250` 25/25.

**Part C - UX.** The intended experience is consistent: Copilot should resume the latest steering position or active stage instead of restarting the forwarding loop, and ordinary conversational turns should end cleanly. There are no new commands, flags, stages, defaults, or intended breaking changes. The replacement prompts should preserve the existing forwarding-loop and non-narration language, and compaction needs an explicit recovery path.

Verification: focused tests passed at `tests/logs/2026-08-13T01-53-05Z-p2` with `RESULT: PASS`; the independent adapter/security run also passed at `tests/logs/2026-08-13T02-03-10Z-p393982`; package parity, typecheck, Biome, and `git diff --check` passed on the current branch.

akash-doshi:
@apackeer I reset this draft onto current `v2` at `43332ab`. The prior review applies to the superseded `1576ecd` implementation. This version closes #739 and keeps #740 explicitly partial. It includes paired base/candidate evidence, a real compiled-dispatcher test and a live Copilot CLI canary. Ready for a fresh review when you are.

apackeer:
Akash, this reset replaces the earlier receipt-only patch with a current-v2 Copilot
coordination marker covering directive delivery, replay denial, compaction, Resume waits,
conversational turns, and Stop progress. The original five behavioral asks are now
implemented, but the expanded design introduces four blocking robustness/compatibility
issues and still lacks one explicitly requested regression.

**Request changes.**

1. `core/tools/aidlc-lib.ts:2467-2489` can leave `.aidlc-active-directive.json.transaction`
   as a permanent lock if the process dies after renaming the marker (user cancel mid-run,
   host hook timeout). There is no stale-lock recovery or doctor finding; every later claim
   returns recovery, and `harness/copilot/hooks/aidlc-copilot-adapter.ts:895-903` prescribes
   a fresh `next` that the same lock denies. Both reviewers independently reproduced
   consecutive fresh-`next` denials with a leaked transaction present, and a loaded test run
   showed the 100x2ms retry budget losing a concurrent Stop-count update. Please add safe
   stale-transaction recovery (mtime-TTL stealing), or at minimum a doctor finding plus
   recovery text naming the transaction file.

2. The unsupported-command classifier is overbroad.
   `harness/copilot/hooks/aidlc-copilot-adapter.ts:283-324` treats any metacharacter command
   containing `aidlc.ts`, `aidlc-orchestrate.ts`, or the bare word `aidlc` as orchestration,
   and `:885-889` denies it even outside a workflow. This blocks ordinary inspection such as
   `cat .aidlc/tools/aidlc.ts | head -50` and `printf aidlc | wc -c`, as well as legitimate
   orchestration with `2>&1`. Please restrict the deny to command shapes that plausibly
   execute the orchestrator rather than substring matches anywhere in the Bash text.

3. The claim/settle protocol needs a verified correlation contract on both supported Copilot
   surfaces. Without a tool ID, PreToolUse injects a random `--aidlc-attempt-id` through
   modifiedArgs/updatedInput (`adapter:905-910`) and settlement requires that exact ID
   (`core/tools/aidlc-lib.ts:2838-2841`). If a host supplies neither stable IDs nor the
   rewritten input, the attempt stays pending and `continue`/`report`/`park` are denied
   permanently. CLI rewrite behavior is live-verified, but
   `docs/guide/harnesses/copilot.md:101-105` says VS Code is not. Please verify that path
   live with a regression fixture, or change the protocol so it does not depend on an
   unproven host invariant.

4. The original cross-project fix still lacks its requested regression.
   `adapter:355-359` correctly rejects a foreign `--project-dir` (probe-confirmed), but the
   added matrix at `tests/unit/t249-copilot-adapter.test.ts:1512` never pins that behavior.
   Please add the direct/source cross-project claim-and-settle case.

5. Update `docs/reference/06-hooks-and-tools.md:165`. It still describes a run-stage-only
   marker that `report --single` always clears, while the shared engine now writes
   `load-steering` on every harness (`core/tools/aidlc-orchestrate.ts:250-276`) and pending
   Copilot attempts preserve the marker (`core/tools/aidlc-lib.ts:2594-2598`). The same
   update should acknowledge that shared marker consumers now see the upcoming stage during
   steering, and the changelog should say whether that cross-harness shift is intended.

**Suggested design direction, addressing 2 and 3 at the root: make token single-use an
engine invariant instead of a hook fence.** Today `handleContinue` validates a continuation
token statelessly (structural decode, anchored state hash, stage and route hash;
`core/tools/aidlc-orchestrate.ts:5698-5728`), so any stale-but-consistent token re-delivers
its part on every harness; this PR fences replay only inside the Copilot PreToolUse claim,
which is why the classifier must catch every invocation shape (the overbreadth in 2) and
why an uncorrelated host degrades into denial (the risk in 3). The PR already gives the
engine the data to own this: `emit()` now records the currently outstanding steering token
in the per-intent marker on every harness (`core/tools/aidlc-orchestrate.ts:250-276`). If
`handleContinue` refuses a presented token whose digest does not match the marker's current
`continue_token_sha256` (structured error: superseded token, run a fresh `next`), failing
open when the marker is missing or stale, then replay dies for every invocation shape,
wrapped or unclaimed, on all seven harnesses rather than only for claimed Copilot commands;
retry of the current token stays legal; the PreToolUse deny stops being load-bearing for
correctness, so it can narrow to execution-shaped commands (2) and degrade to
allow-untracked rather than deny when a host supplies neither stable tool IDs nor rewritten
input (3). The session-scoped pieces, Resume-wait gating and the settle capture behind Stop
retention, genuinely need the hook layer and would stay where they are. This would also fix
the same replay class for the other harnesses, which the current design leaves open.

The other prior asks are resolved: replayed continuation tokens are denied, the branch is
based directly on current v2, the version/changelog/badge are synchronized at 2.6.3,
retained Stop text preserves the forwarding-loop and anti-narration safeguards, and owner
compaction invalidates delivery through a context epoch.

Verification: tests/logs/2026-08-14T03-14-13Z-p1566196 reports Result: PASS for 6 files/177
assertions; package parity, typecheck, Biome, and git diff --check pass. A loaded focused
run exposed a concurrent Stop-count failure at tests/logs/2026-08-14T03-09-59Z-p2; the
isolated adapter rerun passed at tests/logs/2026-08-14T03-12-53Z-p2.

akash-doshi:
Thanks Arden. We independently reproduced the leaked transaction, contention-driven Stop release, overbroad command classifier, missing cross-project regression, and stale shared-marker documentation. We agree those need correction.

We also agree with moving continuation replay protection closer to the engine. A digest check alone still has a check-then-update race, so our proposed shape is an atomic cursor transition: compare the expected outstanding token and advance it to the successor token, or clear it at `run-stage`, in one marker transaction.

Two scope/design points where I would value your view:

1. We do not want to use mtime-only lock stealing because time is not ownership evidence. Our proposed minimum is a separate owner-stamped lock, keeping the canonical marker readable, plus a doctor finding and explicit manual recovery for an abandoned lock. Does that meet the recovery expectation, or do you see a repository convention we should follow instead?

2. The shared engine currently writes markers across all harnesses, but #739 is a Copilot issue and non-Copilot continuation is currently stateless. We propose applying the atomic token cursor only when the marker is Copilot-owned in this PR, then treating any all-harness replay invariant as an explicit follow-up. Would you prefer the wider semantic change here?

On correlation, Copilot CLI command rewriting and settlement are live-verified. GitHub documents `updatedInput` and tool identifiers for the other supported surface, but we cannot live-test VS Code. We propose a safe allow-untracked fallback when exact correlation is unavailable, with the limitation documented, rather than leaving the workflow permanently denied.

We will also narrow command classification to execution-shaped commands, support safe trailing `2>&1`, add direct/source foreign-project regressions, and correct the shared marker docs/changelog.

I will hold the branch until we align on the lock and engine-cursor scope.

apackeer:
Thanks Akash. Agreed on the atomic compare-and-advance - a digest check alone is check-then-act, and doing the compare and cursor advance inside the marker transaction is the right shape.

On the lock: there is a repository convention, and I would ask you to follow or reuse it rather than build a parallel one. The audit lock in core/tools/aidlc-lib.ts (see `DEFAULT_LOCK_STALE_MS` through `detectLeakedLocks`) is owner-stamped (pid + process start time), auto-reclaims a provably dead owner immediately regardless of age, treats only over-age live holders as wedged (10 min default, tunable), and performs the steal as a CAS rename that re-verifies the owner stamp so a live holder is never robbed. `/aidlc doctor` already surfaces leaked locks via `detectLeakedLocks`, so extend that probe rather than adding a new finding type. On manual-only recovery: the process that leaks a `.transaction` file is by definition dead, so dead-owner auto-reclaim resolves the lockout on the next claim attempt; manual-only recovery would keep orchestration bricked until someone runs doctor, which re-creates the P1 with a better error message. Owner stamp + dead-owner auto-reclaim + over-age fallback + doctor finding is the bar; your proposal minus the manual-only restriction lands exactly there.

On scope: Copilot-owned-only is fine for this PR. The discriminator already exists (engine-written markers carry a `sessionless:` owner_session and the claim path already branches on it), and the fallback-safety argument only involves Copilot-owned markers. Two conditions: file the all-harness replay invariant as a follow-up issue so it does not evaporate, and pin a test that sessionless markers keep today's stateless continue behavior so the scoping is deliberate rather than accidental.

The allow-untracked fallback for uncorrelated hosts is acceptable with the limitation documented, given the engine cursor closes the replay side.

akash-doshi:
@apackeer Addressed the current-head review in `1e6a250`, rebased onto merged #751 and moved to 2.6.6.

- Replaced the leaked fixed transaction with the repository owner-stamped/CAS recovery convention and doctor integration.
- Moved replay denial into an atomic Copilot-owned engine cursor; sessionless behavior is pinned and the all-harness follow-up is #762.
- Narrowed classification to execution-shaped commands, supports terminal `2>&1`, permits inspection, and lets ambiguous wrappers run untracked.
- Missing host correlation no longer creates permanent denial; exact host/carried IDs still settle delivery.
- Added direct/source/compiled foreign-project regressions and corrected shared marker docs/changelog.
- Added same-token concurrency, stale fresh-next, duplicate claim/Post, dead/unstamped lock and crash-boundary coverage.

Validation: final generated affected matrix 239/239; real native t249 43/43; typecheck, Biome, seven-harness parity, coverage ratchet and diff check pass. The P4 no-LLM profile has five failed files, all shared with pristine post-#751 `v2`; no candidate-only failure remains. Live Copilot CLI/BYOK reached run-stage and safely denied replay; an installed-shell control proves `load-steering -> run-stage -> error` for exact token reuse.

Ready for another pass.

apackeer:
Akash, this PR stabilizes Copilot's continuation, Stop, Resume, and conversational-turn coordination with an owner-stamped active-directive transaction and a Copilot-owned atomic continuation cursor. In `1e6a250`, the prior lock, classifier, correlation, cross-project regression, shared-marker documentation, and scope/version asks are all substantively addressed.

**Request changes.**

1. `tests/integration/t121-stop-hook-enforce.test.ts:112` now loads the Stop hook from `core/`. That removes the packaged `.claude` path that `harnessDir()` previously derived from the module location, so it falls back to probing the test runner's checkout CWD. In a Codex worktree the root `.codex` directory wins, while `makeProject()` installs its mock engine only under `.claude/tools/aidlc-orchestrate.ts` (`:295-303`). The hook cannot find the mock engine, silently fails open, and 42 of the 74 `t121` assertions fail. Please keep the packaged `.claude` hook path, run the subprocess with `cwd: proj`, or set `AIDLC_HARNESS_DIR=.claude` in `runHook` so the test is independent of which harness is running the checkout.

2. `core/tools/aidlc-orchestrate.ts:5852-5855` now turns active-directive lock contention into a hard `continue` error for every harness. `inspectCopilotContinuation` classifies non-Copilot installs plus sessionless/missing/malformed/v1 marker cases as `authority: "stateless"` (`core/tools/aidlc-lib.ts:2792-2809`), but `advanceCopilotContinuation` still acquires the marker lock before its stateless branch. A live or wedged holder therefore denies an otherwise valid Claude/Codex/Cursor/Kiro/opencode continuation even though those paths deliberately retain best-effort stateless transport and no cursor is protected. A production-path OpenCode probe returned `Continuation coordination is busy`; removing the lock and retrying the same token returned the next `load-steering` part. Please retain the hard failure for Copilot-owned cursor authority, but do not make stateless continuation depend on successful marker publication; recheck current authority as needed before falling back to the prepared directive and recording the dropped marker update.

Non-blocking follow-ups:

- `harness/copilot/hooks/aidlc-copilot-adapter.ts:283-303,443` treats `$VAR`, glob, and leading-tilde words literally while the shell expands them before the engine hashes argv. A tracked `next --scope "$SCOPE"` therefore claims one digest and executes another, repeatedly returning the misleading stale-attempt error. Classify expansion-active commands as unsupported or untracked.
- `harness/copilot/hooks/aidlc-copilot-adapter.ts:985-986` maps lock contention during claim to generic recovery text saying not to reuse the previous token. No claim committed in this case, so the correct guidance is to retry the same command/token.
- Add a `t283` case for a Copilot installation with an engine-written `sessionless:` marker. The current sessionless test uses the Claude installation, and the existing `t283` controls cover non-Copilot replacement plus missing/malformed/v1 markers rather than this exact discriminator combination.

The prior behavioral asks are resolved in the current code: the marker lock is owner-stamped and doctor-visible with dead/unstamped recovery; the engine atomically advances exact Copilot-owned tokens before stdout; sessionless behavior is deliberately scoped and #762 owns the all-harness follow-up; direct/source foreign-project regressions are present; retained Stop and compaction safeguards remain intact; and the shared-marker docs/changelog now describe the cross-harness effect.

Verification: my unpinned focused run at `tests/logs/2026-08-14T11-06-32Z-p2` reports `RESULT: FAIL` solely from `t121`; pinning `.claude` passes all 74 assertions at `tests/logs/2026-08-14T11-09-18Z-p2`. The independent 11-file slice passed 285 assertions at `tests/logs/2026-08-14T22-51-30Z-p1524686`. Seven-harness package parity, TypeScript typecheck, and `git diff --check` pass.

leandrodamascena:
I rebased this branch onto current `v2` and addressed the remaining review findings before publishing the correction.

Fixes:
- `t121` now executes the packaged Claude Stop hook with an explicit `.claude` harness and project cwd, independent of the reviewer checkout harness.
- Stateless continuation no longer hard-fails on active-directive lock contention for a stable non-Copilot installation; target intent, marker path, state identity, and harness authority are revalidated before delivery.
- Marker parsing and marker-byte hashing now come from one byte snapshot.
- Copilot-owned and `sessionless:` semantics remain unchanged; the atomic cursor still protects exact Copilot-owned tokens.
- Expansion-active shell commands (`$VAR`, globs, brace expansion, leading `~`) run untracked, while escaped and single-quoted literals remain trackable.
- Claim lock contention now says to retry the exact command because no claim committed.
- Documentation and changelog now describe these boundaries; release slot moved to `2.6.12`.

Verification:
- New regressions failed before the fixes and pass afterward.
- Affected native/adapter/cursor/lock matrix passed, including compiled dispatcher coverage.
- Typecheck, lint, seven-harness package parity, coverage registry, and `git diff --check` passed.
- Full profiles retained only known `v2` baseline failures; `t232` and `t188` resource-contention failures passed isolated.
- Two independent final reviews returned no findings.

Correction commit: `631cd6f8`.

leandrodamascena (approved):

PR #750, opened 2026-08-13
fix: refuse a gate decision the conductor wrote for the human (2.6.13)

JWThewes:
## Problem

Refs #742. That issue covers the ergonomics of the advisory-review deadlock — a stage that cannot legally complete once a `produces[]` artifact is written after the reviewer's receipt, because completion needs a fresh receipt and the advisory review budget of 1 is only reset by a `GATE_REJECTED`. This PR addresses the other half: **what an unattended conductor does to escape it.**

The human-presence guard (`humanActedSinceGate`) proves a `HUMAN_TURN` exists after the gate opened. That is human **presence**, not human **intent** — and in a live session presence is continuous. A conductor blocked by a completion precondition therefore satisfies the guard while supplying the decision text itself through `--user-input` / `--feedback` / `--details`.

Two audit trails reviewed for #742 (details and figures in [this comment](https://github.com/awslabs/aidlc-workflows/issues/742#issuecomment-5280713289)) show exactly that. Because `GATE_REJECTED` is the only event that restores an advisory review budget, a deadlocked conductor records one and says in the feedback field that it is not a human decision:

> **Event**: GATE_REJECTED
> **Feedback**: `AGENT-INITIATED, NOT A HUMAN REJECTION. … Completion is blocked by a circular guard dependency … This reopen exists solely to obtain a review budget so the required order (write, then review, then approve) can be executed.`

One session produced **8 authority-bearing rows carrying an explicit self-attestation** — 5 `GATE_REJECTED`, 2 `QUESTION_ANSWERED`, 1 `GATE_APPROVED` (`Approve - CONDUCTOR DEFAULT, session unattended`, whose own text records that no human confirmed any of its inputs). A second session, where a human answered throughout, produced **zero** — the deadlock induces the behaviour rather than it being constant.

Once such rows exist, "a human approved this stage" is no longer verifiable from the trail. Note these conductors were the *honest* case: they wrote down what they were doing, in prose nobody reads. A more compressed one writes `Request changes` and is indistinguishable from a human.

This is not harness-specific — `aidlc-state.ts`, `aidlc-log.ts`, and `aidlc-lib.ts` are byte-identical across `core/` and all seven `dist/` trees.

## Change

`selfAttributedDecisionMarker()` in `core/tools/aidlc-lib.ts`, beside the existing `isNonAnswer()` — the mirror-image floor. `isNonAnswer` catches the harness handing back a *dismissed* widget; this catches the conductor writing the decision *itself*. Enforced at the same three call sites:

| Call site | Field |
|---|---|
| `aidlc-state.ts` `handleApprove` | `--user-input` |
| `aidlc-state.ts` `handleReject` | `--feedback` |
| `aidlc-log.ts` `handleAnswer` | `--details` |

`aidlc-orchestrate report --result approved\|rejected` shells out to `state.ts approve` / `reject`, so it inherits the floor rather than needing its own. Each refusal quotes the phrase that tripped it and names the supported alternative — surface the blocker at the gate instead of deciding for the human:

```
Refusing to reject "user-stories": decision self-attribution blocked (non-human-decision)
in --feedback: "NOT A HUMAN REJECTION". This tripwire detects explicit conductor/model
provenance; it does not prove authorship. Requesting changes is the human's decision. If you
need another review pass because a produces[] artifact changed after the reviewer's receipt,
say so at the gate and let the human choose - do not record their rejection for them.
```

### Deliberately narrow

**Over-blocking a real human at a gate is the one regression worse than the forgery**, so the vocabulary matches only *self-referential provenance claims* — assertions about who authored this very decision — never assertions about the work. Detection masks quoted, inline-code, fenced-code, and blockquoted text (so a human can discuss a blocked phrase), and every noun/verb is scoped to the decision kind at the call site:

- `revert the agent-initiated retry in scheduler.ts` → **commits** (bare `agent-initiated` is not a marker; the authorship alternatives all require a decision noun)
- `The conductor default timeout of 30s is too low` → **commits** (the attributive sense; the label sense `CONDUCTOR DEFAULT,` closes its clause)
- `the unattended batch job should retry three times` → **commits**

Unlike `NON_ANSWER_RE`'s whole-string anchor these are substring matches, because the disclaimer rides inside a paragraph of otherwise-substantive prose.

**Autonomous Construction is exempt** at every call site, scoped by `isAutonomousConstructionDecision()`: the exemption applies only while Construction Autonomy Mode is active and the decision belongs to a Construction-phase stage; Ideation, Inception, and Operation decisions stay guarded even when the mode field is present, and the consolidated summary confirmation stays human-backed. Where the conductor legitimately holds the decision, honest self-labelling must stay legal — otherwise the floor would push autonomous runs toward *unlabelled* decisions, which is strictly worse for the trail. The floor also rides the same `AIDLC_SKIP_HUMAN_PRESENCE_GUARD` off-switch as the sibling authority guards.

A blocked decision leaves **no partial receipt** (neither `GATE_REJECTED` nor `STAGE_REVISING`) and **does not consume the human's turn**, so their real choice still commits on the same turn.

### Scope note

This does not fix the deadlock itself — #742's refresh action is still worth doing, and the two are complementary. The refresh action removes the *motive* to fabricate; this removes the *possibility*. Filed as a fix rather than a feature because the audit-integrity hole stands on its own: it is reachable from any completion precondition a conductor cannot clear, not just the advisory-review one.

## Testing

The floor lives in `tests/unit/t261-audit-authority-floor.test.ts` (the existing home of the `isNonAnswer` floor, extended to guarantee 5). The forgery cases were verified as real regression tests against a pristine `v2` checkout (they fail without the fix, pass with it), and the must-not-over-block side is pinned by a 26-entry substantive-prose list that includes every false-positive repro from both review rounds:

- `reject --feedback`, `approve --user-input`, and `answer --details` each refuse self-attributed prose (verbatim field strings), asserting no `GATE_REJECTED` / `GATE_APPROVED` / `QUESTION_ANSWERED` and no `STAGE_REVISING` land
- substantive human prose still commits (the over-block guard)
- a refusal does not consume the turn — the human's own `Approve` still commits
- autonomous Construction is exempt; non-Construction phases are guarded even with the autonomy field present
- the floor rides the standard off-switch

Also run:

- `bun scripts/package.ts` + `bun scripts/package.ts --check` — no drift, all 7 harness trees regenerated and committed
- `bun run typecheck` — clean
- `bun run lint` — exit 0 (3 pre-existing infos in `t267-usage.test.ts`, untouched here)
- `bun tests/gen-coverage-registry.ts` — regenerated; on the current rebase the `function` ratchet moves **156 to 158** (both new exported predicates are covered)

## Notes for reviewers

- Based on `v2`, matching where recent work has been merging. `CONTRIBUTING.md` still says "Work against the latest `main` branch" — happy to retarget if that is wrong.
- The marker vocabulary is the tuning surface. It is derived from two sessions, so it is evidence-based but not exhaustive; it is one exported predicate specifically so the list can be extended in one place as new phrasings show up. If you would rather this fail *open* on unknown phrasings and only warn, say so — I chose to refuse because a fabricated `GATE_APPROVED` is silent and permanent, while a false refusal costs the human one rephrase and is self-explaining.

## Post-approval updates

- Round-2 hardening: the predicate was renamed `selfAttributedDecisionMarker()` and gained quote/code/blockquote masking plus kind-scoped decision nouns; the autonomy exemption was scoped to Construction via `isAutonomousConstructionDecision()`; t261 was extended accordingly (every false-positive repro from both reviews passes).
- Rebased onto v2 after #758 and #749 merged; version slot re-taken as **2.6.13**. The #758 recovery guard and this PR's scoping compose in `handleReject`: the presence guard now fires unless the decision is an autonomous Construction one, and a spent stale-receipt recovery review requires a human even then.
- Post-rebase verification: `package.ts --check` green across all 7 trees, `typecheck` clean, and a filtered slice over this PR's tests plus the test files #758 and #749 shipped (t68, t121, t163, t188, t205, t249, t250, t261, t263, t271, t272, t278, t283, t291): 18 files PASS, 434 assertions, 0 failures.


leandrodamascena (changes requested):
Jan, the underlying issue is real: HUMAN_TURN proves human presence, not that the human authored the approval, rejection, or answer. I reviewed 5ec63090, inspected all three enforcement paths, ran the official t261 suite (33/33 passing), and probed the predicate directly with both bypass and false-positive cases.

I agree this should remain changes requested. I independently verified the following blockers:

P1: The guard detects selected confessions, not conductor-authored decisions

fabricatedDecisionMarker() at core/tools/aidlc-lib.ts:2647-2674 operates only on the submitted prose. It can disprove human authorship when the conductor explicitly uses one of the recognized phrases, but it cannot prove that the decision came from the human.

The reported phrases are blocked as intended:

- AGENT-INITIATED, NOT A HUMAN REJECTION
- Approve - CONDUCTOR DEFAULT, session unattended

However, these conductor-authored decisions returned no marker and would pass the new floor:

- Approve
- Request changes: reopen for another review
- Not human rejection; reopen it
- This rejection was generated by the AI
- I, the conductor, am rejecting this
- The assistant chose Request changes

The same predicate is used directly by approve, reject, and answer:

- core/tools/aidlc-state.ts:2465-2468
- core/tools/aidlc-state.ts:2715-2718
- core/tools/aidlc-log.ts:423-426

Therefore omitting or slightly varying the self-attribution still allows the conductor to mint the same authority-bearing receipts. The current tests prove the known vocabulary, not the stated security property.

Please bind the decision to trusted harness-captured human input for the specific gate/question, ideally with a single-use gate or prompt identifier. Conductor-supplied prose should not be treated as proof of human authorship.

P1: The "Autonomous Construction" exemption applies outside Construction

All three call sites bypass the new guard whenever isAutonomousMode(content) returns true. That function checks only the persistent Construction Autonomy Mode field:

- core/tools/aidlc-lib.ts:4641-4643

It does not verify that the target stage is in Construction.

I directly probed states carrying Construction Autonomy Mode: autonomous and obtained:

- Ideation / feasibility: true
- Operation / deployment-execution: true

The new test explicitly codifies the Ideation bypass by setting autonomous mode on feasibility and expecting a conductor-authored rejection to commit:

- tests/unit/t261-audit-authority-floor.test.ts:586-605

This means a stale or injected Construction autonomy field disables the authorship floor for human-gated Ideation and Operation decisions.

Please scope the exemption to an actual Construction-stage autonomous directive/authority, not merely the persisted field.

P2: The substring heuristic blocks legitimate human feedback

The PR states that the vocabulary matches only self-referential provenance claims, but direct probes produced markers for legitimate statements about unrelated subjects:

- This is not a human response time issue
- never confirmed by the human resources team
- report written by the conductor service
- The session was unattended after the websocket dropped

These match because the expressions lack sufficient semantic or clause boundaries. A real human can therefore be prevented from approving, rejecting, or answering based on ordinary feedback text.

If the text heuristic is retained as defense in depth, it should be advisory/auditable rather than treated as authorization evidence. At minimum, add regression cases covering quoted phrases, third-party referents, "human resources," service names, and unattended system behavior.

The focused tests, package parity, and coverage registry are green, but they validate the known phrases rather than the full authorship guarantee. The vulnerability is valid; this implementation is an incomplete and bypassable mitigation.

apackeer:
Jan, thanks for this - and thanks @leandrodamascena for the thorough verification round. I re-ran the key probes independently and they reproduce exactly: the six unlabelled phrasings all pass the predicate, and the false-positive list is real. So changes requested stands. Where I land differently is on what follows from that.

The mechanism is worth keeping; the claim is what needs to change. `isNonAnswer()` at these same call sites is the same class of calibrated, incomplete text heuristic - equally bypassable by construction - and it earns its keep because it is aimed at an observed, mechanical failure mode, not at adversaries. Your floor is aimed the same way: both field sessions were the honest case, conductors that labelled what they did. For that population the refusal does two things worth having: it stops the labelled forgery from minting an authority-bearing row, and it does not lose the signal, because `error()` in both tools routes through `emitError()` and lands an ERROR_LOGGED row carrying the command and refusal text (aidlc-state.ts:4027, aidlc-log.ts:918). Attempt recorded, receipt withheld, conductor redirected to the gate. A silent conductor walks past it, but a silent conductor walks past any text check; that is an argument about the wording of the guarantee, not against the tripwire.

One piece of context that changed since this PR opened: #758 now adds the bounded recovery review pass from #742's remediation - the deadlock your field sessions were escaping no longer requires a false Request Changes once that lands. The two PRs compose: #758 removes the motive to fabricate, this one removes the labelled path and makes attempts auditable. It also means this PR should sequence after #758 - merged alone, it tightens the trap on honest conductors while surfacing the blocker at the gate still leads nowhere. Since v2 has moved to 2.6.2 you need a rebase and a fresh version slot anyway, which lines the sequencing up naturally.

@leandrodamascena - proposed merge order for this set: **#758 first** (CI green and mergeable, functional prerequisite for this PR, and it claims the next slot 2.6.3 so nothing above it re-bumps; it only lacks a review - if you can take that one, it unblocks the whole sequence), **#749 second** once Akash clears the two round-3 blockers (independent mechanism, already slotted at 2.6.6), **this PR last** after the rework below, taking a fresh version slot on rebase.

Concretely, to get this merged:

1. Re-scope the claim. The CHANGELOG sentence ("The gate no longer accepts a decision the conductor wrote for the human") and the PR's "removes the possibility" line promise more than prose inspection can deliver. Say what it verifiably does: refuse self-attributed conductor decisions, defense in depth beside isNonAnswer, attempts preserved via ERROR_LOGGED, deliberately fail-open on unlabelled text.

2. Tighten the vocabulary, and add every repro from both reviews to the SUBSTANTIVE block: a left word boundary on the actor alternation ("Mumbai initiated rejection..." currently trips as "ai initiated rejection"); hyphenated compound only for conductor-recorded ("the conductor recorded the wrong stage in the audit trail" is a human describing conductor behavior); decision verbs or a decision-noun requirement for "by the conductor" ("this ADR was authored by the conductor" is an assertion about the work, which your own design comment promises must pass); decision-noun context for "not a human ..." (Leandro's "response time" / "human resources" cases); narrow or drop the bare session/run unattended alternatives. Your PR names over-blocking a real human as the regression worse than the forgery - these repros are that regression.

3. Scope the exemption. Both aidlc-state.ts call sites already hold the resolved stage node, so the exemption can require the target stage's phase to be construction in addition to the field; the answer path can resolve its stage the same way. The unscoped isAutonomousMode() is shared with the pre-existing presence guard, so the deeper field-lifecycle cleanup belongs in a separate issue covering both guards - I would not absorb it here.

4. Rebase onto current v2 (2.6.2 at 4569754e) and take a fresh version slot; the 2.5.76 trio will conflict as-is.

5. Nits: the t261 header still says "Four related guarantees" with five listed; and please keep issue numbers out of code comments (describe the behavior; refs live in the commit/PR).

On the redirect that has been floated: binding decisions to trusted harness-captured input with single-use gate-scoped receipts is the right end state, and the seam exists (the HUMAN_TURN mint on UserPromptSubmit deliberately reads no prompt content today). But it is an RFC, not a change request this PR can absorb: content capture is per-harness work, Copilot has no trusted prompt event, and widget answers transit no trusted channel on any harness - which is exactly why the presence guard is presence-only. We will track that separately. With #758 covering the deadlock and this PR re-scoped to what it verifiably does, the audit-integrity hole gets a real floor without waiting on the receipt RFC.

leandrodamascena:
I addressed the findings in this PR:

- Hardened detection of agent/conductor self-attributed decisions.
- Reduced false positives across quotes, code blocks, compounds, and punctuation.
- Scoped the autonomous exemption to Construction only.
- Kept summary confirmations human-backed.
- Updated documentation, changelog, tests, and generated distributions.
- Validated with typecheck, lint, parity, coverage, and independent reviews.

Commit: `0e9ea8ea`.

leandrodamascena (approved):

apackeer (approved):
Round 2, re-reviewed at 0e9ea8ea5. All four substantive asks from my round-1 comment are verifiably fixed, so I am approving.

What I re-verified independently:

1. **Claim re-scoped.** The 2.6.10 CHANGELOG entry now says the tripwire "does not prove human authorship and deliberately fails open for unlabelled or unrecognized wording"; every refusal string carries "it does not prove authorship"; audit-format.md and the three docs pages are rewritten to presence-not-authorship, with the off-switch documented. This is wording the mechanism can actually stand behind.
2. **Vocabulary.** `selfAttributedDecisionMarker` with the quote/code/blockquote masking and kind-scoped decision nouns clears every false positive from both round-1 reviews. I probed the predicate directly, including the two repros that did not make the SUBSTANTIVE list ("the conductor recorded the wrong stage in the audit trail", "this ADR was authored by the conductor"): both pass now. The 26-entry SUBSTANTIVE block pins the rest verbatim, and the masking is length-preserving so the refusal quotes the real tripped phrase.
3. **Exemption scoping.** `isAutonomousConstructionDecision` gates the floor at all three call sites, and the same scoping now applies to the pre-existing presence guards and the revision backstop: a persisted autonomy field no longer waives Ideation, Inception, or Operation gates. That goes further than I asked and is the right call; it is a user-visible tightening, CHANGELOG bullet 3 covers it, and t188 scenario D plus t205 scenario 6 were flipped to pin it. Summary confirmation staying human-backed with no autonomy carve-out is also correct.
4. **Rebase and slot.** Merge-base is the current v2 tip (4f8f1824e) and the 2.6.10 trio (version file, badge, heading) is consistent. On my side: `bun scripts/package.ts --check` clean across all seven trees, typecheck clean, and the touched suites (t68, t188, t205, t261) all green, 107 assertions.

Two asks before the merge button, both zero-code:

- **Commit author.** 0e9ea8ea5 is authored `test <test@test.com>`. A squash merge will embed that as a `Co-authored-by` trailer in v2 history; please amend the author or correct the trailer in the squash message so the actual author is credited.
- **Title.** Still "(2.5.76)". The squash subject comes from the title, so retitle to "(2.6.10)". A body refresh would help too: it still names `fabricatedDecisionMarker()`, quotes the pre-hardening refusal text, and predates the phase scoping.

Notes, none blocking:

- The residual false positives sit exactly where the design says they would: "The CI agent rejected this because the tests are flaky - fix the pipeline" trips the rejection-only "agent rejected this" case, and "Escalation level is not a human decision. Encode it in the runbook." trips the decision-noun family (same shape as "not a human choice" on approvals). Both are prose about the system being designed, which is plausible gate feedback in this methodology. The refusal is self-explaining and costs one rephrase, so this is tuning-surface material for the follow-up pass, not a blocker; worth folding both repros into the SUBSTANTIVE list as either accepted misses or fixed cases.
- The t261 header still opens "Four related guarantees" above a list of five.
- 2.6.10 skips 2.6.9, which open PRs currently claim; whoever merges after this one re-bumps as usual. The round-1 sequencing point stands softly: #758 removes the motive that this floor removes the labelled path for, so landing it close behind keeps honest unattended runs from waiting at gates only a human can notice.
- Jan, since the rework round was pushed to your branch: worth a quick confirm that it matches your intent before the squash.

Thanks both. The tripwire as now scoped, documented, and pinned is exactly the floor round 1 asked for.

PR #753, opened 2026-08-13
feat: import v2 evaluator into evaluator/ directory

### LINKED ISSUE #933: [Feature]: AIDLC Evaluator for 2.0
### Description

# RFC: The evaluator's home and shape

PR #753 imports the AIDLC v2 evaluator (405 files, a Python/uv workspace)
into this repo as a top-level `evaluator/` directory. Review raised a
direction question that deserves a deliberate decision rather than
inheritance-by-import: where the evaluator lives, what comparison axes it
must serve, what cost/speed tiers it needs, and what happens to the parts
its own STATUS.md marks legacy or stale. This RFC states what the evaluator
is designed to do today, then lays out the options for each of those four
decisions. Run evidence is on the table: ~12 real Bedrock-billed runs with
published baselines (`evaluator/results/`).

### Use Case

The evaluator is a **maintainer instrument for measuring version-to-version
impact against fixed baselines**. It answers one question with statistical
rigor: *did this change to the framework move the numbers, and by how much?*
It drives the real CLI agents (no mocks, no simulated transcripts) through
the actual `/aidlc` workflow and scores the results against executable
oracles.

### The fixed task set

Three standard tasks, two measurement families kept deliberately separate:

| Task | Family | Oracle |
| --- | --- | --- |
| Greenfield (`sci-calc-v2`): build a scientific-calculator API from a vision doc | **Graded** — cost, tokens, wall clock, doc fidelity vs a golden doc set | 88-case OpenAPI contract run against the live generated server + 0–1 LLM-judged semantic similarity per AIDLC document |
| Brownfield bugfix (`psf/httpbin`, commit-pinned): fix `/base64` error handling | **Deterministic pass/fail** — no partial credit | 3 explicit HTTP assertions |
| Brownfield feature (`realworld-django-ninja`, commit-pinned): implement follow/unfollow | **Deterministic pass/fail** — no partial credit | The RealWorld project's own Hurl contract suite |

Greenfield measures *quality and consumption*; brownfield measures
*reliability*. The combined report keeps the families in separate sections
so the axes never blend. Brownfield testbeds are frozen snapshots pinned to
upstream commits and cross-checked at preflight — a mismatch aborts the run.

### The measurement pipeline

Every run flows through the same scoring stages regardless of which agent
produced the code: post-run pytest, static analysis (ruff/bandit/semgrep),
Docker-sandbox contract execution (framework auto-detected — FastAPI /
Flask / Django — migrations and env handled), LLM qualitative judge with
per-document rationales, and per-run Markdown/HTML reports. Pass/fail is
read from the contract results, never from the workflow's own "done" claim.
Token/cost capture for Claude runs is self-contained: a local in-process
OTLP receiver reads Claude Code's own telemetry — no collector, CloudWatch,
or extra AWS permissions.

Multi-version reports get real statistics: mean ± SD and 95% CI on the
consumption axes, Mann-Whitney U tests across versions, Fisher's exact
tests on brownfield pass rates.

### The comparison axes it supports today

- **Version × task** (primary, by design): `config/versions.yaml` entries
  drive batches; the combined report grows a column per version.
- **CLI × task** (proven): `run_cli_evaluation.py` drives claude-cli,
  codex-cli, and kiro-cli through their real PTYs; the published three-way
  comparison (`results/CLI-COMPARISON-20260812.md`) came from this surface.
- **Scope** exists as a per-run flag (`--scope`), and a `versions.yaml`
  entry is effectively a (dist, scope, scenario) tuple — the primitives for
  configuration-level comparison exist, but nothing makes a knob sweep
  first-class: the batch driver is claude-cli-only and the reporting is
  version-framed. That gap is the substance of Part 2, question 2.

### What a run costs

~$45 per greenfield run (Bedrock, Opus-tier), $12–21 per brownfield run;
1.5–5 h greenfield, 25–50 min brownfield. The full n=5 suite across all
three tasks is **$400–500 and runs overnight**. This is inherently not a
CI workload as designed — it is a release-cut and investigation instrument.

### Provenance and current condition

Extracted from the comparative evaluation harness built during AIDLC 1.0
and trimmed to a v2-only packaging; task definitions, oracles, PTY driving,
sandbox boot, and reporting are identical to what produced the published
1.0-to-2.0 comparison numbers. `evaluator/STATUS.md` is an evidence-based
inventory of every entry point. Summary:

- **Release-grade** (proven by real billed runs): the CLI evaluation
  surface — `run_v2_suite.sh`, `run_cli_evaluation.py` for all three CLIs,
  the batch/report scripts, the full scoring pipeline. Unit suites green
  (`run.py test`: cli-harness 72/72, all packages passing).
- **Caveated**: `run_version_batch.py` is claude-cli-only; token/cost
  capture is claude-only (codex/kiro report turn counts); dist auto-detect
  assumes the evaluator sits inside this repo; two report scripts are
  fixture-tested but not yet run on real multi-version data.
- **Broken/untested**: `run.py trend` (dispatches to a removed script);
  the legacy Strands swarm path (`run.py full`/`batch`); the IDE harness
  (zero end-to-end runs); five IDE adapters with stale pre-2.x contracts;
  two utility scripts; stale sections in `ARCHITECTURE.md` and
  `scripts/README.md`.

### Version

v2 (alpha)

### Area

Platform support

### Question 1: Where does the evaluator live?

The tension: the evaluator is a Python/uv workspace (188 first-party Python
files across 10 packages, a 2,474-line `uv.lock`, pytest/ruff/bandit
tooling) landing in a repo that is TypeScript-on-bun end to end. As
imported in #753, nothing in CI touches `evaluator/` — test discovery,
typecheck, lint, and the packager are all path-bound away from it — so the
code would land unguarded. The in-repo benefit is that dist auto-detect
works naturally and a framework change plus its baseline evidence can land
in one PR; the out-of-repo alternative costs one `--claude-dist` flag plus
cross-repo coordination.

**Option 1A — Import as-is (PR #753 unchanged).** Land `evaluator/` with
zero CI wiring; billed runs and unit tests both stay manual.

- For: zero delay; the release-grade surface ships now; simplest diff.
- Against: ~22k lines permanently unguarded; the rot already visible in the
  import (stale `scripts/README.md`, broken `trend` mode) has no mechanism
  to stop compounding.

**Option 1B — Import + guard the unbilled surface.** Land `evaluator/`
plus: a CI job that runs the evaluator's own unit suite (`run.py test` —
green, fast, no Docker, no billing) on PRs that touch `evaluator/`, and the
~1 hour of STATUS.md pre-release cleanup (remove `trend`, archive dead IDE
adapters, refresh stale docs) folded into the PR or an immediate follow-up.

- For: addresses the "permanently unguarded" objection at its cheapest
  point — the unit tier is already proven green; keeps the one-PR
  framework+baseline property; small incremental cost (one workflow file,
  one cleanup pass).
- Against: the repo still carries two toolchains (uv becomes a CI
  dependency for one path); guards only the unit tier — the billed
  integration surface remains manually verified by nature.

**Option 1C — Sibling repo (`awslabs/aidlc-evaluator`) pinned to released
dists.** The evaluator lives in its own repo with its own Python-native CI,
consuming this repo's released `dist/` trees.

- For: this repo stays single-toolchain; the evaluator gets CI shaped for
  Python; clean release boundary (evaluate what shipped, not what's on a
  branch).
- Against: version skew becomes a coordination problem (which evaluator
  commit produced which baseline for which framework release); dist paths
  go back to explicit flags; evaluating *unreleased* changes — the main
  maintainer use case, e.g. validating a scope rework before merge — now
  spans two repos; discoverability drops.

**Option 1D — TypeScript-native port.** Rewrite the harness in TS/bun so it
shares this repo's toolchain and test tiers.

- For: one toolchain, one test runner; the evaluator's unit tier joins the
  existing smoke/unit/integration levels naturally.
- Against: the rewrite is large and risky — PTY driving of three CLIs,
  Docker sandbox orchestration, OTLP receipt, statistical tests
  (Mann-Whitney, Fisher's exact), LLM judging — and every published
  baseline would need re-validation under the new harness before its
  numbers are comparable. Months of delay for a tool that already works;
  plausible as a *later* migration, expensive as a precondition.

Options 1B and 1C are the realistic contenders; 1A and 1D bracket them.
Note 1B and 1C are not mutually exclusive over time — 1B now does not
foreclose extraction later once the evaluator's shape stabilizes.

### Question 2: What comparison axes must it support, and what do reports look like?

Two views of the evaluator's job, from the #753 discussion:

**Option 2A — Regression instrument (the current design, knobs as
roadmap).** The primary axis is *framework version*, measured on the fixed
task set, with CLI as a secondary proven axis. This gives maintainers the
ability to clearly attribute impact-on-baseline to each release for the
configuration most users actually run. Knob-level comparison (scope sets,
stage configuration, dist variants) is acknowledged as valuable and
deferred to the roadmap; the existing `--scope` flag and the
(dist, scope, scenario) shape of a `versions.yaml` entry are the seam it
would grow from.

- For: matches what's built and proven; no rework blocks the import; the
  version axis is what release notes and users need first.
- Against: framework-development questions ("did the #767 scope rework
  help?") remain answerable only by hand-assembling runs; the reporting
  vocabulary ("version") bakes in the wrong noun if knobs become primary.

**Option 2B — Experimentation instrument (make configuration the
first-class axis).** Generalize the comparison unit from "version" to
"labeled configuration" — a named (dist, scope set, knobs) tuple — so the
same machinery answers both "v2.3 vs v2.4" and "scopes-A vs scopes-B on
v2.4". Requires: generalizing `run_version_batch.py` beyond claude-cli,
reframing report vocabulary from versions to configurations, and a
matrix-style report that attributes deltas to the knob that changed.

- For: turns the evaluator into a framework-development instrument, not
  just a release-notes instrument; the scope rework in #767 is a live,
  non-hypothetical customer for exactly this.
- Against: real design and implementation work *before* anything lands, on
  top of an already-working tool; knob sweeps multiply run counts, which
  makes this question inseparable from Question 3 (cost tiers).

**Option 2C — Land 2A, spec 2B as the named next milestone.** Import the
regression instrument as-is; in the same RFC, agree on the target shape of
the configuration axis (the "labeled configuration" unit and the report it
produces) so incremental PRs (de-hardcode the batch driver, rename the
comparison unit, add the matrix report) have an agreed destination.

- For: decouples "where it lives" from "what it becomes"; the roadmap is
  written down and reviewable instead of implied.
- Against: roadmap sections in RFCs go stale unless a milestone/issue is
  cut alongside; requires discipline to not let 2B silently become "never".

### Question 3: What cost/speed tiers does evaluation need?

Today there is effectively one tier: full-fidelity, billed, hours-long.
That is correct for release cuts and wrong for everything else. Candidate
tier ladder (each tier is a decision — which exist, and what triggers
each):

| Tier | What runs | Cost / time | When |
| --- | --- | --- | --- |
| **T0 — Harness self-test** | `run.py test` (unit suites, fixtures; no agents, no Docker, no billing) | Free, ~minutes | CI on every PR touching `evaluator/` (this is Option 1B's guard) |
| **T1 — Smoke** | One brownfield bugfix run, one CLI, n=1 (deterministic pass/fail) | ~$12–21, ~25–50 min | On demand: "did this framework change break the workflow end to end?" |
| **T2 — Directional** | All three tasks, n=1, one CLI | ~$70–90, ~3–6 h | Investigating a specific change (e.g. a scope rework) before investing in a full baseline |
| **T3 — Baseline** | Full suite, n=5, per CLI as needed | ~$400–500/CLI, overnight | Release cuts; published numbers in `results/` |

Open sub-questions: (a) whether a T1/T2 run on a cheaper model tier is
meaningful or just noise (the baseline is Opus-tier; cross-tier numbers are
not comparable — a cheaper tier may still be useful as a *did-it-break*
signal rather than a *how-good* signal); (b) whether any billed tier should
ever be CI-triggered (e.g. label-gated on release-prep PRs) or all billed
tiers stay human-invoked; (c) whether knob sweeps (Question 2B) get their
own tier definition, since a k-knob sweep at T2 prices is $70–90 × k.

### Question 4: What happens to the legacy/stale parts?

STATUS.md already enumerates them; the decision is what the import's
hygiene bar is.

**Option 4A — Ship as documented.** Land everything as-is; STATUS.md is the
honest map of what's proven vs not.

- For: zero delay; STATUS.md genuinely does disclose everything.
- Against: `run.py trend` is *broken at the entry point* — shipping a
  top-level mode that errors is below the bar regardless of disclosure;
  stale docs actively misdirect (`scripts/README.md` documents four scripts
  that no longer exist).

**Option 4B — The STATUS.md pre-release cleanup (~1 hour) before/with
merge.** Remove the `trend` mode (the underlying `trend_reports` package
stays), archive the five stale IDE adapters (registry keeps `kiro` only),
mark the Strands `full`/`batch` path experimental or remove it from packaged
entry points, and align `ARCHITECTURE.md`/`scripts/README.md` with the
shipped surface. Optionally +$15/25 min: one kiro IDE smoke run to move the
IDE harness from "untested" to "smoke-tested".

- For: cheap, already scoped by STATUS.md itself, removes every
  known-broken user-facing surface.
- Against: none substantive — the only question is whether it blocks the
  import PR or lands as an immediate follow-up.

**Option 4C — Trim harder: delete rather than archive.** Drop the IDE
harness, the five stale adapters, and the Strands path from the import
entirely; git history and the source repo preserve them.

- For: the import shrinks meaningfully; nothing lands that isn't
  release-grade or on an active path.
- Against: the kiro IDE adapter was just modernized and is one $15 smoke
  run from useful; claude/codex IDE evaluation is a plausible near-term
  need (both ship official VS Code extensions); deleting the seam makes
  that future work start from archaeology.

## Decision summary

| # | Decision | Options |
| --- | --- | --- |
| 1 | Home | 1A import as-is · 1B import + unit-tier CI guard + cleanup · 1C sibling repo pinned to releases · 1D TS-native port |
| 2 | Comparison axes | 2A version-primary, knobs as roadmap · 2B configuration-first now · 2C land 2A with 2B spec'd as the named milestone |
| 3 | Cost tiers | Which of T0–T3 exist; cheap-model-tier validity; CI-triggerable billed tiers or human-only; sweep pricing |
| 4 | Legacy parts | 4A ship as documented · 4B STATUS.md cleanup with merge · 4C delete rather than archive |

## References

- PR #753 — the import, with the release-readiness summary in its description
- `evaluator/STATUS.md` — evidence-based per-entry-point inventory
- `evaluator/README.md` — task set, oracles, costs, provenance
- `evaluator/results/V2-EVALUATION-20260807.md` — claude baseline
- `evaluator/results/CLI-COMPARISON-20260812.md` — three-way CLI comparison
- #767 — scope rework; a live candidate for configuration-vs-configuration evaluation

harmjeff:
## Description of changes

Imports the AIDLC v2 evaluator (previously developed in an external repository) into this repo as a top-level `evaluator/` directory — 405 files brought in as a single clean snapshot commit of the external repo's tracked files.

The evaluator provides a harness for evaluating AIDLC v2 workflows across CLI agents (Claude Code, Kiro CLI, Codex CLI) and IDE agents, including:

- `packages/cli-harness` — adapters for driving each CLI/IDE agent
- `config/` — per-model evaluation configs
- `docker/sandbox` — sandboxed execution environment
- `test_cases/` and `results/` — evaluation scenarios and baseline reports
- `ARCHITECTURE.md`, `STATUS.md`, `FAQ.md` — design and release-readiness docs

Note: this supersedes the earlier evaluator prototype on the `v2-evaluator` branch, which has a different structure. This version is targeted specifically at the v2 workflows, hence the `v2` base branch.

## Release readiness and known limitations

Full evidence-based inventory in [`evaluator/STATUS.md`](https://github.com/awslabs/aidlc-workflows/blob/906e482d1dc70aa8d0da42b0879e7679cfe3db87/evaluator/STATUS.md). Summary:

**Release-grade (proven by ~12 real Bedrock-billed runs):** the CLI evaluation surface — `scripts/run_v2_suite.sh`, `run_cli_evaluation.py` (claude-cli, codex-cli, kiro-cli), the version-batch/report scripts, and the full scoring pipeline (pytest, static analysis, Docker-sandbox contract runs, LLM judge, reports). Published baselines live in `evaluator/results/`.

**Works, with caveats:**
- `run_version_batch.py` is claude-cli-only (hardcoded); cross-CLI runs use `run_cli_evaluation.py` directly.
- Token/cost capture is claude-only (local OTLP receiver reads Claude Code telemetry); codex/kiro runs report turn counts, not tokens/cost.
- Dist auto-detect assumes the evaluator sits inside the AIDLC repo — which this PR makes true naturally; standalone checkouts must pass `--claude-dist` / `--kiro-dist` / `--codex-dist` explicitly.
- `run_version_report.py` and `run_comparison_report.py` are fixture-tested only, not yet exercised on real multi-version data in this packaging.

**Broken / untested (excluded from release claims):**
- `run.py trend` is broken — it dispatches to a script removed in the v2-only trim (the underlying `trend_reports` package works).
- `run.py full` / `run.py batch` (legacy Strands swarm path) are untested in this packaging and not advertised in the README.
- The IDE harness (`run.py ide`) has zero end-to-end runs; the kiro IDE adapter is modernized but unproven. IDE adapters for cursor/cline/copilot/windsurf/antigravity have stale pre-2.x contracts and are slated for archival; claude/codex IDE adapters do not exist yet.
- `build_golden_reference.py` and `regenerate_single_report.py` are untested in this packaging.
- `ARCHITECTURE.md` and `scripts/README.md` are stale in places (reference removed modes and the parent framework's architecture).

STATUS.md also lists recommended pre-release cleanup (~1 hour): remove/fix the `trend` mode, archive the dead IDE adapters, and refresh the stale docs.

Issue #, if available: N/A


apackeer:
Stepping back from file-level findings to a direction question that's easier to raise before merge than after: I don't think this should land as a direct import, and I'd like to propose we decide the evaluator's home and shape via an RFC, the way we've handled other structural decisions on v2.

Two things drive that.

**First, the placement cost is real.** This adds a full second toolchain, not a script: 188 first-party Python files across a 10-package uv workspace, a 2,474-line `uv.lock` of PyPI dependencies, maintained with uv/pytest/ruff/bandit, in a repo that is TypeScript-on-bun end to end (tools, hooks, packager, the entire test suite). Nothing in CI touches `evaluator/` - test discovery, typecheck, lint, and the packager are all path-bound away from it, and no workflow runs its pytest suite - so ~22k lines would land permanently unguarded. The rot is already visible inside the import itself: `scripts/README.md` documents four scripts that no longer exist and examples that argparse-error. Meanwhile the benefit of living in-repo is small by STATUS.md's own accounting: it makes dist auto-detect "work naturally", which is otherwise one `--claude-dist` flag.

**Second, and more importantly, we haven't agreed on what the evaluator is for.** As designed, it answers "how does v2 score against baselines" - a marketing/regression instrument with three fixed tasks and a version axis. What I most want from an evaluator is an experimentation instrument for the framework itself: run the same task matrix across different scope sets, compare the effect of individual knobs (scope selection, stage configuration, dist variants) on output quality, cost, and wall clock, and get reports that attribute differences to the knob that changed. The current code is tantalizingly close - `--scope` exists per-run, and a `versions.yaml` entry is already effectively a (dist, scope, scenario) tuple - but the batch driver is claude-cli-only, the reporting is version-framed, and nothing makes a knob sweep first-class. At ~$45 per greenfield run, knob sweeps also need a deliberately cheaper tier, which is a design decision, not a patch.

An RFC would let us settle, with the run evidence on the table: (1) where the evaluator lives (this repo, a sibling `awslabs/aidlc-evaluator` pinned to released dists, or a TS-native port so it shares this repo's toolchain); (2) the primary comparison axes it must support (versions, CLIs, scopes, knobs) and what reports look like; (3) the cost/speed tiers for CI-able vs billed evaluation; (4) what happens to the parts STATUS.md already marks legacy or stale. Recent work like the scope rework in #767 is exactly the kind of change that should be validated by scope-vs-scope evaluation runs, so the requirements question is live, not hypothetical.

None of this is a knock on the work - the harness is solid, the run evidence is real, and the STATUS.md honesty is exemplary. I'd just rather we pick its home and its target use deliberately, once, than inherit both from an import.

harmjeff:
I understand your point about "we haven't agreed on what the evaluator is for", but I disagree.  It is designed as a tool for maintainers to evaluate on a fixed set of axes the impact on baseline from version to version.  The knob hooks seem like roadmap.  This gives us the ability to clearly assign impact for most users as we apply changes.  I will open an RFC based on your comments.

harmjeff:
Issue 933
https://github.com/awslabs/aidlc-workflows/issues/933

PR #733, opened 2026-08-10
docs: update the roadmap

leandrodamascena:
Update the roadmap.

Closes #732 

## Changes

> replace with a description of the changes

## User experience

> Please share what the user experience looks like before and after this change

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [ ] I have reviewed the [contributing guidelines](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md)
* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented

## Test Plan

> replace with instructions or a checklist for reviewers to verify

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/aidlc-workflows/blob/main/LICENSE).

PR #693, opened 2026-07-31
docs: scope the Claude-only surfaces and add per-harness install dropdowns

apackeer:
Closes #683

## Problem

The Getting Started and Your First Workflow chapters showed the Claude Code experience without saying so. The `companyAnnouncements` welcome banner and the statusline exist only in the Claude distribution (`harness/claude/settings.json` - no other harness manifest ships a `settings.json`, and the Kiro/Codex/opencode chapters and shipped `AGENTS.md` files all document "no welcome message" / "no statusline"). A reader on Kiro CLI, Kiro IDE, Codex CLI, or opencode followed a walkthrough that promised output their harness never renders, then watched the workflow jump straight into the composer with no banner - the confusion reported in #683.

The issue's suggested fix (copy `settings.json` into every dist) would not work: `companyAnnouncements` is a Claude Code settings key with no render equivalent on the other harnesses, so the copy would be dead weight. This PR fixes the docs instead.

## Changes

**`docs/guide/01-getting-started.md`**
- A note at the top: the walkthrough shows Claude Code; the methodology is identical on every harness but prerequisites, configuration, and some surfaces (welcome banner, statusline) are not, with a pointer to the harness chapters.
- Installation Step 1 is now five per-harness `<details>` dropdowns (Claude Code expanded by default). Each holds that harness's copy commands verbatim from its chapter, and links onward for the harness-specific steps (Codex's git-repo requirement and hook-trust pre-seed, opencode's load-bearing `opencode.json` blocks, Kiro's prerequisites).

**`docs/guide/02-your-first-workflow.md`**
- A note at the top: transcripts show Claude Code; on the other harnesses there is no welcome banner and no statusline (use `/aidlc --status` and the progress line at each gate), and Codex CLI invokes `$aidlc`.
- The welcome-banner paragraph now states inline that `companyAnnouncements` is a Claude Code setting with no equivalent elsewhere - on other harnesses no banner appears and the workflow begins directly with initialization.
- Both statusline passages are scoped to Claude Code with the cross-harness fallback.

**`zensical.toml`**
- Enabled `md_in_html` so the dropdown bodies render as markdown on the docs site. GitHub renders the same `<details>` blocks natively and ignores the `markdown="1"` attribute.

## Verification

- Zensical build: same 14 pre-existing issues as the v2 baseline, zero new. All five dropdowns render with highlighted fences and resolved links in the built HTML.
[...]

PR #651, opened 2026-07-24
docs: fix Mermaid parse errors from semicolons in sequence diagrams

mazyu36:
# Summary

Five sequence diagrams in the docs fail to render on GitHub. Mermaid treats `;`
as a statement separator even inside message and Note text, so a line like
`O->>ST: Report approved; engine marks [x] and routes` is cut at the semicolon
and the remainder fails to parse (`Expecting 'SOLID_OPEN_ARROW' ... got 'NEWLINE'`).

## Changes

Replaced the semicolon with an em dash in the five affected lines — the same
style these diagrams already use (e.g. `Run Bolt A (walking skeleton) — questions,
design, code-gen`). Nothing else changed.

- `docs/reference/01-architecture.md` — Conductor Inline Stage Execution / Conductor Subagent Delegation
- `docs/guide/02-your-first-workflow.md` — Subagent Delegation
- `docs/reference/03-orchestrator.md` — agent activation (Note line) / Construction autonomy ladder

## User experience

Before: the five diagrams show GitHub's "Unable to render rich display" parse
error instead of a diagram. After: all five render.

## Checklist

* [x] I have reviewed the [contributing guidelines](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [ ] Changes are documented

## Test Plan

- Rendered all five fixed diagrams with Mermaid 11.16.0 (the version GitHub
  reports in the error) — all parse and draw.
- Reviewers can verify without cloning: open the three files on this branch
  and check the diagrams render.
- `bun test tests/smoke/` and the docs-related unit tests (t132, t174, t239,
  t246) pass. Docs-only change, so no `dist/` regeneration is involved.

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/aidlc-workflows/blob/main/LICENSE).

PR #645, opened 2026-07-22
docs: reconcile the phantom per-intent reverse-engineering record folder

iuryeng:
# Summary

Removes a phantom per-intent `<record>/inception/reverse-engineering/` folder that three surfaces still advertise but nothing ever writes, and documents where reverse-engineering output actually lives (the space-level per-repo CodeKB). This is the documentation drift behind the confusion in #632 — the empty per-intent folder tells users the artifacts belong there, while the engine deliberately writes them to `aidlc/spaces/<space>/codekb/<repo>/`.

Not marked `Closes #632`: whether the space-level placement itself should change is a design call for maintainers — this PR only reconciles the documented tree with the implemented one.

## Changes

The engine treats reverse-engineering exclusively as a codekb stage — `KNOWN_CODEKB_STAGES` in `aidlc-orchestrate.ts` and `aidlc-state.ts`, and the stage frontmatter declares `outputs: aidlc/spaces/<active-space>/codekb/<repo>/`. Three surfaces still said otherwise:

- **`core/aidlc-common/stages/initialization/workspace-scaffold.md`** — the Step 3 ensure-exists list scaffolded an empty `reverse-engineering/` under `<record>/inception/`. Removed, with a short note explaining the deliberate absence (so it doesn't get "fixed" back).
- **`core/tools/data/scaffold/inception/reverse-engineering/.gitkeep`** — the matching empty skeleton dir in the scaffold data. Removed.
- **`docs/guide/14-artifacts-reference.md`** — the record tree listed `reverse-engineering/ (conditional: brownfield)` and the guide never mentioned codekb at all. Removed the tree line, annotated the stage-artifact table row with the real location, and added a "**The 9 reverse-engineering deliverables are not in the record dir**" paragraph mirroring the existing "Team knowledge is not in the record dir" one. It states the semantics as implemented: one store per repo that each brownfield rerun *overwrites* (last write wins — the condition is "Always rerun for freshness" and no short-circuit exists), the record dir receiving only the stage's own `memory.md` diary on demand, and codekb writes carrying the `codekb > <repo> > <name>` audit breadcrumb.

- **`tests/harness/fixtures.ts`** — the `withReArtifacts` option documented itself as copying into `inception/reverse-engineering/` while the code seeds codekb. Same drift class, one comment line.

No engine code changes and no command or flag changes, but the shipped `dist/` trees do change (installed stage prose, plus an installed skeleton directory that goes away), so this ships as **2.5.14** with a CHANGELOG entry and the README badge.

[...]

PR #754, opened 2026-08-13
docs: reconcile planned Bolt with the shipped stage-major walk (2.6.56)

### LINKED ISSUE #734: [Docs]: The Bolt definition describes the opt-in unit-major walk, and no Bolt boundary is recorded on the default path
# AI-DLC Version

v2 (`v2` branch) at `2ce654d1` (2.5.62). Read from `core/` and from `dist/kiro`; the Bolt
definition row is byte-identical across all six `dist/*` harnesses (md5
`f77ed6a5f3fac47e2ab1fa6995ff8fee`), so nothing here is harness-specific.

# AI-DLC Phase

CONSTRUCTION. Documentation only — I am not reporting a code defect and I am not asking
for a behaviour change.

---

## Summary

I could not explain "Bolt" to someone else from the documentation, so I traced it through
the engine. The definition and the default execution model describe two different things.

The glossary defines a Bolt as **one pass through stages 3.1–3.5 for a Unit**, and the
Construction reference draws that as *the* flow. But the default walk is stage-major — one
stage runs for every Unit before the next stage runs — and the Bolt-major walk the
definition describes is opt-in behind `Construction Iteration: unit-major`. On the default
path there is also no Bolt boundary in the audit trail: the tool that emits
`BOLT_STARTED` / `BOLT_COMPLETED` is only ever invoked from the autonomous-swarm path.

So a reader who follows the docs forms a model of Construction that the engine does not
run by default, and cannot find the Bolt anywhere in the resulting audit log.

The fix I am proposing is a wording change in three files. **I am not proposing that Bolt
be removed** — two things genuinely need the name (a batch of Units, and a Unit's build in
its own worktree), and the tooling around them is substantial.

## Relation to prior work

This is a follow-up to #503, not a repeat of it. #503 reported that Construction runs
stage-major, so the approved Bolt plan's skeleton-first sequencing was never honored; it was
closed as completed by #512 ("opt-in unit-major iteration for construction design stages",
2.2.13), which added the `Construction Iteration: unit-major` knob.

What I am reporting is what that resolution left behind in the documentation: the Bolt
definition still describes the unit-major walk, without saying that it is the opt-in one.
Sections 4 and 5 below are separate from #503 entirely.

### 1. Three places define a Bolt as the opt-in walk

| Location | Text |
|---|---|
| `core/aidlc-common/protocols/stage-protocol.md:922` | "One execution of Construction stages 3.1–3.5 for a Unit (or small group of dependency-linked Units) … The first Bolt is the **walking skeleton**" |
| `docs/guide/glossary.md:12` | "The unit of Construction execution: one pass through stages 3.1–3.5 for a Unit…" |
| `docs/reference/04-stages/construction.md:36-38` | "A Bolt is one pass through stages 3.1–3.5 for a Unit or small group of dependency-linked Units." |

The same file that carries the first definition says, 394 lines earlier:

> **Unit-major iteration (opt-in).** By default the walk above is stage-major: a design
> stage runs for every Unit, then the next design stage runs for every Unit, and
> code-generation runs last for every Unit.
> — `stage-protocol.md:528`

`unit-major` activates only when the state file records that exact string
(`aidlc-orchestrate.ts:1151`, which narrows anything else to stage-major).

### 2. "The first Bolt is the walking skeleton" is implemented as "the first Construction stage"

`isSkeletonGateStage` resolves the skeleton gate to the first in-scope Construction EXECUTE
stage, not to a Bolt (`aidlc-orchestrate.ts:1227`). `isFirstConstructionStage` reuses it so
that, per its own comment, "'the first Bolt' means the same stage to the spoken line as it
does to the gate" (`:471`).

Under the default stage-major walk that stage runs for **every** Unit, and its gate is
presented once, on the re-entry after the last Unit settles (`stage-protocol.md:524`). What
the walking-skeleton gate covers by default is therefore every Unit's first design stage,
not "the thinnest end-to-end slice that proves the architecture".

### 3. "driven by `bolt-plan.md`" is not how the batches are formed

`docs/reference/04-stages/construction.md:34-36` says Construction "executes **Bolt by
Bolt**, driven by `bolt-plan.md` (Bolt sequence + walking-skeleton marker) from stage 2.8".

The batches are recomputed from `unit-of-work-dependency.md` — stage 2.7, not 2.8
(`aidlc-orchestrate.ts:1209-1217`, whose own diagnostic reads "recomputed N unit batch(es)
from unit-of-work-dependency.md"). And bolt-plan's walking-skeleton marker can be overridden
at runtime by the active-space `team.md` stance; that override is surfaced only as a
`PRACTICES_OVERRIDE` row with `Reason: bolt-plan-marker-conflict`
(`knowledge/aidlc-shared/audit-format.md:185`, treated as expected behaviour at
`aidlc-utility.ts:2369` and `:2410-2412`).

Stage 2.8 is a gated stage, so a human approves `bolt-plan.md` — including a marker that
may be overridden later, by a precedence rule documented only in the audit-event table.

### 4. Bolt does not correspond to a level in the data model

The runtime has two levels: a Unit, and a batch of Units. `resolveBoltDag` yields
`string[][]` whose flattened elements are consumed as Unit names
(`aidlc-orchestrate.ts:3110`, `:3473`, `:3643`). `aidlc-bolt start --name` takes one name
per Bolt, derives one slug, and forks one worktree; a comma-separated batch with
`--worktree` is rejected outright (`aidlc-bolt.ts:181-196`).

Meanwhile `docs/guide/glossary.md:47` defines a parallel batch as "a group of **Bolts**",
which inverts the nesting the engine uses. And I found no structure anywhere that groups
several Units into one Bolt, so the definition's parenthetical "(or small group of
dependency-linked Units)" has no counterpart in the code. If such a structure exists and I
missed it, that would be worth documenting rather than removing.

### 5. On the default path, nothing records a Bolt

`aidlc-bolt.ts` owns and emits the Bolt lifecycle events (`:4`, and the `BOLT_STARTED`
emission at `:245`). Its only caller is `aidlc-swarm.ts` (`:486`, `:596`, `:834`, `:836`).
`stage-protocol.md` assigns ownership — "Construction bolts use `aidlc-bolt.ts`" (`:639`) —
but never places a call site: neither it nor `skills/aidlc/SKILL.md` says where in the walk
the conductor should run `aidlc-bolt start` or `complete`, and its only other reference to
the tool is `set-autonomy` (`:196`).

The swarm path requires all of: a Construction stage with `for_each: unit-of-work` and
`mode: subagent`, **not** the skeleton-gate stage, `Construction Autonomy Mode: autonomous`,
and not-unit-major (`aidlc-orchestrate.ts:3081-3091` and `:3054-3078`). A default gated run
satisfies none of the last three, so it emits no Bolt events at all — while
`audit-format.md:158` describes `BOLT_STARTED` as "Orchestrator begins a Bolt".

## What I am inferring, not measuring

Taking 4 and 5 together, the only reading consistent with the code is that **a Bolt is one
Unit's build, in its own worktree**. I am stating that as an inference from the structures
above, not as something the codebase says in those words.

## Proposed wording

Five lines, replacing the definitions in the three files listed in §1. This matches what the
code already does, so no code change is implied.

- **Unit** — the work item. Unchanged.
- **Batch** — the group of Units whose dependencies are satisfied and that build
  concurrently. The unit of parallelism and of gating.
- **Bolt** — one Unit's build: one worktree, one `BOLT_STARTED` / `BOLT_COMPLETED` pair.
  Independent of how many stages have run, and never a container for several Units.
- **Walk order** — a separate axis. Stage-major is the default (a stage runs for every Unit,
  then the next stage). Unit-major is opt-in via `Construction Iteration: unit-major` (a Unit
  runs through every stage, then the next Unit). A Bolt means the same thing under both; only
  when each stage runs for it changes.
- **Walking skeleton** — the first gate in Construction, i.e. the gate on the first in-scope
  Construction stage. Not "the first Bolt".

Two smaller corrections follow from the same pass: `docs/guide/glossary.md:47` should read
"a group of Units", and `docs/reference/04-stages/construction.md:34-36` should attribute the
batches to `unit-of-work-dependency.md` (2.7) and say that bolt-plan's marker is advisory
against the `team.md` stance.

If the intent is instead that Bolt-major should be the *default* walk — which is what the
docs currently describe — then this is a behaviour question rather than a wording one, and I
am happy to reframe the issue that way.

## What I did not verify

I read this statically and did not confirm it with a live run. Specifically unconfirmed:
that a default gated run's audit log contains no `BOLT_*` rows. That is cheap to check on any
existing run — `grep BOLT_ ` over the intent's audit shards — and I did not have a run
available at the time of writing.

szsunyuan:
## Summary
- Addresses #734: keep Bolt as the 2.9 planned delivery slice (one or more Units, DoD, confidence hypothesis, ownership) and document the current stage-major walk next to that definition.
- Default walk is stage-major; `Construction Iteration: unit-major` is opt-in. Runtime batches come from `unit-of-work-dependency.md` (2.7). `bolt-plan.md` is planning, not the walk source. `BOLT_*` events fire per Unit on the swarm / worktree path; `SWARM_COMPLETED` closes the batch.
- No engine or walk-order change. Bolt-major protocol ceremony is destaged to non-imperative design notes. Walking-skeleton stage gate, ladder procedure, halt-and-ask, and the shipped per-unit / wave / unit-major / loop-back blocks stay executable.
- Diagrams, worked examples, and the 2.9 handoff now show stage-major. P2/P3 leftovers in the same files (final swarm Code Generation gate, unit-major stage-gate cascade, audit When-column, `org.md` → `team.md` → `project.md` skeleton stance) are included.
- Version is 2.6.56 (re-slotted past current `v2` 2.6.55, including the #767 protocol-module layout).

Engine walk alignment stays later — this PR does not close #734.

## Test plan
- [x] `bun scripts/package.ts --check`
- [x] `bun test tests/unit/t68-version-changelog-sync.test.ts`
- [x] `bun test tests/unit/t302-protocol-modules.test.ts`
- [x] `bun test tests/integration/t47-construction-bolts.test.ts`
- [x] `bun test tests/integration/t34-stage-protocol-structure.test.ts`
- [x] `bun test tests/integration/t305-build-test-loopback-prose.test.ts`
- [x] A reader of `stage-protocol-construction.md` + `04` / `03-orchestrator` / `diagrams.md` / `16` / Inception 2.9 sees one current model (stage-major), with Bolt-major labeled planned


apackeer (changes requested):
Thanks @szsunyuan - this faithfully implements #734's proposal, the mechanics were exemplary (the v2 merge and re-slot to 2.6.3 were handled exactly right), and during review we verified every engine fact the new wording rests on. The direction is where we have to push back; full reasoning is in the comment on #734. Short version: the verified facts show the default runtime does not yet carry the planned Bolt boundary - they do not establish that a Bolt therefore means one Unit. Bolt-as-planned-delivery-slice is deliberate (2.9 bundles Units into Bolts with a Definition of Done, confidence hypothesis, and ownership; when this divergence was first raised in #503, the resolution #512 moved the engine toward that model, not the docs away from it). The redefinition also collides with those contracts inside this diff: delivery-planning.md:87 now says a Bolt is never a container for several Units, while the same stage asks at :108/:115 which Units the Bolt bundles, and the lead persona still says to bundle (aidlc-delivery-agent.md:30-32).

What we would gladly merge - and most of your file list carries over, on this same PR if you like: keep the existing Bolt definition and document the current reality alongside it:

- the default walk is stage-major; `Construction Iteration: unit-major` is the opt-in;
- `BOLT_STARTED` / `BOLT_COMPLETED` are currently emitted only on the swarm/worktree path, so a default gated run's audit contains none;
- runtime batches are recomputed from `unit-of-work-dependency.md` (2.7); `bolt-plan.md` remains the planning artifact, its walking-skeleton marker advisory against the team.md stance;
- under the default walk, the walking-skeleton gate covers every Unit's first design stage.

If you rework in that direction we'll review promptly. If we haven't heard from you in about two weeks, or you'd rather not carry it, we'll close this PR and fold the notes into the docs pass that follows #767 (which restructures stage-protocol.md into modules and would have forced a rebase here anyway). Either way, thank you - this and #734 are exactly the kind of careful contribution we want.

szsunyuan:
Thanks @apackeer — that distinction was the piece I had backwards. Redirected on this same PR.

Bolt stays the 2.9 planned delivery slice: one or more Units, with a Definition of Done, a confidence hypothesis, and ownership. The four runtime facts you listed are now documented next to that definition (glossary, stage-protocol Terminology, Construction / Inception references). 2.9 questions, the delivery-agent bundle lines, Bolt-gate procedure, and the mermaid are unchanged. No engine change.

#734 left open for the later walk work after #767. No version bump — wording only. Happy to adjust anything that still reads as a redefinition.

apackeer (changes requested):
Thanks for the quick turnaround - this round lands the redirect. I re-verified the reworked definitions against the engine and they all hold: stage-major default with the opt-in `Construction Iteration: unit-major` knob (stage-protocol.md's own protocol text), `BOLT_*` emitted only on the swarm / worktree path (`aidlc-swarm.ts` is the only shipped caller of `aidlc-bolt start --worktree` / `complete --merge`), runtime batches recomputed from `unit-of-work-dependency.md`, and the skeleton gate derived as the first in-scope Construction EXECUTE stage (`isSkeletonGateStage` in `aidlc-orchestrate.ts`). Title and body now describe the actual change, the branch is mergeable, `bun scripts/package.ts --check` is clean, and t47 (12/12) + t68 (7/7) pass at 4e350fa94.

One blocker, and it is mechanical:

**Restore the version bump.** The docs-only exemption does not cover this PR. #761 (the cited precedent) touches exactly one file under `docs/` and nothing under `dist/` - that is the shape the exemption exists for (same family as #733, #693, #651, #596, all docs/README only). This PR changes 35 shipped files under `dist/` (5 core `.md` files projected across the 7 harnesses). Merged history is uniform: I scanned the 40 most recent dist-touching commits on `v2` and every one bumps, and the on-point case is #645 (`d2ae3cca8`), a `docs:`-prefixed PR that changed shipped stage prose and bumped CHANGELOG + badge + `aidlc-version.ts`. The reasoning matters more than the rule here: without a bump, one version number names two different shipped `stage-protocol.md` contents depending on when a user copied their `dist/<harness>/` tree, there is no CHANGELOG notice, and the upgrade instruction your dropped `[2.6.3]` entry carried ("re-copy your dist shell") is lost - and this PR exists precisely because installed prose diverged from reality (#734). The entry text you removed was good: restore it under a slot past the `v2` tip at rebase time (2.6.17 as I write this; a couple of open PRs claim the next slots, and t68 plus the conflict-trap note in AGENTS.md cover the race), bump the README badge and `core/tools/aidlc-version.ts`, and rerun `bun scripts/package.ts`.

Worth folding into the same push:

1. `stage-protocol.md` now contradicts itself. The "Construction gate patterns" section (~L177-207, deliberately unchanged) still describes the walking-skeleton gate as a Bolt-level gate covering "the Bolt's design artifacts and generated code together", tied to `BOLT_COMPLETED` - while your new Terminology rows (L938-942) and the `audit-format.md` note correctly say the gate is the first in-scope Construction EXECUTE stage and that `BOLT_*` does not fire on a default run. The same one-line status-note treatment you gave `audit-format.md` would resolve it.
2. "covers every Unit's first design stage" (Terminology, glossary, both guide pages, the Construction reference, and now the PR body) overclaims for 4 of 9 scopes: the skeleton-gate stage resolves to code-generation for poc/bugfix/security-patch and nfr-requirements for infra - not a design stage. The adjacent "first in-scope Construction EXECUTE stage" clause is right everywhere; qualify or drop the design-stage clause.
3. `docs/guide/02-your-first-workflow.md` still shows the "Construction: Bolt 1 - notification-core (walking skeleton)" banner directly under the corrected paragraph, and nothing in the shipped flow prints that banner. Given #734 came from comparing docs to real transcripts, the quickstart should show what a default run actually prints (or drop the block).

Minor, take or leave: `04-phases-and-stages.md` paraphrases the ladder prompt as "continue autonomously, or keep gating?" but the shipped option label is still "Gate every Bolt" - quote the real strings when describing prompts users will see.

Everything else from the redirect is addressed. With the bump restored (and ideally 1-3 above), this is ready.

szsunyuan:
Thanks @apackeer — I appreciate how carefully you checked this, and I learned a lot from the write-up (especially where the docs-only exemption actually cuts, and the four-scope check on the skeleton gate).

Restored the bump at 2.6.18 after merging current v2. The Bolt-gate section now has the same kind of status note as `audit-format.md`. Dropped the design-stage clause; the gate is the first in-scope Construction EXECUTE stage. Quickstart shows the real first-Construction line. Ladder labels quoted as shipped.

Happy to adjust anything that still reads off.

apackeer (approved):
@szsunyuan this closes it out - approved. The a20f66f03 follow-up addresses every item from my last review, and I verified each against the head:

- 2.6.18 is consistent across `CHANGELOG.md`, the README badge, `aidlc-version.ts`, and all seven generated harness copies, and the changelog entry carries the re-copy upgrade instruction.
- The Construction gate section now distinguishes the intended Bolt-major ceremony from today's default stage-major behavior and states the swarm/worktree-only `BOLT_*` boundary.
- The cross-scope "first design stage" claim is gone everywhere it was introduced, and t47 now pins the universally correct "first in-scope Construction EXECUTE stage" wording (and rejects the old clause).
- The quickstart block is now the shipped first-Construction narration verbatim - nice touch using the real engine line - and the ladder choices are quoted exactly as shipped.
- The original redirect stays intact: Bolt remains the planned one-or-more-Unit 2.9 slice, with stage-major default / unit-major opt-in, 2.7 runtime batches, and the advisory bolt-plan marker documented alongside.

Verification: t47 (12/12) and t68 (7/7) green at a20f66f03, plus a broader focused runtime/version/vocabulary slice (6 files, 64 assertions, all passing); `bun scripts/package.ts --check` passes for all seven harness trees.

Two non-blocking notes:

- Open #767 also claims the 2.6.18 slot. Whichever PR lands second rebases and re-slots per the conflict-trap note in AGENTS.md; t68 will catch the duplicate heading.
- I have asked @leandrodamascena for a second opinion before this merges; approving now so my earlier changes-requested does not block.

Thanks for the careful follow-through across the rounds on this one.

leandrodamascena (changes requested):
Yuan, I reviewed `a20f66f03` deeply against the current engine behavior. The direction is correct: Bolt remains the planned Delivery Planning slice, while the shipped default runtime is stage-major and derives batches from the Unit dependency DAG.

The focused tests, package parity, typecheck, and lint pass. However, several authoritative surfaces still contradict that corrected model.

**P1: The canonical protocol still instructs an unshipped Bolt-major flow**

`core/aidlc-common/protocols/stage-protocol.md:175-202,510-522`

The new status sentence calls the procedure “intended,” but the operative instructions immediately below still tell the conductor to:

- gate a Bolt’s combined design and generated code;
- emit `BOLT_COMPLETED`;
- collect questions by Bolt;
- present subsequent Bolt/batch gates.

Under the default runtime, the first Construction gate covers only the first in-scope Construction stage across all Units. Code Generation has not run, and no `BOLT_*` event exists.

A status sentence does not safely override present-tense instructions in the canonical, model-consumed protocol. Please move the future ceremony out of the active protocol or clearly mark every unimplemented subsection as non-executable design intent.

**P1: Major user and reference documentation remains internally contradictory**

Representative locations:

- `docs/guide/04-phases-and-stages.md:232-340`
- `docs/reference/03-orchestrator.md:438-490`
- `docs/reference/diagrams.md:166-189`
- `docs/guide/16-worked-examples.md:303-336`
- `docs/reference/04-stages/inception.md:1311-1336`

The corrected introductions say stage-major, but the diagrams, worked examples, and handoffs still describe Construction as Bolt-by-Bolt, driven by `bolt-plan.md`, with stages 3.1–3.5 and gates per Bolt.

Please update or explicitly label these future-state diagrams/examples. Leaving both models in the same document does not resolve #734 for readers.

**P1: Stage 2.9 is still described as controlling runtime order**

`docs/reference/04-stages/inception.md:1117-1121,1304-1335`

The engine does not consume `bolt-plan.md` for Unit grouping or runtime order. It computes batches from `unit-of-work-dependency.md`:

- `core/tools/aidlc-orchestrate.ts:1313-1341`

The economic sequence, multi-Unit Bolt grouping, DoD, confidence hypothesis, and ownership are currently approved planning content, not executable runtime boundaries. The handoff must state that clearly.

**P2: Batch-level gate claims do not match swarm settlement**

- `core/aidlc-common/stages/construction/code-generation.md:281`
- `docs/reference/04-stages/construction.md:71-75,120-126`
- `docs/guide/glossary.md:48`

The swarm does not present a gate after each intermediate batch. It converges all DAG batches and then emits one final Code Generation stage gate. `t201` pins this final-only behavior.

**P2: Autonomy and unit-major behavior are described incompatibly**

`docs/guide/glossary.md:11` says autonomous mode skips the remaining Construction stage gates. Unit-major disables swarm routing but retains the per-stage gate cascade:

- `core/tools/aidlc-orchestrate.ts:1711-1725`
- `core/aidlc-common/protocols/stage-protocol.md:534`

Please document the interaction explicitly rather than making both unconditional claims.

**P2: Bolt audit rows still describe batch-level events**

`core/knowledge/aidlc-shared/audit-format.md:193-194`

The swarm invokes `aidlc-bolt start/complete` per Unit. `SWARM_COMPLETED`, not `BOLT_COMPLETED`, closes the batch.

**P3: Walking-skeleton stance resolution is underspecified**

The new text says the Bolt marker is advisory against `team.md`, but actual resolution follows:

```text
org.md → team.md → project.md
```

A project-level stance can override the team-level statement.

The new terminology is directionally correct, but the repository still presents both the intended Bolt-major model and the shipped stage-major runtime as current behavior. I recommend **Changes Requested** until the remaining operational, diagram, example, event, and handoff surfaces are reconciled.

szsunyuan:
Thanks @leandrodamascena for the thorough review — I went through this against the engine and learned a lot. The P2/P3 facts check out, and you're right that a reader can still meet both models as if they were current.

The earlier rounds with @apackeer had landed on keeping Bolt as the 2.9 planned slice, documenting the current runtime next to it, and leaving the Bolt-major procedure / diagrams / worked examples for the later walk work after #767. I followed that on a20f66f03.

Your P1s go further, and I can see why. Before I touch more files I want to make sure I implement the right size: do you want that full reconciliation on this PR, or should those ceremony surfaces stay for the #767 / #734 pass?

Happy to do whichever you prefer.

leandrodamascena:
Thanks for checking before widening the diff.

Now that #767 has merged, I recommend completing the reconciliation on this PR rather than deferring it again. The dependency that justified waiting is gone, and the current branch now needs a rebase onto the protocol-module layout anyway.

The key distinction is that the planned Bolt model can remain documented, but it cannot remain as present-tense executable instruction in model-consumed protocol or as the apparent current runtime in diagrams and worked examples.

Please keep the scope focused on reconciling authoritative surfaces:

1. Rebase onto current `v2` and update the new Construction protocol module introduced by #767.
2. Move the future Bolt-major ceremony out of active executable instructions, or label every affected subsection explicitly as non-executable future-state design.
3. Update the active diagrams, worked examples, and the 2.9 handoff so they describe the shipped stage-major runtime and identify `bolt-plan.md` as planning content rather than the runtime ordering source.
4. Correct the remaining runtime details:
   - swarm uses one final Code Generation gate, not one gate per intermediate batch;
   - unit-major suppresses swarm but retains the per-stage gate cascade;
   - `BOLT_*` rows are per Unit/worktree, while `SWARM_COMPLETED` closes the batch;
   - skeleton stance follows `org.md → team.md → project.md`.
5. Re-slot the version: `2.6.18` is already merged, and this branch is currently conflicting.

If that reconciliation is intentionally out of scope, the alternative is to narrow this PR to terminology-only wording that makes no claim to resolve #734, leave #734 open, and remove the active-protocol changes. I would not merge the current middle state because readers and conductors still encounter both models as current behavior.

szsunyuan:
Thanks @leandrodamascena — the fuller reconciliation is on this PR now.

Merged current v2 (#767 protocol-module layout) and re-slotted to 2.6.56. Bolt-major ceremony in stage-protocol-construction.md is destaged to non-imperative design notes (combined design+code gate, subsequent Bolt-level gates, within-Bolt steps 1–7). Walking-skeleton stage gate, ladder procedure, halt-and-ask, and the shipped per-unit / wave / unit-major / loop-back blocks stay executable.

Diagrams, worked examples, and the 2.9 handoff now describe stage-major and treat bolt-plan.md as planning. Same-file leftovers from the review (04 / 03 / 02 closer / 08 / inception notes / construction overview / audit When-column) are in the follow-up commit.

Happy to adjust anything that still reads as current Bolt-major.

leandrodamascena (approved):
Thank you for the work on clarifying the Bolt model and for the original contribution.

I reviewed the current head `ed9bf2445454f029640849e92dd997c3023881b0` against `c06482d267c62afc6f87cbd23ba0414e8fad5d0b`. The direction was already established through maintainer review, and the takeover remained within that accepted protocol and documentation scope.

The takeover reconciles the terminology with the shipped runtime:

- Bolt remains the sprint-like Construction planning iteration, distinct from a Unit, worktree, swarm, or runtime batch.
- Stage-major execution does not consume `bolt-plan.md` as an ordering or grouping boundary; runtime batches remain driven by the Unit dependency graph.
- `BOLT_STARTED` and `BOLT_COMPLETED` are documented as per-Unit/worktree swarm-path events, while `SWARM_COMPLETED` closes the runtime batch.
- Existing stage-level gates, unit-major behavior, and walking-skeleton precedence remain intact; future Bolt-major ceremony is explicitly non-executable.
- The branch was updated to the latest `v2`, preserving the intervening `2.6.85` release and publishing this change as `2.6.86`.

I pushed the validated takeover through merge commits `cc56a6096` and `ed9bf2445`. Validation passed with 148 focused assertions, package parity, typecheck, lint with only the three pre-existing informational notices in `t267`, and whitespace checks. GitHub CI, deterministic integration/e2e, documentation build, Markdownlint, Gitleaks, Semgrep, Bun audit, and changelog checks are all green on the reviewed head.

I found no remaining code, protocol, compatibility, or test blocker.

**Review event: APPROVE**

PR #758, opened 2026-08-14
fix: one bounded recovery review pass after a receipt-invalidating write (2.6.9)

apackeer:
## Problem

Fixes #755. Refs #742.

A stage that declares a reviewer deadlocks when a declared `produces[]` artifact is written after the terminal review receipt:

- `freshReviewReceipts` invalidates the receipt (fingerprint mismatch, or a later `ARTIFACT_CREATED`/`ARTIFACT_UPDATED` row), so completion refuses: *no fresh REVIEW_COMPLETED is recorded... Invoke the reviewer and record the verdict*.
- The review budget scan (`reviewAttemptSummary`) never resets on that invalidation, so a new request refuses: *review request 2 exceeds this stage's review budget (1)... do not re-invoke the reviewer; quote its findings at the approval gate*.
- `--retry-pending` and `--verdict` also refuse (the pair was already matched).

The two guards instruct each other's forbidden action - a circular loop with no tool path forward. #698/#702 added the review-freeze PreToolUse hook to prevent the invalidating write, but Kiro IDE cannot run it (IDE hook payloads carry empty `tool_input`, so a targeted pre-write freeze has no path to inspect) - it is the only harness without that backstop, which is where #755 was hit. The gate-revision backstop does not cover this either: it requires the write to land after gate-open and after a `HUMAN_TURN`.

## Change

One bounded, deterministic recovery pass, plus refusal messages that name the real way out:

- `core/tools/aidlc-lib.ts` - `freshReviewReceipts` now reports `stageStale` / `unitStale` plus `stageStaleProgress` / `unitStaleProgress` (the next request ordinal and whether the recovery was already spent). Staleness is set only when a TERMINAL receipt is invalidated (fingerprint mismatch with a usable recorded fingerprint, or an event-clear of an actually-set verdict), so the normal adversarial revise-then-re-review loop can never trigger it; it is cleared when a later terminal receipt lands and floored by the same attempt boundaries as everything else. A receipt whose request was recovery-marked is terminal on either verdict.
[...]

PR #759, opened 2026-08-14
fix: drop the duplicate **Timestamp** line from audit blocks (2.6.14)

### LINKED ISSUE #715: [Bug]: Audit blocks from park / unpark / practices-override carry `**Timestamp**` twice — renderAuditBlock owns the key but does not filter it out of `fields`
# AI-DLC Version

v2 (`v2` branch). Reproduced on the current `v2` tip (`8b20390b`), `dist/kiro-ide`. Harness-neutral — `core/tools/aidlc-audit.ts` and `core/tools/aidlc-state.ts` are shared, so every dist carries it.

# AI-DLC Phase

Any. The three affected events fire from `park` / `unpark` / `practices-promote`, which are reachable in every phase.

---

## Summary

`renderAuditBlock()` (`core/tools/aidlc-audit.ts:255`) always writes `**Timestamp**` and `**Event**` itself, then iterates `entry.fields` verbatim without filtering those two reserved keys. Three callers also pass `Timestamp` inside `fields`, so their audit blocks carry `**Timestamp**` **twice**:

- `core/tools/aidlc-state.ts:840` — `WORKFLOW_PARKED`
- `core/tools/aidlc-state.ts:863` — `WORKFLOW_UNPARKED`
- `core/tools/aidlc-state.ts:2780` — `PRACTICES_OVERRIDE` (inside `fail()`)

The values are identical, so a reader that parses block-by-block is unaffected. A reader that scans the whole file breaks: the `**Timestamp**` count no longer matches the block count, so any position-based pairing of fields to events is off by one from the first park onward.

## Reproduction

Pristine `v2` tip (`8b20390b`), `dist/kiro-ide`, one `park` and one `unpark`:

```bash
bun .kiro/tools/aidlc-utility.ts intent-birth --scope poc --project-dir .
bun .kiro/tools/aidlc-orchestrate.ts next --project-dir .
bun .kiro/tools/aidlc-orchestrate.ts park --project-dir .
bun .kiro/tools/aidlc-state.ts unpark --project-dir .
```

Resulting audit blocks:

```markdown
## Workflow Parked
**Timestamp**: 2026-08-05T11:41:08Z
**Event**: WORKFLOW_PARKED
**Stage**: intent-capture
**Timestamp**: 2026-08-05T11:41:08Z

---
## Workflow Unparked
**Timestamp**: 2026-08-05T11:41:08Z
**Event**: WORKFLOW_UNPARKED
**Timestamp**: 2026-08-05T11:41:08Z
```

File-wide counts on that shard:

| Marker | Count |
| --- | --- |
| `^## ` (blocks) | 18 |
| `**Event**` | 18 |
| `**Timestamp**` | **20** |

`PRACTICES_OVERRIDE` takes the same shape — `fail()` passes `{ Reason, Timestamp: isoTimestamp() }`.

## Why it matters

To be precise about the blast radius: **the engine's own readers are not affected.** `auditField()` (`aidlc-state.ts:131`) and `findAllEvents()` split the shard into blocks first and take the first match inside a block, so the duplicate line is simply shadowed. `aidlc-runtime.ts summary` is safe for the same reason.

What breaks is any reader that treats the shard as a flat sequence of lines — which the format invites, since every field is a top-level `**Key**: value` line and the block separator carries no structure beyond `---`. I hit this building an external read-only dashboard over `audit/*.md`: zipping `**Timestamp**` occurrences against `**Event**` occurrences gave 20 against 18, and every field after the first park was attributed to the wrong event. Nothing surfaces the mismatch — the shard stays well-formed and `doctor` passes.

So this is a format-contract bug rather than a live engine bug: the audit trail is the documented machine-readable source of truth, and a duplicated reserved key means the file no longer says what its own schema implies. Fixing it also removes a footgun for the next caller, which is the second half of the suggested fix below.

## Suggested fix

Two layers, because fixing only the callers leaves the same mistake available to the next one:

1. **`renderAuditBlock()`** — skip the reserved keys when iterating `entry.fields`:

   ```ts
   for (const [key, value] of Object.entries(entry.fields)) {
     if (key === "Timestamp" || key === "Event") continue;
     ...
   }
   ```

2. **The three callers** — drop `Timestamp` from the `fields` object:
   - `WORKFLOW_PARKED` → `{ Stage: currentSlug }`
   - `WORKFLOW_UNPARKED` → `{}`
   - `PRACTICES_OVERRIDE` → `{ Reason: reason }`

Note `WORKFLOW_UNPARKED`'s local `const ts` must stay — the following `setField(content, "Last Updated", ts)` uses it.

I have been carrying this fix locally since 2.5.11 and can open a PR if the two-layer shape looks right.

warren830:
# Summary

Audit blocks carried a duplicate `**Timestamp**:` line. `renderAuditBlock` writes the timestamp itself, but did not filter the key out of caller-supplied `fields`, so the three emitters that passed it rendered two identically-marked lines in one block.

The consequence is not cosmetic for anything that reads `audit.md` as a whole file: a reader that pairs `**Timestamp**` occurrences with `**Event**` occurrences desynchronises at the first parked workflow and misattributes the time of every later event.

Fixes #715.

## Changes

Fixed at both layers, which is what the issue proposes:

* **Render layer** - `core/tools/aidlc-audit.ts`: new `EMITTER_OWNED_FIELD_KEYS` (`Timestamp`, `Event`) is skipped in the `fields` loop, so no caller can produce a second line.
* **Caller layer** - `core/tools/aidlc-state.ts`: `park`, `unpark`, and the `practices-promote` write-failure path stop passing the redundant field. Both local timestamp values are still used for `Last Updated` / `Parked`, so nothing else moves.
* **References** - `core/knowledge/aidlc-shared/audit-format.md` gains an "Emitter-Owned Fields" section (field order on the `WORKFLOW_PARKED` row corrected to `Timestamp, Stage`), and the `docs/reference/12-state-machine.md` contract paragraph now documents the emitter-owned field contract, the accepted-but-ignored compatibility behavior of `--field Timestamp=...`, and what historical duplicate blocks mean for flat readers. The renderer guarantee is scoped to structured blocks: free-form `append-raw` blocks carry the emitter's timestamp line and no `**Event**:` line, so they sit outside it.

`Timestamp` deliberately stays **out** of `RESERVED_FIELD_KEYS`. Adding it there is the tempting wrong fix: it would break the public `audit append --field Timestamp=...` contract that `t261-audit-authority-floor` pins. The value is now dropped at render instead of refused, so that test stays green - only its stale rationale comment is updated.

## User experience

The rendered timestamp value is unchanged; only the redundant second line is gone. Existing `audit.md` shards need no migration: block-aware readers are unaffected, and flat readers should split on `---` and use the first emitter-owned timestamp in each block (or deduplicate historical timestamp fields).

## Test Plan

Four new tests in existing files whose `covers:` headers already carry the right tokens, so no coverage-registry regeneration:

* `tests/unit/t111.test.ts` - render-layer contract (a caller-supplied `Timestamp` leaves exactly one `**Timestamp**` and one `**Event**` in the block, the surviving line is the emitter's, sibling fields undisturbed), the public `handleAppend` accepted-but-ignored path, and a documentation pin on the updated reference contract.
* `tests/unit/t17.test.ts` - the reporter's actual symptom: across `park` + `unpark`, the file-wide `**Timestamp**` count equals the block count.

Verified on the current head (rebased onto post-#750 `v2`):

```
bun scripts/package.ts && bun scripts/package.ts --check   # exit 0, all 7 harness trees in sync
bun run check                                              # exit 0 (parity + typecheck + lint)
bash tests/run-tests.sh --debug -P 8 --filter '^(t111\.|t17\.|t261-|t68-|t48-)'   # 6 files, 230 assertions, 0 failed
```

Version re-slotted to **2.6.14**: `v2` shipped 2.6.9 (#758), 2.6.12 (#749), and 2.6.13 (#750) since this branch was cut, and 2.6.14 is the next free slot.

## Acknowledgment


apackeer (changes requested):
Warren, this PR fixes #715 by filtering emitter-owned audit fields at the shared renderer, removing the three redundant `Timestamp` caller fields, and regenerating every harness distribution. I reviewed the issue and the implementation against `origin/v2`; the duplicate structured-event fields are reproducible there, and the code change is narrowly scoped.

**Part A - Direction: sound.** The problem exists, `core/tools/aidlc-audit.ts` is the correct harness-neutral enforcement point, and the caller cleanup in `core/tools/aidlc-state.ts` closes the known sources without changing the public command surface. `bun scripts/package.ts --check` passes for all seven harnesses.

**Part B - Review: request changes.**

1. `docs/reference/12-state-machine.md:451` is stale after this change. It still says park/unpark carry a caller-supplied `Timestamp` and rely on the emitter's line being first. The PR now removes those fields and drops public `--field Timestamp=...` values at render time. Please update this contract paragraph to match the new behavior.

The independent review's three notes do not add blocking findings: the accepted-but-dropped `Timestamp` value is deliberate and tested, 2.6.4 is the correct current version slot, and `append-raw` is a pre-existing path outside the structured renderer.

Verification: `bun scripts/package.ts --check` passed; focused tests passed twice at `tests/logs/2026-08-14T08-06-47Z-p2` and `tests/logs/2026-08-14T08-08-24Z-p7894`, each with 4 files and 202 assertions; `git diff --check` passed. Version, changelog heading, and README badge are synchronized, and no new numbered test file was introduced.

**Part C - UX impact.** There are no command, flag, prompt, stage, default, or error changes. New structured audit rows stop duplicating timestamps; `audit append --field Timestamp=...` remains accepted but now silently ignores the supplied value. Historical audit data remains byte-unchanged and requires no engine migration.

leandrodamascena (approved):
Warren, thanks for the clear report and narrowly scoped fix. I reviewed b70eaaab, reproduced the issue on the base branch, inspected the shared renderer and all three known callers, and ran the focused audit/version suites.

The bug is real. On the current base, a parked workflow produces one structured audit block containing:

- two `**Timestamp**:` lines;
- one `**Event**:` line.

The engine’s own readers are safe because they split by block and take the first timestamp. The impact is on external/read-only consumers that process a shard as a flat field sequence, where timestamp/event pairing becomes misaligned after the first affected event.

The implementation direction is sound:

- `renderAuditBlock()` now owns `Timestamp` and `Event` exclusively and prevents a caller from rendering a duplicate;
- park, unpark, and the practices-promote failure path no longer pass redundant timestamp fields;
- the audit-format table reflects the corrected field order;
- all seven generated harness trees are in parity;
- the focused runner passed 202/202 assertions;
- package and coverage-registry checks are green.

I found only the following non-blocking follow-ups:

**P2: A caller-supplied Timestamp is accepted but silently discarded**

The public audit CLI still accepts:

`--field Timestamp=<value>`

but `renderAuditBlock()` now drops that field without reporting that it was ignored:

- `core/tools/aidlc-audit.ts:365-377`
- `core/tools/aidlc-audit.ts:410-413`

This preserves the existing CLI acceptance test, but a successful generic field operation normally implies that the supplied value was recorded.

A clearer contract would be either:

- reserve and reject `Timestamp` with an explicit message explaining that the emitter owns it; or
- support an explicit, validated timestamp-override mechanism if historical/import use cases require one.

If silent dropping remains intentional for compatibility, please document it clearly in the CLI/reference contract.

**P2: Historical duplicate blocks still affect flat readers**

The changelog says existing shards need no migration because the duplicate value was redundant. That is true for block-aware readers, including the engine.

However, the external flat-reader failure that motivated #715 remains present in historical shards. This PR changes only future writes; it does not repair or reinterpret existing duplicate blocks.

Please clarify the release note:

- block-aware readers require no migration;
- flat readers must parse by `---` block and select the emitter-owned timestamp, or deduplicate historical timestamp fields.

A destructive rewrite of append-only audit history is not required, but the compatibility guidance should accurately describe the remaining reader obligation.

**P3: The canonical state-machine reference still describes the old behavior**

`docs/reference/12-state-machine.md:451` still says:

- park/unpark rows carry caller-supplied `Timestamp`;
- safety relies on the emitter timestamp appearing first and parsers taking the first match.

After this PR, those callers no longer pass the field and the renderer drops any supplied value.

Please update that paragraph to describe the new emitter-owned field contract.

These are documentation and API-clarity follow-ups rather than correctness blockers. The duplicate-field bug is reproduced, the shared fix closes future structured writes, and the focused validation is green.

Approving with the follow-ups above.

leandrodamascena:
I rebased this branch onto current `v2` and addressed the remaining review findings.

- Updated the canonical state-machine reference to the emitter-owned `Timestamp`/`Event` contract.
- Documented that public `--field Timestamp=...` remains accepted for compatibility but is intentionally ignored.
- Clarified historical-reader behavior: block-aware readers need no migration; flat readers must split on `---` and use the first timestamp per block or deduplicate old rows.
- Added regressions for the public append path and the documentation contract.
- Re-slotted the release to `2.6.11` and regenerated all harness distributions.

Verification:
- Focused audit/version suite: 204/204 assertions passed.
- Typecheck, lint, package parity, coverage registry, and `git diff --check`: passed.
- Audit emitter integration checks: 26/26 passed.
- Full smoke+unit: only the known `v2` baseline failures in `t248` and `t255`; all other files passed.
- Two independent final reviews: no findings.

Correction commit: `5474cbd0`.

apackeer (approved):
Round-2 re-review at 7ad99d9b9 (the 2.6.14 head, rebased onto post-#750 v2). My round-1 ask is addressed: the docs/reference/12-state-machine.md contract paragraph now documents the emitter-owned field contract, the accepted-but-ignored --field Timestamp=... compatibility behavior, and the flat-reader guidance for historical shards, with the renderer guarantee correctly scoped to structured blocks (free-form append-raw blocks sit outside it).

Verified on this head: the change-lines are identical to the round-1 reviewed diff apart from the version re-slot to 2.6.14 and the scoping sentence; bun scripts/package.ts --check passes for all seven harness trees; bun run check exits 0; the focused suite (t111, t17, t261-audit-authority-floor, t68-version-changelog-sync, t48) passes 273/273 assertions, with t68 confirming the shipped version, latest changelog heading, and README badge agree at 2.6.14. Approving.

PR #760, opened 2026-08-14
fix: bound the conductor's context-park to a measured figure (2.6.5)

### LINKED ISSUE #547: [Bug]: Conductor parks workflow prematurely claiming 'context is heavy' at 37% usage (63% free)
## Description

The AI-DLC conductor (the orchestrator persona running the forwarding loop) repeatedly claims "context is getting heavy" or "context is heavy" and parks the workflow at inter-stage boundaries, forcing the user to resume in a new session. However, `/context` shows only 37% of the 1M context window is used — 62.9% (629k tokens) remains free.

This creates an extremely disruptive experience: the user must `/aidlc --resume` every 1-2 stages, losing flow and wasting time on the unpark/resume handshake.

## Steps to Reproduce

1. Start an enterprise-scope workflow (32 stages) on Claude Code with Opus 4.6 (1M context)
2. Complete stages normally through ideation + inception
3. Observe the conductor adding text like "Context is getting long — let me park here" or "Parking here — context is heavy" after each stage approval
4. Run `/context` immediately after parking
5. Observe: only 37% used, 63% free

## Expected Behavior

The conductor should NOT park due to context pressure when >50% of the context window remains free. Parking for context should only happen when genuinely approaching limits (e.g., >85% usage). The conductor should continue running stages until either:
- The user asks to stop
- Context genuinely approaches capacity (>85%)
- The workflow reaches a natural `done`

## Actual Behavior

The conductor parks after nearly every stage with messages like:
- "Context is getting heavy — let me park here"
- "Parking here — context is heavy and the next stage is substantial"
- "Context is getting long, let me park at this clean boundary"

Meanwhile `/context` shows 370.9k/1M (37% used). The conductor has NO mechanism to check actual context usage — it's guessing based on conversation length, not actual token count.

## AI-DLC Version

v2 (alpha)

## Release / Commit

Latest (installed via `.claude/` workspace shell, July 2026)

## AI-DLC Phase

Not phase-specific — happens across Ideation and Inception stages equally.

## Platform / IDE

Claude Code

## AI Model

global.anthropic.claude-opus-4-6-v1[1m] (Opus 4.6, 1M context)

## Environment

- OS: macOS Sequoia (Darwin 25.5.0)
- Shell: zsh
- Terminal: Warp
- Claude Code: latest

## Additional Context

Screenshot showing the behavior:
- Conductor says "Requirements approved. Parking here — context is heavy and User Stories (next) is another substantial stage with a reviewer loop."
- Immediately after, `/context` shows 370.9k/1M tokens (37%) with 629.1k free space

The root cause appears to be that the conductor persona has NO access to actual context metrics. It's making a judgment call based on perceived conversation length rather than querying the real token count. The SKILL.md instructs to park "when you are running low on context mid-loop" but provides no mechanism to CHECK context — the conductor is guessing conservatively and wrong.

Possible fixes:
1. Expose a `/context`-equivalent API the conductor can call before deciding to park
2. Add a guardrail: "never park for context reasons until you've been explicitly told context is >80%"
3. Remove the conductor's discretion to park for context entirely — only park when the user asks or the system actually compresses

warren830:
# Summary

The conductor parks workflows because context *feels* heavy, and it has no way to know whether it is. The orchestrator skill licensed parking "when you are running low on context mid-loop" — an unmeasurable trigger. The conductor cannot read its own token count, so it guessed from conversation length, and guessed conservatively.

The reporter measured 37% of a 1M window used, 63% free. The framework's own statusline paints that same reading **green**: `core/hooks/aidlc-statusline.ts` `contextColor()` turns yellow at ≥50 and red at ≥75. So the conductor was parking while the framework's own instrumentation said there was nothing wrong — and the user paid an `/aidlc --resume` handshake every one or two stages for it.

Fixes #547.

## Changes

Implements option 2 from the issue — a numeric guardrail rather than removing the capability:

* Parking for context now requires a usage figure **the harness actually surfaced**, at or above **80%**. That floor sits above the statusline's own red line (75), so the two can never contradict each other.
* Absent such a figure, the conductor **keeps running stages**. Having no number is not evidence of pressure.
* Parking on how context *feels* is named and forbidden, with the measured counter-example in the prose so the next reader sees why the bound exists.

Applied to all seven harness orchestrator skills (`claude`, `codex`, `copilot`, `cursor`, `kiro`, `kiro-ide`, `opencode`). Each file's surrounding prose is preserved exactly — `codex` legitimately has no "Tell the user their work is saved" sentence, and still doesn't.

**Deliberately unchanged:** the user-initiated park, and the rule that parking beats fabricating a `done`. This removes one bad trigger, not the feature — a conductor that genuinely needs to stop still has the safe exit.

## User experience

Long workflows stop interrupting themselves. An enterprise-scope run that previously parked every one or two stages now continues until the user asks to stop, the harness reports real pressure, or the workflow reaches `done`.

## Checklist

- [x] I have read the contributing guidelines
- [x] I have performed a self-review of my code
- [x] I have tested my changes
- [x] I have documented my changes (CHANGELOG entry)

## Test Plan

New `tests/smoke/t291-context-park-bounded.test.ts` (43 assertions, mechanism `none`, asserts against authored `harness/` source rather than `dist/`, following the t250 house pattern). Per harness it pins:

* the unbounded `running low on context` clause is gone and cannot regress back in
* parking on how context feels is forbidden
* the skill states the conductor cannot measure its own context window
* **a number survives** — `at or above 80%` plus the surfaced-figure requirement. A bound with no figure is not a bound, so this is the assertion that actually protects the fix.
* the conductor is told to keep going when it has no figure
* the user-initiated park and the anti-fabrication rule are still present

The `HARNESSES` list is an explicit literal, not a glob, so adding a harness without extending this guard fails loudly instead of silently skipping.

Verified locally (bun 1.3.11; CI pins 1.3.14):

```
bun scripts/package.ts && bun scripts/package.ts --check   # exit 0, 7 trees in sync
bun tests/gen-coverage-registry.ts --check                 # OK (fresh, guards green, ratchet held)
bun run check                                              # exit 0
bun tests/run-tests.ts --smoke --unit --parallel 8
```

Suite result is **identical to the pre-change baseline on this branch point** — the same 2 files fail (`t248-codekb-scope-diff`, `t255-workspace-sync`, both unrelated to orchestrator prose; `t255`'s are ~5s timeouts on live git remote queries and look environmental here) with the same 9 assertions. `t119` (SKILL.md 500-line ceiling) still passes with headroom.

Version bumped to **2.6.5**: #754 and #758 both claim 2.6.3, and #759 claims 2.6.4.

## Related

A durable follow-up — plumbing the `context_window.used_percentage` the statusline hook already receives into something `aidlc-orchestrate next` can surface, so `park` could *refuse* below the threshold rather than relying on prose — is a larger change and intentionally out of scope here. This PR closes the reported regression without it.

## Acknowledgment


apackeer (dismissed):
Warren, this PR addresses #547 by tightening context-based workflow parking across all seven harness orchestrator skills. It requires a harness-surfaced usage figure at or above 80%, keeps user-requested parking and the anti-fabrication rule intact, and regenerates the corresponding distributions.

**Direction: sound.** The reported problem exists on `origin/v2`: every orchestrator skill licenses parking when context merely seems low, while the only numeric context signal is consumed by the statusline hook and never reaches the conductor or orchestration engine. The per-harness skill layer is the correct authored surface, and applying the policy to all seven harnesses matches the scope of the existing rule.

I have one change request:

1. `tests/smoke/t291-context-park-bounded.test.ts:36` claims its explicit seven-name list will fail when a new harness is added without extending this guard, but the test only iterates that fixed list. An eighth manifest-backed harness would be silently skipped and the test would remain green. Please derive the set from the repository's manifest-discovered `HARNESS_MATRIX`, or compare the literal against the discovered set so the stated exhaustiveness guarantee is real.

UX-wise, this is consistent with the existing workflow: commands, flags, stage names, errors, user-requested parking, and resume syntax are unchanged. Users simply stop receiving premature park/resume interruptions unless the harness has surfaced at least 80% usage.

Non-blocking follow-up: this remains a prose guardrail, so nothing deterministic makes `park` reject a below-threshold request. The larger engine plumbing is reasonably out of scope here, but please file the follow-up described in the PR body before #547 closes so it remains tracked.

Verification was otherwise clean: `bun scripts/package.ts --check` passed for all seven distributions; our targeted run at `tests/logs/2026-08-14T09-04-31Z-p2` reports `Result: PASS` for 4 files and 72 assertions; the independent reviewer also passed the full smoke tier at `tests/logs/2026-08-14T09-08-57Z-p177266` with 14 files and 1657 assertions; the coverage registry is fresh; and the `2.6.5` version, CHANGELOG heading, and README badge are synchronized.

leandrodamascena (approved):
Thanks, Warren. I reviewed the current head `1e10b8efb` against `b44b9c9c1`.

The direction is aligned with issue #547 and the prior maintainer review, which explicitly accepted the measured 80% guardrail across all shipped harnesses.

I integrated the current `v2` branch without rewriting your original commit and pushed merge commit `1e10b8efb`. During the integration I:

- resolved the version conflict as `2.6.89`;
- regenerated all harness distributions;
- renamed the smoke test from `t291` to `t330` to avoid the current test-number collision;
- addressed the requested review change by deriving the covered harnesses from `HARNESS_MATRIX`, so newly added manifest-backed harnesses cannot be silently skipped.

I found no remaining implementation blockers.

Validation passed: the focused contract test (43 tests), version and conductor-parity tests (31 tests), the complete smoke tier (14 files, 1,736 assertions), package drift check, coverage registry, typecheck, and lint. All GitHub CI, Markdownlint, and Security Scanners checks are green on this exact head.

The remaining architectural limitation is unchanged: the 80% threshold is a prose guardrail rather than deterministic enforcement in `park`. I did not find a dedicated follow-up issue for that engine plumbing, so one should be filed before #547 closes.

Issue #530, opened 2026-07-07
[Feature]: cost-preview refinements (per-unit clause, greenfield-effective counts) and a first unattended-gate story


apackeer:
## AI-DLC Version

v2 branch, HEAD `7c02d48` (2.2.6). Items 1-2 are follow-ups to the open scope-cost-preview PR (#508, 2.2.11) and land after it merges; item 3 is independent of it.

## AI-DLC Phase

Initialization (the scope routing ask) for items 1-2; any gated stage for item 3.

## Summary

Three small, related gaps around "the human at the decision point sees an accurate picture". The first two are the known non-blocking refinements already listed in #508's body; this issue makes them trackable so they do not evaporate when that PR merges. The third is a verified absence in the gate mechanics.

**1. The per-unit clause shows on scopes that SKIP units-generation.** The cost line appends a per-unit ceremony clause ("N stage(s) repeat per unit of work in Construction") computed from the grid's EXECUTE stages, but bugfix/poc/refactor/security-patch/infra SKIP units-generation, where code-generation degrades to a single pass on a cold start. The clause overstates ceremony in the safe direction for consent, but a preview that is knowably wrong for 5 of 9 stock scopes invites distrust of the numbers that ARE right. Fix: suppress (or annotate) the clause when the grid's `units-generation` entry is not EXECUTE; `ScopeCostSummary` already carries `perUnitStages` separately from `gates`, so this is presentation-layer only.

**2. The cold-start confirm shows nominal, not effective, counts for greenfield.** Greenfield projects drop reverse-engineering at birth, so effective counts run 1 stage / 1 gate lower for the 8 scopes that mark it EXECUTE. Project type is unknown at confirm time today, but the `detect` scan (2.2.0) is cheap and read-only; running it before the confirm (or annotating "32 stages, 31 on greenfield") would make the two numbers agree. The scope-change surface already uses the adjusted grid, so the cold-start ask is the only surface showing the stale number.

**3. There is no unattended-gate story.** A workflow that reaches an approval gate simply stops until the human returns: verified absent from `core/` on HEAD - no idle-time awareness, no notification seam, no nudge, and the one adjacent mechanism (`aidlc-state.ts park`, the `WORKFLOW_PARKED` audit event) is manual-only. Long runs routinely cross a human absence (overnight, meetings), and today that reads as a hung run: the statusline shows the same stage for hours and nothing distinguishes "waiting for you" from "stuck". Smallest useful increment, in order of ambition:
[...]

Issue #427, opened 2026-06-22
[Bug] Stop hook overrides approval gates in Claude Code — agent auto-approves and runs unattended


GerardDeLapelinDumont:
## Summary

In the Claude Code harness, the Stop hook prevents the agent from pausing at stage approval gates. The agent explicitly reasons "the stop hook won't let me pause for your answer" and then auto-approves and continues — running the entire workflow lifecycle unattended with zero human gate approvals.

This appears distinct from #426 (Kiro-specific).

## Environment

- **AIDLC v2 commit:** `f3ce1b8` (branch `v2`, ~v2.0.2)
- **Claude Code version:** 2.1.185
- **LLM provider:** Bedrock (us-west-2)
- **Scope detected:** `poc` (7 stages, Minimal depth)

## Steps to Reproduce

1. Init a v2 project in Claude Code and start a workflow (`/aidlc <description>`).
2. Let the workflow reach the first stage approval gate (e.g., end of intent-capture).
3. The agent asks "Approve or Request Changes?", then the Stop hook fires with: _"The AIDLC workflow has a pending step… finish the forwarding loop."_
4. The agent cannot pause → auto-approves → continues to the next stage.
5. Repeats at every subsequent gate.

## Expected Behavior

The agent pauses at each approval gate and waits for explicit human input (Approve / Request Changes) before proceeding.

## Actual Behavior

The agent runs the **entire** lifecycle (intent-capture → requirements-analysis → code-generation → build & test) with **zero human gate approvals**. The Stop hook's "keep going until done" logic overrides the gate's "wait for human" state.

The agent's own output makes the failure explicit: _"The stop hook won't let me pause for your answer — I'll report the stage as approved (you haven't objected) and continue."_

## Forensic Evidence

`audit.md` records `STAGE_AWAITING_APPROVAL` → `GATE_APPROVED` → `STAGE_COMPLETED` with identical timestamps at every gate — i.e. a machine auto-approve in zero (or one) elapsed seconds, not a human decision. Every one of these gate events also carries `Recovered: true`:

```
── intent-capture ──────────────────────────  (all events Recovered: true)
STAGE_AWAITING_APPROVAL  2026-06-22T21:17:31Z
GATE_APPROVED            2026-06-22T21:17:31Z
STAGE_COMPLETED          2026-06-22T21:17:31Z

── requirements-analysis ───────────────────
STAGE_AWAITING_APPROVAL  2026-06-22T21:18:13Z
[...]

PR #766, opened 2026-08-14
fix: an unattended driver must not mint the human-presence token (2.6.108)

warren830:
# Summary

The human-presence gate can be satisfied by something that is not a person.

`hooks/aidlc-record-human-turn.ts` records a `HUMAN_TURN` on `UserPromptSubmit` — the evidence `humanActedSinceGate()` spends to authorize an approval or an interview answer. But the hook has **no evidence for the assertion it is making**: the event carries no signal about who submitted, and the hook reads no stdin ("Presence-only: the prompt text is irrelevant"). Its only guard is "does a state file exist".

That is sound while every prompt comes from a person. It stops being sound the moment something drives `/aidlc` on a schedule — an overnight runner, a CI job, a cron. Each cycle submits a prompt, so each cycle mints fresh, spendable presence.

**This is not a regression of #427 — it is the premise #427's fix was verified against.** That verification reads *"At every approval gate the agent paused and ended its turn; walking away (no input) did not auto-approve"*, which holds precisely because no prompt means no presence. An unattended driver submits prompts, so "walking away" no longer implies "no new human turn".

## Evidence

Measured on a real unattended run against a stock v2 install (10 detached cycles, no human involved at any point):

```
HUMAN_TURN rows in the ledger: 10   (every one minted by a runner-submitted prompt)
gate resolutions:              0
humanActedSinceGate():         true
```

The new t188 case then proves the **consequence** rather than assuming it. Reverting only the hook change turns it red with `Expected: not 0` — i.e. **without this change the gate approves** on a turn minted solely by an unattended prompt. The remaining barrier today is the conductor's prose ("NEVER self-answer questions"), not a mechanism.

## Changes

`AIDLC_UNATTENDED=1` suppresses the mint. One env check in one hook.

* Covers **both** of the hook's paths — the typed prompt and the answered `AskUserQuestion` widget — since an unattended driver should mint from neither.
* **Consistent with doctrine already in the engine.** An unattended autonomous Construction run "has no human at the gate" (`aidlc-utility.ts`), which is why it already refuses scope changes and plan re-shapes, and why `aidlc-state.ts`'s autonomy guard refuses `park`. This closes the one path where an unattended turn still manufactured a human.
* **Fail direction is deliberate:** the flag can only ever WITHHOLD authority. If it leaks into an interactive shell, approvals get refused until it is unset — annoying, and safe. The inverse mistake cannot be undone, because the ledger is append-only.
* **Nothing changes for an interactive session.** With the flag unset the mint is byte-for-byte what it was, which is what the "an attended one still does" half of the new test pins.

### One choice I am flagging rather than making silently

The `.aidlc-human-turn` **marker** is still written. It is not an authority signal, and suppressing it would change the Stop hook's conversational carve-out — separate behaviour, with its own tests, that I have not exercised here. It is a one-line follow-on if you want it; I would rather you decide than have it ride along.

Two further design calls that are yours, not mine:

1. **Skip vs. distinct event.** I skip the mint. The alternative is recording something like `RUNNER_TURN` that `humanActedSinceGate` does not accept, which keeps the trail richer at the cost of a new `VALID_EVENT_TYPES` member plus the state-machine chapter tables under the same-commit rule. `SESSION_STARTED` already records that a session happened, so I took the smaller surface.
2. **Naming.** `AIDLC_UNATTENDED` reuses the word the codebase already uses for this condition. It is the symmetric half of the existing `AIDLC_SKIP_HUMAN_PRESENCE_GUARD` (which turns the guard *off*); what was missing was a way to say "this prompt is not a person's".

## User experience

An operator running AI-DLC unattended sets `AIDLC_UNATTENDED=1` in the driving process, and gates then refuse exactly as they do when nobody has acted — the workflow parks and waits for a real human instead of being able to walk past its own approval. Interactive users see no change.

Related: #530 ("a first unattended-gate story") is where this belongs as a design thread; I have posted the finding there.

## Checklist

- [x] I have read the contributing guidelines
- [x] I have performed a self-review of my code
- [x] I have tested my changes
- [x] I have documented my changes (CHANGELOG + `docs/reference/06-hooks-and-tools.md` hook table, same commit per the Documentation Policy)

## Test Plan

Two cases added to `tests/unit/t188-human-presence-gate.test.ts`, whose `covers:` header already carries `file:hooks/aidlc-record-human-turn.ts`, so no coverage-registry change. Both spawn the **real hook as a process** (the env read is the contract under test) with `AIDLC_PROJECT_DIR` pointed at the fixture:

* *an unattended prompt mints NO HUMAN_TURN; an attended one still does* — the flag is the only difference between the two halves, which is what stops the test passing for the wrong reason (a hook that never mints at all).
* *a gate REFUSES on a turn minted only by an unattended prompt* — the consequence, asserted end to end through the real `approve` CLI.

**Red-green proven:** stash the hook change, regenerate `dist/`, and both fail — the second with `Expected: not 0`, i.e. the gate approved. Restore, and 24/24 pass.

Verified locally (bun 1.3.11; CI pins 1.3.14):

```
bun scripts/package.ts && bun scripts/package.ts --check
bun tests/gen-coverage-registry.ts --check
bun run check
bun tests/run-tests.ts --smoke --unit --parallel 8
```

The first three exit 0. The suite result matches the baseline on this branch point: the same 2 files fail (`t248-codekb-scope-diff`, `t255-workspace-sync`), both unrelated to presence. `t255`'s failures are ~5s timeouts on live git remote queries and its count moves with load — 16 when the file runs alone here, 7-8 under `--parallel 8` — so it is environmental. `t188` is 24/24.

Version bumped to **2.6.6**: 2.6.3 is claimed by #754 and #758, 2.6.4 by #759, 2.6.5 by #760.

## Acknowledgment


apackeer:
Warren, this PR addresses a real gap in the human-presence guard: a scheduled driver can submit the same prompt event as a person and thereby mint the `HUMAN_TURN` token that approval and interview gates trust. The proposed fail-closed `AIDLC_UNATTENDED=1` declaration is a proportionate way for the driver, which is the component that knows it is unattended, to withhold that authority.

STAMP:    tests/logs/2026-08-14T18-07-21Z-p2
TRACES:   /home/ubuntu/src/aidlc-workflows/.claude/worktrees/pr-766/tests/logs/2026-08-14T18-07-21Z-p2/{sdk,tui,kiro-acp}-drive-*.ndjson  (0 files)
SUMMARY:  tests/logs/2026-08-14T18-07-21Z-p2/summary.txt  +  failures.txt
RESULT:   unit . 31 pass/0 fail . reds: none . live vars set: none . invariant grep hits: 0 (path-excluded).

**Part A - Direction: sound-but-misplaced**

The problem exists on current `origin/v2`, and the fail direction is sound. The implementation boundary is incomplete, though: the repository has five production mint implementations, not one shared hook. Claude uses the changed hook, and Cursor/opencode forward to it, but Codex, Copilot, Kiro CLI, and Kiro IDE append `HUMAN_TURN` directly in their adapters. This policy should live in a shared mint predicate/helper used by every seam, or be applied explicitly to every adapter.

**Part B - Review: request changes**

1. `core/hooks/aidlc-record-human-turn.ts:68` does not protect the direct adapter mints at `harness/codex/hooks/aidlc-codex-adapter.ts:597`, `harness/copilot/hooks/aidlc-copilot-adapter.ts:625`, `harness/kiro/hooks/aidlc-kiro-adapter.ts:236`, and `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:239`. I confirmed behaviorally that the packaged Codex adapter still records one `HUMAN_TURN` with `AIDLC_UNATTENDED=1`. Please apply the contract across all shipped harnesses and add focused coverage for each independent mint path.

2. The leaked-flag failure mode is opaque. `core/tools/aidlc-state.ts:2485` and the `aidlc-log` answer/summary refusal paths tell the user to type or acknowledge again. With `AIDLC_UNATTENDED=1` still set, that cannot mint presence, so the advice loops forever. Please append a conditional diagnostic naming the flag on these refusal paths.

3. `CHANGELOG.md:6` says an unattended driver can no longer satisfy the gate, but this is an opt-in declaration: an unflagged driver still mints, even after the cross-harness gap is fixed. Please describe the conditional contract explicitly. Also update `docs/guide/07-interaction-modes.md:76` to document the declaration and how an operator returns to interactive mode.

Non-blocking release note: open PR #749 currently also declares 2.6.6. The version, README badge, and CHANGELOG heading are internally synchronized here, but whichever PR merges second needs the documented rebase/re-bump.

**Part C - UX impact**

This adds one operator environment setting and no new command, stage, or default. With the flag honored, unattended prompt cycles no longer grant approval/interview authority and the workflow waits for a person; interactive sessions remain unchanged. The marker used by the Stop-hook conversational carve-out is intentionally unchanged. The submitted code currently delivers that UX only on Claude, Cursor, and opencode, and a leaked flag produces an unexplained retry loop, so the release behavior is not yet consistent or diagnosable.

Verification: focused t188 plus t68 passed 31/31; package drift, coverage registry, and static checks also passed. No e2e or live-service variables were needed.

apackeer (dismissed):
Requesting changes for the cross-harness, diagnostic, and documentation gaps described in my review comment.

leandrodamascena (approved):
Thanks for working on this @warren830

PR #767, opened 2026-08-14
feat!: classic and express scopes, classic implicit default (AWS_AIDLC_DEFAULT_SCOPE overrides), conditional protocol modules (2.6.18)

apackeer:
## What this PR does

Answers the recurring community feedback that AI-DLC's default run is too long
and too token-hungry (several community forks have appeared to strip it down,
e.g. the `slim` fork). Three changes, each independently useful. No existing
scope is removed or renamed and no state migration is required, but this
revision **declares one behavior change**: the implicit default scope is now
`classic` (see section 1 and the 2026-08-19 response below).

Rebased onto the current v2 tip (v2.6.17, #792); ships as 2.6.18. This
revision addresses the 2026-08-17, 2026-08-18, and 2026-08-19 reviews in
full; the point-by-point responses are below. It also ships the live Kiro
IDE pre-merge gate for Windows as well as macOS (section 4).

### 1. `classic` scope: the v1-style lifecycle, now the implicit default

- NEW `classic` scope: everything except the 7 Ideation stages, 26/33 in the
  plan, only 8 of them unconditional, which is v1's spine. v1 had no Ideation
  phase; users migrating from v1 get the flow they know.
- `classic` carries the production Standard test floor and
  `review_cap: advisory`. It has **no routing keywords**: it is selected by
  explicit name (`--scope classic`), per-project default, or the implicit
  fallback below.
- **`workshop` is retained unchanged**: same grid (byte-identical parity
  fixtures), same `testStrategy: Minimal` override, same
  `workshop`/`lab`/`training` keywords. In-flight workshop workflows and
  overlay upgrades keep working with no migration; `--scope workshop` remains
  valid.
- `feature` is untouched (all 33 stages) and remains fully available —
  explicitly (`--scope feature`, `/aidlc-feature`) or as the project default
  via `AWS_AIDLC_DEFAULT_SCOPE=feature`.
- **Default change (declared):** the implicit default scope is now `classic`.
  Exactly two things control the implicit default: the
  `AWS_AIDLC_DEFAULT_SCOPE` env var (which overrides) and a single hard-coded
  `DEFAULT_SCOPE = "classic"` constant. The engine's resolution fallback,
  `/aidlc-init`, and a direct low-level `intent-create` without `--scope`
  (previously a hardcoded `poc` that ignored the env var) all resolve the env
  var first and fall back to `classic`; the shipped Claude `settings.json`
  sets the var to `classic`; the `--help` "(default)" marker follows the
  resolver. Explicit `--scope`, keyword detection, and the compose offer are
  unaffected — rich or unmatched prose still enters the compose offer before
  any workflow is created, so nothing silently defaults where an ask existed.
  In-flight workflows keep their persisted scope. Teams that want the
  previous behavior set `AWS_AIDLC_DEFAULT_SCOPE=feature`. The changelog
  entry, the `feat!` title, and an e2e journey (t301: freeform text with no
  keyword falls back to `classic`) all declare/pin this.

### 2. `express` scope: the lightweight run, plus composer-led sizing

- New 10-stage requirements-to-deploy grid: initialization, conditional
  reverse-engineering, requirements-analysis, code-generation, build-and-test,
  and a conditional deploy tail (deployment-pipeline, deployment-execution,
  observability-setup). `depth: Minimal`, reviewers off via
  `review_cap: none`, ships as `/aidlc-express`.
- The zero-Unit path is now first-class engine behavior: Code Generation runs
  one stage-level iteration (no Bolt, walking skeleton, ladder, per-Unit
  receipts, or swarm ceremony), artifacts land under stage-level paths with
  no synthetic unit segment, and the stage contracts that consume skipped
  producers (Code Generation, Deployment Pipeline, Deployment Execution,
  Observability Setup) name express explicitly with requirements/workspace
  fallbacks; the engine classifies those absences as designed
  (`expected: true`), and the conditional tail reports skipped when no
  deployable target exists. Since the 2026-08-19 round the zero-Unit routing
  is **structural** (any plan whose effective grid excludes Units Generation),
  not keyed to the express scope name.
- Testing floor made consistent everywhere: express uses the Minimal
  strategy's requirement-driven new-test floor (one test per requirement,
  happy-path floor per component); org rules, the scope doc, and the stage
  instructions now agree.
- The swarm path is structurally unreachable in express (it skips
  units-generation, so no unit DAG can exist), keeping the scope's cost
  ceiling honest.
- The cold-start compose offer anchors its examples with live counts
  (`express = 10 of 33 stages, classic = 26, feature = all 33`), so the
  lightweight options are visible at the moment users choose; express and
  classic also join the composer's matched-scope candidates.

### 3. Protocol modules: smaller fixed session context, conditioned on the directive

- Rare-path machinery moved from the always-loaded stage protocol and all
  seven harness orchestrator skills into four conditionally loaded modules:
  `stage-protocol-reviewer.md`, `stage-protocol-swarm.md`,
  `stage-protocol-ensemble.md`, `stage-protocol-construction.md`. Stubs and
  each SKILL's protocol-loading list say when to load what; run-stage and
  invoke-swarm directives carry a validated `protocol_modules` hint computed
  on the existing review-class/mode/phase seams (a `review_cap: none` scope
  never lists `reviewer`), and every harness's invoke-swarm instruction loads
  every emitted module, reviewer included. Since the 2026-08-19 round the
  generated stage and scope runners also load the base protocol plus every
  directive-named module before acting (asserted against the rendered runner
  output in t302).
- The Construction module's ceremonies are conditional on the actual
  directive: reviewer work only when `directive.reviewer` is present,
  skeleton/ladder only when a real Unit DAG and applicable skeleton stance
  exist, and no Bolt/per-Unit ceremony on a zero-Unit directive.
- Swarm settlement is deterministic across sessions: when every batch
  converges, the settle directive carries `swarm_settled: true`, drops the
  reviewer fields, and pins `protocol_modules` to
  `["construction", "swarm"]`; the marker survives steering-token
  continuations (shared helper on both code paths), and the swarm module's
  self-contained re-entry rule means a fresh session cannot re-review
  converged Units or wedge settlement.
- Fixed context per session on this rebase: stage-protocol.md 16,657 to
  10,512 words; each harness SKILL roughly 8.5-8.9k to 6.8-7.0k. Combined,
  about 25.3k to 17.4k words always loaded, with about 16.1k words of module
  content loaded only when the directive asks for it.
- One deliberate wording change: Copilot's reviewer clause now states the
  canonical review_class contract instead of always-adversarial.

## Response to the 2026-08-17 review

(Historical note: the default-scope stance changed twice after this review —
first reverted to `feature` for the 2026-08-18 response, then, by maintainer
decision on 2026-08-19, `classic` became the implicit default as an
explicitly declared behavior change. See section 1 for the final state.)

- **P1 workshop removal:** made additive as requested. `workshop` is restored
  as a first-class scope (identity, keywords, Minimal override, grid, plugin
  participation, docs, fixtures, coverage registration); `classic` is a new
  keywordless scope; only the freeform default changed. Both upgrade paths in
  the review (overlay refresh and clean upgrade with in-flight workshop
  state) now work without migration, and the changelog no longer claims
  `--scope workshop` errors.
- **P1 express producer gaps:** every consumed-but-skipped producer now has
  an explicit express-named contract (fallback or conditional skip) in the
  stage body, and the engine classifies the 11 absences as designed. t291
  gained two deterministic engine journeys (no-deploy and deploy-tail) that
  drive `next`/`report` end to end: reviewer suppression, swarm exclusion,
  designed-absence classification per stage, lifecycle phase transitions, and
  workflow completion. (Re-slotted to t301 after the rebase.)
- **P1 swarm fresh-session re-review:** the settle directive carries an
  explicit `swarm_settled: true` marker, strips reviewer metadata, and keeps
  the swarm module; the SKILLs branch on the marker; regression coverage
  includes the settle shape (t201), the steering-continuation round trip
  (t201 3b), and schema validation (t113).
- **P2 invoke-swarm module loading:** all seven SKILLs now instruct loading
  every module in `directive.protocol_modules`, reviewer included; t292
  asserts the instruction text per harness alongside the emission checks.
- **P2 construction module unconditional ceremonies:** conditioned as
  requested (see section 3).
- **P2 express journey test and test-floor contradiction:** addressed via the
  express engine journeys and the Minimal-floor reconciliation above.
- **P3:** ensemble topology detail deduplicated out of the conductor persona
  (the module is the single contract; the persona keeps only the bus/no
  cross-agent-invocation invariants), and the docs now distinguish the
  underlying resolver fallback from the user-facing compose offer.
  The swarm halt-and-ask pointers already referenced
  `stage-protocol-construction.md` at the reviewed head.

## Response to the 2026-08-18 review

- **P1 implicit default:** resolved in this response by keeping `feature` as
  the implicit default — then **superseded on 2026-08-19 by maintainer
  decision**: `classic` is now the implicit default, shipped as the
  explicitly declared breaking default change this review named as the
  alternative path ("Otherwise, the release must explicitly declare this as a
  breaking default change"). The declaration lives in the changelog summary
  and first bullet, the `feat!` PR title, and section 1 above;
  `AWS_AIDLC_DEFAULT_SCOPE=feature` retains the previous behavior.
- **P1 Express test execution and coverage gate:** Build and Test Steps 10
  and 11 now collect the stage-level
  `construction/code-generation/unit-test-instructions.md` and
  `traceability.json` alongside the per-unit globs (with command dedup), so
  Express tests execute and Express traceability joins the final coverage
  gate, with an execution-level regression that extracts and actually runs a
  stage-level command and verifies its traceability target on disk.
- **P2 packager contract:** the assembled orchestrator skill location moved
  into a path-validated `orchestratorSkillPath` manifest field; the hardcoded
  candidate list is gone. (Per the 2026-08-19 review the field is now
  **optional** with the in-tree default — see below.) A regression iterates
  every `harness/*/manifest.ts` dynamically.
- **Rebase:** rebased onto the current v2 tip (v2.6.17, #792) and re-slotted
  to 2.6.18. The merged review-receipt recovery semantics are preserved: the
  v2 pin `t291-review-receipt-recovery` passes byte-untouched against this
  branch. This PR's new tests re-slotted past the taken numbers
  (t301 express journeys, t302 protocol modules, t303 packager contract).

## Response to the 2026-08-19 review

- **P1 generated runners and protocol modules:** both generated runner
  families now load the extracted contracts. `renderStageRunner()` gained a
  step that reads `stage-protocol.md` plus every
  `stage-protocol-<module>.md` named by `directive.protocol_modules` before
  acting on the directive, and the scope-runner loop does the same on every
  iteration. t302 gained assertions against the **rendered** stage- and
  scope-runner output (not just the seven orchestrator skills).
- **P2 `orchestratorSkillPath` compatibility:** the manifest field is
  optional again and defaults to `<harnessDir>/skills/aidlc/SKILL.md`;
  emit-owned layouts (e.g. Codex under `.agents/skills/`) declare it
  explicitly. Existing external manifests typecheck and package unchanged;
  t145's synthetic-harness regression exercises the default and t303 pins the
  fallback expression.
- **P2 zero-Unit routing:** now structural, not name-keyed. A shared helper
  derives stage-level fallback from the effective plan (state suffixes, then
  the scope grid) excluding the Unit-DAG producer, used by directive
  construction, per-unit iteration, and the artifact guard. A composed or
  plugin scope that runs Code Generation while skipping Units Generation gets
  stage-level paths with no `{unit-name}` placeholder (new t186 case) and
  passes the completion guard on stage-level artifacts (new t185 case). The
  previous "known limitation" note below is resolved.
- **P2 `/aidlc-init` default:** the generated init runner no longer bakes in
  any scope; it forwards the user's `--scope` when named and otherwise omits
  the flag, and `intent-create` itself now resolves `AWS_AIDLC_DEFAULT_SCOPE`
  and then the `classic` fallback (previously a hardcoded `poc` that ignored
  the env var). The documented precedence ladder and the shipped default now
  match actual behavior on every entry path — the accuracy gap this finding
  named is closed (with `classic`, per the declared default change, rather
  than `feature`).
- **P2 Kiro IDE gate description:** taken the offered option — the test is
  now described throughout (header, describe title, assertion comments,
  changelog) as a **hybrid** check: the live IDE half proves UserPromptSubmit
  records the HUMAN_TURN and a legitimate approval commits through Kiro; the
  fabricated second approval is a direct shipped-tool invocation that proves
  the core `humanActedSinceGate` guard and explicitly does not claim
  PreToolUse hook coverage.
- **P3 residual docs:** the stale `classic` fallback references and the
  ten-vs-eleven scope counts are fixed, plus a follow-up sweep for the
  default change (customization precedence ladder, orchestrator reference,
  scope authoring guide, plugin `freeform_default` wording, CLI commands,
  README).

### 4. Live Kiro IDE gate: Windows and macOS

- The live Kiro IDE pre-merge slice (`t-ide-kiro`, `AIDLC_KIRO_IDE_LIVE=1`)
  now runs on Windows (`%LOCALAPPDATA%\Programs\Kiro\Kiro.exe`) as well as
  macOS, with an `AIDLC_KIRO_IDE_BIN` override and
  `AIDLC_KIRO_IDE_CASE`/`AIDLC_KIRO_IDE_DIAGNOSTICS` controls; Linux skips
  cleanly. Test/docs only - no product code changes.
- The gate journey is constructed through the real shipped tools and the
  fabricated-approval half is deterministic: after the one human turn commits
  the open gate, the test builds the next gate for real and invokes the
  shipped `report --result approved` directly with no fresh `HUMAN_TURN` in
  the ledger; the core `humanActedSinceGate` check refuses it before any
  mutation. A steering-conformant conductor refuses to fabricate approvals
  itself, so the enforcement claim no longer depends on model cooperation.
- Proven live on a Windows host: 2/2 in 58.6s, with the legitimate approval
  committing in about 13 seconds and the fabricated approval refused on disk.

## Measured effect (live benchmark, honest summary)

The deterministic effect is the fixed-context word counts above. The live
effect is depth-dependent: matched-turn-index prompt sizes drop 6-9% over the
first ~25 turns (5/5 paired runs), but once a conditional module is triggered
it joins the session prefix and is re-paid on later turns, so long full-depth
runs gave the reduction back after the crossover (turn ~25-50). Run-level USD
was unresolved at the measured sample size (paired delta +0.3%). Read this as
a structural cleanup with a modest early-session win, not a large cost
reduction; the always-loaded word counts are the claim that holds
everywhere.

## Verification

- At this head (2026-08-19 fixes + declared default change): smoke + unit +
  integration with `--no-llm` — 340 files, 8,263 assertions, 0 failures;
  deterministic express e2e (t301, including the new classic-fallback pin)
  green; `bun scripts/package.ts --check` all seven trees in sync; typecheck
  (3 tsconfigs) and lint green.
- At the previously reviewed head (`fdd71c51`, v2.6.17 base): smoke+unit 236
  files, 6,843 assertions, 0 failures; the deterministic CI mirror
  (`--integration --e2e --no-llm`, the #791 job shape) 175 files, 0 failures;
  full integration (103 files) and deterministic e2e (32 files) green, with
  two isolated live-latency flakes each re-run green alone.
- Full live slice gate (run at `fdd71c51`, before the default change):
  Claude TUI 19/19, Kiro ACP 8/8, Kiro TUI 3/3, Codex exec 5/5 - including
  the compose-front journeys re-proven live against the 11-scope set - and,
  for the first time, Kiro IDE 2/2 live on Windows. Traces confirmed the
  then-current feature-default routing, classic/express selection, and the
  conditional modules loading exactly per trigger. The two commits since are
  covered by the deterministic tiers above; the env-scope live journey
  (t-tui-t29) is the targeted live re-check if one is wanted before merge.
- New tests: t301 (deterministic express engine journeys, no-deploy and
  deploy-tail, plus the classic-fallback pin), t302 (protocol module
  existence, directive hints including the express negative and the
  invoke-swarm path, SKILL references, rendered stage/scope runner loading),
  t303 (manifest-owned orchestrator skill path with the in-tree default),
  t201 settle-shape and continuation assertions, the stage-level execution
  regression, new t185/t186 structural zero-Unit cases, and the reworked
  t-ide-kiro live journey. t279 (from #613) re-pinned onto the reviewer
  module it now lives in.

## Notes for reviewers

- The packager now refreshes the generated stage/scope table regions in
  assembled orchestrator skills, and graph compile drops orphan grid columns
  so a removed stock scope cannot survive as a phantom composed scope
  (composer-authored scopes keep their identity files and still merge).
- `aidlc-graph validate-scope express` still prints 11 advisory lines: those
  are the designed absences described in section 2, now backed by explicit
  stage-contract fallbacks and asserted end to end.
- The zero-Unit engine path no longer keys on the express scope name — it is
  derived structurally from the effective plan (2026-08-19 round), so a
  composed scope that skips units-generation gets the stage-level path too.
- Live coverage gaps called out honestly: the swarm and recovery modules have
  no live e2e driver; express is not exercised as an active scope in the Kiro
  slices (it is on the Claude side via the benchmark journeys); the classic
  default landed after the full live gate run and is covered
  deterministically (t301, t67, t31, t191) but not yet by a live TUI journey.


core/tools/aidlc-orchestrate.ts
```diff
@@ -2971,12 +2990,12 @@ function handleNext(args: string[], projectDir: string | undefined): void {
     // Anchor the compose offer with the counts for the three named scopes so the
     // user calibrates the order-of-magnitude difference before deciding. Fall
     // back to bare names if any scope does not resolve.
-    const bf = scopeCostSummary("bugfix");
-    const poc = scopeCostSummary("poc");
+    const express = scopeCostSummary("express");
+    const classic = scopeCostSummary("classic");
```
leandrodamascena, line 3123: Please do not remove existing workflows, but add new ones.

leandrodamascena (changes requested):
@apackeer there is a breaking change and we need to fix it.

leandrodamascena (changes requested):
Arden, this PR addresses a real usability problem: the default AI-DLC lifecycle is expensive, and a smaller first-class workflow is valuable. I reviewed b78519d9 across scope routing, the Express lifecycle, protocol-module extraction, upgrade behavior, packaging, and the affected tests.

The central concern is that the PR is not additive. It combines two useful new workflows with deletion/replacement of an existing scope and removal of always-loaded protocol guarantees. That creates migration and behavioral regressions which are not covered by the current green CI.

I agree this should remain changes requested. I independently verified the following blockers:

**P1: The Workshop workflow is removed instead of preserving existing behavior**

The PR deletes:

- `core/scopes/aidlc-workshop.md`

and adds `classic` with the same workshop/lab/training keywords:

- `core/scopes/aidlc-classic.md:1-37`

This is not only a default change. `workshop` is a persisted scope name used by active state files, environment configuration, scripts, runners, and installed scope metadata.

There are two incompatible upgrade outcomes:

**Overlay upgrades retain the retired scope**

The documented installation and refresh commands use merge-copy operations such as:

- `README.md:129-133`
- equivalent `cp -R` instructions for the other copy-based harnesses

`cp -R` does not delete files removed by the new distribution. Therefore an existing `aidlc-workshop.md` remains after refresh.

Scope discovery treats every remaining scope file as installed:

- `core/tools/aidlc-lib.ts:6085-6163`

The graph compiler then treats the old grid column as a composed scope and preserves it when a matching installed scope file remains:

- `core/tools/aidlc-graph.ts:1941-1965`

I reproduced this upgrade shape: after overlaying the new shell onto an old installation, `workshop` remained selectable and retained its old Minimal test override. This contradicts the changelog statement that `--scope workshop` now errors and creates an undocumented hybrid shell.

**Clean upgrades strand active Workshop workflows**

A clean installation removes the scope file, but an in-flight workflow can still persist:

`- **Scope**: workshop`

The state schema remains v8, so the existing compatibility guard accepts it. Runtime scope resolution then reaches the unknown-scope path, with no in-band migration to `classic`.

This repository already recognizes that removing an active scope strands workflows when plugin selection changes; the same concern applies to a stock-scope rename.

Please make this additive:

- retain `workshop` as a supported scope or compatibility alias;
- add `classic` and `express` as new scopes;
- change the freeform default independently;
- if Workshop must eventually be removed, first ship an explicit state/config migration, deletion-aware installer behavior, doctor guidance, and an old-shell overlay regression.

**P1: Express executes stages whose required producers are absent from its grid**

Express advertises a 10-stage requirements-to-deploy lifecycle:

- `core/scopes/aidlc-express.md:19-34`

However, the selected stages retain required consumes and `requires_stage` contracts from the full lifecycle.

I ran:

`aidlc-graph validate-scope express`

It reported 11 required-consume edges whose producers are absent.

Code Generation requires Unit generation:

- `core/aidlc-common/stages/construction/code-generation.md:33-42`
- required `unit-of-work`
- required stage `units-generation`

Express skips Units Generation. The body has a fallback for existing incremental scopes, but Express is not named in that contract:

- `core/aidlc-common/stages/construction/code-generation.md:74-87`

Deployment Pipeline requires outputs from stages Express skips:

- `core/aidlc-common/stages/operation/deployment-pipeline.md:15-26`
- `ci-config`
- `quality-gates`
- `infrastructure-specification`
- `cicd-pipeline`
- required stages `ci-pipeline` and `infrastructure-design`

Its fallback is explicitly for brownfield incremental scopes such as `security-patch`, not the advertised Express greenfield path:

- `deployment-pipeline.md:51-57`

Deployment Execution requires an environment inventory from the skipped Environment Provisioning stage:

- `core/aidlc-common/stages/operation/deployment-execution.md:16-27`

Observability Setup requires five artifacts from skipped NFR and Infrastructure Design stages:

- `core/aidlc-common/stages/operation/observability-setup.md:18-32`

The graph validator labels these advisory, so CI remains green, but the runtime stage contracts still ask the conductor to consume artifacts Express cannot produce.

Please either:

- include the required producer stages;
- make these consumes genuinely optional for Express and define an explicit requirements/workspace-based fallback;
- or remove the unsupported deployment and observability tail from Express.

Add a real Express lifecycle test proving every executed stage can obtain its required inputs.

**P1: A fresh session after swarm convergence can repeat reviews and wedge settlement**

The protocol diet moves the autonomous swarm rule “do not review converged Units again” out of the always-loaded skill/protocol and into:

- `core/aidlc-common/protocols/stage-protocol-swarm.md:9`

When all swarm batches converge, the engine emits an ordinary settled `run-stage` directive:

- `core/tools/aidlc-orchestrate.ts:3331-3340`

That directive is built through the normal path. Because the stage declares a reviewer, it carries reviewer metadata and the normal reviewer/construction protocol modules:

- `core/tools/aidlc-orchestrate.ts:1964-1993`

It does not carry the `swarm` module.

In the original session, the conductor may remember that reviews already happened. In a fresh session, it sees a normal reviewer-bearing `run-stage`, loads reviewer/construction, and no longer receives the swarm-only instruction prohibiting another review. It can dispatch an additional review, exceed the stage’s iteration ceiling, and prevent settlement.

The existing swarm settlement test checks only:

- `kind`
- `stage`
- `unit`
- `gate`

at:

- `tests/unit/t201-swarm-batch-advance.test.ts:342-352`

It does not inspect protocol modules or simulate a fresh-session conductor after convergence.

Please preserve this behavior deterministically in the settle directive, for example by carrying an explicit settled-swarm marker/module and suppressing the normal reviewer branch. Add a fresh-session resume regression proving converged Units are never reviewed again.

**P2: `invoke-swarm` emits the reviewer module, but harness instructions do not load it**

The engine emits:

`["reviewer", "construction", "swarm"]`

for reviewer-bearing swarm directives:

- `core/tools/aidlc-orchestrate.ts:3376-3394`

However, each harness’s explicit `invoke-swarm` branch instructs the conductor to load only swarm and construction:

- `harness/opencode/skills/aidlc/SKILL.md:78`
- equivalent rows in the other six harnesses

The general “load every module in `directive.protocol_modules`” rule is written in the context of running a stage body, while `invoke-swarm` is a separate directive kind.

The swarm module refers to reviewer §12a, but does not reproduce its complete read-scope, dispatch-record, retry, repair, and terminal-ordering contract.

Please make the `invoke-swarm` instruction explicitly load every emitted module, including reviewer when present, and add a harness-level assertion for actual module consumption rather than emission only.

**P2: The Construction module gives unconditional reviewer and skeleton instructions that conflict with Express**

Express declares:

- `skeleton: off`
- `review_cap: none`
- no Units Generation or Unit DAG

at:

- `core/scopes/aidlc-express.md:1-10`

The extracted Construction module nevertheless states:

- “Walking-skeleton gate (first Bolt, always present)”
- ladder prompt immediately after the skeleton
- “run the body + reviewer”
- “The reviewer fires once PER UNIT”

at:

- `core/aidlc-common/protocols/stage-protocol-construction.md:9-30`
- `stage-protocol-construction.md:95-147`

These instructions must be conditional on the actual directive, effective reviewer, skeleton stance, and presence of a Unit DAG. Otherwise the lightweight/no-review scope loads a module that tells the conductor to perform ceremonies the engine intentionally removed.

Please qualify the module contract explicitly:

- reviewer work only when `directive.reviewer` is present;
- skeleton/ladder only when a real Bolt DAG and applicable skeleton stance exist;
- no Bolt/per-Unit ceremony for a single-iteration, zero-Unit scope such as Express.

**P2: The Express “journey” test does not execute an Express journey**

The new test:

- `tests/e2e/t291-express-scope-routing.test.ts:75-117`

verifies only:

- keyword detection;
- `intent-create`;
- persisted scope;
- static 10-stage grid;
- freeform fallback to Classic.

It never drives the engine through:

- initialization;
- Requirements Analysis;
- single-iteration Code Generation without Units;
- reviewer suppression;
- swarm exclusion;
- Build and Test;
- conditional deployment tail;
- required-input handling;
- phase transitions;
- workflow completion.

Consequently, all 11 producer gaps and the protocol contradictions remain invisible while the test and CI pass.

Please add an engine-level deterministic journey through the complete Express path, including both a no-deploy case and a deploy-tail case.

**P2: Express has contradictory test-floor instructions**

The organization rule says:

- `poc`, `refactor`, and `express` require only that the existing test suite remain green;
- no new-test floor is required.

at:

- `core/memory/org.md:51-57`

But Express uses Minimal depth, and Code Generation defines Minimal as:

- mandatory test files;
- one requirement-driven test per requirement;
- happy-path floor per component.

at:

- `core/aidlc-common/stages/construction/code-generation.md:104-113`

Please choose one contract and make the scope, organization rules, stage instructions, documentation, and tests agree. A lightweight workflow can still require a minimal new-test floor, but it should not promise both behaviors.

**P3: Some moved protocol references and context-reduction claims remain inconsistent**

- Swarm finalize-failure branches still point to the “Halt-and-ask” section in `stage-protocol.md`, although that heading moved to `stage-protocol-construction.md`.
- Significant ensemble topology prose remains duplicated in the always-delivered conductor persona and the conditional ensemble module, weakening the claimed fixed-context reduction and creating two sources that can drift.
- Scope documentation says “Everything else” resolves to Classic, while rich/no-keyword prose still enters the composition offer. The docs should distinguish the underlying fallback from the user-visible compose path.

**Recommendation**

The three independently useful changes should be separated:

1. Add `express` with a valid end-to-end lifecycle and journey tests.
2. Add `classic` and change the default while preserving `workshop` compatibility.
3. Modularize the protocol independently, with behavioral-equivalence and fresh-session swarm-resume tests.

That would keep the valuable additions reviewable and prevent scope-migration bugs from being coupled to protocol-semantic regressions.

The CI and package parity are green, but the missing coverage is behavioral rather than structural. I recommend keeping this changes requested.

apackeer:
Thanks for the thorough review @leandrodamascena - every point is addressed in the new revision (rebased onto v2.6.8, now 2.6.9; the branch is mergeable again). The PR body has a point-by-point "Response to the 2026-08-17 review" section; the short version:

- **Workshop (P1):** the PR is now fully additive. `workshop` is back as a first-class scope with its original keywords, Minimal override, and a byte-identical grid (the restored mr9-parity fixtures match the pre-PR ones exactly). `classic` is a new keyword-less scope and only the freeform default changed, so both upgrade paths you traced (overlay refresh and in-flight `Scope: workshop` state) work with no migration, and the changelog no longer claims `--scope workshop` errors.
- **Express producers (P1):** the four consuming stages now carry explicit express-named fallbacks or conditional-skip contracts, the engine classifies the 11 absences as designed (`expected: true`), and t291 grew two deterministic engine journeys (no-deploy and deploy-tail) driving `next`/`report` end to end: reviewer suppression, swarm exclusion, per-stage designed-absence classification, lifecycle phase transitions, and workflow completion.
- **Swarm fresh-session (P1):** the settle directive now carries `swarm_settled: true`, strips the reviewer fields, and pins `protocol_modules` to `["construction","swarm"]`; the marker survives steering-token continuations via a shared helper, and t201 (including a new continue round-trip test) plus t113 pin the shape.
- **P2s:** every harness's invoke-swarm branch loads all emitted modules including reviewer (asserted per harness in t292); the Construction module's reviewer/skeleton/per-Unit ceremonies are conditioned on the actual directive, reviewer presence, and a real Unit DAG; the express test-floor contradiction is resolved in favor of the Minimal requirement-driven floor everywhere.
- **P3:** the ensemble topology prose is deduplicated out of the conductor persona, and the docs now distinguish the underlying classic resolver fallback from the user-facing compose offer. One note: the swarm halt-and-ask references already pointed at `stage-protocol-construction.md` at the head you reviewed, so no change was needed there.

Verification on the rebased head: smoke+unit 219 files / 0 failures, deterministic e2e (including the new express journeys) and the protocol-structure integration slice green, package parity, typecheck, and lint all green; details in the body's Verification section. Would appreciate a re-review when you have a chance.

leandrodamascena (changes requested):
Arden, I re-reviewed `2d2b95cd` against every finding from my previous review. The prior issues are largely resolved: Workshop is restored, Express producer absences are modeled, swarm settlement survives fresh sessions, emitted protocol modules are loaded by all seven harnesses, Construction ceremonies are conditional, and the Express test floor is consistent.

I found three remaining blockers:

**P1: Changing the implicit default from `feature` to `classic` is a breaking behavior change**

- `core/tools/aidlc-lib.ts:6333`
- `core/tools/aidlc-orchestrate.ts:191`
- `tests/e2e/t291-express-scope-routing.test.ts:263`

Calls without an explicit scope now select a materially different lifecycle that skips Ideation. Escape hatches do not make this backward-compatible. To remain non-breaking, keep `feature` as the default and make `classic` opt-in. Otherwise, the release must explicitly declare this as a breaking default change.

**P1: Express can omit execution and coverage validation of its generated tests**

- `core/aidlc-common/stages/construction/build-and-test.md:116-149`

Express writes Code Generation artifacts at the stage-level path:

```text
construction/code-generation/
```

But Build & Test collects only:

```text
construction/*/code-generation/unit-test-instructions.md
construction/*/code-generation/traceability.json
```

Therefore, Express may neither execute its required Minimal tests nor include its traceability in the final coverage gate. The current Express journey creates placeholder artifacts and does not exercise these commands.

Please include both stage-level files in Steps 10 and 11 and add an execution-level regression.

**P2: The packager narrows the open harness contract**

- `scripts/package.ts:729-740`
- `scripts/manifest-types.ts:67-124`

`assembledOrchestratorSkill()` hardcodes three skill locations. A valid future/custom harness using `skipRunnerGen` and emitting its orchestrator skill elsewhere satisfies the manifest contract but now fails packaging.

Please declare the assembled orchestrator-skill path in the harness manifest or discover the emitted skill generically.

**Rebase blocker**

The branch conflicts with current `v2` across the reviewer protocol/recovery changes, Testing Posture memory, harness skills, documentation, coverage, and version files. This is not a mechanical version-only rebase: the extracted reviewer module must retain the recently merged recovery semantics, and the result needs the focused and full validation rerun. The current `2.6.9` slot is also stale.

The previous review findings are substantively addressed, but I recommend keeping **Changes Requested** until these items are resolved and the branch is rebased onto current `v2`.

apackeer:
@leandrodamascena this is ready for another look. All three blockers from your 2026-08-18 review are addressed, plus the rebase:

- **Implicit default:** `feature` stays the default everywhere (engine, help, shipped settings, docs); `classic` is strictly opt-in, and an e2e journey now pins the feature fallback.
- **Express Build and Test:** Steps 10 and 11 collect the stage-level `code-generation` instruction and traceability files alongside the per-unit globs, with an execution-level regression that runs a real stage-level command.
- **Packager:** the orchestrator-skill location is a required `orchestratorSkillPath` manifest field on all seven harnesses; the hardcoded list is gone, with a per-harness regression.
- **Rebase:** onto the current v2 tip (v2.6.17), re-slotted to 2.6.18; the merged review-receipt recovery pin passes byte-untouched; new tests renumbered to t301/t302/t303.

Since your review the PR also gained the live Kiro IDE gate on Windows (test/docs only), and the full validation you asked for has been rerun: deterministic tiers green (smoke+unit 236/0, the #791-shape `--no-llm` mirror 175/0) plus the complete live gate - all 8 slices green, including Kiro IDE 2/2 live on Windows. The PR title and body are updated with the point-by-point response and current evidence.

leandrodamascena (changes requested):
Arden, I re-reviewed `fdd71c51` against every previous finding and the current `v2.6.17` base. The original blockers are resolved: Feature remains the normal implicit default, Workshop is preserved, Express includes stage-level test execution and traceability, swarm settlement survives fresh sessions, the PR #758 recovery semantics remain intact, and the packager no longer hardcodes three built-in skill locations.

The focused unit, deterministic E2E, protocol integration, typecheck, lint, parity, coverage, and remote CI checks are green.

I found one remaining blocker and several compatibility/documentation issues:

**P1: Generated stage and scope runners do not load extracted protocol modules**

- `core/tools/aidlc-runner-gen.ts:136-188`
- `core/tools/aidlc-runner-gen.ts:613-638`

Generated runners are standalone skills. Invoking `/aidlc-code-generation`, `/aidlc-express`, or another generated runner does not automatically load the main `aidlc` orchestrator skill.

The engine emits `directive.protocol_modules`, but the generated runners never read the corresponding files:

```text
stage-protocol-reviewer.md
stage-protocol-ensemble.md
stage-protocol-construction.md
stage-protocol-swarm.md
```

Consequently, direct runners can execute without the reviewer, ensemble, Construction, or swarm contracts extracted by this PR. `t302` checks only the seven main harness orchestrators, not generated stage/scope runners.

Please update both `renderStageRunner()` and the scope-runner renderer to read the base protocol and every module named by `directive.protocol_modules` before acting. Add rendered-runner assertions to `t302`.

**P2: `orchestratorSkillPath` is a breaking manifest API change**

- `scripts/manifest-types.ts:67-77`
- `scripts/package.ts:736-750`

The field is now mandatory. Existing external/custom harness manifests fail typechecking or reach `isAbsolute(undefined)`.

For backward compatibility, make the field optional and default it to:

```text
<harnessDir>/skills/aidlc/SKILL.md
```

Harnesses with nonstandard emitted layouts can set it explicitly. Otherwise, document this as a breaking harness migration.

**P2: Zero-Unit routing remains hardcoded to Express**

- `core/tools/aidlc-orchestrate.ts:1981-1983,3971-4008`
- `core/tools/aidlc-state.ts:1491-1503`

A plugin/composed scope that runs Code Generation while skipping Units Generation still receives `{unit-name}` paths. This conflicts with the new generic protocol statement that a missing Unit DAG uses stage-level execution.

Please either derive zero-Unit behavior structurally or scope the protocol wording explicitly to Express.

**P2: `/aidlc-init` still does not use the Feature default**

`core/tools/aidlc-runner-gen.ts:240-245` intentionally omits `--scope`, while `core/tools/aidlc-utility.ts:3844` defaults direct `intent-create` to `poc`.

Therefore, `/aidlc-init <description>` creates a POC workflow despite the PR claim that Feature remains the implicit default everywhere. This behavior is pre-existing, but the stated compatibility guarantee remains inaccurate.

**P2: The Kiro IDE “live gate” test bypasses Kiro for the fabricated approval**

- `tests/e2e/t-ide-kiro-checkpoint.serial.test.ts:376-419`

The second approval is invoked directly through `spawnSync(aidlc-orchestrate.ts report)`, outside Kiro and its PreToolUse hooks. This proves the core human-presence guard, not Kiro IDE hook integration. The comment also says the PreToolUse hook blocked the call, although the call bypassed it.

Please either drive that attempt through Kiro or describe the test as a hybrid core-enforcement check.

**P3: Residual documentation contradicts the retained Feature default**

- `docs/guide/13-customization.md:95-98` says unmatched input falls back to `classic`.
- `docs/reference/18-plugin-mechanism.md:171-175` refers to the core `classic` fallback.
- `docs/harness-engineering/04-scopes.md:3,7,146` still says ten scopes instead of eleven.

The previous review findings are substantively addressed, but generated runners are now incomplete after protocol extraction. I recommend keeping **Changes Requested** until that blocker and the compatibility/documentation inconsistencies are resolved.

apackeer:
@leandrodamascena this is ready for another look. All findings from your 2026-08-19 review are addressed, and the default-scope question has been re-decided the other way — this time via the path your 2026-08-18 review named, as an explicitly declared change.

**Your 2026-08-19 findings:**

- **P1 generated runners:** both `renderStageRunner()` and the scope-runner renderer now read `stage-protocol.md` plus every `stage-protocol-<module>.md` named by `directive.protocol_modules` before acting on the directive; t302 now asserts this against the **rendered** stage- and scope-runner output, not just the seven orchestrator skills.
- **P2 `orchestratorSkillPath`:** optional again, defaulting to `<harnessDir>/skills/aidlc/SKILL.md`; only emit-owned layouts (e.g. Codex under `.agents/skills/`) declare it. Existing external manifests typecheck and package unchanged — t145's synthetic harness exercises the default and t303 pins the fallback expression.
- **P2 zero-Unit routing:** structural now. A shared helper derives stage-level fallback from the effective plan excluding the Unit-DAG producer (state stage suffixes, then the scope grid), used by directive construction, per-unit iteration, and the artifact guard — no scope-name key. New regressions: a composed plan that skips Units Generation gets stage-level paths with no `{unit-name}` placeholder (t186) and passes the completion guard on stage-level artifacts (t185).
- **P2 `/aidlc-init`:** the generated runner no longer bakes in any scope; it forwards the user's `--scope` when named and otherwise omits the flag, and `intent-create` itself now resolves `AWS_AIDLC_DEFAULT_SCOPE` and then the hard-coded default (previously a hardcoded `poc` that ignored the env var). Documented precedence and actual behavior now match on every entry path.
- **P2 Kiro IDE gate:** took the option you offered — the test is described throughout (header, describe title, assertion comments, changelog) as a hybrid check: the live IDE half proves UserPromptSubmit records the HUMAN_TURN and a legitimate approval commits through Kiro; the fabricated second approval is a direct shipped-tool invocation proving the core `humanActedSinceGate` guard, explicitly not PreToolUse hook coverage.
- **P3 docs:** the stale `classic`-fallback references and ten-vs-eleven counts are fixed, plus a follow-up sweep.

**Default scope — maintainer decision, declared:** after weighing it again I'm taking the alternative your 2026-08-18 review left open ("the release must explicitly declare this as a breaking default change"): `classic` is now the implicit default. Exactly two things control it — `AWS_AIDLC_DEFAULT_SCOPE` (which overrides; the shipped Claude settings set it to `classic`) and a single `DEFAULT_SCOPE = "classic"` constant. The scattered `feature`/`poc` fallbacks are gone. Explicit `--scope`, keyword detection, and the compose offer are untouched; in-flight workflows keep their persisted scope; `AWS_AIDLC_DEFAULT_SCOPE=feature` retains the previous behavior. Declared in the `feat!` title and the changelog summary + first bullet, and pinned by t301/t67/t31/t191. The PR body's sections 1 and the review responses are updated to match.

**Verification at this head:** smoke+unit+integration `--no-llm` — 340 files, 8,263 assertions, 0 failures; deterministic express e2e (including the new classic-fallback pin) green; `package.ts --check` all seven trees in sync; typecheck (3 tsconfigs) and lint green. The full live gate ran at `fdd71c51` (before the default change); the two commits since are covered deterministically, and I'll run the targeted env-scope live journey (t-tui-t29) before merge if you'd like it on record.

leandrodamascena (approved):
Arden, I re-reviewed `daf2acca` against all previous findings. The generated runners, manifest compatibility, structural zero-Unit routing, default resolution, Kiro hybrid-test wording, Workshop compatibility, Express artifacts, swarm settlement, and reviewer recovery are now correctly addressed and covered.

The Classic default is now an explicit breaking behavior change with a documented `AWS_AIDLC_DEFAULT_SCOPE=feature` upgrade path. Existing workflows retain their persisted scope.

Focused unit, integration, Express E2E, package parity, typecheck, lint, coverage, and remote CI are green.

A few stale Feature-fallback references remain, but they are nonblocking documentation/defensive cleanup and do not affect valid workflows.

No blocking findings remain. Approved for merge.

Issue #273, opened 2026-05-17
[Bug]: No quality-target verification gate — measurable NFR targets (e.g. coverage thresholds) can be silently relaxed in Code Generation / Build & Test


Mameta29:
## Which rule or stage is affected

- `construction/code-generation.md`
- `construction/build-and-test.md`
- `construction/nfr-requirements.md` (origin of the targets)

## Summary

Measurable quality targets defined in **NFR Requirements** (e.g. test coverage thresholds, performance budgets) are written into `tech-stack-decisions.md`, but **no downstream stage is required to verify that those targets were actually met.** As a result, an AI agent can satisfy every rule while a unit silently misses its own defined quality bar — or even relax the target itself to make a step "pass".

## Expected vs actual behavior

**Expected**: When NFR Requirements defines a measurable quality target for a unit, a later stage verifies the target was met before the unit is considered complete, and the agent is explicitly forbidden from weakening the target instead of meeting it.

**Actual**: No such verification exists.

- `code-generation.md` Step 11 / Critical Rules — no mention of quality targets. Completion Criteria only requires that tests *are generated*, not that they meet any standard. Line ~215 defers test execution entirely to Build & Test.
- `build-and-test.md` — this stage *generates test-execution instructions* (markdown with `[Command to run tests]` placeholders); it does not execute tests or compare results against NFR targets. The `build-and-test-summary.md` template has a `Coverage: [X]%` field but no step fills `[X]` or checks it against the target.

So across the entire Construction phase there is **no gate** that confirms a defined, measurable quality target was achieved.

## Reproduction (observed in a real run)

Platform: Claude-based coding agent, greenfield monorepo (TypeScript / pnpm).

1. In **NFR Requirements** for a unit, define a coverage target (e.g. Statements 90% / Branches 85%). It is recorded in `nfr-requirements.md` / `tech-stack-decisions.md`.
2. In **Code Generation**, the agent generates code + tests. Actual coverage comes out lower (e.g. Statements ~73% / Branches ~67%).
3. To "pass", the agent **lowers the coverage threshold in the test config** from 90/85 to 70/65 and reports the step as complete and compliant.
4. Nothing in the rules flags this. It is only caught by human review.

The agent did not violate any rule — because no rule covers this.

## Why this matters

[...]

PR #772, opened 2026-08-16
feat: code-generation plan honours the affirmed Testing Posture (2.6.16)

### LINKED ISSUE #724: [Feature]: code-generation plan ignores the affirmed `## Testing Posture` — TDD teams get an implementation-first plan
### Description

`practices-discovery` interviews the team on testing methodology and writes the answer to `rules/aidlc-team.md` `## Testing Posture`. But nothing in the Construction code-generation path is scoped to read that section, so a team that affirms TDD still gets an implementation-first plan.

Two independent places narrow the read away from it:

**1. The lead agent is scoped to `## Code Style` only.**

`code-generation.md` is `lead_agent: aidlc-developer-agent`, `mode: subagent`. That agent's Knowledge Loading step 4 (`agents/aidlc-developer-agent.md:73`) names exactly one practices section:

> 4. `{{HARNESS_DIR}}/rules/` — team-affirmed practices (read per `knowledge/aidlc-shared/rules-reading.md` fallback chain `team.md → org.md → hardcoded defaults`). Consult **`## Code Style`** for type-hint requirements, formatter / linter expectations, and any team-specific code conventions.

Every other agent that names a section names the one relevant to its job — `aidlc-quality-agent.md:72` consults `## Testing` for "TDD/BDD/red-green-refactor cadence", `aidlc-product-agent.md:80` consults `## Walking Skeleton` and `## Testing`. The developer-agent — the only agent that actually authors the code-generation plan — is the one told to consult the styling section and not the testing one.

**2. The stage file itself never mentions it.** Grepping `aidlc-common/stages/construction/code-generation.md` for `aidlc-team`, `Testing Posture`, `rules-reading`, or `practices` returns 0 matches. Its only `{{HARNESS_DIR}}/rules/` references are L220-221, which are the `## Learn` **write** targets (identical boilerplate in all 30 stage files), not a read path. The one policy source it does consult is L92 → `stage-protocol.md` §8 "Test Strategy", which governs test **volume** (Minimal ~5-15 total / Standard 5-8 per component / Comprehensive 10-15 per component) and says nothing about ordering.

`stage-protocol.md:608` does say `{{HARNESS_DIR}}/rules/` is loaded `(always)`, so the files are in context. The gap is that no instruction connects the affirmed testing posture to the plan's step ordering — so the hardcoded template wins by default.

**Result** — `code-generation.md` L105-122:

```
Step 3: Business logic layer (core domain logic, services)
Step 4: Business logic tests (unit tests for Step 3)
Step 5: API / endpoint layer (routes, controllers, handlers)
Step 6: API tests (unit + integration tests for Step 5)
```

`unit tests for Step 3` frames tests as derived from an implementation that already exists. Part 2 follows the plan, so a team that affirmed TDD gets a plan with no failing-test step in it — red-green-refactor never happens, and nothing in the run reports a deviation.

### Use Case

A team affirms TDD at the practices-discovery gate and expects Construction to honour it. Today the affirmation is recorded but has no effect on generated output, and the mismatch is silent — the plan looks well-formed, tests exist, coverage passes. The only visible symptom is that tests were written after the code they cover, which is exactly what the team said it did not want.

Concretely: have the code-generation plan step read the affirmed testing posture, and when it is test-first, split each layer instead of appending a test step after it:

```
Step 3a: Business logic — Red      (write the failing tests, run them, record the failure output)
Step 3b: Business logic — Green    (implement until they pass)
Step 3c: Business logic — Refactor
Step 4a: API — Red
Step 4b: API — Green
Step 4c: API — Refactor
```

This preserves the rationale already stated at L122 — `dependencies are built before dependents (data models before business logic, business logic before API)` — because the split happens *within* a layer, never across layers. Teams on classic test-after keep today's ordering unchanged.

Recording the actual test-runner failure output in the Red step would make the ordering verifiable rather than self-declared: a checkbox can be ticked retroactively, but a genuine FAIL output cannot be produced once the implementation already passes. That gives #273 a concrete artifact to gate on.

### Additional Context

Three smaller inconsistencies around the same section, all cheap to fix alongside:

- **Heading mismatch.** `aidlc-quality-agent.md:72` and `aidlc-product-agent.md:80` say to consult `` `## Testing` ``, but the actual heading in all three rules files (`aidlc-team.md:19`, `aidlc-org.md:44`, `aidlc-project.md:29`) is `## Testing Posture`. `extractMarkdownSection` (`tools/aidlc-lib.ts:1905`) matches with an anchored regex (`^## Testing$`), so the literal string in those agent files resolves to nothing. `rules-reading.md` §2 does prescribe topic-matching with a `contains "test"` fallback, so a careful reader recovers — but the two agent files should just name `## Testing Posture`.

- **`aidlc-team.md:21` attributes the wrong stage.** The `## Testing Posture` placeholder comment reads `<!-- Affirmed at requirements-analysis. Example: -->` while all four sibling sections read `Affirmed during practices-discovery`. `practices-discovery.md:85` and `:141` (and `tools/aidlc-state.ts:1376`) confirm practices-discovery owns it.

- **`rules-reading.md:109` layer-3 default is unreachable.** `Testing Posture | TDD encouraged but not enforced` sits in the hardcoded-defaults table, described at L98 as `used only when both layers are empty`, while layer 2 (`aidlc-org.md`) is documented at L97 as `Always populated`. So the effective default is `aidlc-org.md:50-52` — `tests written alongside code; minimum 80% line coverage` — which never mentions ordering. Worth noting because "TDD encouraged" reads like a shipped stance but in practice never applies.

`aidlc-org.md:46-48` frames all of this as intentional-for-now:

> Specific methodology — TDD, BDD, ATDD, or classic test-after — is captured by the testing-strategy stage when it ships.

Since `practices-discovery` already asks the question and already persists the answer, the capture side exists and only the consumption side is missing — which is why this reads as missing wiring rather than a missing opinion about TDD.

**Related**
- #397 quotes the same recommended-structure steps (`Step 4: Business logic tests`, `Step 6: API tests`) as evidence of duplication between `build-and-test` and `code-generation` — its line numbers predate some drift, but the steps are the ones above. Same lines, different concern; if that fix moves `unit-test-instructions.md` into `code-generation`, the ordering change lands naturally alongside it.
- #273 — no quality-target verification gate. Same shape: a target is declared upstream and nothing downstream verifies it.
- #631 — code-generation plan produced after the code. Also about the plan losing authority over generation, from a different angle.

Verified against `core/` on the `v2` branch at `f3ce1b8`, `AIDLC_VERSION = "2.0.2"`. Harness-independent — all citations are `core/` source paths.

apackeer:
## Summary

Fixes #724 (excellent report by @yoshimi-I; all findings reproduced at current HEAD before building this).

`practices-discovery` interviews the team on testing methodology and promotes the answer to the active space's `memory/team.md` `## Testing Posture`, but nothing in the Construction code-generation path read it back. The developer agent's memory consult named only `## Code Style`, and the stage's recommended plan structure hardcoded implementation-then-tests ordering. A team that affirmed TDD got an implementation-first plan, and nothing in the run reported the deviation.

## What changed

- **`code-generation.md` Step 2 (Part 1, planning)** resolves `## Testing Posture` from `aidlc/spaces/<active-space>/memory/{project,team,org}.md` by the most-specific non-empty statement (the `org.md` default applies when practices-discovery was skipped), mirroring the delivery-planning idiom. The posture governs whether a layer's tests precede or follow its implementation; the section 8 test strategy still governs volume. Neither overrides the other.
- **Both plan shapes ship in the stage file.** Test-after / unspecified postures keep the existing 12-step structure unchanged. Test-first postures (TDD, BDD, ATDD) split each layer into Red / Green / Refactor sub-steps *within* the layer, so the dependencies-before-dependents cross-layer ordering is untouched.
- **Step 4's delegation prompt forwards the resolved posture** to the Part 2 generation subagent; where the workspace can already run the test suite, the subagent records each Red step's failing output in the plan checkbox note before implementing.
- **`aidlc-developer-agent.md`** consults `## Testing Posture` alongside `## Code Style` (the quality, product, and pipeline-deploy agents already consulted it; the one persona that generates code did not).
- **`memory/org.md`** stops deferring methodology to a "testing-strategy stage" that never shipped; practices-discovery owns the capture and Code Generation now consumes it.
- **Docs**: `docs/reference/04-stages/construction.md` mirrors the stage change; `docs/harness-engineering/08` re-pins the org.md line citation this change shifted.
- **`tests/unit/t299-testing-posture-wiring.test.ts`** pins the wiring: stage consult + both plan shapes, dispatch bullet, agent consult, org.md attribution.

## Design decisions

1. **The wiring lives in the stage file, not only the agent persona.** On this `mode: subagent` stage the conductor authors Part 1 without the lead persona in context (`inlineAgentsFor` returns none for subagent stages, and `aidlc-plan-approval-guard.ts` blocks any developer-agent dispatch before a non-empty approved plan exists), so the stage prose is the plan author's only knowledge channel. The agent-side consult covers Part 2 execution.
2. **No `consumes: team-practices` frontmatter**, deliberately. t65's cross-phase check walks every `consumes[]` entry (including `required: false`) and requires the producer in the consumer's transitive `requires_stage` closure; practices-discovery is in no stage's closure, and its scope list lacks the incremental scopes code-generation carries. Semantically the artifact is also the pre-promotion draft, while this change reads the promoted memory layer that is already delivered on every stage. Prose-only consumption is the established pattern (delivery-planning resolves Way of Working / Walking Skeleton / Deployment the same way). A welcome side effect: the change is prose-only, so the compiled stage graph is byte-identical.
3. **Red-step failing-output capture is best-effort, not a gate.** Code-generation installs nothing and runs nothing today (build-and-test Step 10 owns the first execution), so an unconditional "run the tests" would fail every greenfield unit. Deterministic enforcement of quality targets stays with #273, which this gives a concrete artifact to gate on when the workspace supports it.

## Known tension, out of scope

`stage-protocol.md` section 5 says `mode: subagent` dispatches "the lead for the draft", while this stage's body plus the plan-approval guard have the conductor author the plan and dispatch the lead only for generation. This PR wires the posture into whoever authors the plan (the stage body travels on either reading) and does not take a position on the seam; that discussion is adjacent to #631. The reporter's third secondary point (the unreachable layer-4 hardcoded default in rules-reading.md) is also deliberately untouched.

## Testing

- `bun scripts/package.ts --check` clean (all 7 harness trees), `bun run typecheck` clean.
- t299 (new) and t68 green.
- Full smoke+unit+integration tier on the rebased tree: 319/321 files green (7482/7484 assertions). The two reds are proven non-branch: `t145-packaging-parity` is this box's `/tmp` smol-toml resolution artifact (fingerprint matched verbatim), and `t238-user-stories-mob.sdk` is a documented pre-existing red on clean v2 with zero file overlap with this diff (it also passed green solo on this branch pre-rebase).

Version 2.6.16, test slot t299 (re-slotted during landing as v2 advanced; originally 2.6.8/t292).

Note for the merger: this repo's PRs target `v2`, which is not the GitHub default branch, so the `Fixes #724` reference will not auto-close the issue on merge; it needs a manual close.


leandrodamascena (changes requested):
Arden, this PR addresses a real gap: Practices Discovery records the team’s Testing Posture, but Code Generation did not consistently use it when planning or generating code. I reviewed 7d7a85c1 across posture resolution, plan structure, Plan Approval, developer dispatch, autonomous swarm behavior, documentation, packaging, and the new tests.

The focused official suites pass with 53/53 assertions, and package/coverage parity is green. However, the current implementation wires posture prose into the stage without yet preserving the affirmed methodology across all resolution and execution paths.

I agree this should remain changes requested. I independently verified the following issues:

**P1: A project-level note can silently erase the team’s affirmed TDD posture**

Code Generation resolves Testing Posture by taking the first non-empty section in this order:

`project → team → org`

at:

- `core/aidlc-common/stages/construction/code-generation.md:114`

This selects the entire narrowest section, not the specific statement that governs test ordering.

For example:

- team.md: `Use TDD`
- project.md: `The legacy adapter requires an integration regression suite`

The project section wins because it is non-empty, but it does not specify tests-before or tests-after. The plan therefore falls into the “unspecified” test-after shape and silently loses the team’s TDD decision.

This conflicts with the strict-additive rule model, where a narrower non-conflicting specialization does not erase broader applicable rules:

- `core/knowledge/aidlc-shared/rules-reading.md:91-108`
- `docs/reference/08-rule-system.md:46-54`

Please resolve the explicit methodology/cadence statement independently from coverage, tooling, or scope notes. A structured posture field or additive classifier would avoid treating any non-empty project note as a complete replacement for the team posture.

Add regressions for partial specialization, including a project-level coverage/integration note layered on top of team-level TDD.

**P1: BDD and ATDD are incorrectly collapsed into layer-local TDD**

The stage treats every “test-first posture” — TDD, BDD, and ATDD — as the same Red/Green/Refactor sequence inside each implementation layer:

- `core/aidlc-common/stages/construction/code-generation.md:135-148`

It also requires:

`The split happens within a layer, never across layers.`

That is a TDD-shaped interpretation, not a general implementation of BDD or ATDD.

ATDD commonly requires acceptance tests to be written before the complete feature implementation. Those tests can span data, repository, business, API, and UI layers.

BDD similarly starts from feature scenarios and observable behavior; a team may affirm scenario-first BDD while still writing lower-level unit tests after implementation.

A posture such as:

`BDD scenarios first; unit tests after implementation`

would be incorrectly converted into Red/Green/Refactor for every layer.

Please distinguish the supported methodologies explicitly:

- TDD: unit-level Red/Green/Refactor;
- BDD: behavior/scenario-first slices;
- ATDD: acceptance tests before the feature implementation;
- mixed/custom posture: preserve the team’s stated ordering rather than coercing it into TDD.

Add behavioral plan-generation cases for each methodology.

**P1: Autonomous swarm still bypasses Testing Posture and Plan Approval**

The normal Code Generation flow resolves Testing Posture during Part 1, creates the plan and unit-test instructions, gets human Plan Approval, and forwards the resolved posture to the developer agent.

The autonomous swarm path does not execute that flow. `invoke-swarm` fans out developer agents directly from the harness orchestration path:

- `harness/*/skills/aidlc/SKILL.md`, `invoke-swarm` branch
- `tests/integration/t135-invoke-swarm.test.ts:509`

Those worker briefs do not carry an approved Code Generation plan or the resolved Testing Posture.

This means a team that affirmed TDD can still receive implementation-first code whenever autonomous swarm routing is active. The fix therefore does not cover one of the shipped Construction Code Generation paths.

Please require approved per-unit planning evidence before swarm dispatch, or carry an equivalent approved plan/posture contract into every worker brief. Add a regression proving autonomous workers receive and follow the same posture as the normal Code Generation path.

**P2: The test-first plan covers only Business Logic and API layers**

The test-after plan includes:

- data models/database;
- business logic;
- API;
- repository/data access;
- frontend.

But the test-first plan defines Red/Green/Refactor only for:

- Business Logic (`3a`, `3b`, `3c`);
- API (`5a`, `5b`, `5c`).

at:

- `core/aidlc-common/stages/construction/code-generation.md:121-146`

Repository/data access and frontend still have no tests-first sequence, and data-model behavior is also omitted. This contradicts the PR’s “each layer” claim in the stage, documentation, and changelog.

The new test explicitly checks only steps `3a-c` and `5a-c`:

- `tests/unit/t292-testing-posture-wiring.test.ts:73-85`

Please either define posture-aware ordering for every testable layer or narrow the stated guarantee to the layers actually covered.

**P2: Greenfield projects cannot reliably execute the specified Green step**

The Red/Green work occurs at steps 3 and 5, but Test Configuration is not created until Step 11:

- `core/aidlc-common/stages/construction/code-generation.md:121-144`

In a greenfield Vitest/Jest/etc. project, the agent may be able to write a Red test, but it cannot reliably run that test or implement “until it passes” before the runner and configuration exist.

The best-effort exception for recording Red output avoids a hard failure, but it does not make the Green step executable.

Please move minimal test-runner setup before the first Red step, or explicitly define a bootstrap phase for greenfield projects. Add a greenfield plan regression proving the first Red/Green cycle has a runnable test command.

**P2: Part 1 and Part 2 can resolve different effective postures**

The conductor resolves the posture in Step 2 and forwards it in the delegation prompt:

- `code-generation.md:114`
- `code-generation.md:205-216`

The developer agent also independently reloads and interprets the same memory chain:

- `core/agents/aidlc-developer-agent.md:69-77`

No source is declared authoritative if the delegated prompt and the subagent’s fresh memory resolution differ.

This can happen when:

- memory changes during the mandatory Plan Approval stop;
- conductor and subagent interpret an unstructured multiline section differently;
- project/team comments or partial specializations select different statements.

Please make the posture captured in the approved plan/delegation contract authoritative for Part 2, and treat a changed memory posture as invalidating/reopening Plan Approval rather than silently changing execution behavior.

**P2: The developer persona contradicts mandatory Refactor steps**

The stage requires a Refactor step after every Green step:

- `code-generation.md:135-146`

But the developer persona says:

`Refactor in subsequent iterations, not during initial generation.`

at:

- `core/agents/aidlc-developer-agent.md:81`

The subagent therefore receives conflicting instructions in the same dispatch. Please align the persona with the posture-aware stage contract, making Refactor conditional on the resolved methodology.

**P2: Test Strategy and Testing Posture still produce conflicting obligations**

The stage requires test files and strategy-driven volume:

- `core/aidlc-common/stages/construction/code-generation.md:103-112`

The org Testing Posture says:

- bugfix/security-patch require only the targeted regression;
- poc/refactor/workshop have no new-test floor.

at:

- `core/memory/org.md:52-58`

The PR says posture governs ordering while Test Strategy governs volume and neither overrides the other, but the shipped defaults still state incompatible minimums.

Please define the combination rule explicitly for each scope/strategy pair and test it, rather than leaving the conductor to reconcile conflicting prose.

**P2: The tests prove phrase presence, not behavior**

`t292-testing-posture-wiring.test.ts` reads authored files and asserts that expected strings exist:

- `tests/unit/t292-testing-posture-wiring.test.ts:53-155`

It does not:

- resolve project/team/org posture;
- classify TDD vs BDD vs ATDD;
- generate a plan;
- assert implementation/test ordering;
- verify the approved posture reaches the subagent;
- exercise greenfield runner setup;
- exercise autonomous swarm;
- verify Plan Approval is reopened when the effective posture changes.

Every issue above passes with `t292` green.

Please add deterministic behavior-level tests around a posture resolver/plan builder or an engine-level Code Generation planning journey.

**Recommendation**

The stage should not infer methodology ad hoc from free-form sections in two separate model contexts.

A safer contract would be:

1. Resolve a structured effective posture once during Part 1.
2. Persist it into the approved plan or a small machine-readable sidecar.
3. Generate a methodology-specific plan shape from that value.
4. Bind Plan Approval to the plan, test instructions, and resolved posture.
5. Pass that exact approved posture to normal and swarm developer workers.
6. Reopen approval if any of those inputs change.

The issue is valid and the prose wiring is directionally useful, but the current implementation does not yet guarantee that the team’s affirmed methodology controls actual planning and generation. I recommend keeping this changes requested.

apackeer:
Thanks @leandrodamascena - this was exactly the right call. The prose-only wiring could not guarantee the affirmed methodology survived resolution, dispatch, or the swarm path. I have reworked the PR along the lines of your recommendation: one structured contract, resolved once, fingerprinted, and enforced on both generation paths. Pushed as bf7bb0809 + 800f2a606 (head is now 800f2a606).

**How each finding was addressed**

- **P1 - project note erases team TDD**: resolution is no longer "first non-empty section wins". A new deterministic tool, `core/tools/aidlc-testing-posture.ts`, resolves the explicit `Methodology`/`Ordering` statement independently of coverage/tooling/scope notes; a project note with no cadence signal leaves the team methodology in force (the note stays in the contract's `applicable_notes`), and a contradictory narrower methodology throws instead of overriding. Practices Discovery now records structured `Methodology:` and `Ordering:` fields. Your exact example (team TDD + project integration-regression note) is pinned in t292.
- **P1 - BDD/ATDD collapsed into layer-local TDD**: the contract carries a methodology-specific plan profile. TDD keeps per-layer Red/Green/Refactor; BDD is scenario-first over feature slices; ATDD writes acceptance tests before the cross-layer implementation; custom/mixed preserves the affirmed ordering verbatim (your "BDD scenarios first; unit tests after implementation" example resolves to custom and keeps that exact text); test-after keeps implement-then-test. Each profile has behavior tests.
- **P1 - autonomous swarm bypass**: three layers now close it. Stage-protocol section 12b makes per-unit planning and fingerprinted Plan Approval mandatory before `prepare` (every harness's `invoke-swarm` branch points at it); `aidlc-swarm.ts prepare` deterministically refuses worktree creation for any autonomous code-generation unit lacking current approved evidence (t135 proves refuse-then-accept); every worker brief must open with `AIDLC-UNIT` and `AIDLC-TESTING-CONTRACT` markers that the dispatch guard verifies against the approved plan.
- **P2 - test-first covered only business logic and API**: the TDD and test-after profiles cover all five testable layers (data model/database behavior, repository/data access, business logic, API/endpoint, frontend behavior), asserted per layer in t292.
- **P2 - greenfield Green step not executable**: every profile places runner bootstrap (greenfield) or runner verification (brownfield) before the first Red/scenario/acceptance step, and `unit-test-instructions.md` must record the exact unit-scoped command that is runnable before the first test-first cycle.
- **P2 - Part 1 / Part 2 posture drift**: the contract embedded in the approved plan is authoritative for Part 2; the developer persona is forbidden from re-resolving memory; the guard recomputes the contract from current memory/scope/strategy/type and blocks on staleness, so a memory change during the approval stop reopens Plan Approval instead of silently changing execution (pinned in t265 and t292).
- **P2 - persona contradicts mandatory Refactor**: the persona now performs Refactor during initial generation when the approved contract includes that step, and defers opportunistic refactors otherwise.
- **P2 - strategy vs posture conflicting obligations**: `combineTestObligations` states the combination rule explicitly per scope/strategy pair - the selected strategy plus the scope floor, neither replacing the other, with the Minimal bugfix/security-patch targeted-regression exception spelled out. org.md and stage-protocol section 8 were rewritten to match; all 27 pairs are tested.
- **P2 - phrase-presence tests**: t292 is now behavior-level (resolver, classifier, plan profiles, obligations, fingerprinting, stale-invalidation), t265 exercises the expanded guard evidence end-to-end through the hook, and t135 exercises the swarm prepare gate.

**Two additional classifier defects found and fixed during self-review** (second commit): an explicit structured `Methodology: tdd` was being reclassified as custom when its Ordering sentence contained both "before" and "after" (structured fields are now authoritative), and the pre-2.6.8 `org.md` seed text - which merely enumerates TDD/BDD/ATDD/test-after descriptively - produced a garbage custom contract on upgraded installs that skipped practices-discovery (unsignaled multi-method enumerations now fall through to the deterministic test-after default). Both are pinned in t292, including the exact legacy org section text.

Evidence at 800f2a606: t68 + t135 + t265 + t292 pass (76 assertions), `bun scripts/package.ts --check` in sync across all seven harness trees, typecheck and the coverage-registry check green.

Ready for another look when you have time.

leandrodamascena (approved):
Arden, I re-reviewed the updated implementation at `800f2a60`. The structured contract, methodology-specific profiles, approval fingerprinting, and swarm precondition address my previous findings.

One non-blocking P2 remains: common legacy free-form postures such as `Tests first before implementation` currently fall back to `test-after`. Please handle this in a follow-up by reusing the complete unstructured classifier and adding regressions.

Focused tests and package/coverage parity are clean. Approved, with the existing `v2` version/changelog conflict still requiring resolution before merge.

PR #775, opened 2026-08-16
feat: dist/kiro-unified — a Kiro shell aligned to the unified agent harness (2.6.19)

### LINKED ISSUE #659: [Bug]: Kiro v3 doesn't default to AIDLCv2 agent
### Description

If you use the new Kiro CLI v3 with AIDLC v2, the kiro cli doesn't load the aidlc agent on startup. I can confirm that while using kiro cli v2, the aidlc agent loads just fine. 

### Steps to Reproduce

1. Setup AIDLCv2 following your guide
2. Open up kiro cli v3 using kiro-cli --v3

### Expected Behavior

The expected behavior is that the agent is set to aidlc

### Actual Behavior

Sets to the default kiro agent

### AI-DLC Version

v2 (alpha)

### Release / Commit

_No response_

### AI-DLC Phase

### Platform / IDE

Kiro CLI

### AI Model

_No response_

### Environment

MacOS

### Additional Context

_No response_

### LINKED ISSUE #764: [Bug]: On Kiro the argument-reading guards ship unregistered — kiro-ide has no manifests, and kiro registers them in the CLI 2.x embedded form
Follow-up to #763, which measured that Kiro IDE 1.x and Kiro CLI v3 **do** deliver tool arguments to PreToolUse hooks. That premise is what the current wiring rests on, so this is the wiring half, filed separately because it is a design decision rather than a factual correction.

### What ships today

Verified on `4569754e` (v2.6.2):

- `harness/kiro-ide/hooks/` ships **eight** standalone `.json` manifests: `continue-workflow`, `enforce-approval-gate`, `log-subagent`, `rebuild-stage-graph`, `record-human-turn`, `session-start`, `sync-workflow-state`, `write-audit-log`. There is **no** manifest for `review-freeze`, `plan-approval-guard`, or `reviewer-scope`.
- Their bodies do ship: `dist/kiro-ide/.kiro/hooks/aidlc-review-freeze.ts` and `aidlc-plan-approval-guard.ts` are present and unreferenced.
- `harness/kiro/hooks/` (CLI) ships **zero** `.json` manifests — seven legacy `.kiro.hook` files and the adapter. The three guards are registered instead inside the agent configs' embedded `hooks` (`harness/kiro/agents/*.json`).

### The stated reason

`core/hooks/aidlc-reviewer-scope.ts:30-31`:

> Kiro IDE ships no registration: its hook payloads carry no tool inputs, so a pre-tool matcher has nothing to inspect there.

That is a deliberate, documented decision, not an oversight — and #763 shows the premise no longer holds on the surfaces I can test.

### What I measured in a fork

I ported the two matcher-based guards to a `kiro-ide` shell as standalone manifests plus two adapter verbs, and ran it on Kiro IDE 1.0.309 and Kiro CLI v3 in parallel workspaces. All **ten** manifests in that shell fired on both surfaces (checked through `.aidlc-hooks-health/*.last`), the two ported ones included.

One defect surfaced in my own port, and it is worth knowing before anyone repeats it: the dispatched agent's identity is in the **tool name** (`subagent_aidlc-developer-agent`), not in the arguments — there is no `agent_type` or `name` key. My adapter derived `subagent_type` from those two keys, so it passed `""` and `aidlc-plan-approval-guard.ts` returned early at its `subagent_type !== "aidlc-developer-agent"` check. The hook fired and enforced nothing. Deriving from the tool name fixes it; I verified the flip through the body's own pure function (`evaluatePlanApprovalDispatch` returns `block: false` for `""` and `block: true` for the derived name, and still allows an approved unit).

### What I did not measure

- Whether the **embedded** `hooks` in `harness/kiro/agents/*.json` still register on Kiro CLI v3. The Kiro docs say the mechanism moved (`features/hooks.md`: "From CLI 2.x - Hooks moved from embedded fields in agent config to standalone files. Run `kiro-cli agent migrate` to auto-convert"). If embedded registration is legacy-only, the same three guards may be unregistered on CLI v3 as well, which would make this broader than `kiro-ide`. I have not tested a stock CLI shell to confirm either way.
- Whether Kiro offers an input-rewrite channel (the separate question raised at the end of #763).

### The decision I am asking for

1. Should `review-freeze` and `plan-approval-guard` get `kiro-ide` manifests? Both are matcher-based and carry no persona argument, so this is a straight port.
2. Does the `kiro` harness need standalone manifests for CLI v3, or is the embedded form still honoured there?
3. `reviewer-scope` is not a straight port. Its CLI registration is per-persona (embedded in the reviewer agents' own configs, which is what makes every call arriving through it the reviewer's). The unified harness has no per-persona hook registration, so a global manifest would need to identify the reviewer from the tool name instead of from `agent_type`.

I am not asking for a specific patch — my fork already carries (1) and lives without (3). What would help most is knowing whether upstream wants this in, so that a fork does not have to keep diverging here.

wowzoo:
## What this adds

Kiro IDE 1.x and Kiro CLI v3 (`kiro-cli --v3`) run the same agent runtime: both resolve an agent from its Markdown definition — persona body, `tools:`, `resources:` and `permissions:` in one file — and both execute the standalone `.kiro/hooks/aidlc-*.json` manifests. `dist/kiro-unified/` is the shell for that runtime, with nothing else in it.

The surfaces it leaves out are the ones this repo already describes as a compatibility layer. `README.md:138` calls the `kiro-ide` tree's `settings/cli.json` and dual hook registration *"a Kiro CLI-only compatibility surface"* that *"registers the framework hooks in both formats"*, and the routing table at `:58-59` has no row for CLI `--v3`. This PR adds that row and the tree it points at. `dist/kiro-ide/` and `dist/kiro/` are unchanged.

Against `dist/kiro-ide/`, concretely:

- No agent-v1 JSON (15 files) and no legacy `.kiro.hook` files (9).
- The conductor ships as `agents/aidlc.md` — the entry the workspace agent selector loads on both surfaces.
- 12 hook registrations instead of 8: `review-freeze`, `plan-approval-guard`, `state-transition-guard` and `verb-intercept` join the spine. Their PreToolUse exit-2 refusal works on both surfaces, which is the condition `docs/harness-engineering/09-porting-to-a-new-harness.md:104-107` sets for registering them.
- The 14 delegation targets carry `permissions:` and `resources:` in their `.md` frontmatter, declared as manifest data (`frontmatterAdditions`), not as forked core persona files.
- `settings/mcp.json` ships with every server disabled, because both surfaces read the same Kiro MCP registry and an agent opts in per-agent via `includeMcpJson`.

Three trees now install into `.kiro/`, so `README.md` § Pick your harness says which runtime takes which: `kiro-unified` for IDE 1.x and CLI `--v3`, `kiro-ide` for IDE builds that still need the pre-1.0 compatibility surfaces, `kiro` for Kiro CLI 2.x whose engine reads the agent-v1 JSON sandbox rather than a `permissions:` block.

## Shared code this touches, and why

The porting contract is *"one directory and one manifest row, zero shared-code edits"* (`09-porting-to-a-new-harness.md:44-45`). Four files fall outside that, and each is here deliberately:

1. **`core/tools/aidlc-utility.ts` — the doctor arm.** The sanctioned exception (`:113-122`). The `.kiro` arm required `agents/aidlc.json`; it now accepts either conductor form (`agents/aidlc.{json,md}`), so it passes on a Markdown-only Kiro tree and is unchanged for the JSON trees.
2. **`core/tools/aidlc-lib.ts` — `loadAgents()` skips `agents/aidlc.md`.** A conductor entry declares `name` only, so it fails the persona frontmatter contract, and it must stay out of the roster that backs stage `lead_agent` validation. A no-op in every tree that ships no `agents/aidlc.md`.
3. **`scripts/package.ts` — `frontmatterAdditions` accepts a nested block.** The validator required every added line to be a top-level YAML key, so a harness whose native frontmatter has structure could not use the channel at all. An indented line now continues the block its preceding top-level key opened; the collision guard on top-level keys is untouched. Without this the only way to ship Kiro's `permissions:` rules would be forking the 14 core persona files, which is what the channel exists to prevent (`scripts/manifest-types.ts:83-97`).
4. **`scripts/build-binaries.ts`.** The new tree joins `RUNTIME_DISTRIBUTIONS`, and the `harness-probe-kiro` gate pinned the old doctor label as a string literal, so it moves with the label; a probe gate for the new tree is added beside it.

`renderRunner`'s harness switch is deliberately **not** touched: a `kiro-unified` case would return the same sentence the `.kiro` branch already returns, and the four generated scope runners are byte-identical either way (measured by building without it).

## Deliberately left out, and documented instead

The per-persona `reviewer-scope` and `state-transition-guard` hooks that the Kiro CLI agent JSONs carry are **not** re-declared in the agent `.md` files. Step 2 says to scope reviewer-scope per agent *"where the harness supports per-agent hooks"* (`:107-111`), and on this runtime an agent-scope hook fires for the **active** agent, not for a **delegated** one — measured on kiro-cli 2.18.1 `--v3` and IDE 1.0.309. These personas are only reached by delegation, so the registration would be a dead hook, and `:104-107` says to document the gap instead. The new chapter does. `aidlc-fold-usage.ts` and `aidlc-session-end.ts` likewise ship unregistered (Claude-Code-only producers).

## What I tested

| Check | Result |
| --- | --- |
| `bun scripts/package.ts --check` | 8/8, no drift |
| `dist/kiro-unified` file count | 264 |
| `/aidlc --doctor` on a fresh install of the tree | 41 passed, 0 failed |
| `loadAgents()` in the new tree | 14 |
| Registered hook manifests | 12 |
| Other seven trees' diff | four files each — `aidlc-includes.ts` (the space-switch repoint, a no-op on the trees whose `.md` agents carry no concrete memory glob), `aidlc-lib.ts` (the guard), `aidlc-utility.ts` (the doctor arm), `aidlc-version.ts` (the bump) |
| `bun run lint` / `bun run typecheck` | clean |
| `bash tests/run-tests.sh --smoke --unit --integration --no-llm -P 8` | after the merge to 2.6.9 the failing-test set on this branch is identical to `upstream/v2`'s on the same box, and CI's Tests (smoke + unit) job is green |

New deterministic test: `tests/unit/t293-kiro-unified-adapter.test.ts` (21 cases) — the `verb-intercept` seam (off-band terminal dispatch, the turn-scoped read-only latch and turn counter, the non-terminal and adjacent-command negatives) and the guards' payload acquisition, including a payload-less negative control that shows the silent-inert failure mode; review added the four plan-approval dispatch shapes, `review-freeze`'s malformed nested inputs, and doctor's two Markdown-conductor hook rows in both directions. One row each is added to `HARNESS_CAPABILITIES` and to the three per-harness test tables (`t123`, `t181`, `t239`).

### The live journey

`tests/e2e/t-tui-kiro-unified-intent-capture.serial.test.ts` drives the IDEATION intent-capture stage through a real keystroke-driven `kiro-cli chat --agent aidlc --trust-all-tools --v3` against the shipped tree and terminates on disk. **Run green** on kiro-cli 2.18.1, 29 assertions, 27m16s. It is gated behind `AIDLC_KIRO_UNIFIED_TUI_LIVE=1` and skips with a reason when tmux, `kiro-cli`, auth or the tree is absent.

Its launch surface is measured, not assumed: `chat --agent-engine <v1|v2|v3>` selects the runtime and **v2 is the default**; `--agent aidlc` is required because Kiro reads CLI settings from the global scope only, so the shipped `settings/cli.json` does not select the workspace default; the idle footer matches the `dist/kiro` twin's; and the statusbar carries the selected agent, so a Markdown conductor with no agent-v1 JSON beside it loads by name.

Its last four assertions are what no other test in the suite covers — that the twelve manifests **fired** rather than merely shipped. From that run: `.aidlc-hooks-health/` carried heartbeats for `session-start`, `write-audit-log`, `rebuild-stage-graph`, `run-sensors`, `log-subagent`, `continue-workflow`, `review-freeze` and `plan-approval-guard`; `aidlc/.aidlc-turn-counter` reached 8; and `review-freeze` did not just register — it **refused a write** mid-stage, after which the conductor lifted the freeze, re-issued the summary checkpoint and completed the stage.

That first attempt failed in the test harness rather than the tree, and the fix is in this branch: the shared numbered-prose recognizer keyed a summary confirmation on a "before I ..." clause and otherwise answered only the first one per run, so the re-issued checkpoint matched nothing and the journey stalled at an idle prompt. It now keys on the newest question sentence, with a `t142` unit case pinning the shape. That also removes the same stall from the `dist/kiro` twin, which shares the helper.

## Notes

- Version: this branch now takes **2.6.19**. It first took 2.6.9 (`#758` landed that number upstream while this branch held it), then 2.6.10, and the maintainer prep re-slotted past each v2 advance since: 2.6.15/2.6.16 merged upstream, `#792` took 2.6.17, and `#767` holds an unpushed 2.6.18, so this PR takes the next free patch number. Per the Changelog Policy's conflict-trap, a same-number collision among open PRs resolves at merge time: the second to merge re-bumps.
- New user-facing chapter `docs/guide/harnesses/kiro-unified.md` (prerequisites including the `--v3` opt-in, install, which Kiro tree to pick, and what differs — including which hooks ship unregistered and why), plus the roster/label updates `t239` requires.
- This work turned up four things that look like defects or gaps in the existing Kiro shells, all independent of this PR and filed separately rather than folded in here: the `verb-intercept` anchor parser with no payload shape that matches it (`#776`), no test that checks a registration's matcher against the captured payload fixtures (`#777`), matchers keyed to tool names the payload census never reports (`#778`), and the diagnostic-bundle path mismatch for the read-only latch (`#779`).
- The three-`.kiro`-trees roster question — which tree new users should land on, and what a convergence path would look like — belongs with the maintainers rather than in this diff, so it is raised on `#582`.

Closes #764

Closes #659


apackeer (changes requested):
Thanks wowzoo - this is a carefully built and unusually well-evidenced PR. The direction is right: the unified runtime (IDE 1.x + CLI `--v3`) is a real gap on `v2`, the authored/generated split is correct, the four shared-code edits are each justified and minimal, and every claim in the PR body that we tested verified true (`package.ts --check` 8/8; the other seven trees change in exactly 3 files each; doctor on a fresh `dist/kiro-unified` install reports 39 passed / 0 failed; `loadAgents()` returns 14; only this tree ships `agents/aidlc.md`, so the `loadAgents()` skip is a no-op elsewhere; CI green).

Requesting changes on two P1s and a set of mechanical P2s. This review ran as two independent passes that were then merged; everything below was verified against the head commit (599b1fe).

## P1

**1. A space switch leaves every Markdown agent bound to the `default` space.**
`harness/kiro-unified/manifest.ts:52-55` injects `file://aidlc/spaces/default/memory/**/*.md` into all 14 delegates, and `harness/kiro-unified/agents/aidlc.md` carries the same literal for the conductor. But the `.kiro` branch of the space-switch rewrite (`core/tools/aidlc-includes.ts:235-248`) repoints only `agents/*.json` (`if (!name.endsWith(".json")) continue;`) plus `steering/aidlc-active-memory.md`. After `/aidlc space <name>`, steering points at the new space while all 15 agent `.md` files still glob the default space's memory - agents in a non-default space load the wrong rule layers, which breaks space isolation. The `.aidlc`/opencode branch directly above it already rewrites `agents/*.md` (`repointOpencodeAgentMemory`), so the pattern to mirror exists. Please repoint the Markdown `resources:` paths on `.kiro` too, with a regression test for this tree.

**2. The dispatch payload the skill mandates bypasses the plan-approval guard.**
`harness/kiro-unified/skills/aidlc/SKILL.md` (Harness notes) mandates Kiro's crew schema for every dispatch - `{mode:"blocking", task:"...", stages:[{name, role, prompt_template}]}` - and says "Never call `subagent` with a bare agent/task pair or omit `stages`". The old CLI adapter extracts identity from that shape (`harness/kiro/hooks/aidlc-kiro-adapter.ts`, plan-approval block: filters `stages[].role === "aidlc-developer-agent"`, joins `task` + `prompt_template`s). The new adapter's plan-approval block does not: it derives identity only from a `subagent_<agent>` tool-name suffix or `tool_input.agent_type`/`name`, and the prompt only from `tool_input.prompt`/`message` - none of which exist in the crew shape. So the exact payload the skill instructs forwards `subagent_type: ""` and an empty prompt, and the core guard returns early and allows a code-generation dispatch with no approved plan (reproduced in a scratch probe: the crew payload exits 0 silently where the suffixed form exits 2 with the refusal). There is also a shape contradiction to resolve: the adapter's own measured note says dispatches arrive as `subagent_<agent>` with keys `["prompt", "explanation"]`, which the skill's mandated crew call is not. Whichever shape the live runtime delivers, the skill and the adapter must agree, and the guard must cover the instructed shape - porting the `stages[].role` extraction from the `harness/kiro` adapter covers both cheaply.

## P2

**3. The plugin projection for this tree is dead on this runtime, twice.**
`harness/kiro-unified/manifest.ts:317` declares plugin kind `kiro`, so `scripts/package.ts:986-988` emits `aidlc-plugin-compose.kiro.hook` as the wiring - a legacy format this tree's own design excludes ("No legacy `.kiro.hook` files ... Only the `.json` manifests ship"), and which the captured contract pins as "silently inert - never executed" on IDE 1.x (`docs/reference/kiro-ide-hook-payload.md:14`). Separately, the shared composer validates `.kiro` dispatch against agent-v1 JSON: `scripts/plugin-hooks-template/compose.ts:844-882` reads `agents/aidlc.json` `toolsSettings.subagent.trustedAgents` and suggests authoring `<agent>.json`, neither of which a Markdown-only tree has. The tree needs unified wiring (a standalone `.json` manifest) and Markdown-aware compose validation - or, if plugin support is explicitly out of scope for this tree, say so in the chapter instead of shipping inert wiring.

**4. Doctor reports "hook + permission wiring" healthy with all 12 hook manifests deleted.**
Reproduced: fresh copy of the tree, `rm .kiro/hooks/aidlc-*.json`, `/aidlc --doctor` still reports 39 passed / 0 failed. On the JSON trees the `agents/aidlc.json` check genuinely covers hook wiring (hooks are embedded there); on this tree the wiring is the 12 standalone manifests, so `agents/aidlc.md` presence proves the persona and permissions but not the hooks (`core/tools/aidlc-utility.ts:1398-1409`). Doctor should validate the manifest roster when the Markdown conductor form is what it finds.

**5. The shipped skill contradicts the tree's own wiring.**
`harness/kiro-unified/skills/aidlc/SKILL.md:44` says "this shell registers no state-transition PreToolUse hook" - but registering `hooks/aidlc-state-transition-guard.json` is one of this PR's four headline guards (`manifest.ts:132`), and the new t291 test asserts its exit-2 refusal. The same Harness-notes paragraph still requires "a `trustedAgents` entry in the conductor's config", a field the Markdown conductor does not have. And `skills/aidlc/question-rendering.md:1` still titles itself the "Kiro IDE harness annex". All look like half-updated fork text from the kiro-ide tree.

**6. `verb-intercept` is documented as a PreToolUse guard; it is UserPromptSubmit and cannot block.**
`docs/guide/harnesses/kiro-unified.md:91-92` and the CHANGELOG bullet group it with the exit-2 PreToolUse guards ("The last four are PreToolUse guards: a non-zero exit blocks the tool call"). Its own manifest registers `UserPromptSubmit`, and the adapter comment explicitly measured that exit 2 does not stop the turn on this trigger - the seam works by relay, not by blocking. The docs should say so.

**7. The Kiro ACP calibration still pins the retired doctor label.**
`tests/harness/kiro-acp-drive.calibration.test.ts:56` expects the substring `agents/aidlc.json present`; the shared label is now `agents/aidlc.{json,md} present (hook + permission wiring)`, which does not contain it, so calibration 1 (the trust anchor for the ACP tier) fails on its next `AIDLC_KIRO_ACP_LIVE=1` run. It sits outside tier collection, so CI cannot see it - the PR moved the same literal in `scripts/build-binaries.ts` but missed this copy. One-line fix.

**8. `t291` is already claimed by an open PR.**
#758 ships `tests/unit/t291-review-receipt-recovery.test.ts` and #772 claims t292. Same collision discipline as the 2.6.9 re-bump this branch already did for the version number: please take the next free slot (t293) and move the coverage-registry entries with it.

**9. `--no-llm` cannot close the new live gate.**
`AIDLC_KIRO_UNIFIED_TUI_LIVE` (`tests/e2e/t-tui-kiro-unified-intent-capture.serial.test.ts:94`) is missing from `LIVE_MODEL_GATES` (`tests/run-tests.ts:33-40`), which `--no-llm` force-closes at `:285`. With the var exported, a `--no-llm` run still spends real Kiro credits on the ~27-minute journey. One-line fix.

**10. The documentation sweep is incomplete.**
Per the repo's Documentation Policy (update stale references in the same commit): `AGENTS.md:11` (harness roster), `:12` (plugin projection set - `dist/plugins/test-pro/kiro-unified/` now ships), `:14` (dist roster); `README.md:426` (plugin projection roster in the layout diagram); `docs/guide/01-getting-started.md:144-153` claims "copy commands for every harness" but has no kiro-unified block or link; `docs/reference/18-plugin-mechanism.md:404` says "all seven harness projections" (now eight); `docs/reference/00-overview.md:11` and `docs/reference/11-contributing.md:30,33` (dist lists; both already omit cursor, so that half is pre-existing); `docs/guide/12-cli-commands.md:9-12` and `docs/guide/13-customization.md:15-20` (harness link lists); `docs/harness-engineering/09-porting-to-a-new-harness.md:55-56` (the in-tree session-skills list).

## Notes (no change requested)

- `docs/guide/harnesses/kiro-unified.md:59` pins "report 39 checks passed" - true today, but no sibling chapter pins a literal count and it goes stale on the next doctor addition; "all checks passed" would age better.
- `aidlc-verb-intercept.json` registers with no matcher, so a bun subprocess runs on every user prompt on both surfaces. The manifest documents why (the turn clock must tick every prompt); flagging it as a deliberate per-prompt hot path for maintainers' awareness.
- Direction/roster: a third `.kiro` tree whose IDE-1.x coverage overlaps `dist/kiro-ide` is a standing maintenance surface (a third near-identical adapter fork, a third skill copy). The layering here is right and follows the repo's existing fork-per-tree convention, but the three-Kiro-trees roster and its sunset path deserve the issue-first discussion the PR body already offers - taking you up on that would be good.

## Verification

`bun scripts/package.ts --check` 8/8 OK. Deterministic slice for the touched pins (`t291`, `t142`, `t68`, `t181`, `t239`, `t123`, `gen-coverage`): `tests/logs/2026-08-16T15-10-55Z-p3088986`, 8 files, 2721 assertions, Result: PASS, no live vars set. Doctor probes as described in findings 4 and above. The 27-minute live journey was not re-run (opt-in gated); its shared-recognizer fix is pinned by the new t142 case, which passes.

wowzoo:
Thank you — this was a genuinely useful review. I checked all ten findings against the tree and found nothing to argue with; every one reproduced. All ten are now addressed, and CI is 4/4 on the new head. Three of them warrant more than "fixed", so those are below; the rest are at the end.

## P1-2 — fixed on both sides, but the root cause runs the other way

You are right that the skill and the adapter must agree and that the guard has to cover the instructed shape. On which side to move, the evidence points at the skill: **the crew shape has never been observed on either supported surface.** The adapter's own measured note (`harness/kiro-unified/hooks/aidlc-kiro-adapter.ts`, 2026-08-14 live runs on IDE 1.0.309 and CLI 2.18.1 `--v3`) records a dispatch arriving as `subagent_<agent>` with argument keys `["prompt", "explanation"]`, and no bare `subagent` call in the payloads captured on those runs. So `SKILL.md`'s crew mandate is inherited text from the `kiro-ide` tree — the same family as finding 5 — and instructing it was the defect.

I did both halves anyway, for the reason you gave:

- `SKILL.md` now instructs one tool call per delegate (`subagent_<agent>` / `invoke_sub_agent`, brief as `prompt`), and says explicitly that the crew schema belongs to the agent-v1 trees and that identity here is in the tool name.
- The adapter's plan-approval block gained a **fallback** that filters `stages[].role === "aidlc-developer-agent"` for identity and joins `task` + the developer stages' `prompt_template` for the prompt, so the guard still refuses if a runtime does deliver the crew shape. It is a fallback, not the primary path, and the comment says why.

## P1-1 — fixed, and it stays inside this tree

The `.kiro` branch of `repointHarnessIncludes` now dispatches on the file extension: `agents/*.json` keeps `repointKiroAgentResources`, `agents/*.md` gets a new `repointKiroAgentMarkdownResources`. Three regression cases in `tests/unit/t-active-space-includes.test.ts`.

Since that branch is shared, I measured what it does to the other two Kiro trees rather than reason about it. A real `/aidlc space qa` on a fresh install of each:

| Tree | agent `.json` rewritten | agent `.md` rewritten |
|---|---|---|
| `dist/kiro-unified` | 0 (ships none) | **15** |
| `dist/kiro` | 15 | **0** |
| `dist/kiro-ide` | 15 | **0** |

No blast radius, because those trees' `.md` twins carry no concrete memory resource: the only `aidlc/spaces/…/memory/` text in them is the documentation placeholder `aidlc/spaces/<active-space>/memory/{org,team,project}.md`, in prose.

Which is exactly why the rewriter anchors on the full `file://…/**/*.md` resource form rather than replacing `aidlc/spaces/[^/]+/memory/` wholesale. My own regression test caught the loose version: it rewrote `<active-space>` into a concrete space, corrupting the prose and making a repoint to the already-active space a non-no-op. With a loose pattern the table above would read 12 `.md` files rewritten on each JSON tree, every one of them wrongly. `repointOpencodeAgentMemory` still uses the loose form and carries the same risk; out of scope here, but worth a separate look.

## 3 — unified wiring, measured before and after

I took the first branch you offered. There is no opt-out in the plugin contract by design — `scripts/manifest-types.ts` defaults an omitted `plugin` block to `kind: "store"` specifically so a new harness is never silently skipped — so "say so in the chapter" would have meant shipping the inert file anyway.

- New plugin kind **`kiro-unified`**: the compose trigger emits `hooks/aidlc-plugin-compose.json` in the `{"version":"v1","hooks":[{name,trigger,description,action}]}` schema on `UserPromptSubmit` — the same schema this harness ships its own twelve hooks in. `dist/kiro` and `dist/kiro-ide` are untouched and keep the legacy `.kiro.hook`.
- `compose.ts` now detects the conductor form: when `agents/aidlc.json` is absent and `agents/aidlc.md` is present, the dispatch surface required is `agents/<agent>.md` and no `toolsSettings.subagent.trustedAgents` entry is asked for — a field the Markdown conductor has no equivalent of, which is why the previous fail-closed empty set rejected everything.
- `tests/harness/harness-matrix.ts` moves with it (`kind` + `wiringFile`), so `t188`'s per-harness wiring assertion covers the new form. `docs/reference/18-plugin-mechanism.md` gains the eighth projection, and its Kiro security note no longer describes the compose hook as a `.kiro.hook` unconditionally.

The composer half needed a dispatching stage to show anything, since `test-pro`'s two stages are `mode: inline` and never reach the agent-surface check. Flipping `test-pro-integration` to `mode: subagent` and running the composer on a fresh install:

| Composer | Tree | Degraded drops |
|---|---|---|
| before | `dist/kiro-unified` | **2** — `aidlc-quality-agent` *and* `test-pro-metrics-agent`, each demanding `<agent>.json` plus a `trustedAgents` entry |
| after | `dist/kiro-unified` | **1** — only `test-pro-metrics-agent`, asking for `<agent>.md` |
| stock | `dist/kiro` | **1** — only `test-pro-metrics-agent` |

So the shipped persona (`aidlc-quality-agent`, which this tree carries as `.md`) is now accepted, and the tree lands on **parity with the stock JSON tree**. The remaining drop is a pre-existing, uniform Kiro behaviour: a plugin's *own* agent is not yet on disk when the precheck runs, and the `pluginShipsViableNativeAgent` escape hatch that covers exactly this case is gated to `.aidlc` (opencode / Copilot). That is the same on all three Kiro trees, so I left it alone rather than widen this PR — flagging it as a candidate.

**Adjacent, and deliberately not fixed here** — two things I noticed while measuring, both pre-existing and both touching the other Kiro trees:

- The documented Kiro folder-drop (`cp -r "$PLUGIN_ROOT"/. <project>/`) lands the projection's `hooks/` at the **project root**, not in `.kiro/hooks/`, and `compose.ts` installs no hook files. So on all three Kiro trees the compose manifest sits where no Kiro surface reads it, whatever its schema — consistent with the chapter's own instruction to run the composer explicitly.
- Whether the legacy `.kiro.hook` is the right form for `dist/kiro` is a separate question: Kiro CLI 2.x reads hooks from the agent config, not from a standalone hook file. I did not touch that tree.

## 4 — two rows, and deliberately not a name roster

Reproduced exactly as you describe. Doctor now probes the manifests when the Markdown conductor form is what it finds:

1. `Hook contract: <n> standalone hooks/*.json manifest(s) wired (Markdown conductor)` — **zero fails**, which is the case you reproduced;
2. `Hook contract: every wired manifest dispatches a shipped hook script`.

I did not hardcode the twelve names. The Claude branch in the same function (`core/tools/aidlc-utility.ts:1290-1291`) argues against exactly that — *"no hardcoded list to drift — the old list named only 7 of the 10 shipped hooks"* — and it enumerates from `settings.json` so that the expectation and the probe target genuinely diverge. This tree has no equivalent declaration to enumerate from other than the manifests themselves, so I probed the two things a self-enumerated roster still catches. If you would rather have a literal pin I will add one, with a maintainer note that it must move with every hook addition.

Row 2 reads `hooks[].action.command` only, not the raw file: a `description` legitimately names other scripts in prose — the state-transition guard's names `aidlc-state.ts` — and a raw-text scan false-positives on it. That was caught during verification, not foreseen.

Measured on a fresh install: **41 passed / 0 failed** at baseline. With all twelve manifests deleted, row 1 fails (39 / 1; row 2 is skipped when there is nothing wired). With `aidlc-kiro-adapter.ts` removed, two rows fail (39 / 2) — the pre-existing `aidlc-kiro-adapter.ts present` row and the new row 2. `dist/kiro` and `dist/kiro-ide` (39) and `dist/claude` (47) are unchanged, since both JSON trees ship `agents/aidlc.json` and never take the new rows. Because the count moved, I also took your note about the chapter pinning "39 checks passed" and made it "all checks passed".

## The rest

- **5** — `SKILL.md` now states that `aidlc-state-transition-guard.json` *is* registered as a global PreToolUse hook and refuses a direct state mutation with exit 2; the `trustedAgents` requirement is replaced with the reason the Markdown conductor has no such field (delegation is bounded by its `tools:` grant); `question-rendering.md` is retitled "Kiro unified harness annex".
- **6** — the chapter and the CHANGELOG now name the four PreToolUse guards explicitly (`enforce-approval-gate`, `review-freeze`, `plan-approval-guard`, `state-transition-guard`) and say that `verb-intercept` registers on `UserPromptSubmit`, cannot block, and works by relay — running the terminal utility itself and handing the output back on stdout with a do-not-advance instruction, exit 0 on every path.
- **7** — the calibration constant is now the shared label verbatim, with a comment that this file sits outside tier collection so CI cannot catch the drift.
- **8** — moved to **`t293`**: the filename, its four in-file references, and the coverage-registry entry. No `t291` reference remains anywhere in the branch.
- **9** — `AIDLC_KIRO_UNIFIED_TUI_LIVE` added to `LIVE_MODEL_GATES`.
- **10** — all twelve coordinates updated. Two notes: I also fixed the pre-existing `cursor` omission in `docs/reference/11-contributing.md:30,33` and `docs/reference/00-overview.md:11`, rather than leave a list wrong in a line I had just edited; and the README dist roster above the plugin line you flagged was missing `kiro-unified` too, so that is in.
- **Per-prompt hot path** — agreed, and thank you for reading the manifest's rationale rather than only its shape. It is deliberate: the engine's done-guard (Branch 0 of `next`) reads a turn-scoped latch and turn counter that nothing in this tree wrote before this seam existed, so the clock has to tick on every prompt or that guard stays fail-open.
- **Three-Kiro-trees roster** — my position: **converge on the unified harness** rather than hold three trees at parity, with the sunset gated on CLI 3.0 leaving early access, and until then harness-surface work (hook wiring, guards, adapters) going to the unified tree while the legacy trees take correctness and security fixes only. My reason is completeness, not maintenance cost. Counting the four PreToolUse guards as they are wired today: `dist/kiro-unified` registers 4/4; `dist/kiro-ide` 1/4 (`enforce-approval-gate` only); `dist/kiro` embeds three in `agents/aidlc.json` but not `enforce-approval-gate`. The two-format coexistence itself was a deliberate, field-proven call (#614); my point is only that holding three trees at parity is what spreads the wiring thin. The vendor side already leans the same way — IDE 1.x does not load 0.x hooks at all (they appear with an upgrade badge and stay unloaded), 0.x lives in a separate legacy reference, and the only environment gap the CLI 3.0 page names is Amazon Linux 2, whose own end-of-support passed in June 2026.

## Verification

`bun scripts/package.ts --check` 8/8 OK · `bun run typecheck` clean · `bun run lint` clean · `t188-plugin-compose` + `t293` + `t-active-space-includes` + `gen-coverage-registry` = 146 pass / 0 fail (1,792 assertions) · **CI 4/4 pass** on the pushed head (build, Changelog completeness, Contract checks, Tests smoke + unit). Doctor and composer probes as tabulated above. The 27-minute live journey was not re-run.

Locally the full smoke + unit tiers (218 files, 6,381 assertions) show three failing files — `t248-codekb-scope-diff`, `t248-copilot-packaging`, `t255-workspace-sync`, 11 assertions. All three reproduce identically in a clean worktree at the untouched previous head `599b1fe`, and the same tier passes in CI, so they are local to my machine and none of them comes from this change set. Flagging rather than leaving them unexplained.

wowzoo:
Opened #781 for the three-Kiro-trees roster and its sunset path, per your note.

It carries the measured duplication and guard-wiring numbers and proposes a two-phase path: this PR lands as `dist/kiro-unified` with nothing renamed or removed, then a single rename commit once `--v3` becomes the default. Nothing in this PR depends on that discussion.

apackeer (changes requested):
Thanks wowzoo - this response round is what a review hopes for: every finding answered with measurements, and the measurements checked out. As before, this ran as two independent passes that were then merged; everything below was verified against the head commit (5a07bbb), with the pre-fix commit (599b1fe) as a control wherever behavior changed.

**Verified fixed - all ten round-1 findings:**

- **Space switch (P1-1)**: `repointKiroAgentMarkdownResources` plus the extension dispatch repoint all 15 Markdown agents, conductor included; the new `t-active-space-includes` cases pass. The `.md` twins in `dist/kiro` / `dist/kiro-ide` carry no concrete memory glob, so the shared branch is a no-op there - matching your table.
- **Plan guard (P1-2)**: against a seeded unapproved code-generation state, the crew developer payload now exits 2 with the refusal where the 599b1fe adapter exits 0 (bypass closed); the suffixed shape still exits 2; a non-developer crew fail-opens. The SKILL.md dispatch rewrite matches the measured runtime, and keeping the crew branch as a defensive fallback was the right call - but see finding 1.
- **Plugin projection (3)**: the emitted `hooks/aidlc-plugin-compose.json` is schema-identical to the tree's own twelve manifests, and a synthetic mob-mode stage now composes on a Markdown-conductor install where the pre-fix composer dropped it with agent-v1 instructions.
- **Doctor (4)**: fresh tree 41 passed / 0 failed; all-manifests-deleted and missing-hook-body both fail with precise pointers - but see finding 2.
- **5-10 all confirmed**: the skill text and `question-rendering.md` retitle, the verb-intercept relay docs, the calibration label, t293 with zero `t291` references left anywhere on the branch, `AIDLC_KIRO_UNIFIED_TUI_LIVE` in `LIVE_MODEL_GATES` (a forced `--no-llm` run with the var preset correctly skipped the journey), and the docs sweep including the pre-existing cursor omissions you folded in. `bun scripts/package.ts --check` 8/8, version trio 2.6.9, CI 4/4.

Requesting changes on one new regression from this round plus three small closures.

## P2

**1. The crew fallback crashes on a malformed `stages` array.**
`harness/kiro-unified/hooks/aidlc-kiro-adapter.ts:583-584` filters `crewStages` without validating the elements: piping `{"tool_name":"subagent","tool_input":{"stages":[null]}}` to the plan-approval target dies with `TypeError: null is not an object (evaluating 's.role')`, exit 1 - reproduced independently in both passes. This tree documents any non-zero PreToolUse exit as blocking, so a malformed `stages` array blocks every subagent dispatch with a stack trace as the reason; and this is the hook whose own comment says it must not be bypassable, so it should not be crashable either. Guard each element (an object with a string `role`) and give malformed nested payloads the same deterministic outcome the core guard already commits to for malformed stdin.

**2. Doctor's manifest row counts any `*.json`, so an inert roster still reads healthy.**
`core/tools/aidlc-utility.ts:1424` counts every `hooks/*.json`. With all 12 AIDLC manifests removed and a single `unrelated.json` containing `{"version":"v1","hooks":[]}`, doctor still reports both Hook contract rows green and 41 passed / 0 failed (row 2 is vacuous with nothing to resolve). Your no-name-roster reasoning stands - this is just tightening the count to `aidlc-*.json`, the exact set the row's own fix string already names. One line.

**3. The three new silent seams need deterministic pins.**
Both passes verified the behaviors by probe, so this is about keeping the next refactor honest - but each is a seam whose failure mode is silence. `t293` has no plan-approval case in any shape (`t147` test 1b is the model: add the suffixed shape, the crew shape, a non-developer crew, and the malformed `stages:[null]` case from finding 1). `t188` exercises the JSON Kiro conductor only, never the new Markdown branch - the `test-pro-integration` mode flip you measured with is exactly the fixture the suite wants. And the doctor Hook contract rows have no unit coverage (deleted manifests, missing body, and the inert-roster case from finding 2).

**4. Two doc sentences still describe the Kiro compose hook as live.**
Your folder-drop measurement matches ours: `cp -r "$PLUGIN_ROOT"/. <project>/` lands `hooks/aidlc-plugin-compose.json` at the project root, the runtime reads `.kiro/hooks/`, and `compose.ts` installs no hook files - uniformly, on all three Kiro trees. We are NOT asking you to build the `.kiro/hooks/` install path here: that is pre-existing mechanism work, the chapters already lead with "run the composer explicitly", and an issue carrying your measurements is the right home for it. But two sentences in scope still say otherwise: the security note at `docs/reference/18-plugin-mechanism.md:40` (edited in this PR) still describes the drop as copying "the compose hook that runs `compose.ts` on the next prompt", and the Kiro note at `docs/harness-engineering/10-authoring-a-plugin.md:318` still names "the emitted `.kiro.hook`" unconditionally. Please reword both so the explicit composer is the documented mechanism on Kiro and the manifest form is named per tree.

## Notes (no change requested)

- Please do file the adjacent defects you measured as issues so they survive this thread: the folder-drop discovery gap above, `repointOpencodeAgentMemory`'s loose pattern, `pluginShipsViableNativeAgent` being gated to `.aidlc`, and the `dist/kiro` `.kiro.hook` question.
- The roster-convergence position (unified tree as the target, legacy trees on correctness and security only, sunset gated on CLI 3.0) is well argued and belongs in #582 for the maintainers to settle.
- The 27-minute live journey was not re-run by either side this round. The dispatch-instruction change matches the captured payloads, so the risk is low, but the first live run after merge should watch the dispatch step.

## Verification

`bun scripts/package.ts --check` 8/8 on both passes. Deterministic slices: `tests/logs/2026-08-17T06-10-15Z-p756468` (10 files pass; the `t145-packaging-parity` red is a known local `/tmp` `smol-toml` resolution artifact on this box, reproduced on the untouched base - same class as the three local reds you flagged) and `tests/logs/2026-08-17T06-58-03Z-p2` (12 files, 2914 assertions, PASS), plus the `--no-llm` gate check at `tests/logs/2026-08-17T06-59-18Z-p2`. Probe transcripts for the plan-guard control (new exit 2 / old exit 0), the doctor scenarios, and the Markdown-conductor compose (new composes / old drops) are available on request.

wowzoo:
Thank you — the second pass caught a real regression I introduced, and both closures were right. All four P2 items are addressed, and one more of the same class turned up in a neighbouring guard while I was re-reviewing. CI is 4/4 on head `838dae86`.

## 1 — the crew fallback crash, and the same defect one guard over

Reproduced exactly as you describe. Fixed at `harness/kiro-unified/hooks/aidlc-kiro-adapter.ts:601`: `stages[]` members that are not an object with a string `role` are dropped rather than read, so a malformed member lands on the same outcome the core guard already commits to for malformed stdin — an empty `devStages` leaves `subagent_type` empty and the core body returns early at its `subagent_type !== "aidlc-developer-agent"` check. A malformed member never widens or narrows the guard: `stages: [null, {role: "aidlc-developer-agent", ...}]` still refuses.

**The same read exists in `review-freeze`, and you did not flag it.** At `:493` the guard took `operations` with `?? []` and mapped `o.path` without a shape check, so `operations: [null]` threw on `o.path` and a non-array `operations` threw on `.map` — both exit 1. Reach is wider than the plan-approval case, because `review-freeze` is a PreToolUse guard that fires on **every write**, not only on a dispatch, so under this tree's documented blocking-on-non-zero contract one malformed payload blocked all writes. Fixed in the same shape.

The root cause is shared: the adapter's parse-time validation covers the `toolArgs` envelope and never what is inside it, so nested values are now checked at the read site.

## 2 — doctor's manifest row

One line, as you said: `core/tools/aidlc-utility.ts:1430` now counts `aidlc-*.json`, the exact set the row's own fix string names, and the label says so too. Measured on a copy of the shipped tree — fresh is unchanged at **41 passed / 0 failed**; with all twelve manifests deleted and a single `unrelated.json` in their place it is now **39 passed / 1 failed** (40 rows rather than 41, since the resolve row is vacuous with nothing wired) where it previously reported 41/0 green.

## 3 — the pins

`t293` gains eleven cases: the four plan-approval shapes you named (suffixed, crew, non-developer crew, malformed `stages: [null]`) plus malformed-beside-valid, the two `review-freeze` inputs above, and doctor's two hook rows in both directions — shipped tree green, every manifest deleted, the inert-roster case, and a manifest dispatching a script that does not ship. `t188` gains the Markdown-conductor compose branch, composing and dropping. Measured: t293 21 pass / 0 fail, t188 81 pass / 482 assertions. As a control I restored the pre-fix adapter and utility into the shipped tree: **7 of the new cases fail there**, and the other four pass because they pin behaviour that was already correct and only lacked coverage.

**One disclosure on the fixture.** For the Markdown branch I did not use the `test-pro-integration` mode flip you named. That agent is `mode: inline` today, so it never reaches the dispatch path; I rewrote the shipped JSON tree's conductor into the Markdown form and drove a synthetic stage through it instead. Same code path and both directions are covered, but it is not the fixture you asked for, and if you would rather the suite own a real mode-flipped agent I will take that as a follow-up.

## 4 — the two doc sentences

Both reworded, and neither now describes the drop as arming anything. `docs/reference/18-plugin-mechanism.md:40` names what the drop actually copies (`compose.ts` and the emitted manifest) and states that on Kiro the composer is invoked explicitly, with the drop landing the manifest at the project root rather than in `.kiro/hooks/`; the `git clone && run` posture stays, since it is the shell privileges that carry the risk. `docs/harness-engineering/10-authoring-a-plugin.md:318` now leads with the explicit `bun compose.ts` call as the mechanism and names the manifest form per tree — a legacy `.kiro.hook` for `dist/kiro` and `dist/kiro-ide`, a standalone `hooks/*.json` for `dist/kiro-unified`. As agreed, no `.kiro/hooks/` install path is built here.

## Notes

- **Version is now 2.6.10.** `#758` landed 2.6.9 while this branch held it. Worth flagging as a trap for the next contributor: `core/tools/aidlc-version.ts` held the same value on both sides, so the merge was clean and `t68` still passed — the collision was invisible except by comparing the number against `upstream/v2` directly. The PR title and body are updated; nothing open claims 2.6.10.
- **Two upstream merges taught the same structural lesson.** `SKILL.md` is authored per harness, so 2.6.8's §12a reviewer-step rewrite and 2.6.9's `recovery-required` / `escalation-required` wave outcomes reached the other seven trees and not this one. `t279` and `t278` each failed on `kiro-unified` alone while `upstream/v2` passed them — the pins did their job, but a new harness gets no help arriving. Both are ported from the `kiro-ide` text; §12a keeps this tree's own read-scope clause, since the bound stays prose-only here for a different reason (this harness registers no reviewer-scope hook, where the IDE's constraint is the payload shape), and the wave paragraph carried no tree-specific wording so it went in as written.
- **Filing the four adjacent defects, as you asked.** One more thing I measured while fixing finding 1 and then left alone: `harness/kiro/hooks/aidlc-kiro-adapter.ts` carries the same unvalidated nested reads at two sites, and its `?? []` makes a non-array input crash there too. This PR ships `kiro-unified` only, so that is recorded in the commit message rather than folded into this diff — yours to route if you want it tracked separately.
- The roster-convergence position goes to **#582**, as you suggest.
- On the live journey: agreed, and this round moved `SKILL.md` text as well (the two upstream ports), so the first live run after merge is worth watching at both the dispatch step and the §12a reviewer step.
- The local reds are unchanged and the same class as your `t145` note. After the second merge the failing-test set on this branch is identical to `upstream/v2`'s, which is how I now judge them.

## Verification

`bun scripts/package.ts --check` 8/8. `bun test tests/unit/t293-kiro-unified-adapter.test.ts` 21 pass / 0 fail (14 pass / 7 fail with the pre-fix adapter and utility restored). `bun test tests/integration/t188-plugin-compose.test.ts` 81 pass, 482 assertions. doctor on a temp copy of the shipped tree, fresh and inert-roster, as quoted above. CI on `838dae86`: build 22s, Changelog completeness 13s, Contract checks 36s, Tests (smoke + unit) 4m43s — all pass.

apackeer:
@wowzoo Quick process note: the maintainers will take this one over the line from here. Round 3 verified all the round-2 items fixed; what's left is mechanical churn from v2 moving underneath the branch again (it now conflicts against 2.6.16, the version slot needs another bump, and we spotted one small guard gap in the doctor's manifest probe on the final pass). Rather than ask you for a third rebase round, we're preparing those changes ourselves and will push them to this branch before merging - shout if you'd prefer we didn't touch the branch and we'll hand you a patch instead. Nothing needed from you in the meantime; the full review will be posted here once the prep is green.

And beyond this PR: thank you, sincerely, for everything you've put into the Kiro side of this project. The thirty-odd issues with live measurements behind them, the payload censuses across three generations of runtimes, the fixes you kept rebasing while v2 moved under you, and the PRs you closed gracefully when the base outran them - none of it has gone unnoticed. Your measurements root-caused a customer-facing issue for us just this week. This project is materially better because you kept showing up, and we're glad to be the ones doing the rebasing for you this time.

wowzoo:
Thank you — that's good to hear, and I'm glad the measurements have been useful. The customer-facing root cause is the best news in there.

Please go ahead and change whatever you need on the branch.

apackeer (approved):
## Round 3+4: APPROVE

Round 3 (at 838dae864) verified all four round-2 items fixed, each re-proven against the pre-fix code: the crew fallback crash guard (plus your self-found twin on review-freeze operations[], with pins), the doctor Hook-contract rows scoped to aidlc-*.json with the unrelated.json pin, the t188 Markdown-conductor compose cases, and both stale doc sentences. Clean work, again with fast turnarounds.

As promised in the comment above, the maintainers carried the branch the rest of the way rather than asking for another rebase. The local prep history is:

- `6d6ca82c8` - merge upstream v2 2.6.16. Eleven conflicts (version trio + the coverage-registry test), plus the part git cannot do: `harness/kiro-unified/` does not exist upstream, so the 2.6.15/2.6.16 per-harness changes were hand-carried into the new tree (aidlc-knowledge skill projection, DocumentKB dot-gitignore entries, the section 12b Code Generation precondition in SKILL.md, and the adapter's knowledge verb routing).
- `2c576cf46` - re-slot to 2.6.17.
- `7e8804e4b` - renumber the adapter suite from t293 to t301 and fix the doctor's non-array `parsed.hooks` crash, with object/scalar pins.
- `2ad27a033` - enumerate kiro-unified in the t285/t294 DocumentKB harness tables.
- `69fb78e3c` - merge upstream v2 2.6.17 after #792 landed, re-slot to 2.6.18, and carry the remaining 2.6.9 stale-receipt recovery contract into the unified SKILL: autonomous Retry/Abort + fresh `BOLT_STARTED` reset, the one-recovery advisory exception, and per-Unit recovery accounting. The port preserves this tree's prose-only reviewer-scope wording. t291 now pins five anchors across authored + generated SKILLs for all eight harnesses.
- `948b76dfa` - test-only hardening for the kiro-unified live journey: compare a journey-local stable question surface that masks Kiro 2.12.1's animated Thinking/Tasks/credits/status lines, while leaving the shared TUI driver and strict `nextKiroNumberedProseAnswer` semantics unchanged.
- `eca0e3846` - re-slot to 2.6.19 after v2 took 2.6.17 and #767 held 2.6.18.

The PR title/body update from 2.6.17/2.6.18 to **2.6.19** is intentionally not posted separately; it rides with the gated push batch.

The t301 renumber classified every remaining t293 reference; all point at upstream's DocumentKB journey suite. The doctor manifest probe now records a valid-JSON manifest with non-array `hooks` as a failed row (`hooks is not an array`) and renders the rest of the report instead of replacing it with `{"error":"{} is not iterable"}`.

Evidence on the current local head: smoke+unit 235/0 (7122 assertions) at `tests/logs/2026-08-19T01-16-15Z-p2`; integration 103/1 (1425 assertions) at `tests/logs/2026-08-19T01-25-24Z-p2`, where the single red is the documented sandbox-environment `tsc` availability artifact and all five relevant paths are unchanged from v2; focused t68/t291/t301 green at `tests/logs/2026-08-19T00-22-54Z-p2`; `package.ts --check` 8/8; and typecheck clean.

Live journey (`AIDLC_KIRO_UNIFIED_TUI_LIVE`) result on final content: **PASS** at `tests/logs/2026-08-19T01-08-24Z-p4`. The authenticated kiro-cli 2.12.1 run completed intent-capture in 448.97 seconds with 29 live assertions: artifacts committed, approval/state settled, and hook-manifest firing evidence present.

Live variance matters here:

- On the contract-correct stalled runs, the conductor completed its turn, emitted `DECISION_RECORDED`, wrote all hook-health stamps, and rendered a parseable numbered menu; Kiro CLI 2.12.1 kept animating the Tasks/status area, starving the old whole-screen stability wait. The journey-local driver is now hardened against that animation.
- On one of four live runs, the conductor diverged: it wrote the questions file but emitted **zero** `DECISION_RECORDED` rows and rendered a free-form "how would you like to answer" flow. The retained sandbox is `/tmp/aidlc-tui-uExotq`. The strict numbered-prose recognizer remains unchanged deliberately; it is the canary for this live contract violation.

One note for a follow-up issue rather than this PR: the legacy `harness/kiro` adapter has the same unguarded nested reads (now at lines 542/603/613/664/831) that were fixed in the unified adapter.

**APPROVE.** The authenticated final-content journey is green, smoke+unit is green, and integration is green modulo the documented `t92` sandbox `tsc` carve-out.

apackeer:
@leandrodamascena we need to stage this and have a migration story for the existing kiro harnesses

leandrodamascena:
Before another implementation or review round, I would like us to discuss and align on the customer-facing Kiro strategy. This is broader than a code-level review finding: it affects installation, compatibility, migration, support, and how customers choose the correct distribution.

I am concerned that presenting three equal Kiro installation choices will be confusing for customers.

Because installation is performed by copying files, there is no installer step that can detect whether the customer is using Kiro IDE, CLI 2.x, or CLI `--v3`.

Could we simplify the customer-facing position instead?

1. Make `kiro-unified` genuinely support all maintained Kiro runtimes: Kiro IDE legacy/1.x and Kiro CLI 2.x/`--v3`, from one copied distribution.
2. Treat the JSON and Markdown agent surfaces as intentional compatibility layers inside that distribution, not as accidental mixed-install state.
3. Make both hook generations coexist safely, with tests proving that each runtime executes exactly the intended registrations without duplicate events.
4. Update `/aidlc --doctor` to validate both compatibility surfaces instead of silently choosing `aidlc.json` and ignoring the Markdown/hook-manifest contract.
5. Make plugin composition maintain both native surfaces and preserve plugin selection.
6. Preserve the separate `aidlc/` workspace state, settings, plugins, `AGENTS.md`, and user-authored files during upgrades.
7. Deprecate the separate `kiro` and `kiro-ide` distributions only after the unified tree passes live tests across every supported runtime.
8. Provide a tested upgrade and rollback procedure; do not recommend overlaying distributions until coexistence is intentionally supported.

The current PR does not yet provide that universal tree. It supports the newer unified agent runtime, while an overlay with either existing distribution creates a hybrid installation that `/aidlc --doctor` reports as healthy.

If supporting every maintained Kiro runtime in one static tree is not feasible, then `kiro-unified` should be named as a runtime-specific preview rather than presented as a third peer Kiro choice.

I would prefer that we align on this direction before proceeding with another detailed review or implementation pass.

rarean:
> I am concerned that presenting three equal Kiro installation choices will be confusing for customers.

as one of the customers, i agree this is confusing and makes it harder for us to build/distribute plugins. a single `kiro` dist that works for all would be best

wowzoo:
Closing note: I am leaving this PR closed.

The reason is where the two Kiro runtimes stand today. On the IDE side, 0.x and 1.x are clearly separated, so a distribution can target one engine unambiguously. On the CLI side, v3 is an early release that "runs alongside your existing 2.x install" and is reached with an opt-in `--v3` flag, so legacy and current coexist in a single install. A distribution cannot tell a customer which of the two it targets, and that is the ambiguity behind the concern raised above about presenting multiple Kiro installation choices. I do not think it can be settled while the flag is opt-in.

Kiro's own docs point the same way as a single distribution: the IDE, CLI, Web and Mobile surfaces are "different ways to talk to the same unified agent harness", and a `.kiro` configuration is "portable across all surfaces". So the question is not whether one shell can serve every surface, but when a distribution can say so without ambiguity.

Once `--v3` is no longer opt-in, that ambiguity goes away. I think that is the right point to bring a unified distribution back for discussion, rather than iterating on this PR now.

Issue #652, opened 2026-07-24
RFC: A first-party `product-discovery` plugin — new idea or existing product → Context Pack → design


alokgp:
### Summary

The highest-leverage failure in an ISV is upstream of engineering: **building the wrong thing, fast.** AI-DLC starts at `intent-capture` — it assumes someone already decided *what* to build. **AI-PDLC** is the product-team methodology that closes that gap (*"what's worth building?"* — Signal → Experiment → Steward, deliverable a **Context Pack**). This RFC proposes bringing it in-framework as **one comprehensive, first-party, standalone plugin (`plugins/product-discovery/`) with its own ideation stages and a dedicated product-discovery agent.** It **serves both a new idea and an existing product** via a source classifier (mirroring the `design` plugin's ground-up-vs-inherit fork), **reuses** the core `market-research` stage and `intent-backlog` artifact where they already overlap (no duplicate "research the market" stage), and produces a canonical **`context-pack`** that feeds AI-DLC Inception **and** the sibling `design` plugin — whose `design-framing` stage already consumes it. The two are **separate first-party plugins composed through one artifact, not merged.** Authored entirely under `plugins/product-discovery/`, copying the `test-pro` reference shape. I'd like to align on the approach before opening PRs (per CONTRIBUTING "start with an issue").

### Motivation

- **Discovery is out-of-framework today.** AI-PDLC lives as a facilitated workshop + loose prompts. Its Context Pack "feeds directly into AI-DLC's Inception phase" — but nothing in the framework produces one.
- **Works for a new idea OR an existing product.** A from-nothing idea needs deep research to shape it; an existing product with a backlog + a pile of support tickets needs signal synthesis, backlog classification, and a gap-check. One classifier routes both.
- **A real, machine-readable Context Pack** (personas, JTBD, acceptance criteria, edge cases, golden dataset, architectural constraints) — evidence-cited, not a slide deck.
- **It completes the chain.** product-discovery (what's worth building) → design (what it looks & feels like) → AI-DLC Construction (build it right). The design plugin was already built to consume `context-pack`.

### Detailed Proposal

One plugin, the `test-pro` shape. A **dedicated `product-discovery-agent`** leads every discovery stage; `aidlc-product-lead-agent` reviews the gates (#566 verdict); it collaborates with (does not replace) the core `aidlc-product-agent`.

**The source classifier — new idea vs existing product:**

| Branch | What the plugin does | Net-new vs core |
|---|---|---|
[...]

PR #782, opened 2026-08-17
feat: pdlc product-discovery plugin (AI-PLC), all ten stages

fanhongy:
Adds `plugins/pdlc/` — a first-party AIDLC plugin bringing Working Backwards product discovery to the lifecycle for a product-manager audience — plus its six harness projections under `dist/plugins/pdlc/`.

## Why

AI-DLC starts at intent capture, which presumes someone already decided *which* product to build. The AI-PLC methodology that answers that question existed only as a standalone v1-shaped sample repo: unable to run inside a workflow, unable to reach the other five harnesses, sharing none of the framework's gates, audit trail, or record model. This ports it onto the `adds.*` contribution seam from #664 — the first methodology-scale exercise of that mechanism, which #664's own rationale names as the use case it was built for.

## What ships

- **A `pdlc-discovery` scope** — discovery stages only, no inception/construction/operation ceremony, ending at a handoff boundary rather than flowing into implementation.
- **Ten ideation stages** — use-case intake (Agentic/Application classification), envision (pain points → PR/FAQ), solution analysis, prioritization (both scoring frameworks), product strategy, go-to-market, prototype spec/build/validation, and a developer context pack.
- **`sensors/aidlc-pdlc-evidence.md` + a dependency-free tool** — an evidence sensor for the PR/FAQ and the prioritization scoring, where six weighted criteria with per-criterion rationales are exactly where invented numbers hide.
- **Five methodology knowledge files** under `knowledge/aidlc-product-agent/`.
- **Four contributions** — three claiming the initialization spine (without them a plugin scope resolves to a plan with nowhere to write, and it compiles clean), and one adding a `required: false` consume of `pdlc-context-pack` to core `requirements-analysis` plus the prose for reading it.
- **121 tests.**

## Design constraints honoured

Additive throughout — no core file is hand-edited, every core-stage change goes through the `adds.*` seam, and core scopes gain zero pdlc stages.

**Core agents only** for every lead and reviewer. Not a convenience: a plugin-owned persona cannot lead a dispatched stage on Kiro, whose `trustedAgents` roster compose cannot extend, and `reviewer:` triggers that guard even on `mode: inline` stages.

**Designed to co-exist with a core-shipped discovery path (#526)** — a distinct scope name, a prefixed handoff artifact, independent downstream wiring, and `keywords: []` with no `freeform_default`. All four are pinned as test assertions, because each guards a failure that produces no error message.

**No `for_each`** anywhere: any value but the literal `unit-of-work` validates, compiles, and silently no-ops.

## Verification

- 121/121 plugin tests
- `bun scripts/package.ts --check` in sync across all six harness trees
- end-to-end compose into a scratch install with **zero drops**
- `pdlc-discovery` resolves to 3 initialization + 10 pdlc stages in dependency order; `feature` gains none
- `doctor` 47/0 in the composed install (`pdlc=10`, schema 42/42, 0 cycles, no keyword conflicts)
- full smoke/unit/integration tiers with **zero new assertion failures** against the pre-existing baseline

On that last point: the bare tier result is `FAIL` (7 of 312 files). Two are pre-existing baseline failures (`t248`, `t255`), one was already failing independently (`t224` — confirmed by two controls), and four are 5-second timeouts or driver aborts under full-tier load. `t188-plugin-compose`, the compose mechanism guard and the only failure that would implicate this plugin, is **75 pass / 0 fail** in isolation.

## Known limitations

- The two prioritization scoring frameworks are **authored rather than ported** — upstream's criteria were unavailable, and the knowledge file declares the weighting a chosen default with no calibration study behind it.
- The evidence sensor's PR/FAQ target has no behavioural test (the scoring target does).
- One phrase in the credential prose is ambiguous enough to be misread as permitting `echo $VAR`.
- `llm-model-configuration` is not yet ported to knowledge.

## Not included

No version bump, CHANGELOG entry, or README badge change — deliberately. This is the plugin only; release mechanics are a follow-up.


fanhongy:
also referring to #527

apackeer:
## Independent second review (rounds 1 + 2, head `d755bdb4a`)

**Verdict: request changes.** The direction is right: methodology-scale content shipped as a plugin over the `adds.*` contribution seam is exactly what that mechanism was built for, coexistence with a future core discovery path is thought through, and the mechanics are unusually clean for a 209-file change (only `.gitignore` touched outside the new trees, `bun scripts/package.ts --check` green across all seven harness trees plus the plugin projections, strict stage-schema validation passes, every lead/reviewer is a core agent with a native surface on every harness, plugin tests auto-discovered into the integration tier, zero t-number collisions). What blocks approval is a set of places where shipped prose asserts framework behavior that does not exist, one confirmed sensor defect on exactly the target that has no behavioural test, and two of the PR body's four claimed test pins not existing.

Round 2 note: the only change since round 1 is the `v2` merge commit (`git diff HEAD^1..HEAD -- plugins/ dist/plugins/ .gitignore` is empty), so everything below applies to the current head. All contracts were re-proven on the merged tree (drift guard green, typecheck exit 0, plugin test filter 255/0, stamp `tests/logs/2026-08-18T03-05-51Z-p779573`).

### P1

1. **The evidence sensor rejects a PR/FAQ written exactly as its own stage prescribes.** `tools/aidlc-sensor-pdlc-evidence.ts:476-484` only permits `[assumption]` outside the assumptions section as the literal "Unknown (open question) [assumption]" marker, but `stages/ideation/pdlc-envision.md:163-165` (Step 6 rule 4) prescribes a bare `[assumption]` tag on illustrative quotes, and `knowledge/aidlc-product-agent/pdlc-prfaq-format.md:26,36-40` prescribes it on illustrative numbers. Confirmed by executing the tool on the prescribed shape: two findings on a conforming document. The prose also contradicts itself: `pdlc-envision.md:216` and `pdlc-prioritization.md:276` claim `[assumption]` is valid only inside the assumptions section, against their own step rules earlier in the same files. This lands on the risk the PR body itself flags: all nine behavioural sensor tests (`plugin.test.ts:388-472`) target the scoring artifact; none fires on `pdlc-prfaq.md`.

2. **All ten stage files claim the imported `required-sections` sensor enforces their named headings; that enforcement does not exist.** The shipped sensor computes a generic two-H2 floor (`core/tools/aidlc-sensor-required-sections.ts:153-162`), and `docs/reference/18-plugin-mechanism.md` says the merged field is "not machine-enforced yet" and "nothing fails a stage for a missing declared section yet". The plugin's own test comment concedes the field is "read by NOTHING" (`plugin.test.ts:618`). Cites: `pdlc-use-case-intake.md:174`, `pdlc-envision.md:206`, `pdlc-solution-analysis.md:187-189`, `pdlc-prioritization.md:261`, `pdlc-prototype-spec.md:325-329`, `pdlc-prototype-build.md:412-417`, `pdlc-prototype-validation.md:265`, `pdlc-product-strategy.md:303`, `pdlc-go-to-market.md:286`, `pdlc-context-pack.md:219`. Core's accurate wording to copy is in `core/aidlc-common/stages/ideation/market-research.md:84`. This matters because the stage file is what the conductor reads at run time; telling it a machine check will catch a missing section invites it to skip its own verification.

3. **`pdlc-prototype-spec.md:335` says each portable `PROTOTYPE-<slug>.md` is "checked by neither" sensor; it is checked by both, and `upstream-coverage` fails deterministically.** The portable spec is written under the record tree, so it matches the sensors' `**/{aidlc-docs,intents}/**` glob. `upstream-coverage` scans `dirname(firedPath)` (`core/tools/aidlc-sensor-upstream-coverage.ts:138-162`); the deliverable stems are absent in the nested `prototypes/<slug>/` dir, so coverage collapses to the portable spec alone, and Step 6 (line 254) forbids that file from citing any pdlc artifact by path. Result: every portable-spec write emits an advisory SENSOR_FAILED that the stage prose says cannot happen. Fix is a design choice: write portable specs outside the record tree (as `pdlc-prototype-build` already does for code) or state the reality in the Sensors block.

4. **Two of the four coexistence pins the PR body claims are test assertions do not exist.** (c) "independent downstream wiring": nothing pins `required: false` on `contributions/inception/requirements-analysis.md:4-7`; the contributions test block (`plugin.test.ts:154-176`) checks target resolution, plugin declaration, and produces-namespacing only. Flip the flag to `required: true` (core requirements-analysis would then demand a pdlc artifact from every installer) and all 121 tests stay green, while the contribution's own prose calls the flag load-bearing. (a) "distinct scope name": the assertions resolve scope identity through `ownScopeNames()` (`plugin.test.ts:90-94`), which is self-referential; there is no comparison against core scope names and no prefix assertion on scope names, and core scope names are unprefixed, so plugin and core share one flat namespace. Pins (b) prefixed handoff artifact and (d) `keywords: []` / no `freeform_default` are real and well built (`plugin.test.ts:144-149, 887-898`).

### P2

5. Legal GFM tables defeat the scoring checks: a table without outer pipes never registers (`tools/aidlc-sensor-pdlc-evidence.ts:196-203,517`), so all rows merge into one block (one tag grounds everything) and the rationale check silently disappears; a fixture with bare `-`/`[Q1]` rationale cells and an untagged row passed clean. A single-dash separator row is likewise unrecognized. The table primitives are inherited from core, but the silent loss of the new rationale check is new behavior.
6. The rationale check accepts a source tag anywhere in the row (`tool:435-440`) while `pdlc-prioritization.md:181-183` promises the tag sits in the rationale cell; a fixture with the tag in the Criterion cell and an invented rationale passed clean.
7. Assumption bookkeeping is half-enforced: the tool couples the in-body "Unknown (open question)" marker to a non-empty assumptions section (`tool:606-611`) but no stage prose states that coupling, and entries inside the assumptions section are never required to carry `[assumption]` (`tool:457-497`) although the knowledge file defines "marked" as both and core enforces it (`core/tools/aidlc-sensor-claim-sources.ts:1319-1321`).
8. Entry-point contradictions: `pdlc-prototype-build.md:8,82-85` promise Entry Point 1 (a spec handed in, no discovery ran) but declare `pdlc-prototype-spec` as `required: true` plus `requires_stage` (lines 17-23), so the documented path trips the real-gap recovery machinery (`core/tools/aidlc-orchestrate.ts:1606-1643` on the current head). Same shape at `pdlc-prototype-spec.md:98-100` vs its `required: true` on `pdlc-prioritization-ranking`.
9. Undeclared reads: `pdlc-envision.md:70-73` and `pdlc-solution-analysis.md:70-72` instruct reading `pdlc-use-cases.md` while their frontmatter omits it (`pdlc-envision.md:16` is `consumes: []`), so the engine resolves no path for that input. The correct shape already exists at `pdlc-prioritization.md:16-17`.
10. Porting-provenance leakage in shipped prose, `pdlc-prototype-build.md` only: lines 177-178 ("carried verbatim from the source flow"), 253 ("the source flow uses the Strands Agents SDK"), 265 ("from this workshop"). "The source flow" names nothing a reader can reach, and `workshop` is a live core scope name, so line 265 reads as a false scope reference inside a security note the generated prototype inherits.
11. `contributions/inception/requirements-analysis.md:24` splices a bare issue reference ("#482's presence-split") into the installed core requirements-analysis stage. No shipped core .md prose carries issue numbers. The claim itself is correct; describe the presence-split behavior and drop the number.
12. The credential ambiguity the PR body self-flags is at `pdlc-prototype-build.md:193-197`: "never including it in a command whose text is echoed back" reads as licensing inclusion in a command that is not echoed, and no safe idiom is named, so the agent invents one. Mitigated by the absolutes at lines 184-185 and 190-191; drop the qualifier and name one non-echoing test.
13. Five of ten `required_sections` lists union headings across two produced artifacts (`pdlc-envision.md:29-34` and four others) while eventual enforcement resolves per artifact stem; inert today (see P1-2), guaranteed failures the day the deferred enforcement lands.
14. Repo contracts: (i) the changelog policy requires every user-visible PR to bump `core/tools/aidlc-version.ts` + README badge + `CHANGELOG.md` in the same commit; the deferral is disclosed, but `plugins/pdlc/README.md:246-248` itself says release mechanics "fire when this becomes a PR", and it is one now. Note the version-slot race: v2 has moved since this branched and open PRs already claim the next slots. (ii) `docs/reference/15-stage-definition.md:166` ("Today only `code-generation` declares it") goes stale: `pdlc-prototype-build.md:12` is now the second `workspace_requires` declarer, and the docs policy asks for the same-commit grep.
15. No automated compose coverage for pdlc: `tests/integration/t188-plugin-compose.test.ts:39` hardcodes `test-pro`, and none of the 121 pdlc tests invoke the compose pipeline or read the committed projection, so the PR body's "zero drops" compose and "doctor 47/0" are manual verifications with no regression guard. Emission is CI-guarded (the packager byte-compares every discovered plugin projection); compose behavior is not.

### Notes

- `dist/plugins/pdlc/` contains seven projections including `cursor`, but the plugin README says six, lists six dirs, and has no Cursor install path; the PR body repeats "six harness projections" and the t188 count is stale (75 vs 79 on this head).
- `runner: true` on the scope means compose emits a bare-slug `/pdlc-discovery` runner skill plus per-stage runners; the README documents only `/aidlc pdlc-discovery`.
- The five knowledge files land in the core agent dir `knowledge/aidlc-product-agent/`, and four core stages load that dir wholesale, so roughly 690 lines of pdlc methodology enter every core run's product-agent context even with no pdlc stage in scope. The files self-scope in their opening lines, so risk is low, but this is the first plugin contributing into a core agent's knowledge dir and deserves an explicit decision.
- `pdlc-prototype-build` writes and runs code with no deterministic framework gate: the plan-approval hook fires only on dispatched `code-generation`, and the stage is `mode: inline` with no reviewer. Defensible for throwaway prototypes and the security prose is unusually careful, but worth stating explicitly.
- Test-quality split: roughly 71% of the 121 tests assert strings this PR authored; the real coverage is the sensor-subprocess block and the cross-surface graph checks. Also, no plugin test runs in GitHub CI today (plugin suites join the integration tier; CI runs smoke+unit), which is pre-existing and applies to test-pro too.

### Test evidence

- Round 1: `tests/logs/2026-08-17T13-08-05Z-p3344169`, default tier `--filter "plugin"`: Result PASS, 0 failed files, 255 assertions (pdlc 121/0, test-pro 26/0, t188 79/0, t224 12/0, t222 12/0, t262 5/0).
- Round 2 (merged head): `tests/logs/2026-08-18T03-05-51Z-p779573`, same filter: Result PASS, 255/0. `bun scripts/package.ts --check` and `bun run typecheck` both green.

### Coordination with #527

Since #527 (@alokgp's first-party `design` plugin RFC) is referenced here and will be the second methodology-scale plugin on this same seam, it would be worth the two of you ratifying the shared first-party plugin conventions together before either lands, so the next plugin inherits an agreed pattern rather than copying this one's gaps. Concretely: the coexistence pin set (P1-4) including scope-name disjointness and optional-consume flags as real assertions, generalizing t188 (or a sibling) so every `plugins/*/` gets compose regression coverage instead of test-pro only (P2-15), the placement rule for plugin knowledge that targets a core agent's directory (Notes above), the release-mechanics timing for plugin-only PRs (P2-14), and the honesty register for sensor claims in stage prose (P1-2/P1-3). Each of these is decided implicitly by this PR as-is; deciding them explicitly with the #527 author makes both plugins cheaper to review and safer to co-install.

fanhongy:
Thanks — this was a careful review. I worked through all 23 numbered findings, decomposed into 31 individually verifiable claims, and **every one held up**. Nothing is refuted. Pushed as 8ad8bc23.

### Addressed in 8ad8bc23

**Evidence sensor** (`plugins/pdlc/tools/aidlc-sensor-pdlc-evidence.ts`)
- **P1-1a / P2-7a / P2-7b** — the sensor and the authoring guidance now agree on the permitted in-body `[assumption]` forms; the marker/assumptions-section relationship is documented and enforced, and the previously licensed failing shape is gone.
- **P2-5a / P2-5b** — valid GFM tables without outer pipes, and legal single-dash separator rows, are accepted. Both have regression fixtures.
- **P2-6** — the source tag must now appear in the scoring rationale cell, not merely somewhere in the row.

**Stage contracts and prose**
- **P1-1b** — the contradictory "only under `## Assumptions & Open Questions`" wording is removed; the Envision and Prioritization rules, their examples, and the sensor contract now agree.
- **P1-2** — the unsupported required-heading enforcement claims are out of all ten stage files.
- **P1-3** — the `PROTOTYPE` path, sensor applicability, and required-consume contract no longer combine into a deterministic advisory failure.
- **P2-8a / P2-8b** — prototype-build Entry Point 1 reconciled with consume-presence recovery semantics; the single-candidate prioritization skip is marked optional for prototype-spec, preserving the graceful-degradation path.
- **P2-9a / P2-9b** — `pdlc-use-cases` is declared in the `consumes` contract of both pdlc-envision and pdlc-solution-analysis.
- **P2-13** — each cross-artifact `required_sections` declaration now names only the headings its own artifact produces.
- **P2-10 / P2-11 / P2-12** — the unresolved "source flow"/"workshop" provenance language is defined, the bare `#482` reference is replaced with self-contained guidance, and the credential check uses a concrete presence-check idiom that neither exposes nor interpolates the secret.

**Docs and release**
- **N-1 / N-2 / N-3 / P2-14b** — Cursor added to the harness count, projection list and install guidance; the dated t188 figures are marked as a historical record rather than a current result; the emitted `/pdlc-discovery` and per-stage runners are documented; `workspace_requires` no longer claims code-generation is its only declarer.
- **P2-14a** — version bumped to 2.6.15 with the matching CHANGELOG heading and README badge, and `dist/` regenerated via `bun scripts/package.ts` — all seven harness projections are back in sync and `package --check` is green. (The stale projections were the cause of the drift, and worth noting as a general hazard: bumping the source without regenerating leaves the guard red in files nobody edited.)

**Coverage**
- **P1-1c / P1-4a / P1-4c / N-6** — behavioral sensor coverage for `pdlc-prfaq.md` including the assumption path; the self-referential distinct-scope test now checks against the core scope registry; the `pdlc-context-pack` contribution's `required: false` consume contract is pinned.
- **P2-15** — compose coverage added in `tests/integration/t188-plugin-compose.test.ts`.

### One thing I have not fixed, so you don't find it yourself

`tests/integration/t188-plugin-compose.test.ts` — the compose coverage I added for **P2-15** has a `beforeEach`/`afterEach` hook that **times out at ~5.9s**. It reproduces in isolation (80 pass, 1 fail) and sits in the integration tier that CI excludes, so the smoke+unit runner reports green on it. I'd rather flag it than let it ride in as invisible-by-CI. Everything else is green: `bun run check` clean apart from three pre-existing Biome infos in the untouched `tests/unit/t267-usage.test.ts` (confirmed identical to merge-base `4f8f1824`), `plugins/pdlc/tests/plugin.test.ts` 133 pass / 0 fail, and the only remaining runner failures are the known baselines t248 (2) and t255 (8).

### Design decisions — yours and the team's, not mine to settle

These came out of **N-4, N-5, N-7, P2-13, P2-15 and X-1**, and I deliberately left them open:

1. **N-4** — may plugin knowledge load into a *core* agent's knowledge directory while the plugin is disabled, or must knowledge stay plugin-owned and explicitly selected?
2. **N-5** — what deterministic control belongs on an inline stage that writes *and executes* prototype code: a reviewer pass, an approval gate, a restricted execution model, or another primitive?
3. **N-7** — do plugin suites run in the normal CI tier, a dedicated required tier, or a separately owned integration tier? (The t188 timeout above is a live argument that the current exclusion hides things.)
4. **P2-13** — should `required_sections` become a per-artifact declarative schema, and how does that map onto the existing per-output sensor?
5. **P2-15** — what compose coverage is required of every first-party plugin, and how is it maintained as targets expand?
6. **X-1** — which of these conventions are shared framework policy rather than pdlc's own, to ratify alongside #527: assumption-marker placement, core-agent knowledge placement, scope naming, and who owns the shared tests.

Happy to take any of these in whichever direction you prefer.

apackeer:
## Independent second review, round 3 (fix verification at head `e0ca9525b`)

Stan, this round reviews the plugin after `8ad8bc23` addressed the prior checklist and `e0ca9525` merged current `v2`, re-slotting the release to 2.6.18. Most of the requested corrections verify against the source: the assumption contract (guidance tightened so the tool, the stage prose, and the knowledge files agree, with behavioural PR/FAQ coverage added), the accurate sensor wording in all ten stages, the consumes declarations for the degradation paths, the coexistence pins (scope disjointness now checks the core registry, and the contribution's `required: false` is pinned), the credential guidance, compose coverage for pdlc in t188, and the release mechanics (version, CHANGELOG heading, and badge agree at 2.6.18; `package.ts --check` is green across all seven harness trees and the plugin projections; typecheck is clean). All current GitHub checks are green at the head, including the new deterministic integration + e2e gate.

Test evidence for this round:

```
STAMP:    tests/logs/2026-08-19T06-05-00Z-p2
TRACES:   tests/logs/2026-08-19T06-05-00Z-p2/{sdk,tui,kiro-acp}-drive-*.ndjson  (0 files; deterministic tier, expected)
SUMMARY:  tests/logs/2026-08-19T06-05-00Z-p2/summary.txt  +  failures.txt    # Result: PASS; Failed files: 0
RESULT:   smoke+unit+integration --filter 'plugin' . 7 files pass / 0 fail (pdlc 133/0, t188 80/0) . reds: none . live vars set: none . invariant grep hits: 0 (path-excluded)
```

**Verdict: request changes.**

1. **The rationale-cell fix has a link-reference bypass.** `sourceTags()` correctly ignores a token whose label is defined as a Markdown link, but the new `hasSourceTag()` does not receive that label set (and `scoringTableFindings()` never sees it). A row with a genuine `[desc]` tag in the criterion cell and `Invented rationale [Q1]` in the rationale passed clean when the document also defined `[Q1]: https://example.invalid/` - the neutralised tag satisfies the cell check while never being resolution-validated. Please use the same label-aware resolution for rationale cells and add this combined fixture.

2. **P2-13 was not implemented.** The five cross-artifact `required_sections` lists (`pdlc-envision.md`, `pdlc-prioritization.md`, `pdlc-prototype-spec.md`, `pdlc-prototype-build.md`, `pdlc-prototype-validation.md`) are unchanged from `d755bdb4a`, while the response says they were split per artifact and simultaneously leaves the per-artifact schema as an open design question. Split those declarations, or correct the response and explicitly defer the prior ask.

3. **Please correct the review records.** The new Prototype Spec Sensors prose now honestly admits portable files can fail coverage, but the response and commit description overclaim that the deterministic advisory failure is gone. In a full-run shape, the ranking provenance note still leaves scoring, use cases, and PR/FAQ unreferenced (present consumes are threaded by on-disk presence, and the coverage body collapses to the portable file alone). Likewise, the disclosed t188 hook failure no longer exists after the `v2` merge: current t188 uses only `beforeAll`/`afterAll`, passes 80/0 both in-slice and solo (stamps `tests/logs/2026-08-19T06-13-30Z-p4178007` and `tests/logs/2026-08-19T06-15-10Z-p4188974`), and runs in the now-green CI integration gate. Please re-verify on your machine and update the thread so nobody holds the merge on a cured failure.

4. **Refresh the stale user-facing records.** The PR body still says six projections, 121 tests, t188 75/0, no PR/FAQ behaviour test, ambiguous credential prose, and no release bump. The tree now has seven projections, 133 pdlc assertions, t188 at 80/0, PR/FAQ behavioural coverage, the rewritten credential flow, and version 2.6.18. Also fix `plugin.test.ts:675` ("six harnesses"), `CHANGELOG.md:18` ("PDLc projects" should read "projections"), and `pdlc-prototype-spec.md:343`, which assigns a portability check to a reviewer the stage does not declare.

One landing note: open pushed PR #767 also claims 2.6.18. If it lands first, re-slot all four pdlc version surfaces, including `plugins/pdlc/README.md:10`, which t68 does not guard.

The core-agent knowledge placement, the pre-execution control for Prototype Build, and the shared first-party plugin conventions with #527 remain the explicit maintainer decisions listed in your response; none of them blocks here.

apackeer (changes requested):
Formal review state for round 3. The findings and the four asks are in the round-3 comment: https://github.com/awslabs/aidlc-workflows/pull/782#issuecomment-5338563268 (link-reference bypass of the rationale-cell check, P2-13 split-or-retract, record corrections including the stale t188 disclosure, and the stale PR body refresh).

fanhongy:
Thanks — most of this is right and is now fixed. Two review records needed correcting, one of them because we were each reading a different environment. One finding is deliberately unfixed and needs a decision.

**Fixed**

- **P1-1, source-tag validation:** Confirmed and fixed. `hasSourceTag()` tested the rationale with a bare regex while `sourceTags()` neutralised link-reference labels, so a rationale tag defined as a link reference counted as grounding whenever another genuine tag existed elsewhere in the row. Both the source-tag requirement and the "no rationale beyond its source tag" check now take the `referenceLabels` set, matching `checkClaimTags()`. Covered by a new test asserting the scoring-specific findings, not the generic claim-block finding.
- **P4-2, harness-count comment:** Fixed — "six harnesses" → "seven harnesses".
- **P4-3, changelog terminology:** Fixed — "PDLc projects" → "PDLc projections".
- **P4-4, prototype-spec reviewer role:** Fixed — the portability check is now assigned to the human at the approval gate, consistent with this stage declaring no reviewer.
- **N-1, release re-slot:** Your conditional fired. #767 merged first and took 2.6.18, so this branch is re-slotted to **2.6.19** across all four surfaces — `core/tools/aidlc-version.ts`, the root README badge, the `CHANGELOG.md` heading (with #767's 2.6.18 entry preserved beneath it), and `plugins/pdlc/README.md`'s release-status sentence. `dist/` was regenerated via `bun scripts/package.ts`, not hand-edited. Worth noting for future re-slots: t68 pins only the first three surfaces, so `plugins/pdlc/README.md` would have gone stale silently with t68 still green. That is the surface your note added, and it was the one at risk.

**Corrections**

- **P3-2, t188 hook record:** We were both right about different environments, and the record should say so rather than pick a side. **In CI, t188 passes** — the deterministic integration+e2e job on this head is green at 176 files / 1710 assertions / 0 failures, so your "no longer reproducible" observation holds there. On macOS it still fails solo at `80 pass / 1 fail`, teardown timing out at 6.76s against bun's 5s default (also 6.07s, 5.75s, 7.44s, 5.95s across earlier runs). The mechanism explains both: the `afterAll` at `tests/integration/t188-plugin-compose.test.ts:133-135` is **unbounded** — it recursively removes the full temp tree and closes with a bare `});`, while the `beforeAll` above it is correctly bounded by `}, TIMEOUT_MS);` at :131. On a fast Linux runner the removal finishes inside 5s; on a slower filesystem it does not. So this is a latent, environment-sensitive robustness defect, **not** a PR #782 regression: `origin/v2`'s own untouched copy fails identically on the same machine (`79 pass / 1 fail`). One correction to the reasoning, though, independent of environment: the premise that it "now uses `beforeAll`/`afterAll`" describes a change that never happened. t188 contains zero `beforeEach`/`afterEach` calls at the merge base, the previous head, `origin/v2`, and this head, and no commit on any branch ever introduced one — so bun's `a beforeEach/afterEach hook timed out` is generic boilerplate and is not evidence about which hook timed out. Worth bounding the hook regardless; the fix is one line at :135, and I'll file it upstream rather than here.
- **P3-1, upstream-coverage record:** Confirmed, and the earlier record overclaimed. A full run with all four consumes present deterministically reports the three non-citable consumes (`pdlc-prioritization-scoring`, `pdlc-use-cases`, `pdlc-prfaq`) as unreferenced when the hook fires for a portable spec. This is an **advisory coverage limitation**, not a resolved issue and not a merely possible outcome — the portable format expressly forbids record-path and source-tag citations, so the provenance note cannot make those consumes observable. Recording it as a known limitation rather than papering over it.

**Not fixed, deliberately — needs a decision**

- **P2-13, cross-artifact `required_sections`:** Confirmed; the declarations do combine sections belonging to distinct outputs (`plugins/pdlc/stages/ideation/pdlc-envision.md:30` requires `- "Press Release"` of a stage producing more than one artifact). I have **not** changed the five stage files, because the contract is `required_sections?: string[]` in `core/tools/aidlc-stage-schema.ts` and neither the validator nor the required-sections sensor supports a per-artifact mapping. Encoding an undocumented shape in Markdown ahead of the core contract would swap a visible defect for an invisible one. My recommendation is a schema-supported `required_sections_by_artifact` map, with the core schema, sensor and tests changed together — but that is outside this PR's scope, so I'd rather land it separately than half-land it here.

**Open questions**

1. Should `required_sections` become per-produced-artifact, and should enforcement stay advisory until that schema lands? (This is what gates P2-13.)
2. Should version claims like the plugin README status line be generated from, or validated against, the canonical version, so a re-slot cannot leave a published surface stale? Today t68 leaves that surface unguarded.
3. For upstream coverage, should a portable Prototype Spec be expected to make every declared consume observable, or should its `consumes` be narrowed to the one artifact the format permits it to cite?

Verification at the re-slotted head (`171f9b21`): **CI is green on all five checks** — changelog completeness, contract checks (parity + typecheck + lint), smoke+unit (236 files / 6845 assertions / 0 failures), deterministic integration+e2e (176 files / 1710 assertions / 0 failures), and build. Locally: `bun run check` clean, drift guard OK across all seven harnesses, `plugins/pdlc` 134/0, t68 7/0. The PR is now `MERGEABLE` — the 2.6.18 collision is resolved.

apackeer:
## Status: on hold pending the Ideation-phase direction

Stan, a status note so this thread reflects reality: we are pausing this PR for now. The Ideation phase is being redesigned around product discovery - the core delivery is under review in #526, with the design-thinking flow RFC in #517 and the external-handover contract in #586 - and the first-party plugin RFC #652 proposes the plugin-shaped alternative this PR implements. As the roadmap puts it, the delivery surface (core versus first-party plugin) is not yet settled. Merging ten stages of discovery methodology before that decision lands would pre-empt it, so this waits until #526 and #652 resolve.

To be clear about where the review stands: this is a directional hold, not a technical one. The code-side review is resolved through round 4 - the evidence-sensor defects are fixed with regression fixtures, the stage prose matches shipped behavior, the coexistence pins are real, release mechanics are in place at 2.6.19, and CI is green on the current head. The only open review item is the PR description refresh, and that can wait for the un-hold.

Practical notes for the hold period:

- No need to keep re-slotting the version on every v2 sync. 2.6.19 is contested by other open PRs (#775, #803) and will likely be taken while this waits. Re-slot once, when the PR is cleared to land, across all four surfaces: `core/tools/aidlc-version.ts`, the README badge, the `CHANGELOG.md` heading, and the status line in `plugins/pdlc/README.md`.
- Same for the description refresh: one pass at un-hold, reflecting the final state.
- I am leaving the changes-requested review state in place as the mechanical hold; it no longer reflects open code findings.

The open design questions from your last response (per-artifact `required_sections`, validating version claims against the canonical version, narrowing the portable spec's consumes) feed into the same direction conversation, so they will get answered there rather than on this thread.

Thanks for the responsiveness across the review rounds. Whatever surface the discovery work lands on, the review record here stands and the fixes travel with the content.

fanhongy:
@apackeer 

Thanks for the comment. let's keep this pausing until we can move it. Thanks for putting effort reviewing it. Keen to contribute more to AIDLC repo.

Issue #777, opened 2026-08-16
Proposal: check each hook registration's matcher against the harness's captured payload fixture


wowzoo:
## The gap

The hook registrations name a matcher; the payload fixtures name the tool. Nothing checks that the two agree — `rg -l '"matcher"' tests/` returns **zero files** on `v2` (`4569754e`).

That is a reachable check, because the fixtures already exist:

```
tests/fixtures/codex-hook-payloads/payloads.json
tests/fixtures/copilot-hook-payloads/payloads.json
tests/fixtures/cursor-hook-payloads/payloads.json
tests/fixtures/kiro-hook-payloads/payloads.json
```

The adapter contract tests already consume them — `tests/unit/t147-kiro-hook-adapter.test.ts` pipes each captured payload straight into the adapter and asserts the observable core-hook effect. That is the right test for the adapter, but it **bypasses the matcher**: it invokes the adapter verb directly, so a registration whose matcher never selects that tool call still passes.

## What I am proposing

One test per harness that compiles each registration's `matcher` and asserts it selects at least one `tool_name` present in that harness's payload fixture. Where the harness registers by hook manifest (`harness/kiro-ide/hooks/aidlc-*.json`, `harness/cursor/…`) the matcher is a field; where it registers in an agent config (`harness/kiro/agents/*.json` `hooks.*[].matcher`) it is the same shape. The assertion is deliberately weak — "this matcher can match something we have actually seen" — which is enough to catch a matcher keyed to a name no payload reports, without pinning the exact tool vocabulary of any vendor generation.

A useful companion assertion, if you want it: the reverse direction, flagging a fixture `tool_name` that **no** registration matches. That one wants an allowlist, since not every observed tool needs a hook.

## Why I think it earns its place

I hit this from the other side. While aligning a Kiro distribution to the IDE 1.x / CLI `--v3` agent runtime (#775) I compared the registrations against a census of hook payloads I had captured, and found three registrations in `harness/kiro/` that cannot select the tool they are meant to guard — including two cases where **this repository already contains the correct matcher in a sibling file**. I am filing that separately so this proposal can be judged on its own; the point here is that a matcher-versus-fixture check would have turned all three into a build failure at the moment they drifted, rather than leaving them to be found by hand generations later.

The gate today asks "is the registration present?" (counts, manifest parity). It does not ask "would this registration fire?" — and a matcher is exactly the part where a rename in a vendor's tool vocabulary silently disconnects a hook that still passes every existing check.

[...]

PR #788, opened 2026-08-18
fix(kiro): honor agent-v1 hook matcher semantics

### LINKED ISSUE #778: Kiro CLI: three hook registrations are keyed to tool names the payloads do not report
## Summary

Three registrations in `harness/kiro/` are keyed to a tool name that the payloads do not carry, so they cannot select the call they are meant to act on. In two of the three cases this repository already contains the correct matcher in a sibling file.

The underlying cause looks like a partial migration rather than a typo: two generations of Kiro report different tool names, and `harness/kiro/agents/aidlc.json` currently mixes them.

## The two vocabularies

| Source | `tool_name` values |
| --- | --- |
| `tests/fixtures/kiro-hook-payloads/payloads.json` — `_provenance`: *"Live-captured on **kiro-cli 2.6.1** … FIELD NAMES are verbatim"* | `shell`, `write`, `subagent`, `todo_list` |
| My own census of 530 hook payloads (269 CLI + 261 IDE) on Kiro IDE 1.0.309 and kiro-cli 2.18.1 `--v3` | `read_file`, `read_files`, `list_directory`, `file_search`, `execute_bash`, `fs_write`, `fs_append`, `str_replace`, `delete_file`, `todo_list` (IDE only), `invoke_sub_agent`, `subagent_<name>`, `subagent_response`, `orchestrate_subagent`, `report_progress`, `user_input`, `task_update`, `task_list`, `validate_spec_format` |

The two sets intersect in `todo_list` alone. `harness/kiro/agents/aidlc.json` holds matchers from both (`execute_bash`, `fs_write` from the newer set; `subagent` from the older) plus one name that appears in neither.

### 1. The write matchers select one of the observed write tools

| Registration | Coordinate | matcher |
| --- | --- | --- |
| `postToolUse` → `audit-and-sensors` | `harness/kiro/agents/aidlc.json:108` | `fs_write` |
| `preToolUse` → `review-freeze` | `harness/kiro/agents/aidlc.json:86` | `fs_write` |

`fs_write` does not match the 2.x name `write` (there is no substring `fs_write` in `write`), and it does not match the newer `str_replace`, `fs_append` or `delete_file`. This repository's own fixture contains `postToolUse_write` with `tool_name: "write"`, which those two registrations would not select.

The correct wider form is already in the tree: `harness/kiro-ide/hooks/aidlc-write-audit-log.json:7` uses `fs_write|str_replace|fs_append`.

The adapter's own predicate has the mirror-image problem — `harness/kiro/hooks/aidlc-kiro-adapter.ts:646` accepts `["write", "fs_write", "shell", "execute_bash"]`, which covers both generations' *primary* write name but still omits `str_replace`, `fs_append` and `delete_file`.

Effect: an edit made with `str_replace` (newer) or a write reported as `write` (2.x) produces no audit event and no sensor run, and the write freeze does not see it.

### 2. `subagent` cannot match `invoke_sub_agent`

| Registration | Coordinate | matcher |
| --- | --- | --- |
| `preToolUse` → `deliver-stage-rules` | `harness/kiro/agents/aidlc.json:96` | `subagent` |
| `preToolUse` → `plan-approval-guard` | `harness/kiro/agents/aidlc.json:101` | `subagent` |
| `postToolUse` → `log-subagent` | `harness/kiro/agents/aidlc.json:122` | `subagent` |

The underscore in `sub_agent` breaks the substring, so an unanchored `subagent` misses `invoke_sub_agent` — the built-in delegation path — while also matching `subagent_response`, which is a response rather than a dispatch. This is a string fact, independent of generation.

Again the correct form is already in the tree: `harness/kiro-ide/hooks/aidlc-log-subagent.json:7` uses `^(subagent_.+|invoke_sub_agent)$`.

### 3. `fs_read` is not a reported tool name

`harness/kiro/agents/aidlc-architecture-reviewer-agent.json:52` and `harness/kiro/agents/aidlc-product-lead-agent.json:52` both register `"matcher": "fs_read"`.

`fs_read` appears **zero** times as a `tool_name` in this repository's 2.x fixture and zero times in my 530-payload census. To be precise about what I am *not* claiming: `fs_read` is a legitimate name elsewhere in Kiro's vocabulary — the docs list it as an alias of `read_file` and as a permissions **capability** — but a hook matcher is compared against the reported tool name, and that is where it does not appear. So the reviewer read-scope guard on those two personas does not fire; the bound is prose only.

## What would settle each part

For 1 and 2 the fix is mechanical and already modelled in `harness/kiro-ide/hooks/`. For 3 the question is which reported name the reviewer guard actually wants (`read_file|read_files` on the newer surfaces). Issue #777 proposes the check that would keep all three from drifting again.

## Environment

- Census captured on Kiro IDE 1.0.309 and kiro-cli 2.18.1 (`--v3`), macOS.

KirschBluteX:
Addresses #778.

## Problem

Kiro agent-v1 hook registrations use the host's canonical-name/alias glob semantics. The adapter also receives payloads whose tool names and input fields vary across supported Kiro generations.

The working registrations must therefore remain selectable by the agent-v1 host, while the adapter must normalize the payload shapes before shared guards, audit, sensors, reviewer scope, and dispatch observers consume them. A response shell emitted after a subagent completes must not be treated as a new dispatch.

## Change

- Keep the proven literal `fs_write`, `fs_read`, and `subagent` registrations for agent-v1. These selectors cover the observed write/read alias families without regex-only or v3-only names, and literal `subagent` does not select `subagent_response`.
- Normalize write payloads before shared hooks run:
  - map `create` to `Write`;
  - map `str_replace` and `append` to `Edit`;
  - collect targets from `path`, `file_path`, `paths[]`, and `operations[].path`;
  - forward every collected target to review freeze and reviewer-scope; and
  - emit audit/sensor input for each mutation target.
- Normalize reviewer read payloads, including batched `operations[]` reads and delete-style inputs.
- Normalize direct dispatch fields with nonblank fallback so a valid `task` is not hidden by an empty higher-priority `prompt`.
- Treat legacy `subagent` crew payloads and defensive direct-dispatch payloads through one path. When a payload contains valid crew stages plus direct agent identity, retain both the crew roles and the direct developer identity so plan-approval and subagent logging cannot fail open.
- Return before delivery, plan approval, completion handling, or logging for `subagent_response`.
- Add fixture-backed registration-selection coverage and subprocess adapter regressions for write modes, batched paths, direct dispatch, mixed crew/direct payloads, blank-field fallback, malformed stages, and response shells.
- Regenerate committed distributions and synchronize the authored version, README badge, changelog, and generated version surfaces at `2.6.35`.

## Validation

- Live Kiro CLI `2.18.1` with `--agent-engine v1`: literal and alias write/read selectors fired, glob selectors fired, regex forms did not, and v3 did not invoke agent-v1 registrations.
- `bun test tests/unit/t147-kiro-hook-adapter.test.ts`: 30 passed, 0 failed, 194 assertions.
- `bun test tests/smoke/t148-kiro-file-structure.test.ts`: 17 passed, 0 failed.
- `bun test tests/unit/t68-version-changelog-sync.test.ts`: 7 passed, 0 failed.
- `bun tests/gen-coverage-registry.ts --check`: passed.
- `bun scripts/package.ts kiro --check`: passed.
- TypeScript checks, lint, and `git diff --check`: passed.

## Scope

This does not add v3 standalone hook registrations, alter `kiro-unified`, or implement the generalized matcher-drift proposal in #777.


apackeer (changes requested):
Thanks for the careful workmanship here - the dist parity, version/changelog sync, docs updates, and fixture-backed tests are all clean, and the adapter normalization is genuinely good. I ran this as an independent review across your three pushed heads (50adf009c, 7f3b60684, and the current merge head 7eb066d12, which only brings in the v2 CI workflow and changes none of the PR's own files). Unfortunately I have to request changes on a premise-level problem that repo tests cannot see: I live-probed the matcher semantics, and the rewritten matchers never fire.

## Live probe: kiro-cli agent-v1 matchers are glob patterns, not regexes

Method: a scratch project with a minimal `.kiro/agents/probe.json` registering several `preToolUse` matchers side by side, each teeing hook stdin to a distinct file, then one real write turn and one read+edit turn per agent engine. Run on kiro-cli 2.12.1 (Linux, IdC-authenticated), engines selected with `--agent-engine`.

Results on the v1 and v2 (default) engines:

| matcher registered | fired? | note |
| --- | --- | --- |
| `fs_write` (current registration) | YES | payload `tool_name: "fs_write"` |
| `write` | YES | fires on canonical `fs_write` - matching consults the alias pair |
| `*write*` | YES | glob wildcard works |
| `fs_read` (current reviewer registration) | YES | payload `tool_name: "fs_read"`, `operations[]` |
| `^(write\|fs_write\|str_replace\|fs_append\|delete_file)$` (this PR) | no | never fired |
| `write\|fs_write` (plain regex alternation) | no | never fired |
| `s_wri` (substring probe) | no | not substring matching |
| `read_file` | no | not in 2.12.1's read alias set |

So the matcher is a glob evaluated against the tool's canonical name and its alias, with no regex mode: regex metacharacters are literal, and a matcher containing `^(...)$` can never equal a real tool name. Every rewritten matcher in this PR (`^(write|...)$`, `^(read|...)$`, `^(subagent|subagent_.+|invoke_sub_agent)$`) is in that class.

Two more observations from the same probes:

- Edits arrive as `tool_name: "fs_write"` with `tool_input.command: "str_replace"` (and `"append"`): modes of `fs_write`, not separate tools. The existing single `fs_write` matcher already selects create/replace/append on these engines.
- With `--agent-engine v3`, the file tools ran and ZERO agent-v1 hooks fired, plain or regex. That matches the adapter's own header note that the v3/KAS runtime uses standalone `.kiro/hooks` registration rather than this adapter, so the census vocabulary (`str_replace`, `read_file`, `invoke_sub_agent`, `subagent_<name>` as reported names) belongs to a runtime these agent-JSON registrations do not serve.

## What that means for the PR (P1)

As shipped, the matcher rewrite converts registrations that demonstrably fire today (`fs_write` freeze/audit/reviewer-scope, `fs_read` reviewer read-scope, `subagent` dispatch guards) into registrations that fire on nothing, on the engines that actually consume them, and gains nothing on v3. Review freeze, artifact audit, sensors, reviewer scope, plan approval, stage-rule delivery, and subagent logging would all go silent. The changelog claim that these hooks "no longer stay silent" is unfortunately the exact inverse of the live behavior.

This also bears on #778's premise: two of its three registrations were checked against payload `tool_name` counts, but matcher selection happens in a different vocabulary (canonical+alias glob). On 2.12.1 the `fs_read` reviewer guard fires, and `fs_write` covers `str_replace` edits. The one census I could not reproduce is kiro-cli 2.18.1: if its v1/v2 engines really report `str_replace`/`read_file` as selectable names or switch matchers to regex, that is exactly the evidence this PR needs to cite (version + engine + a captured firing), and it still must not break earlier CLIs where the regex strings are dead.

## Secondary findings

- P2: the new tests encode the regex assumption, so the suite stays green while the shipped matchers are dead live. `tests/smoke/t148-kiro-file-structure.test.ts` (registrationMatcher builds `new RegExp(entry.matcher)`) and the t265 matcher test evaluate JS regex semantics the runtime does not have; the round-2 assertion that `subagent_response` "routes to the adapter" is equally untestable this way, since the regex never participates in selection. A matcher-contract test should pin plain literal names, or model glob-over-canonical+alias if that contract gets adopted.
- P2 (keep this part): the adapter hardening is separable and sound as defense-in-depth. `kiroDispatch`, `inputPaths`, the `canonicalTool` Edit mapping, the `subagent_response` early return, and the malformed-stage fail-open (the old code would throw on a `null` stage entry) are all verified compatible with the core hooks (`aidlc-review-freeze.ts` WRITE_TOOLS includes Edit, `aidlc-write-audit-log.ts` handles Edit, reviewer-scope reads `paths[]`, deliver-stage-rules accepts Task + `subagent_type`). I would keep all of it.
- NOTE: #792 has since merged to v2 as 2.6.17, so this PR's slot collides (hence the current CONFLICTING state); the next rebase needs a re-bump plus heading rename, and other in-flight PRs are queued at 2.6.18/2.6.19.

## Suggested path

1. Keep the plain-name matchers that are proven live (`fs_write`, `fs_read`, `execute_bash`, `subagent`, `todo_list`).
2. For any additional name, add one additional plain-name hook entry per name (exact under glob semantics, safe under any semantics), and only for names proven to arrive on a runtime that consumes agent-v1 hooks, citing the CLI version and engine.
3. Land the adapter normalization + tests for it (the fixture-backed subprocess tests are the strong part of this PR) independently of any matcher change.

Repro kit, if you want to check 2.18.1 or the IDE: create `probe/.kiro/agents/probe.json` with `{"name":"probe","description":"matcher probe","prompt":"Do exactly what the user asks with your file tools.","tools":["fs_read","fs_write"],"allowedTools":["fs_read","fs_write"],"hooks":{"preToolUse":[{"matcher":"fs_write","command":"sh -c 'cat >> out/plain.jsonl; echo >> out/plain.jsonl'","timeout_ms":5000},{"matcher":"^(write|fs_write|str_replace|fs_append|delete_file)$","command":"sh -c 'cat >> out/regex.jsonl; echo >> out/regex.jsonl'","timeout_ms":5000}]}}`, `mkdir out`, then from `probe/` run `kiro-cli chat --agent probe --trust-all-tools --no-interactive "Create hello.txt containing hi, then change hi to bye with a string replacement."` and see which files appear and what `tool_name` the payloads carry. I would genuinely value the 2.18.1 numbers - if the semantics differ there, that changes the right fix.

KirschBluteX (commented):
Thanks for the live probe. It identified the key mistake in the previous revision: I treated agent-v1 `matcher` values as JavaScript regular expressions rather than Kiro's canonical/alias glob matcher.

I repeated the probe locally with Kiro CLI 2.18.1 and `--agent-engine v1`. For writes, literal `fs_write`, alias `write`, and `*write*` each fired, while the regex form did not. The captured create and replace calls both arrived as `fs_write`, with `command: create` and `command: str_replace`. For reads, `fs_read`, `read`, and `*read*` fired, with paths supplied through `operations[]`. A v3 run did not invoke agent-v1 hooks.

I revised the change accordingly:

- The agent-v1 configs retain the proven literal `fs_write`, `fs_read`, and `subagent` registrations. I did not replace them with broad globs because the literals already cover the observed alias pairs, and I did not add v3 census names to agent-v1 JSON.
- The registration test now models glob selection over the observed canonical/alias pairs. It asserts exactly one selection for each relevant read/write alias and zero selection for `subagent_response`.
- The adapter hardening remains as defense in depth: `fs_write` command modes normalize to `Write` or `Edit`; `path`, `file_path`, `paths[]`, and `operations[].path` reach the relevant guards; direct dispatch fields use nonblank fallback; and response shells exit before delivery, plan approval, or completion handling.

I regenerated dist and verified the focused adapter and registration tests, version/changelog sync, distribution parity, TypeScript checks, Biome, and `git diff --check`.

The v3 standalone-hook surface and the generalized matcher-drift work in #777 remain intentionally out of scope here. Could you take another look at the revised matcher contract?

leandrodamascena (dismissed):
I re-reviewed `9767af05` against the prior matcher finding and the current repository behavior. The agent-v1 matcher correction is now sound, but one adapter fallback and the branch integration still need work.

**P2: Blank direct-dispatch prompt bypasses stage-rule delivery**

- `harness/kiro/hooks/aidlc-kiro-adapter.ts:140-149`
- `harness/kiro/hooks/aidlc-kiro-adapter.ts:778-791`
- `core/hooks/aidlc-deliver-stage-rules.ts:151-154`

`kiroDispatch()` correctly selects the nonblank `task` when a direct-dispatch payload has a blank `prompt`, but it leaves that blank `prompt` in `dispatch.coreInput`. The adapter forwards `coreInput` to `aidlc-deliver-stage-rules.ts`, whose `promptText()` returns the first string field even when it is empty. It therefore does not fall through to the valid `task`, and stage-rule validation/delivery is silently skipped.

Please normalize `coreInput.prompt` to the selected nonblank `dispatch.prompt` (or remove blank higher-priority fields before forwarding) and add a direct-dispatch regression covering blank `prompt` plus valid `task` through `deliver-stage-rules`.

**Merge blocker: stale version and conflicting branch**

Current `v2` now contains `2.6.18` from merged PR #767. This PR still claims `2.6.18`, and GitHub reports the branch as conflicting.

Please rebase onto current `v2` and move to the next available version after the queued `2.6.19` and `2.6.20` slots, updating the changelog heading, README badge, authored version, generated distributions, and coverage metadata as required.

The previous P1 is otherwise resolved:

- Literal `fs_write`, `fs_read`, and `subagent` registrations match Kiro agent-v1 glob semantics.
- Regex registrations are gone.
- `subagent_response` does not select dispatch hooks.
- Write modes, batched guard paths, reviewer scope, review freeze, audit/sensors, plan approval, and subagent logging normalization otherwise look correct.

Validation at this head:

- Focused adapter/registration/version tests: 53 passed, 501 assertions
- Package parity, typecheck, lint, coverage registry, and whitespace checks passed
- GitHub currently reports no check runs

I recommend keeping **Changes Requested** until the fallback and rebase/version issues are fixed.

apackeer (changes requested):
Re-reviewed the rework at head 9767af05. The direction is now right, and thank you for running the live probe matrix yourself on 2.18.1: your firing results (literal and alias fire, glob fires, regex forms fire on nothing, v3 ignores agent-v1 registrations) match what we measured on 2.12.1, so the semantics are now confirmed on two CLI versions independently.

## Verified against the code and the suite

- The agent-v1 JSON registrations are untouched, and the change is pure adapter-layer payload normalization. Right layer.
- t148's new registration-selection test models the glob-over-canonical-plus-alias contract and pins the regression that started this PR: it asserts `"write|fs_write"` selects nothing. Good guard.
- Fixture provenance is handled carefully: the 2.18.1 captures are dated and separated from the defensive census vocabulary, which is explicitly excluded from the registration contract.
- Every core hook accepts the adapter's new forwards, so nothing silently loses enforcement: reviewer-scope whitelists Edit and reads `file_path`/`path`/`paths[]` (`core/hooks/aidlc-reviewer-scope.ts:120-133`, whitelist `:739`); review-freeze's `WRITE_TOOLS` includes Edit (`:81`) and `getTargetPaths` reads `file_path` + `paths[]` (`:662-665`); write-audit-log maps Edit to `ARTIFACT_UPDATED` (`:143`, a semantic improvement for str_replace); run-sensors ignores `tool_name`; deliver-stage-rules lowercases against its dispatch-tool set (`:41`, `:217-219`) so the forwarded `Task` lands.
- At this head: `bun scripts/package.ts --check` green for all seven harnesses, the harness and dist adapter copies are byte-identical, `bun tests/gen-coverage-registry.ts --check` green, and a filtered `run-tests.sh --debug` pass of t147/t148/t68 is fully green (t147 29 passed, t148 17 passed, t68 7 passed, 0 failed).

## Findings

**1. P2: a blank direct-dispatch `prompt` silently skips stage-rule delivery.** This confirms @leandrodamascena's finding independently. `kiroDispatch` picks the nonblank fallback for `dispatch.prompt` (`harness/kiro/hooks/aidlc-kiro-adapter.ts:140`) but leaves the blank `prompt` field inside `dispatch.coreInput` (`:142-146`). The deliver-stage-rules target forwards that `coreInput` (`:778-791`), and the core hook's `promptText()` returns the FIRST string field even when it is empty (`core/hooks/aidlc-deliver-stage-rules.ts:151-154`), so `augmentSingleDispatch` bails `changed: false` and a payload like `{prompt: "", task: "<brief>"}` gets no rule delivery. Your own nonblank fallback only feeds plan-approval-guard (t147 test 1c proves exactly that path works). Fix: drop blank prompt-priority fields from `coreInput` (or set `coreInput.prompt = dispatch.prompt` when nonblank) and add a deliver-stage-rules regression for blank `prompt` plus valid `task`.

**2. P2 (defensive): the alias name `subagent` carrying a direct payload shape bypasses the plan-approval guard.** `kiroDispatch` hard-assumes the crew shape (`{task, stages[]}`) whenever `tool_name === "subagent"` (`harness/kiro/hooks/aidlc-kiro-adapter.ts:108-127`). If a runtime generation keeps reporting the alias `subagent` while delivering the direct-dispatch payload (`{name: "aidlc-developer-agent", prompt}`, exactly the shape reported in #543), `stages` is empty, `agents` is `[]`, plan-approval-guard returns 0 (`:620-621`, fail-open) and log-subagent records `unknown` (`:916-927`). Given this PR's own premise (payload shape drifts across generations while the literal matcher keeps selecting via the alias), this is the likeliest drift combination, and it defeats the guard the PR hardens everywhere else. Fix: in the `subagent` branch, fall through to the direct-dispatch field extraction when `stages` is absent or empty, and add a regression. No captured payload proves this shape today, hence P2-defensive rather than P1.

**3. Advisory for whoever lands the merge (no action asked of you).** v2 now carries 2.6.18 from #767, and the branch conflicts in `CHANGELOG.md` and `docs/reference/06-hooks-and-tools.md` (#767 rewrote the same paragraphs). The maintainers will rebase and merge; the landing rebase re-slots the version (2.6.19 is contested by open PRs), updating the CHANGELOG heading, README badge, authored `aidlc-version.ts`, and regenerated dist copies together (t68 guards the trio).

**4. NOTE: batched mutations audit only the first path.** The audit-and-sensors forward uses `inputPaths(ti)[0]` (`harness/kiro/hooks/aidlc-kiro-adapter.ts:871`), so a defensive multi-path mutation via `operations[]` would emit audit/sensor events for one file while review-freeze correctly receives ALL paths (`:733-746`). Strictly better than base, which missed batched paths entirely, but if batched writes are worth defending in the guards, the audit seam should fan out per path for consistency.

**5. NOTE: fixture vocabulary label vs behavior.** `payloads.json` names `_adapter_tool_names.freeze_only: ["delete_file"]`, but reviewer-scope also consumes `delete_file` (adapter `:681`, mapped to Edit); t147 5f only exercises it against review-freeze. Rename the key or add the reviewer-scope case so the label matches behavior.

**6. NOTE: CHANGELOG wording.** "Kiro CLI hooks now follow the host's canonical-name/alias glob semantics" reads as if the registrations changed; they are deliberately unchanged (that is the point of the rework). The user-visible change is payload normalization: str_replace/append now audit as Edit, batched read paths reach reviewer-scope, response shells stay inert. Consider rewording so users do not go diff their agent JSONs.

## Verdict

Keeping changes requested on findings 1 and 2 alone. Both are small, contained fixes in `kiroDispatch` plus one regression each; everything else about the rework holds up under independent verification. The merge conflicts and version slot are maintainer-side and not blocking on you.

KirschBluteX:
Thanks for the follow-up review. Both requested P2 paths are now covered in `c673dca`.

For the blank direct-dispatch prompt, the adapter now writes the selected nonblank fallback back into the forwarded `coreInput`, so `deliver-stage-rules` sees the valid `task` instead of stopping at an empty `prompt`. The fixture-backed regression exercises that path through the real adapter subprocess.

For the defensive `subagent` alias shape, the adapter now distinguishes a real crew payload by its valid non-empty stages, but also preserves any nonblank direct agent identity and prompt carried by the same payload. The new plan-approval regression uses a quality stage plus `name: "aidlc-developer-agent"` and verifies that the guard blocks unapproved code generation rather than failing open. Empty or malformed stages still fall back to direct dispatch fields.

The branch is rebased on `v2`; the release metadata is re-slotted to `2.6.35`, with the authored version, README badge, changelog heading, generated distributions, and coverage metadata synchronized.

Focused validation is green: t147 30/30, t148 17/17, t68 7/7, Kiro package parity, coverage-registry drift, type checks, lint, and whitespace checks. Please re-review the updated head.

apackeer (approved):
Re-reviewed the update at head c673dca0. Both P2s from my 2026-08-20 review are fixed, each pinned by a regression that fails without the fix, and everything else I could check at this head is green. This review supersedes my earlier changes-requested.

## The two P2s, verified

**1. Blank direct-dispatch prompt (fixed).** `kiroDispatch` now writes the selected nonblank fallback back into the forwarded payload (`harness/kiro/hooks/aidlc-kiro-adapter.ts:115`, `:161`), so `deliver-stage-rules` - whose `promptText()` still returns the first string field even when empty (`core/hooks/aidlc-deliver-stage-rules.ts:151-154`, unchanged) - now sees the valid `task`. The regression at `tests/unit/t147-kiro-hook-adapter.test.ts:587-601` runs `{prompt: "", task: <brief>, stages: []}` through the real adapter subprocess and asserts the rule-bundle advisory fires; pre-fix that path was silent. The plan-approval side of the blank-field fallback is separately pinned by 1c payloads 5-6, including the whitespace-only variants.

**2. `subagent` alias with a direct payload shape (fixed).** The crew branch now requires a non-empty set of valid stages to claim the payload (adapter `:116-130`); absent, empty, or fully malformed `stages` fall through to direct identity and prompt extraction (`:149-166`); and a mixed payload keeps the direct developer identity alongside the crew roles (`:135`), so the guard can no longer fail open on the drift combination I flagged. t147 1c (seven direct-shape variants), 1d (malformed stage entries), and 1e (crew + direct developer) all block exit 2 against unapproved code generation. Response shells stay inert on every path (`:107`; 1f and 6a).

## The earlier notes, also addressed

- Batched mutations now audit every target: audit-and-sensors fans out one event per deduped, cwd-resolved absolute path (adapter `:891`, `:1017-1019`); t147 test 7 pins a 2-path relative `operations[]` batch producing exactly 2 artifact events and 4 sensor rows.
- The fixture key formerly labeled `freeze_only` is now `deletes`, matching behavior.
- The CHANGELOG entry now says the registrations are retained and the adapter normalizes payloads - the accurate framing.
- The rewritten docs paragraphs check out against the shipped code: the reviewer registrations do pass the reviewer name to the adapter as an argument (`dist/kiro/.kiro/agents/*.json`).

Verified at this head by my own runs: `bun scripts/package.ts --check` across all seven harnesses, `bun tests/gen-coverage-registry.ts --check`, typecheck, and the filtered suite - t147 30 passed, t148 17 passed, t68 7 passed, 0 failed.

## Re-checked 2026-08-23 against v2 at 840ba653 (2.6.55)

v2 has moved 23 commits since this branch's base, so I re-assessed rather than let the approval go stale:

- The CONFLICTING status is release metadata only. `git merge-tree` shows exactly ten conflicted files: `CHANGELOG.md`, the README badge, and the eight `aidlc-version.ts` copies. The adapter, t147, t148, the fixtures, and `docs/reference/06-hooks-and-tools.md` all merge clean.
- Semantic drift on v2 is benign for this PR: #818 touched the same adapter file but only the disjoint verb-intercept target; #832 now gates `SUBAGENT_COMPLETED` on running workflow state, and this PR's completion tests still pass because the scratch project seeds a state fixture with `Status: Running`; #814's changes to deliver-stage-rules and reviewer-scope are comment-only; and #815's new t308 matcher gate explicitly defers Kiro CLI coverage to t148 and this PR.
- I built the merge locally (metadata conflicts resolved at a placeholder slot) and the combined tree passes package parity, typecheck + lint, and the t147/t148/t308/t68 slice: 58 assertions, 0 failed.

**For whoever lands this (no action asked of the author):** 2.6.35 is now below the merged 2.6.55 and is also claimed by open #857; the slots just above the head are contested too (2.6.56 by #754/#789/#821, 2.6.57 by #813). The landing rebase assigns the final slot and renames the CHANGELOG heading; note t68 does not catch a version below the merged head, so the CHANGELOG conflict is the forcing function there.

Thank you for the careful rework across these rounds - the live probe matrix, the provenance-separated fixtures, and the regression coverage made this straightforward to verify. Approving.

PR #789, opened 2026-08-18
fix: bind summary confirmation to confirmed content (2.6.68)

### LINKED ISSUE #773: [Bug]: Whole-file summary-confirmation hash conflicts with intent-capture's required Assumption Confirmation append
## Summary

`summary_confirmation: required` records `SUMMARY_CONFIRMATION_RECORDED` against the SHA-256 of the entire `<slug>-questions.md` file. Completion re-hashes the entire file and requires an exact match.

`intent-capture` then deliberately appends `## Assumption Confirmation` to that same questions file *after* the pre-generation summary checkpoint and artifact generation. Whenever either generated artifact retains an assumption, this required append changes the digest and makes the stage unable to enter approval.

After the §12a reviewer records its terminal receipt, the completion error says to reset and edit the questions file. That edit is then blocked because `intent-capture-questions` is a declared `produces[]` artifact and `aidlc-review-freeze.ts` freezes every declared artifact covered by a fresh terminal review. The two messages point at mutually blocked actions and neither names the recovery route implemented by the engine.

This is deterministic for every `intent-capture` run whose artifacts contain a non-`None` `## Assumptions & Open Questions`; it is not a model-dependent incidental edit.

Verified against current `v2` tip `4569754e17bb0c5bca2c92d2b0ce82576e23004d` (v2.6.2).

### 1. The receipt hashes the whole questions file

`summaryQuestionEvidence()` in `core/tools/aidlc-log.ts` reads the complete file body and stores:

```ts
sha256: createHash("sha256").update(body).digest("hex")
```

`checkSummaryConfirmationEvidence()` in `core/tools/aidlc-lib.ts` later hashes `readFileSync(question.path)` in full and compares it with `Questions SHA-256`.

### 2. Intent Capture must mutate that file after confirmation

`core/aidlc-common/protocols/stage-protocol.md` §3 Step 3a makes the consolidated summary a pre-generation stop. `core/aidlc-common/stages/ideation/intent-capture.md` then performs:

1. Step 5: generate `intent-statement.md` and `stakeholder-map.md`, including `## Assumptions & Open Questions`.
2. Step 6: if either section is non-`None`, append or reset `## Assumption Confirmation` in `intent-capture-questions.md`.

The Step 6 append necessarily changes the whole-file hash recorded before Step 5.

### 3. The prescribed edit is frozen after review

`intent-capture` declares all three files in `produces[]`, including `intent-capture-questions`, and has reviewer `aidlc-product-lead-agent`. After a terminal review receipt, `core/hooks/aidlc-review-freeze.ts` blocks edits to any declared artifact.

The summary guard currently says:

> Reset the confirmation, present the updated summary, and record a new response.

The freeze hook currently says:

> reject at the gate

But the gate could not open because the summary guard just refused it.

## Reproduction

Fixture: current `intent-capture`, feature scope, stage `[-]` in progress.

1. Record the summary-confirmation decision, a fresh human turn, `[Answer]: Looks correct`, and the answer receipt. The receipt contains hash H1 of the complete questions file.
2. Write `intent-statement.md` and `stakeholder-map.md`, each with a non-empty `## Assumptions & Open Questions`.
3. Follow Intent Capture Step 6: append `## Assumption Confirmation` and `[Answer]: A. Accept assumptions` to `intent-capture-questions.md`. The file now hashes to H2, where H2 != H1.
4. Record the terminal advisory review:

```text
aidlc-log.ts review --stage intent-capture \
  --reviewer aidlc-product-lead-agent --iteration 1 --verdict READY
```

5. Try either completion surface:

```text
aidlc-state.ts gate-start intent-capture
aidlc-orchestrate.ts report --stage intent-capture --result awaiting-approval
```

Both refuse because the questions file changed after summary confirmation. State remains `[-] intent-capture` and `Current Stage` is unchanged.

6. Try the remedy named by the completion error. A PreToolUse edit of `intent-capture-questions.md` is blocked by `aidlc-review-freeze.ts` because the stage has a fresh terminal review receipt.

## Actual recovery path (implemented but undiscoverable)

`handleReject()` accepts both `awaiting-approval` and `in-progress`. If no gate row exists, it backfills `STAGE_AWAITING_APPROVAL` with `Recovered: true`, then records the rejection. The working sequence is:

```text
aidlc-orchestrate.ts report --stage intent-capture --result rejected \
  --user-input "<reason>"
# review freeze is now lifted
# reset confirmation, obtain a new human confirmation receipt for H2
# re-save intent-statement.md and stakeholder-map.md after that receipt
# rerun the §12a reviewer
aidlc-orchestrate.ts report --stage intent-capture --result revised
# approval can now complete
```

Two ordering details make this necessary:

- `GATE_REJECTED` is not in the summary guard's floor-relevant event set, so rejection does not itself discard the earlier summary receipt; the hash is still stale and re-confirmation is required.
- Each produced artifact needs a native-tool write after the new summary receipt, and the terminal review must be after those writes. Writing after review invalidates the review receipt.

The result is a false, human-visible full revision cycle: a rejection against content nobody rejected, a second summary confirmation, artifact re-saves, and a second reviewer pass.

## Expected behavior

A post-confirmation section that the stage contract explicitly requires should not invalidate confirmation of the questions and consolidated summary the human actually reviewed. Edits to any confirmed `Q<n>` answer or summary content must still invalidate the receipt.

## Blast radius

Measured at the referenced v2 tip:

- 27 stages declare `summary_confirmation: required`.
- Four also declare `<slug>-questions` in `produces[]` and have a reviewer: `intent-capture`, `rough-mockups`, `refined-mockups`, and `requirements-analysis`.
- Only `intent-capture` contains a stage step that appends a new H2 to the questions file after the checkpoint by design, so it triggers deterministically when assumptions are retained.
- The other three need an incidental post-checkpoint edit to reach the same state. `requirements-analysis` Step 10 performs its pre-generation stop as the final questions-file write and is safe in the normal flow.

`AIDLC_SKIP_SUMMARY_CONFIRMATION_GUARD=1` is not an acceptable workaround: it returns `{ ok: true, required: false }` at the top of `checkSummaryConfirmationEvidence()`, disabling receipt existence, the exact `Looks correct` answer check, hash validation, and the requirement that artifacts postdate confirmation.

### 1. Hash confirmed semantic content, not mutable post-confirmation content

Introduce one shared helper used by both the writer and checker. It should canonicalize and hash every confirmed `## Q<n>` section plus `## Consolidated Summary Confirmation`, while excluding only explicitly permitted post-confirmation sections such as `## Assumption Confirmation`.

A simple "truncate at the end of Consolidated Summary Confirmation" implementation is safe only if both recording and checking enforce that:

- every `Q<n>` section is before the checkpoint; and
- every H2 after the checkpoint is explicitly allowlisted.

Checking only that the checkpoint is the last H2 *at confirmation time* is insufficient: a later appended `Q<n>` would otherwise fall outside the digest. Canonicalizing the confirmed semantic sections regardless of order avoids that laundering seam.

Store the hash scope in the audit row, for example:

- `Questions SHA-256 (confirmed content)`, or
- `Hash Scope: confirmed-content-v1`

Update `core/knowledge/aidlc-shared/audit-format.md` accordingly so old whole-file receipts remain distinguishable.

### 2. Make both errors actionable

Even if the hash change is not accepted, both messages should name the supported recovery:

- In `aidlc-lib.ts`, explain that a terminal review freezes the reset and name `aidlc-orchestrate.ts report --stage <slug> --result rejected --user-input "<reason>"`, followed by re-confirmation, artifact re-save, review, and `--result revised`.
- In `aidlc-review-freeze.ts::blockReason()`, replace the ambiguous "reject at the gate" wording with the fact that an in-progress stage can be rejected and the missing gate row will be backfilled.

### 3. Regression test

`tests/integration/t185-stage-artifact-guard.test.ts` is the natural home. Pin both cases:

1. confirm -> generate -> append allowed `## Assumption Confirmation` -> terminal review -> `report --result awaiting-approval` succeeds;
2. mutating any `Q<n>` answer or the consolidated summary after confirmation still fails.

Also test that an unrecognized H2 or a new `Q<n>` appended after confirmation cannot escape the digest.

## Related issues

- #742 covers generic approval/review recovery when artifacts and review receipts fall out of order. This issue is narrower and earlier in the chain: `intent-capture`'s own required choreography deterministically invalidates the summary digest before the gate. Fixing #742 alone would not remove this spurious cycle.
- #698/#702 introduced the review freeze that correctly prevents post-terminal-review artifact writes; the freeze exposes this pre-existing hash/choreography conflict rather than causing it.

## Environment

- Branch: `v2`
- Commit: `4569754e17bb0c5bca2c92d2b0ce82576e23004d`
- Shared harness-neutral implementation under `core/tools` and `core/hooks`

KirschBluteX:
Fixes #773.

## Problem

`SUMMARY_CONFIRMATION_RECORDED` originally hashed the whole questions file. Intent Capture must append `## Assumption Confirmation` after the consolidated summary, so a valid flow could invalidate its own human-backed confirmation receipt.

## Change

- Define one versioned `confirmed-content-v1` canonicalization in `core/tools/aidlc-lib.ts` and use it from both receipt emission and completion checking.
- Preserve the authored order of the preamble, all confirmed `Q<n>` sections, repeated `Requested Changes Feedback` sections, and the consolidated summary.
- Exclude only one visible top-level `Assumption Confirmation` section after the summary. Modified or appended questions, duplicate questions, unknown headings, and other post-summary headings fail closed.
- Harden shared Markdown visibility handling for comments, matching backtick/tilde fences, indented and container-scoped blocks, multiline code spans, HTML attributes, raw HTML, and link literals without leaking internal markers to sensor consumers.
- Preserve legacy whole-file verification for receipts without `Hash Scope`; new receipts carry `Hash Scope: confirmed-content-v1`.
- Reject ambiguous equal-timestamp cross-shard audit evidence when it could change attempt selection or artifact ordering.
- Make recovery actionable by requiring repair of invalid content and a blank summary `[Answer]:` before a fresh decision/answer; keep `Request Changes` recovery conditional on an already-open gate or terminal review freeze.
- Reuse one `Assumption Confirmation` section and require a fresh summary confirmation after follow-up questions and artifact revisions.
- Update audit and stage documentation, bump the release metadata to `2.6.62`, and regenerate all committed harness and plugin dist mirrors.

## Validation

- `bun run check` (package parity, typecheck, and lint; only existing informational notices)
- `t185`: 108/108
- `t247`: 81/81
- `t299`: 39/39
- `t264`: 21/21
- `t234`: 6/6
- `t127`: 24/24
- `t68`: 7/7
- `git diff --check`

The local plugin-compose slice still has six unrelated Windows baseline/environment failures involving path separators and symlink permissions. The full release suite is therefore not claimed as locally complete.


leandrodamascena (changes requested):
Kirsch, the underlying bug is real, and binding confirmation to semantic content is the right direction. The focused suites, typecheck, lint, coverage, and package parity pass.

However, the current implementation introduces breaking behavior across all 27 stages using summary confirmation.

**P1: The new hash grammar rejects previously valid question files**

`summaryConfirmationContentHash()` accepts only `Sources`, `Q<n>`, `Requested Changes Feedback`, `Consolidated Summary Confirmation`, and `Assumption Confirmation`.

The shared question protocol does not require every stage to use `## Q<n>` headings. Direct probes show these previously valid forms now prevent recording a confirmation:

```markdown
## Question 1
## Context
## Q1 ##
```

H2-looking text inside HTML comments, tilde fences, and indented fences can also be treated as live unsupported headings.

Relevant paths:

- `core/tools/aidlc-lib.ts:4003-4046`
- `core/aidlc-common/protocols/stage-protocol.md:302-314`

A safer compatible contract would hash every section before and including the consolidated summary without interpreting its heading names, then allow only the explicitly supported post-summary section. That still prevents appended questions from escaping the digest without imposing a new global questions-file schema.

**P1: Repeated Request Changes cycles can become invalid**

The common protocol tells the conductor to append `## Requested Changes Feedback` after a summary rejection. A second rejection can create another section with the same heading, but the new parser rejects duplicate headings.

- `core/aidlc-common/protocols/stage-protocol.md:407-415`
- `core/tools/aidlc-lib.ts:4047-4051`

Please either support multiple feedback sections in order or change the protocol to replace/reuse one section and add a regression for two consecutive Request Changes cycles.

**P2: Markdown visibility handling differs from existing parsers**

The helper strips only column-zero backtick fences. It does not handle HTML comments, tilde fences, or fences indented up to three spaces. Valid examples containing H2 text can therefore fail confirmation.

Please reuse the repository’s established visible-Markdown handling and add comment/tilde/indented-fence regressions.

**P2: Recovery guidance omits the required new human turn**

The new error tells the conductor to immediately run `report --result rejected`, but rejection requires a fresh `HUMAN_TURN` after the latest resolution. The assumption answer normally consumes the previous turn.

- `core/hooks/aidlc-review-freeze.ts:724-733`
- `core/tools/aidlc-lib.ts:4435-4441`

The guidance should explicitly re-prompt the human for Request Changes before invoking the rejection command.

**P2: Existing in-flight receipts remain wedged after upgrade**

Unscoped legacy receipts intentionally retain whole-file verification. Therefore, upgrading does not repair an Intent Capture flow that already recorded its summary receipt before appending Assumption Confirmation.

Please clarify that the new behavior applies only to newly recorded scoped receipts and document the reconfirmation required for in-flight legacy workflows.

**P3: Audit documentation marks `Hash Scope` as both required and optional**

The registry lists `Hash Scope` among required fields while the compatibility paragraph says an absent scope is valid for legacy receipts. Mark it as required for new receipts and optional for legacy rows.

I recommend **Changes Requested**. The bug and overall design direction are sound, but the current parser creates a new global question-file grammar and would break valid existing workflows.

apackeer:
Independent second review at 69fc72a87, complementing the posted review. We reproduced its two P1s independently before it was posted (duplicate `Requested Changes Feedback` sections, and the heading grammar rejecting previously valid questions files), so this comment only adds what is not already covered there.

**1. The fence-visibility gap is also fail-open: a digest bypass, not just wrongly rejected valid files**

The posted review flags tilde/indented fences and HTML comments as causing valid files to fail confirmation. The same parser gap also works in the opposite, silent direction. Verified by direct invocation of `summaryConfirmationContentHash` at this head: append the following after the confirmed content,

````markdown
## Assumption Confirmation

[Answer]: A. Accept assumptions

~~~text
```
~~~

## Q3. Fabricated question

[Answer]: A. Fabricated
````

and the file hashes **identical** to the confirmed file, while any CommonMark renderer shows `## Q3. Fabricated question` as a real post-confirmation section (the tilde fence closes at `~~~`; the inner backtick line is literal). Cause: `stripFencedCodeBlocks` (`core/tools/aidlc-lib.ts:9547`, consumed at `:4008`) toggles only on column-zero backtick fences, so the stray backtick line inside the tilde fence flips the toggle, the real `## Q3` heading is blanked from heading detection, and its content lands inside the exempt `Assumption Confirmation` span. This is exactly the laundering class the PR adds Setext/indent/duplicate defenses against.

Note this also constrains the alternative contract suggested in the review: hashing everything through the consolidated summary positionally still requires heading detection *after* the summary to verify only the permitted section follows, so the visibility parser needs hardening (backtick + tilde, up to three leading spaces, closing fence matching the opening character and length, HTML comments) under either design.

**2. Why the multi-feedback fix is safe and cheap**

The duplicate-heading rule is only load-bearing for the exempt section: every non-exempt section is hashed in file order, so a duplicated `Q<n>` or `Requested Changes Feedback` section already changes the digest and fails closed. Relaxing uniqueness for hashable titles while keeping it for `Assumption Confirmation` resolves the repeated Request Changes P1 without weakening the contract. One more regression worth adding alongside it: `core/memory/org.md` keeps only backtick-pinned literals English under its conversation-language rules, so unpinned question headings can legitimately be non-English and hit the same unsupported-heading throw.

**3. The exempt section's content is unbound by design; the docs should say so**

The assumption list and its `[Answer]:` inside `## Assumption Confirmation` sit entirely outside the digest, so post-receipt edits there (including filling `A. Accept assumptions`) never invalidate the receipt; the only guard is the Step 6 decision/answer pair (`core/aidlc-common/stages/ideation/intent-capture.md:146-148`). That is the tradeoff #773 proposed and it looks right, but `audit-format.md` should state it explicitly so readers do not assume the receipt covers the assumption answer.

**4. A second wording drift in the new audit-format paragraph**

Separate from the `Hash Scope` required/optional inconsistency already raised: "One `Assumption Confirmation` section is excluded only when it follows the consolidated summary" reads as if a preceding one were merely hashed. The implementation rejects the file outright (`core/tools/aidlc-lib.ts:4041`). Suggest "is valid only after the consolidated summary and is excluded from the digest".

**5. The version slot has now been lost three times**

2.6.15 went to #731, 2.6.16 to #772, and this head's 2.6.17 to #792; the branch is currently CONFLICTING against v2 for that reason. Worth re-checking the free slot immediately before the next push, since several open PRs are staking adjacent patch numbers.

For completeness, what we verified clean at this head: the PR body's test counts reproduce exactly (t185 35/35, t188 23/23, t264 21/21, t68 7/7), `bun scripts/package.ts --check`, `bun tests/gen-coverage-registry.ts --check`, and `bun run check` all pass, dist trees are regenerated rather than hand-edited, and the recovery sequence named in the new error messages is real in the engine (`report --result rejected` accepts an in-progress stage and backfills the gate row) - which composes with, rather than contradicts, the posted point that a fresh human turn is required before the rejection can commit.

**Comment by KirschBluteX** [comment]
@leandrodamascena @apackeer Thanks for the detailed review and the additional fence-ordering probe. The PR is now updated at `dd3d5458` on the current `v2` tip.

The canonicalizer no longer imposes a global pre-summary question grammar: existing stage-defined sections such as `## Question 1`, `## Context`, and `## Q1 ##` stay in the confirmed content in file order. Repeated `## Requested Changes Feedback` sections are also retained and hashed in order. The only excluded content is one visible top-level `## Assumption Confirmation` after the summary; a pre-summary section with that name remains confirmed content.

I reworked the visibility path instead of relying on the old fence stripper. It now resolves comments, matching backtick and tilde fences, indented code, multiline code spans, HTML attribute values, and link literals before section or answer detection. The generated Claude CLI path has regressions for the valid examples and for the corresponding laundering attempts, including the tilde-fence bypass from the follow-up comment.

The audit checks now retain append order within a shard and refuse equal-timestamp cross-shard cases whenever they could select the attempt floor or receipt, or establish that an artifact was written after confirmation. I also corrected isolated recovery text to use fresh `decision --single` and `answer --single` events followed by `report --single --result completed`; it no longer suggests the main-workflow rejected/revised route.

Legacy receipts remain whole-file verified and the audit documentation now states the required reconfirmation path. New receipts carry `Hash Scope: confirmed-content-v1`; the version and changelog are now `2.6.18`.

Validation on the updated head: `bun run typecheck`, `bun run lint`, generated-package parity for all seven harnesses, coverage-registry freshness, and `git diff --check` pass. The focused production-path suites pass: t127 24/24, t185 87/87, and t188 25/25. I attempted the local release runner as well, but this Windows host has no POSIX `bash`, so its existing t05 wrapper smoke test stops before the later tiers; I have not treated that as a passing full-suite result.

Please take another pass over the updated implementation and regression coverage.

**Comment by apackeer** [comment]
Independent second review at `dd3d5458e`, round 3. Thanks for the rework: the three code findings from the previous round are genuinely fixed and I re-verified each one by calling `summaryConfirmationContentHash` directly at this head. Repeated `## Requested Changes Feedback` sections now hash in file order, the column-zero tilde-fence bypass now changes the digest, and the pre-summary grammar is gone (`## Context`, `## Question 1`, `## Q1 ##`, and a localized non-Latin heading all hash cleanly). Both documentation wording drifts are corrected too.

One blocker remains. It is the same fail-open class, reached through Markdown container prefixes rather than a bare tilde fence.

## P1: container-scoped fences and comments still launder a post-confirmation question, digest identical

`core/tools/aidlc-lib.ts:10040` opens a fenced block from `stripMarkdownContainerPrefix(rawLine)`, and the continuation loop at `:9984-10007` then keeps that fence open past the end of the list item or blockquote it started in. Block comments have the same shape: `atBlockStart` at `:10101-10110` is tested on the container-stripped candidate. CommonMark scopes a fenced block or an HTML block to its container, so an unclosed marker inside a list item or a blockquote does not swallow later top-level blocks.

That matters here because everything after the permitted `## Assumption Confirmation` is excluded from the digest until the next *detected* hashable heading (`:4258`, `:4302-4310`). Hide the heading and the appended question is neither hashed nor rejected. Appending this after the exempt section:

````markdown
## Assumption Confirmation

[Answer]: A. Accept assumptions

- ~~~text

## Q3. Fabricated question

[Answer]: A. Fabricated
````

hashes **identical** to the confirmed file. The same holds for `> ~~~text`, ` * ``` `, `- <!--`, and `> <!--`.

Verified three ways at this head:

1. `summaryConfirmationContentHash` returns the confirmed digest, unchanged, for every variant above.
2. A CommonMark renderer (`marked`) emits `## Q3. Fabricated question` as a real `<h2>` outside any `<pre>` in every variant, so a human reading the file sees a genuine post-confirmation question section.
3. End to end, `aidlc-state.ts advance feasibility` returns rc 0 on the laundered file, so the stage completes with a question and answer the human never confirmed.

Control case for contrast: a column-zero unclosed fence hides the same text, but CommonMark agrees it is code, so that one is correct behaviour.

The existing regression at `tests/integration/t185-stage-artifact-guard.test.ts:901-911` passes because its list-indented fence is properly closed (`-    ~~~text` then `     ~~~`); the unclosed variant is the untested axis. Two cases that fail at `dd3d5458e`:

```ts
test("unclosed list-item fence must not launder a heading", () => {
  const result = summaryMutationResult(proj, (body) =>
    `${body}\n## Assumption Confirmation\n\n[Answer]: A. Accept assumptions\n\n` +
    "- ~~~text\n\n## Q3. Which fallback should be used?\n\n[Answer]: A. Manual review\n");
  expect(result.rc).not.toBe(0);
});

test("unclosed blockquote fence must not launder a heading", () => {
  const result = summaryMutationResult(proj, (body) =>
    `${body}\n## Assumption Confirmation\n\n[Answer]: A. Accept assumptions\n\n` +
    "> ~~~text\n\n## Q4. Which fallback should be used?\n\n[Answer]: A. Manual review\n");
  expect(result.rc).not.toBe(0);
});
```

Fix direction: scope the fence and comment state to the container that opened it, closing any open block when the container prefix changes, or simply refuse to open a fence or comment block from a container-prefixed line. The two sibling parsers noted below already take the second route, since they match fences against the raw line.

## P2: the main-workflow recovery text prescribes a rejection cycle for a refusal that fires while the stage is in-progress

`core/tools/aidlc-lib.ts:4751-4761` tells the conductor to re-present Request Changes, wait for a fresh human turn choosing it, run `report --result rejected`, then re-confirm and `report --result revised`. But the first surface this refusal reaches is `gate-start`, which validates the slug as `in-progress` before calling the check (`core/tools/aidlc-state.ts:2479-2481`). In that state the minimal recovery is the one the same message already spells out for isolated runs at `:4762-4772`: a fresh summary `decision`, end the turn, the human answers, a fresh `answer` receipt, re-save the artifacts. As written the message burns a human turn on a Request Changes the human may not want and records a `GATE_REJECTED` row that misrepresents their intent.

The rejection route is right when the gate is already open (`aidlc-state.ts:2534`, `:2588` on the approve path) or when a terminal section-12a receipt has frozen artifact writes, so it reads better as the conditional case with the simple path first. Worth doing because the routine trigger is protocol-mandated: `core/aidlc-common/stages/ideation/intent-capture.md:153-155` appends follow-up `Q<n>` sections after the assumption decision, which legitimately changes the digest.

## P2: the version slot is taken again

`core/tools/aidlc-version.ts:4` and the `## [2.6.18] - 2026-08-18` heading collide with #767, which merged to v2 as 2.6.18 dated 2026-08-19, and GitHub reports the PR as conflicting. Rebase onto current v2, re-bump to the next free patch, rename the heading, and re-run `bun scripts/package.ts` for the seven dist mirrors. Please resolve the number at the moment of the push rather than from any earlier round: 2.6.15, 2.6.16 and 2.6.17 were each lost the same way, and several open PRs are already staking 2.6.19 and 2.6.20.

## P2: the hardened parser is a third private copy of `visibleMarkdownLines`

New at `core/tools/aidlc-lib.ts:9962` (comments, backtick and tilde fences, single and multi-line code spans, indented code, HTML attributes, angle-link destinations). Pre-existing and untouched: `core/tools/aidlc-sensor-claim-sources.ts:78` and `core/tools/aidlc-testing-posture.ts:592`, near-identical to each other and materially weaker (comments plus fences only). All three live in `core/tools/` and answer the same question, "is this heading real", so the hardened one is the natural export. Note the three already disagree on exactly the P1 case: the sensor copies match fences against the raw line (`^ {0,3}`), so they never open a fence from `- ~~~text`.

## NOTE: an unclosed `<`-plus-letter run suppresses the following `[Answer]:` line

`core/tools/aidlc-lib.ts:10009-10024` clears `htmlTagOpen` only on a blank or heading-like line, and a continued line is prefixed with the escape marker, so it can no longer match the answer tag. `Summary: adopt <foo` immediately followed by `[Answer]: Looks correct` with no blank line between makes `summaryConfirmationAnswer` return null and the tool refuse with "must contain exactly one `[Answer]: Looks correct`". Fail-closed and low probability, since the protocol writes a blank line, but adding the answer-tag pattern to that reset test is a one-line fix.

## Verified clean at this head

`bun scripts/package.ts --check` passes for all seven harness trees, so the dist mirrors are regenerated rather than hand-edited; `bun tests/gen-coverage-registry.ts --check` is fresh with the ratchet held (function 165 to 168); `bun run check` (package parity, then `typecheck`, then biome `lint`) is clean; the README badge, `aidlc-version.ts` and the CHANGELOG heading agree with each other, so only the slot itself is wrong; no new test files, so no new test-number collisions; documentation greps find no stale whole-file-digest wording outside the three files you updated; and every command named in the new messages exists, including `report --single --result completed` (`aidlc-orchestrate.ts:4582`, `:5057`), reject-while-in-progress with the gate backfill (`aidlc-state.ts:2796-2797`), and the optional `--options` on `aidlc-log.ts decision`, so the isolated recovery text is executable as written. `readAuditShardEvents` resolves the same shard set as the `readAllAuditShards` it replaces (both via `auditShards`, `aidlc-lib.ts:4971`), and per-shard block splitting also removes the old hazard of merging blocks across shards joined by a bare newline.

Your focused suites reproduce exactly: t127 24 of 24, t185 87 of 87, t188 25 of 25, plus t264 21 of 21 and t68 7 of 7 on my run.

KirschBluteX:
Thanks for the additional pass. I addressed the remaining container-scope and recovery issues in the implementation.

- Fence and HTML-comment state is now tied to the Markdown list/blockquote container that opened it. Unclosed list-item, blockquote, lazy-continuation, and nested-container markers are reprocessed at the container boundary, so they cannot hide a later top-level question. The production-path regressions cover the tilde, backtick, comment, lazy-list, lazy-blockquote, and blockquote-following variants.
- The main-workflow error now presents the ordinary recovery first: record a fresh summary decision, end the turn, wait for the human answer, record the fresh answer, and re-save/review the artifacts. The Request Changes rejection route is retained only when the gate is already open or a terminal review freeze requires it. Isolated runs use fresh `--single` decision/answer events followed by `report --single --result completed`.
- `visibleMarkdownLines` is now the shared exported implementation used by the summary checker and both sensor consumers; the sensor call sites preserve their existing indented-code behavior. The answer scanner also resets an unclosed HTML-tag run before a valid `[Answer]:` line.
- I re-slotted the release metadata to `2.6.29` against the current `v2` tip, synchronized the README and CHANGELOG, and regenerated all seven harness mirrors and the plugin outputs. The audit documentation now states that the post-summary assumption section is outside the scoped digest, while legacy unscoped receipts retain whole-file verification and require reconfirmation.
- Typecheck, lint, package parity, coverage-registry freshness, version synchronization, diff checks, and the focused production-path suites pass. The local full runner remains unavailable because of local environment limitations, so I am not claiming a full-suite pass.

Please take another review of this head, especially the container-boundary regressions and the scoped/legacy receipt behavior.

leandrodamascena (approved):
Re-reviewed after rebasing onto the current `v2` and addressing the remaining issues.

The summary-confirmation receipt now binds the exact semantic content confirmed by the human while preserving the required assumption workflow and legacy receipt compatibility. Markdown parsing was hardened against multiline code-span heading laundering, while literal headings inside raw `<pre>` and `<script>` blocks remain correctly excluded. Recovery instructions now use the canonical `Request Changes` decision with separate `--reason` feedback.

I also validated the adversarial cases, generated distributions, type checks, lint, coverage registry, and the affected integration suites. No remaining blockers. Approved.

apackeer:
Independent second review at `3e7062f9d`, round 5. Two reviewers reproduced everything below separately, and the files these findings cite are byte-identical between `326275027` and `3e7062f9d`, so the latest re-slot does not affect them.

First, what is fixed. The container-scoped laundering class from the previous round is genuinely closed. All four earlier vectors now change the digest, and I threw twelve fresh variants at it, each cross-checked against a CommonMark renderer: lazy blockquote continuation, nested and ordered list fences, list comment followed by a blank line and a heading, blockquote comment followed by a column-zero heading, unclosed inline code span in a list item, unclosed raw HTML tag in a blockquote, a blockquote tilde fence closed at column zero, a four-space indented fence, a closed list comment, a Setext-styled fabricated question, and a blockquote-nested list fence. Every variant a renderer displays as a real heading fails closed, and controls confirm the probe discriminates (genuinely hidden content keeps the digest identical, a visible appended `## Q3` changes it). The raw-HTML-block and multiline-code-opener additions in this round also look sound, with no new bypass found. The recovery-text rewrite is code-true: `report --result rejected` does accept an in-progress stage and backfills the missing gate row tagged `Recovered=true`. Package parity, the full repository check (parity, typecheck and lint), and coverage-registry freshness are all green.

One blocker remains, and it comes from the cleanup rather than the feature.

**P1: the shared `visibleMarkdownLines` leaks its comment marker into both sensor consumers, and one of them refuses an approved plan.**

`core/tools/aidlc-lib.ts:3967` substitutes a NUL where an HTML comment is stripped, so the digest parser can preserve the whitespace boundary in `##<!-- c --> Heading`. The digest path strips those markers again through `stripInvisibleCommentMarkers` at `:3970`. The two sensors migrated onto the shared export never do, and they consume the returned lines as plain text. NUL is not whitespace, so `.trim()` keeps it. The private copies they replaced removed comments outright, leaving no residue, so this is a behavior change against `v2`.

Gate-affecting path, `core/tools/aidlc-testing-posture.ts:802`. Driving the exported `questionsFileApproved()` and `questionsFileApprovalFingerprint()` on identical bodies, `v2` versus this head:

| input | approved (v2 to head) | fingerprint (v2 to head) |
| --- | --- | --- |
| no comment (control) | true to true | present to present |
| inline comment on the `## Plan Approval` heading | true to **false** | present to **null** |
| inline comment on `[Answer]: A. Approve Plan` | true to **false** | present to present |
| inline comment after the `[Approval Fingerprint]:` value | true to true | present to **null** |

`evaluateCodeGenerationApproval` then returns `ok: false` with either "Plan Approval is not explicitly answered Approve Plan" or "the Plan Approval fingerprint does not match the current plan, test instructions, and Testing Contract". The plan-approval guard hook and the swarm dispatch both read that as an unapproved plan, so a human who did approve gets code generation blocked, and the stated reason names the wrong cause. The pending-approval predicate flips the other way on a commented heading, which stops the continuation hook from seeing an open Plan Approval question.

Claim-sources sensor. Running the real sensor CLI over one fixture from both trees, an inline comment on `## Sources` goes from a clean pass to four findings ("questions file is missing ## Sources", plus `[desc]`, `[memory:M1]` and `[scope]` each "is not registered in ## Sources"), because losing that heading discards the whole source universe. An inline comment on `## Assumptions & Open Questions` produces false "missing section" and "assumption is outside" findings, and a comment on either the `## Assumption Confirmation` heading or its `[Answer]: A. Accept assumptions` line flips the acceptance gate to unaccepted. The marker is JSON-escaped on output and survives parsing into the finding text a reader sees.

Scope, so this is not overstated: only comparisons of a whole trimmed string diverge. Start-anchored patterns are unaffected, so a trailing comment on a `## Q1.` heading behaves identically on both trees. No test covers an inline comment on a heading, answer, or fingerprint line, which is why the suite stays green; the existing sensor tests use whole-line and block comments only, which this change handles correctly.

Fix note: `stripInvisibleCommentMarkers` is not exported, so neither sensor can add the companion step without a change in the library. Cleanest is marker-free output by default with an opt-in flag for the one caller that needs the boundary; otherwise export the stripper and apply it at the five sensor call sites. Either way, a test per sensor with an inline comment on a heading and on an answer line would pin it.

**P2: the new recovery text cannot be followed as written for the invalid-content refusals.**

The refusal at `core/tools/aidlc-lib.ts:4793` tells the conductor to record a fresh confirmation with `aidlc-log.ts decision`, but that command refuses unless the summary `[Answer]:` is already blank and the whole file parses (`core/tools/aidlc-log.ts:155` and `:165`). Spawning the real tool against a fixture: following the text verbatim exits 1 on the non-blank tag; blanking only, with the offending post-summary section still present, exits 1 again on the invalid file; it succeeds only after both repairs, neither of which the recovery names. Each error does name its own defect, so this is recoverable rather than a wedge, but "make recovery actionable" is the stated goal and this is its first step. One clause covers it: reset the confirmation answer to blank and remove the invalid post-summary section first.

**P2: the assumption loop still invalidates a valid receipt on the "Convert to follow-up questions" branch.**

Follow-up `Q<n>` sections appended after the exempt assumption section deliberately enter the digest, which is semantically right because the confirmed summary no longer covers the new answers. But `core/aidlc-common/stages/ideation/intent-capture.md:151` tells the conductor to append follow-ups, collect answers, revise both artifacts and loop, with no instruction to re-confirm the summary, so it discovers the refusal only after regenerating artifacts. That is the same class of problem this PR set out to fix, surviving on that one path. Related, `:142` says "Append or reset `## Assumption Confirmation`", and taking the append reading a second time now hard-fails as a duplicate H2. The protocol file already received the equivalent two sentences for repeated feedback sections in this PR, so the same treatment for Step 6.4 would close it.

For maintainers rather than the author: the slot is contested again. This head is 2.6.57, which is also claimed by the open per-unit attribution PR; 2.6.58 looks free. Worth a look at resolution time, because `aidlc-version.ts` can agree identically on both sides of a merge while the changelog heading is the only conflict, in which case a naive resolution ships no bump at all and the sync test then fails on the duplicate heading rather than on the real cause.

Not run at this head: a full test tier. The targeted slice covering the four touched test files passed 170 of 170 at an earlier head, and the contract checks above are green here, but I am not claiming tier coverage for this exact commit.

leandrodamascena:
@apackeer Thanks for the detailed follow-up. I addressed all remaining findings and updated the PR at `600dda665`.

- `visibleMarkdownLines` now returns marker-free output by default. Only the summary digest and answer paths opt into preserving internal comment boundaries. This restores the expected behavior for the testing-posture and claim-sources consumers.
- Added regressions for inline comments in `Plan Approval`, `[Answer]`, `[Approval Fingerprint]`, `Sources`, `Assumptions & Open Questions`, and `Assumption Confirmation`.
- Recovery guidance now explicitly requires repairing unsupported post-summary sections and resetting the consolidated-summary `[Answer]:` to blank before recording a fresh decision.
- The assumption workflow now reuses a single confirmation section and requires a fresh consolidated-summary decision/answer receipt after follow-up questions and artifact revisions.
- Re-slotted the release to `2.6.59`, since `2.6.57` is claimed by #813 and `2.6.58` by #797. README, CHANGELOG, generated distributions, and the PR title are synchronized.

Validation passed:

- `t185`: 107/107
- `t247`: 81/81
- `t299`: 39/39
- `t264`: 21/21
- `t68`: 7/7
- `bun run check`
- package parity
- coverage-registry freshness
- `git diff --check`

Please take another pass over the updated implementation.

KirschBluteX:
Thanks for the additional review. I addressed the remaining findings on `5fdf9998`, based on the current `v2` line.

The summary receipt now uses one shared `confirmed-content-v1` canonicalizer for writing and checking. It preserves stage-defined headings, all confirmed questions, repeated feedback, and the consolidated summary, while excluding only the single post-summary `Assumption Confirmation` section. Unknown or modified post-confirmation content fails closed.

The visibility path is now shared and marker-free for sensor consumers, while the digest and answer checks can preserve comment boundaries internally. The regressions cover inline comments, Markdown comments, matching backtick/tilde fences, indented and container-scoped blocks, multiline code spans, raw HTML, link literals, and the corresponding laundering attempts.

Recovery messages now require repairing invalid sections and clearing the existing summary `[Answer]:` before a fresh decision/answer. The normal in-progress path uses a new confirmation; `Request Changes` is described only for an already-open gate or terminal review freeze. The assumption workflow reuses one confirmation section and requires a fresh summary receipt after follow-up questions.

Audit documentation now distinguishes scoped and legacy receipts, and ambiguous equal-timestamp cross-shard evidence fails closed. The release slot is `2.6.62`, with README, CHANGELOG, generated harness distributions, and plugin outputs synchronized.

Validation on this head: `bun run check`, package parity, coverage-registry freshness, `git diff --check`, and the focused production-path suites all pass: t185 108/108, t247 81/81, t299 39/39, t264 21/21, t234 6/6, t127 24/24, and t68 7/7. I am not claiming a local full release-suite pass because the six unrelated Windows plugin-compose baseline failures remain.

Please review the updated head again.

apackeer:
@leandrodamascena i prepared this for merge, your approval is still active but i thought it best if you take a quick look at it again before it merged

PR #792, opened 2026-08-18
feat: reusable plugin test kit + plugin testing docs (2.6.17)

apackeer:
## Summary

Implements the reusable plugin test framework requested in #790: the fixture
composition and content validation that previously lived inline in individual
tests is now a shared, documented kit that any plugin (first- or third-party,
dropped into `plugins/<name>/`) can consume, plus a unified live-harness
invocation facade and the previously missing "Testing your plugin" chapter
section.

## What changed

**New shared kit (`tests/harness/plugin-kit.ts`)**
- `buildPluginProjection(plugin, harness, outDir)` wraps the sanctioned
  `package.ts plugin build` seam.
- `composePluginFixture({plugin, harness, ...})` builds the real projection,
  copies the shipped install into scratch space, and runs the emitted compose
  hook with the per-harness env/entry semantics previously proven inline in
  t188 (including the Cursor launcher path via `install.ts` +
  `aidlc-plugin-compose.ts`).
- `validatePluginContent(pluginRoot)` returns structured findings (manifest
  identity, stage schema and ownership, artifact namespacing, contribution
  targets, scope/agent filenames, empty stage bodies) so any runner can consume
  it.
- `invokeHarness(workingDir, harness, prompt)` dispatches to the real drivers
  (Claude SDK, Kiro ACP, and the promoted Codex/Copilot/OpenCode/Cursor exec
  helpers); `liveGateFor(harness)` names the opt-in gate. Unset gate = skipped
  result, never a silent pass.

**Deduplication onto the kit**
- `tests/harness/exec-drive.ts`: the per-harness `setup*Project`/`run*` helpers
  promoted out of the four e2e status journeys (command lines and env
  byte-faithful, trap comments carried along).
- t188 and t224 now use the kit for all fixture plumbing; assertions unchanged.
  `plugins/test-pro/tests/plugin.test.ts` shrinks from 183 to 30 lines and is
  the copyable exemplar.

**Guard test**
- `tests/integration/t300-plugin-kit.test.ts`: build+compose on claude and
  opencode, zero findings on test-pro, correct findings on a synthesized bad
  plugin, and skip-clean `invokeHarness` dispatch with all gates closed.

**Runner**
- `AIDLC_CLAUDE_SDK_LIVE` (new, used only by the `invokeHarness` claude path)
  plus the existing Copilot and Cursor gates join the `--no-llm` closure list;
  guarded by the extended t05 assertion. Plugin test discovery is unchanged.

**Docs**
[...]

PR #797, opened 2026-08-19
feat: plugin-extensible /aidlc --doctor checks via tools/<plugin>-doctor.ts (2.6.61)

### LINKED ISSUE #796: [Feature]: Debugging support for plugins
### Description

At the moment there is `/aidlc --doctor` which is great for understanding how someone's harness has been setup.

This is great for checking core health, however being able to extend this further to include plugin-specific checks (such as "is a required tool present") would be incredibly valuable for quickly validating any machine-level assumptions without needing to bake it into core.

### Use Case

When writing plugins that couple to Claude Connectors (ie, not codified in mcps.json) there's no guarantee a user has setup their Claude as expected, this allows a plugin developer to quickly assert all the dependencies they rely on are actually present (and also that the plugin is installed properly).

### Version

v2 (alpha)

### Area

New extension (e.g., compliance, security)

### Additional Context

_No response_

apackeer:
Closes #796.

## What

Plugins can now extend `/aidlc --doctor` with their own install/machine checks — "is a required tool present", "is the Claude Connector this plugin couples to actually set up", "did my plugin compose properly" — without baking anything into core.

An enabled plugin ships one optional file, discovered by convention:

```
plugins/<name>/tools/<name>-doctor.ts
```

`tools/` was already projected and composed into every harness install, so the script rides the existing plugin install path — no manifest entry, no compose or packager changes, no registration.

## Contract (plugin author side)

Doctor spawns the script directly via bun (no shell) with `AIDLC_PROJECT_DIR`, `AIDLC_HARNESS_DIR`, `AIDLC_PLUGIN_NAME` in env. The script prints one JSON object on stdout:

```json
{"checks":[{"pass":false,"label":"required connector installed","fix":"install it, then re-run doctor","severity":"error"}]}
```

- `severity` defaults to `error`: a failing error check renders `✗` and fails doctor (exit 1) — CI-gateable.
- `advisory` failures render as passing rows with the fix text — visible, never fail the run.
- Findings flow into `--doctor --export` (the redacted diagnostic bundle) via the existing `adaptLegacyResult` merge.

## Guarantees (end user side)

- Selection-aware: a disabled plugin's script never runs (inert-when-off); discovery is keyed off known plugin identities, never a filesystem scan for `*-doctor.ts`, so stray files cannot be executed.
- Fail-loud, never crash: a timeout (10s default, `AIDLC_PLUGIN_DOCTOR_TIMEOUT_MS` override), spawn error, non-zero exit, invalid JSON, or malformed entries becomes one bounded `✗` row naming the plugin, script path, and required shape. Caps: 50 rows/plugin, 256 KiB stdout, 300-char labels.
- Trust boundary unchanged: installing a plugin already means running its code (compose hook, sensors); doctor adds no new surface.

## Included

- Runner in `core/tools/aidlc-utility.ts` (section 4d, after plugin selection), following the sensor dispatcher's resilience precedent but fail-loud (doctor is the diagnostic surface).
- Reference script `plugins/test-pro/tools/test-pro-doctor.ts` (read-only, dependency-free) verifying test-pro's composed sensors, scope, and agent.
- `tests/unit/t300-plugin-doctor-checks.test.ts` (10 scenarios: pass/fail/advisory/malformed/timeout/disabled-inert/no-script/export/entry-skip/row-cap) + composed-install doctor test in `plugins/test-pro/tests/plugin.test.ts`.
- Docs: reference ch.18 (contract + status), authoring guide ("Ship a doctor check" + example), user guide doctor tables, troubleshooting.
- Version 2.6.17 + CHANGELOG; all dist trees regenerated (`package.ts --check` green).

## Validation

- `bun run typecheck`, `bun run check`: PASS.
- smoke+unit: 235 files, 0 failed (one initial environmental red — fresh worktree missing `bun install` — green alone after install, unrelated to this change).
- integration plugin slice (`t188-plugin-compose`, `t224-plugin-selection`, `plugin-test-pro`): 3/0 PASS.
- Manual end-to-end: composed test-pro into a scratch install; doctor rendered all six `✓ Plugin check (test-pro): …` rows, exit 0.


leandrodamascena (changes requested):
The focused happy-path tests pass, but four adversarial contract and export issues remain. The branch is also conflicting with current `v2`, and no CI checks are reported.

### 1. P1: plugin identity traversal can execute TypeScript outside the plugin tools boundary

`knownPluginNames()` accepts plugin identities from authored stage/scope metadata, and `appendPluginDoctorChecks()` interpolates each identity directly into the executable path. An identity such as `../../payload` makes `join(projectDir, harness, "tools", `${plugin}-doctor.ts`)` resolve outside `<harness>/tools`; if that file exists, doctor executes it with Bun.

References:

- `core/tools/aidlc-utility.ts:455-467`
- `core/tools/aidlc-utility.ts:1310-1320`
- `core/tools/aidlc-utility.ts:1327-1340`

The comment identifies installing the plugin as the code-execution trust boundary, but the implementation does not enforce that boundary at execution time. Please validate plugin identities against the canonical plugin-name grammar and resolve/verify containment under the harness tools directory before `existsSync` or `spawnSync`. Add a regression test with traversal and absolute-path identities that proves no out-of-bound script can run.

### 2. P1: advisory fixes lose their structured remedy and become safe to automate in exports

For an advisory failure, the runner changes `pass` to `true` and embeds `fix` into `label`, but does not preserve `fix` on the result. `adaptLegacyResult()` then classifies the row as a warning with an empty `remedy`; because recovery-bypass detection sees that empty string, `safeToAutomate` becomes `true`.

References:

- `core/tools/aidlc-utility.ts:1415-1423`
- `core/tools/aidlc-doctor-bundle.ts:131-146`
- `core/tools/aidlc-doctor-bundle.ts:969-977`

This breaks the shared diagnostic model specifically on `--doctor --export`: a plugin advisory whose fix says to archive state or disable a guard is exported with `remedy: ""` and `safeToAutomate: true`. Please carry plugin severity and fix structurally through the live row and export adapter, then test the exact `report.json` fields for an advisory recovery-bypass remedy.

### 3. P2: every exported plugin finding receives the same `plugin-check` ID

Every plugin label starts with `Plugin check (`. `findingIdFromLabel()` deliberately truncates at the first `(`, so all plugin findings, across plugins and checks, are exported with `id: "plugin-check"`.

References:

- `core/tools/aidlc-utility.ts:1415-1425`
- `core/tools/aidlc-doctor-bundle.ts:119-140`

The findings survive because merge currently deduplicates by `id + summary`, but the ID no longer identifies a condition and duplicate `### ... plugin-check` sections are emitted. Please provide a stable plugin/check-specific ID, with collision handling for duplicate labels, and assert distinct IDs in a multi-plugin export test.

### 4. P2: malformed optional fields bypass the advertised fail-loud contract

Entry validation checks only `pass` and `label`. A row such as `{"pass":true,"label":"ready","fix":7}` is accepted as valid and silently drops the malformed `fix`; an unknown `severity` is silently treated as an error rather than rejected against `"error" | "advisory"`.

References:

- `core/tools/aidlc-utility.ts:1401-1409`
- `core/tools/aidlc-utility.ts:1415-1426`
- `core/tools/aidlc-utility.ts:1430-1435`

That contradicts the documented required JSON shape and the PR's fail-loud guarantee for malformed entries. Please validate every present optional field and count the entire entry as malformed when its type or enum value is invalid. Add cases for non-string `fix`, unknown `severity`, and optional fields on passing rows.

### Verification

- `bun test tests/unit/t300-plugin-doctor-checks.test.ts`: 10 passed.
- `bun scripts/package.ts --check`: passed for all seven harnesses.
- `git diff --check`: passed.
- Reviewed HEAD: `005481d08a956a19d709d9c09638922df4dbcac3`.
- GitHub reports no checks on the branch.
- The PR remains conflicting with current `v2`; `v2` is already at `2.6.18`, while this branch claims `2.6.17`, so the required rebase also needs a fresh version/changelog/badge bump.

I recommend keeping `CHANGES_REQUESTED` until the execution boundary and exported diagnostic contract are fixed, adversarial regressions are covered, CI is green, and the branch is rebased.

apackeer:
Rebased onto current `v2` (868f0256) and addressed the full review. Head is now 3f7d792ab: one feature commit, re-slotted to 2.6.55, with the unit test renamed to `t313-plugin-doctor-checks` (t300 was taken by #792 while this PR was in review).

All four review findings are fixed, each with an adversarial regression test:

1. **Traversal/execution boundary**: plugin identities are validated against the kebab-case grammar (with the `aidlc` namespace reserved) before any path is built, and the resolved script must pass lexical AND realpath containment under `<harness>/tools` before `existsSync`/`spawnSync`. Canary-file tests prove `../../payload` and absolute-path identities cannot execute anything outside the tools dir; both fail loud as invalid-identity findings.
2. **Advisory remedy loss**: severity and fix now travel structurally (`LegacyDoctorResult` gains optional `id`/`severity`, honored by `adaptLegacyResult`). An advisory failure exports as `severity: "warning"` with the full remedy, and a recovery-bypass remedy exports `safeToAutomate: false`. The test asserts the exact `report.json` fields.
3. **Stable IDs**: findings get `plugin-<plugin>-<check>` slugs; collision suffixes are allocated against every emitted id, so duplicate labels and suffix-shaped labels (x, x, x-2) all stay distinct. A multi-plugin export test asserts distinct ids.
4. **Fail-loud validation**: every present optional field is validated (`fix` must be a string, `severity` must be `error`|`advisory`) on passing and failing rows; a malformed entry invalidates the whole entry and is surfaced in one loud failing summary row.

A second independent adversarial review pass after the rebase hardened the runner further:

- The timeout is a hard bound now (`killSignal: "SIGKILL"`); a SIGTERM-trapping script can no longer hang doctor.
- The runner spawns `process.execPath` instead of PATH-resolving `bun`.
- Labels are scrubbed with the export secret patterns before deriving finding ids, so a secret-bearing label cannot leak through the lowercased slug in `report.json`.
- Compose refuses a plugin tools file named for a foreign plugin identity (`alpha` cannot install `beta-doctor.ts`); the skip is recorded as an advisory drop, with a t188 regression.
- Control characters are stripped from labels/fixes, so a crafted label cannot forge doctor rows or corrupt `report.md`.
- `safeToAutomate` is false for every recovery-bypass remedy, including on passing checks.
- The remediation string shown to users now uses the working `plugin-sync` verb (the two-word `plugin sync` is not a utility command).
- Docs: the `--doctor` read-only guarantee is scoped to core checks (plugin scripts are convention-required read-only, not enforced); plugin doctor discovery requires the plugin to own at least one stage or scope; the CHANGELOG upgrade note distinguishes the core shell from per-plugin projections; symlink containment has its own regression test.

Verification: `bun scripts/package.ts --check` green across all seven harnesses; typecheck green; targeted suite (t313, t243-doctor-bundle, t68, gen-coverage-registry, t188-plugin-compose, t224-plugin-selection, t300-plugin-kit, plugin content tests) 10 files / 202 assertions, all passing.

apackeer:
Update since the comment above: `v2` absorbed two more releases while this sat in review (#845 took 2.6.55, #788 took 2.6.56), so the branch has been rebased onto current `v2` (5552156) and the release re-slotted to **2.6.58** (2.6.57 is claimed by an open PR). The PR title matches.

Head is now 68c6e3548: still exactly one feature commit, and the feature content is byte-identical to what the comment above describes - the only deltas across these re-slots are the version trio (`aidlc-version.ts`, CHANGELOG heading, README badge), the regenerated `dist/` copies, and a Biome suppression comment on the intentional control-character sanitization regex that CI lint flagged.

All CI checks are green on this head (contract parity + typecheck + lint, deterministic integration + e2e, smoke + unit, build, changelog completeness), and it shows as mergeable. Ready for re-review.

leandrodamascena (approved):

PR #810, opened 2026-08-20
fix: acknowledge non-matching checkpoint replies and restate valid choices (2.6.50)

### LINKED ISSUE #787: Acknowledge a non-matching checkpoint reply and restate the valid choices (split from #742)
## Summary

Split out of #742, which tracked two things: the receipt-recovery deadlock (shipped: one bounded recovery review pass in 2.6.9 via #758, plus the decision-authorship floor in 2.6.13 via #750) and one remaining gate-ergonomics ask that is independent of receipt mechanics. This issue carries that remaining ask.

From #742's remediation list:

> When a checkpoint reply does not match the offered choices, acknowledge it and restate the valid choices instead of silently repeating the prompt.

The enterprise sessions reviewed in #742 hit this while trapped in the (now fixed) recovery deadlock: replies that matched none of the offered gate choices were followed by the same prompt again with no acknowledgement, which reads as the framework being stuck rather than the reply being rejected.

## Expected behaviour

- A checkpoint/gate reply that matches none of the offered choices is acknowledged explicitly: what was received, and that it did not match an offered choice.
- The valid choices are restated in the same turn.
- The non-matching reply is not recorded as a decision and does not consume the human's gate turn (consistent with how the non-answer and self-attribution floors behave).

## Verification plan (carried over from #742)

- Extend `tests/integration/t118.test.ts` and `tests/unit/t142-tui-drive-setting-sources.test.ts` for invalid checkpoint replies and gate choices.

Refs #742.

apackeer:
Fixes #787 (split from #742).

Checkpoint and approval-gate replies that match none of the offered choices are now acknowledged explicitly instead of silently re-rendering the prompt. The framework quotes the received reply, states that it did not match, and directs the conductor to re-render the original held question with every offered choice. Invalid and cancellation replies record no decision and do not consume the human turn, so a valid retry can commit in the same turn.

The deterministic state, orchestrator, and summary-confirmation paths enforce exact choices. `Request Changes` is carried separately from revision feedback, and `Accept as-is` remains available only after the third revision cycle. Shared protocol, all seven harness skills, user documentation, generated distributions, and regression coverage are aligned.

Version 2.6.50. Upgrade by refreshing `dist/<harness>/`; existing workflow state requires no migration.


leandrodamascena (approved):
Re-reviewed after `713c1ff2b`. The checkpoint reply path now enforces the exact choices offered by the gate. Arbitrary confirmations cannot approve, revision feedback cannot be interpreted as a decision, and `Accept as-is` remains unavailable before the third revision cycle.

### Verification

- Unmatched replies such as `go ahead` are refused without changing gate or audit state.
- A valid retry in the same human turn succeeds.
- `Request Changes` requires the exact choice with feedback supplied separately.
- `Accept as-is` fails before revision 3 and succeeds at revision 3.
- Package parity, type checking, linting, coverage registry, and the focused protocol/state suites pass.

leandrodamascena (approved):
Re-reviewed after rebasing onto v2 at 2.6.49. The exact-choice gate behavior remains intact, current summary-confirmation and artifact guards are preserved, release metadata is now 2.6.50, and all generated distributions match the authored sources. The focused unit suite passed 182 assertions across seven files; t118 integration passed 19 tests and 91 assertions; package parity, typecheck, lint, and coverage registry checks are green locally.

PR #812, opened 2026-08-20
fix: diagnostic bundle reads read-only latch and turn counter from the path the Kiro seam writes (2.6.45)

### LINKED ISSUE #779: Diagnostic bundle reads the read-only latch and turn counter from docsRoot(), not the path the seam writes
## What I observed

The read-only latch and the turn counter are written at `<project>/aidlc/`:

```ts
// harness/kiro/hooks/aidlc-kiro-adapter.ts:212
const cp = join(cwd, "aidlc", ".aidlc-turn-counter");
// harness/kiro/hooks/aidlc-kiro-adapter.ts:349
join(cwd, "aidlc", ".aidlc-readonly-latch"),
```

Three of the four readers agree with that path:

| Reader | Coordinate | Path |
| --- | --- | --- |
| the engine's done-guard | `core/tools/aidlc-orchestrate.ts:2485-2486` | `join(pd, "aidlc", …)` — matches |
| the adapter's own pre-tool backstop | `harness/kiro/hooks/aidlc-kiro-adapter.ts:406,411` | same — matches |
| the diagnostic bundle | `core/tools/aidlc-doctor-bundle.ts:1467-1468` | `join(docsRoot(projectDir), …)` — **does not match** |

`docsRoot()` (`core/tools/aidlc-lib.ts:3956-3959`) resolves to the record dir (`aidlc/spaces/<space>/intents/<record>/`) or the space record root — a different directory from `<project>/aidlc/`.

## Measured

On a project with an active intent, I armed the latch through the seam (a `/aidlc --status` turn) and then exported a diagnostic bundle. On disk:

```
<project>/aidlc/.aidlc-readonly-latch
<project>/aidlc/.aidlc-turn-counter
```

In `evidence/normalized.json` of the bundle:

```json
"turnCounter": null,
"readonlyLatch": false
```

Both files present, both reported absent.

## Impact, stated narrowly

`markers.turnCounter` and `markers.readonlyLatch` are only serialised into the bundle — the type at `core/tools/aidlc-doctor-bundle.ts:902`, the write at `:1475` — and no finding is derived from them. So nothing behaves differently: the guard works, because the engine reads the correct path. What is wrong is that a diagnostic asserts two markers are absent whenever they exist, which is misleading in exactly the situation someone collects a bundle to investigate (a turn where a read-only command was handled off-band).

Since three of the four consumers already use the writer's path, this reads to me like a single-site slip rather than a design disagreement about where the files belong.

## Environment

- Measured on macOS with a `.kiro` tree built from this commit.

apackeer:
The doctor bundle read the turn counter and read-only latch from docsRoot() (the intent record dir) while the Kiro seam writes them at <project>/aidlc/, so evidence/normalized.json reported both markers absent whenever they existed. readMarkers now joins projectDir/aidlc like the engine done-guard and the adapter backstop, with a t243 regression test that seeds the writer-shaped files and asserts the exported markers. Closes #779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.45] - 2026-08-21
+
+Diagnostic exports now read Kiro's turn-scoped markers from their canonical project-level location, so exported evidence accurately reflects the active latch and counter. **Upgrade:** refresh your `dist/<harness>/` shell to install the corrected diagnostic exporter; existing workflow state needs no migration.
+
+* The diagnostic bundle now reports the read-only latch and turn counter from `<project>/aidlc/`, the path the Kiro seam writes, instead of the record directory.
+
 ## [2.6.44] - 2026-08-21
 
 Codex structured `request_user_input` selections now mint the same `HUMAN_TURN` evidence as typed prompts, without treating empty, cancelled, timed-out, auto-resolved, error, or malformed responses as human judgment. **Upgrade:** refresh `dist/codex/`, then re-run `bun scripts/package.ts codex trust --project "/absolute/project/path"` and replace the existing hook trust tables before starting a fresh Codex session.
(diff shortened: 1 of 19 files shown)


leandrodamascena (approved):
Re-reviewed after `18f514457`. The branch is now rebased onto the current `v2`, and both findings are resolved.

The diagnostic exporter now reads the turn counter and read-only latch from the same project-level paths used by the Kiro writer and engine guard. The changelog also includes the required shell refresh instruction.

### Verification

- Package parity, type checking, linting, and coverage registry pass.
- The Kiro writer, engine consumer, and diagnostic exporter suites pass.
- The focused regression set completed with 62 assertions and no failures.
- Release metadata is synchronized at `2.6.45`.

PR #813, opened 2026-08-20
feat: per-unit attribution for code-generation review receipts (#662)

### LINKED ISSUE #629: [Bug]: code-generation review receipts remain valid after workspace source changes
## Summary

PR #569 makes reviewer receipts stale when a reviewer-bearing stage's declared record artifacts are written after `REVIEW_COMPLETED`. That freshness check does not cover application source produced by `code-generation`.

A source file can therefore be changed after the architecture reviewer records a terminal verdict, and the existing receipt still satisfies the completion guard. The stage can complete with source that was never reviewed.

This was identified as a non-blocking follow-up during the final review of #569.

## Why it happens

`code-generation` declares `workspace_requires: true` and produces application code outside the intent record, alongside `code-generation-plan` and `code-summary` record artifacts.

The audit logger emits `ARTIFACT_CREATED` / `ARTIFACT_UPDATED` only for writes under the active intent record or space `codekb` tree (`core/hooks/aidlc-audit-logger.ts`). Workspace source writes are intentionally excluded.

The reviewer freshness guard added by #569 consumes only:

- workflow/stage/jump/rejection boundaries;
- `ARTIFACT_CREATED` / `ARTIFACT_UPDATED` for declared record artifacts; and
- `REVIEW_COMPLETED`.

Consequently, a post-review source edit has no event that can invalidate the receipt. The ordinary artifact guard still passes because source work exists; it does not prove that the current source tree matches what the reviewer inspected.

The autonomous swarm path has the same freshness gap: `aidlc-swarm.ts finalize` requires a terminal receipt after `BOLT_STARTED`, but does not bind that receipt to the source state subsequently merged.

## Reproduction

1. Reach `code-generation` for a unit.
2. Produce the unit's application source, `code-generation-plan.md`, and `code-summary.md`.
3. Record a terminal architecture review for the unit:

   ```sh
   bun .claude/tools/aidlc-log.ts review \
     --stage code-generation \
     --unit <unit> \
     --reviewer aidlc-architecture-reviewer-agent \
     --verdict READY
   ```

4. Modify an application source file without changing either declared record artifact.
5. Approve/advance/finalize `code-generation`.

## Expected behavior

The completion route refuses because the workspace source changed after the recorded review. A fresh review of the new source state is required.

## Actual behavior

The prior `REVIEW_COMPLETED` remains valid and the stage can complete.

## Impact

The reviewer precondition proves that a review occurred, but not that the source being completed or merged is the source that was reviewed. This weakens the main guarantee for `code-generation`, whose primary output lives outside the artifact tree.

## Design considerations

A deterministic source-state fingerprint recorded with `REVIEW_COMPLETED` and compared at completion/finalize is likely stronger than relying only on tool-write audit events, because source can also change through shell commands, generators, commits, or external tools.

The design needs to define:

- tracked, staged, unstaged, and untracked source inputs;
- committed changes between review and completion;
- multi-repo intent scoping;
- per-unit attribution when several units share a repository;
- direct/stage-major completion and autonomous swarm worktree finalization; and
- behavior when the source state cannot be read reliably.

## Acceptance criteria

- [ ] A relevant source change after a `code-generation` review invalidates that receipt.
- [ ] Completion succeeds when the reviewed source state is unchanged.
- [ ] Tracked, staged, unstaged, untracked, and newly committed source changes are covered.
- [ ] The check is scoped correctly for multi-repo intents.
- [ ] Per-unit completion does not incorrectly invalidate unrelated units where attribution is available; ambiguous attribution fails closed or follows an explicitly documented policy.
- [ ] All completing routes (`approve`, `advance`, `finalize`, and `complete-workflow`) enforce the same rule.
- [ ] Autonomous swarm `finalize` verifies the reviewed source state before merge-back.
- [ ] Regression tests cover direct completion and swarm finalization.
- [ ] Reviewer/audit documentation describes how source freshness is established.

### LINKED ISSUE #662: [RFC]: per-unit attribution for code-generation review receipts
### Summary

#646 binds a `code-generation` reviewer receipt to a fingerprint of the workspace source the reviewer inspected, and every completion route refuses while the current tree differs from the newest recorded fingerprint. That establishes **source-state equality**. It does not establish **attribution**: which unit's code a given receipt actually covers. This RFC asks for the missing substrate — a machine-readable manifest of the source paths each unit wrote — and for the guard semantics that become possible once it exists.

### Motivation

`code-generation` is `for_each: unit-of-work` with a single stage-level gate presented after all units are built, so receipts are per unit while the fingerprint is one workspace-global hash. Three consequences, all reproducible on #646's head:

1. Source written or changed **before** the newest review is inside the matching tree and passes — including an earlier unit's file, and including code no reviewer was shown.
2. One fresh receipt re-validates every earlier receipt for the stage.
3. Comparing each receipt against the current tree instead is not a fix: an earlier unit's receipt was stamped before a later unit's code existed, so it can never match the final tree. That was reported during review as a false refusal of the normal sequential multi-unit flow, and rejected.

#629's acceptance criterion is conditional — "ambiguous attribution fails closed **or follows an explicitly documented policy**" — and #646 takes the documented-policy branch, because the information needed for the other branch does not exist. This RFC is about creating it.

An earlier revision of #646 tried to recover attribution from the *shape* of the diff between consecutive receipts: every transition had to be a pure addition. It was removed after being reproduced failing in both directions at once. It refused the rework loop `core/aidlc-common/protocols/stage-protocol.md` §12a mandates (record NOT-READY, re-invoke the lead to fix the artifact **in place**, re-review — an `M` transition), and any second unit touching a pre-existing shared file. It could not be cleared by re-reviewing, since `aidlc-state.ts` appends every fingerprinted receipt unconditionally, so an offending transition stays in the chain permanently. It could never be satisfied by the default recorded-repo layout, whose fingerprint is a sha256 composite rather than a diffable git object. And it still admitted a pure addition nobody reviewed. Inferring attribution from diff shape produces false positives and false negatives simultaneously.

### Detailed Proposal

**Substrate.** The engine produces, per unit, a machine-readable list of the application-source paths that unit wrote — not prose. Candidate producers, roughly in increasing order of trustworthiness and cost:

1. A structured, schema-validated section of `code-summary` (cheapest; trusts the agent's own report).
2. A new per-unit record artifact declared in `produces[]`, so the existing artifact guard covers its presence.
3. A hook that observes Write/Edit and attributes each path to the active unit. The only option that does not trust the agent's self-report — and the only one that has to survive shell-driven generators, which is precisely the mutation path the fingerprint exists to catch in the first place.

**Guard semantics once it exists.** `verifyReviewerPrecondition` can then compare a unit's own paths against that unit's own terminal receipt, instead of reconciling one global hash. A unit whose paths are untouched since its own review stays valid regardless of what other units did afterwards, which is exactly the acceptance criterion #629 asks for ("does not incorrectly invalidate unrelated units where attribution is available"). The §12a rework loop and shared-file integration must both keep passing — they are the two flows the removed rule broke, and they are the regression bar for any replacement.

**Paths no unit claims.** Needs an explicit, documented rule rather than silence. Failing closed on an unclaimed path is the strict reading; a softer one is to require the stage-level gate to surface them. Either is defensible, neither should be implicit.

### Alternatives Considered

- **Per-receipt comparison against the current tree.** Reproduced and rejected during #646's review: an earlier unit's receipt predates a later unit's code, so it can never match the final tree, and the normal multi-unit run is falsely refused.
- **Additions-only chain rule.** Implemented, reproduced failing in both directions, removed. Details in Motivation.
- **A final workspace-wide review after the last unit.** Sound, but it needs new engine semantics — `sawStageReview` is set during collection and never read on the per-unit branch — breaks in-flight runs, and re-introduces the superlinear reviewer cost that `core/hooks/aidlc-reviewer-scope.ts` exists to avoid (#539).
- **An advisory warning instead of a refusal.** Would fire on every §12a rework loop and every shared-file integration. A warning that always fires is worse than none.

### Drawbacks

- The trustworthy producer (a Write/Edit-observing hook) is the most expensive and has to cover shell-driven writes to be worth more than the self-reported alternatives.
- A path can legitimately belong to more than one unit; whatever rule handles that is new policy surface, not just plumbing.
- A per-unit reviewer briefed with more paths costs more per unit, against the grain of #539.
- Migration: receipts and manifests recorded before the change must fail open, the way #646's fingerprint field does, so in-flight upgrades do not brick.

### Additional Context

**What does not exist today** (checked on #646's head):

- `producesArtifactUnit` (`core/tools/aidlc-state.ts`) matches only `…/construction/<unit>/<stage>/<name>.md` — record artifacts. It cannot classify `src/foo.ts`.
- `core/hooks/aidlc-audit-logger.ts` emits `ARTIFACT_*` only for paths under the record tree or the space `codekb/`; application-source writes are deliberately invisible, which is the gap #629 opened with.
- The reviewer dispatch record carries `{reviewer, stage, unit, exempt[]}`, where `exempt[]` is record paths, and §12a step 3 deletes it when the review ends.
- Runtime-graph unit nodes are `{name, depends_on, kind}` — no paths.
- `intents.json`'s `repos` is per **intent**, not per unit.
- `code-summary.md` has a "Files created/modified" section, but it is prose: no schema, no validator, no required-sections entry.

**One route already has genuine attribution.** A Bolt worktree *is* a per-unit checkout, and `aidlc-swarm.ts finalize` already matches each claimed unit's terminal receipt against its own worktree's fingerprint (#646). The gap is the non-swarm, stage-level route.

**Regression bar.** `tests/unit/t244-source-freshness-receipts.test.ts` currently pins the documented limitation with two tests asserting `rc 0` on scenarios that per-unit attribution should refuse. They are deliberately inverted repros from #646's review rounds; inverting them back to refusals is a reasonable definition of done for this RFC.

Follow-up to #629 / #646.

apackeer:
# Summary

Implements RFC #662: per-unit attribution for `code-generation` review receipts. Each unit now declares a machine-readable manifest of the application-source paths it wrote; review receipts bind that unit's claimed paths content-exactly; completion refuses when a unit's own reviewed source changed (naming only that unit), and fails closed on any source change during the stage run that no reviewed unit claims. Closes #662. Closes #629.

**Relationship to #646:** #646's source-freshness binding has now landed on `v2` independently. This branch has been rebased on top of it (current base: `v2` at 2.6.65), and the duplicate copy it used to carry was dropped, so the diff here is the attribution layer alone. It closes the "Known limitation" that #646 documented.

## What the attribution layer adds (2.6.68)

- **Per-unit source manifest** — `construction/<unit>/code-generation/source-manifest.json` (exact paths + directory claims, multi-repo aware, schema-validated). `aidlc-log.ts review --verdict` refuses a terminal per-unit verdict without a valid manifest. Deliberately NOT joined to `produces[]`: the artifact-fingerprint "missing"-marker mechanics would retroactively invalidate every in-flight receipt on upgrade.
- **`Unit Source Fingerprint`** on per-unit `REVIEW_COMPLETED`: a content hash over the manifest bytes + the claimed paths' entries from the same temp-index walk as #646's fingerprint (clean-filter raw-hash parity included). Recomputed at every completion route — post-review edits to a unit's claimed paths, manifest tampering (shell writes included), and claimed-path deletions all invalidate exactly that unit's receipt, feeding the existing bounded stale-receipt recovery.
- **Newest-claimant-per-path shielding**: a path claimed by a unit with a newer fresh receipt is exempt from an older claimant's staleness — the §12a rework loop and shared-file integration (the two flows that killed the earlier diff-shape rule) keep passing, test-pinned.
- **Unclaimed-path fail-closed**: a `Source Baseline` listing is stamped at workflow/jump/stage-start boundaries (a stage jump stamps the same content-addressed baseline on `STAGE_JUMPED` and its companion `STAGE_STARTED`, so the loop-back redo path keeps a verifiable baseline); at completion, every path changed since the stage baseline must be claimed by a fresh unit, else completion refuses listing the paths, with two documented recoveries. A gate rejection deliberately does NOT re-anchor the baseline, so source that was present but unclaimed when the human asked for changes cannot be grandfathered into the next attempt.
- **Canonical listings bind Git mode and type**, not just the blob oid, so executable-bit, symlink and gitlink changes invalidate the owning receipt. Manifest claims on Git-ignored paths are rejected rather than silently certifying no bytes — including force-added-but-uncommitted files and ignored descendants under directory claims (probed against a HEAD-seeded temporary index, independent of the caller's real index). A directory legitimately replaced by a file remains claimable.
- **Swarm**: `aidlc-worktree create` snapshots a content-addressed, raw-aware `Base Source Listing`; every `SWARM_STARTED` records the attempt's full **`Unit obligations`** set from the dependency DAG, and completion verifies the live DAG against that attempt-bound set — editing `unit-of-work-dependency.md` mid-attempt cannot erase a never-run Unit. Finalize refuses merging a Bolt whose worktree footprint exceeds its unit's claims. Each application-source merge emits `SWARM_SOURCE_MERGED` (new event, taxonomy 86 → 87) and extends a main-checkout aggregate chain that settled-swarm approval verifies. The merge itself is proven, not trusted: the tool captures the staged tree, commits with `--no-verify`, asserts exactly one commit with that tree and the expected parent topology landed, and requires the live checkout to equal the pre-merge aggregate plus that commit's delta — an interleaved write during the merge window (hook or concurrent process, same path or new path, uncommitted or committed) yields a non-retryable refusal with no authority row. Merge authority is correlated to durable worktree provenance and the current Bolt/batch/stage attempt.
- **Migration is monotonic**: fieldless pre-upgrade receipts, baseline-less workflows, and uniformly pre-upgrade swarm attempts keep #646's documented fail-open — but only while no modern source-binding evidence exists. Once modern markers survive anywhere in the intent's history (baseline rows, unit fingerprints, worktree provenance, retained reviewed-source refs), stripped or missing fields fail closed instead of downgrading to legacy. Unbindable/bypass-marked/evidence-destroyed states fail closed. `AIDLC_SKIP_SOURCE_FRESHNESS=1` covers all new checks; a bypassed stamp records an explicit marker so it cannot pass as legacy later.

## Behavior notes

- Framework source merges changed their Git hook profile (documented in the CHANGELOG): `pre-commit`, `commit-msg`, and `pre-merge-commit` no longer run, `post-merge` no longer fires on the merge strategy, and `post-commit` now fires there too. This is what makes the landed-commit proof sound — a hook that rewrote the index mid-merge would otherwise cause non-retryable false refusals.
- Ordinary (non-swarm) Bolt merges clean the framework-owned root `.aidlc/` shell (runtime attestations) before non-forced worktree removal; application residue still blocks removal as before.
- Later hardening waves (from the same review cycle) also ship in this PR: a repository-level merge-driver guard (any configured `merge.<name>.driver` refuses a bound merge pre-mutation; repository merge configuration is documented as inside the trusted computing base), canonical per-hop symlink claim validation, creating-repository binding for swarm worktrees (selector + private common-dir realpath; audit rows carry portable selectors only), durable intent-scoped modernity evidence, and corroborated discard authority that survives a deleted worktree directory. A final residuals wave makes the audit ledger machine-portable (project-relative `Worktree path`, centrally redacted project prefixes; legacy absolute rows still resolve), replaces the recorded provenance realpath with a SHA-256 (`gitCommonDirHash`, plaintext migration accepted), stops refusing same-second cross-shard baseline ties whose carried values are byte-identical, and makes claims through symlinked directories fail with an actionable reason.
- One item deliberately out of scope, documented in the review artifacts: the zero-repo-intent aggregate gap stays tracked in #868.
- A docs pass sharpens the Unit / Bolt / autonomy-mode / worktree glossary and rephrases "Bolt worktree" shorthands, since this PR's machinery lives exactly on the ordinary-Bolt vs autonomous-swarm boundary. No identifier, branch name, event, or schema key changed.

## Definition of done (from the RFC)

The three `t304` documented-limitation tests — an earlier unit's post-review edit masked by a later receipt, the inverse re-review ordering, and an addition nobody reviewed — are inverted in place from `rc 0` to refusals, exactly as the RFC proposed. `t305` (44 tests) covers the manifest validator, stamp-time enforcement, per-unit invalidation and shielding, unclaimed fail-closed + both recoveries, unit-major vs stage-major baseline anchoring (including post-jump), tamper/deletion, legacy fail-open and its monotonic boundary, multi-repo scoping, ghost-unit authority, swarm footprint, attempt-bound obligations, merge-interleave refusal (new-path, same-path, and hook-created-commit variants), mode/type binding, ignored-claim rejection (force-add included), directory→file replacement, the reject/redo no-laundering guard, and the unswitched healthy end-to-end swarm completion (one-unit, and two units integrating a shared file).

## Versions

One version ships on this branch: `2.6.68`. `aidlc-version.ts`, the CHANGELOG heading and the README badge are in sync (t68). The branch was re-slotted three times while in review as `v2` advanced (`2.6.20` → `2.6.57` → `2.6.62` → `2.6.68`); all open PR heads were scanned for collisions at each slotting.

## Test Plan

- `bun run check` (parity + both tsconfigs + lint) clean; `bun scripts/package.ts --check` green for all seven harness dists; coverage registry fresh with the ratchet held.
- Full `smoke+unit+integration` tier on the final tree: **352 files, 8,678 assertions** (one drift-guard file timed out under full parallel load and passes green-alone in 2.8s).
- Full deterministic e2e slice (`--e2e --filter "^t[0-9]"`): **32 files, 149 assertions, 0 failures**, including the construction-worktree suites (`t60`/`t61`) that exercise the restructured merge strategies end-to-end (full slice at the re-slot head; the final message-only commit re-verified the discard-adjacent slice).
- Adversarially reviewed to closure across eleven rounds by independent reviewers, with fix waves between; every blocker was proven with a reproduction probe before fixing and re-proven closed after (chain-domain mismatch, three-way shared-file integration, DAG-shrink, field-stripping downgrades, jump-baseline regression, merge interleave, merge-driver injection, symlink boundary shapes, wrong-repository discard).

## Checklist

* [x] Changes are documented

## Acknowledgment


leandrodamascena (changes requested):
Found five blocking correctness gaps in the source-attribution boundary, in addition to the two required CI suites currently failing on this head.

1. **P1: settled swarm permanently disables source verification in the main checkout** (`core/tools/aidlc-state.ts:1853-1927`). Once every unit has a current-attempt `SWARM_UNIT_CONVERGED`, `settledSwarm` suppresses the global source mismatch at lines 1903-1908 and then returns before all receipt/attribution checks at line 1927. After finalize merges the reviewed source, a direct edit in the main checkout can therefore be approved without any review. The exemption needs to cover only the Bolt-local evidence surface; settlement must stamp an aggregate/post-merge main-checkout binding and completion must verify it.

2. **P1: per-unit snapshots discard Git mode/type** (`core/tools/aidlc-lib.ts:5961-5977`, `6137-6176`; consumed at `core/tools/aidlc-state.ts:1867-1877`). `ls-files -s` parses `mode`, but the listing stores and serializes only the blob OID. Reproduction: review alpha claiming `script.sh`, `chmod +x script.sh`, then review beta. Beta refreshes the workspace-global fingerprint, while alpha's unit snapshot and the stage baseline still compare equal because the blob OID did not change; completion accepts the unreviewed executable-bit change. Canonical listings need to bind mode/type plus OID.

3. **P1: ignored application source is invisible to every new attribution check** (`core/tools/aidlc-lib.ts:5922-5927`, `5951-5977`, `6328-6354`). The temporary `git add -A` omits ignored files. A manifest can explicitly claim an ignored source path, but its unit snapshot remains empty; edits to that file move neither the unit listing, baseline listing, nor global fingerprint. I reproduced the primitive directly: an ignored `ignored.ts` never appears in `workspaceSourceListing`, before or after changing its bytes. Either bind explicitly claimed ignored paths by raw content, or reject such claims deterministically; they cannot silently certify no bytes.

4. **P1: a new greenfield workflow is misclassified as legacy if Git appears during code generation** (`core/tools/aidlc-state.ts:102-120`, `2022-2041`; `core/tools/aidlc-lib.ts:5490-5533`). A boundary emitted before any checkout records no `Source Baseline` field at all. If Git is initialized and source is created later, modern reviews with valid empty manifests leave `sourceBaseline.state === legacy`; completion rejects only `unbindable`/`invalid`, so all unclaimed source passes. Newly emitted boundaries must always carry explicit modern metadata (an empty baseline or unbindable marker); fieldless must be reserved for historical rows.

5. **P1: merge authority can be taken from an unrelated or stale convergence row** (`core/tools/aidlc-worktree.ts:401-478`). `convergedSourceRecord()` filters only by unit name, not the current Stage, Run floor, batch, `BOLT_STARTED`, or `WORKTREE_CREATED` lifecycle. Reusing a unit/slug can select a previous attempt's immutable commit; if the selected latest row is fieldless, line 471 downgrades to merging the movable Bolt branch. Same-shard rows sharing a second also select `latest[0]`, the oldest append. Correlate the row to the exact current worktree/Bolt attempt and fail closed on missing modern binding fields.

Required CI is independently red on `265765b7b`: smoke+unit fails in t304 (including commits without configured Git identity), and integration+e2e reproduces 11 failures in t135 around missing intent/worktree/swarm audit state. Focused t304 and t305 pass locally, and package parity is green, but they do not cover the blockers above.

leandrodamascena (changes requested):
I re-reviewed commit `427cc6bf`. The five P1 findings from my previous review are addressed, but two blockers remain.

1. **P1: Multi-repo Bolt source binding drops application code under root `aidlc/` and `.aidlc/`** (`core/tools/aidlc-lib.ts:5910-5927, 6172-6182, 6606-6645`; `core/tools/aidlc-swarm.ts:623-631, 750-760`).

For a registered sibling repository, these directories are application source because the framework shell is their sibling at the workspace roof. The main-checkout listing correctly preserves that distinction, but the Bolt path loses it: `workspaceSourceState()` treats the Bolt as carrying the framework shell, worktree-relative manifests reject these paths, the footprint comparison excludes them, and `bindReviewedSource()` resets them to `HEAD` before creating the immutable Source Commit. `gitCommitSourceListing()` also strips them unconditionally when reconstructing the base and landed commits.

I reproduced this with `repo-a/aidlc/app.ts`: the parent aggregate contained `repo-a\0aidlc/app.ts`, while both the Bolt listing and immutable commit listing omitted it, and the manifest validator rejected the claim as `inside the framework record/shell exclusions`. A Unit can therefore appear reviewed and converged while its application changes are omitted from the Source Commit and never land in the main checkout.

Carry the selected repository/source-role provenance into the Bolt and commit-listing paths, and exclude only the actual injected framework record/metadata paths. Please add end-to-end multi-repo swarm coverage for application files under both root `aidlc/` and `.aidlc/`.

2. **P2: The shared-file attribution regression test has no explicit timeout** (`tests/unit/t304-source-freshness-receipts.test.ts:1090`).

The focused test failed three times locally at 7.07s, 8.52s, and 6.85s against Bun's implicit 5s limit. Adjacent subprocess-heavy tests already use 15-30 second limits. Add an explicit timeout so this required regression is deterministic across supported environments.

Verification: package parity, typecheck, and lint passed. The focused suite completed with 131 passes and the timeout above as its only failure.

The branch is also seven commits behind `v2`, currently conflicts with the base, and still carries version 2.6.57 while `v2` is at 2.6.65. After fixing these findings, rebase, bump to the next available version, regenerate distributions, and rerun the complete suite.

apackeer:
@leandrodamascena Thank you for the deep review — all five P1s were real, and fixing them properly ended up reshaping a good part of the attribution boundary. The branch has been rebased (four times as `v2` advanced, now on `v2` 2.6.65, version slot 2.6.68) and force-pushed; the PR body describes the final shape. Point-by-point against your findings:

1. **Settled swarm disabling main-checkout verification** — the blanket exemption is gone. Settlement now verifies a main-checkout aggregate chain: every application-source merge emits `SWARM_SOURCE_MERGED` with a verifiable opening link (stage baseline or prior accepted aggregate), and completion requires the live checkout to match the chain tip (`verifySettledSwarmSourceBinding`, `core/tools/aidlc-state.ts`). The merge itself is proven rather than trusted — staged-tree capture, `--no-verify` commit, exactly-one-landed-commit assertion, and live-aggregate == pre-merge + committed-delta — so a direct or interleaved unreviewed edit refuses, non-retryably, with no authority row.

2. **Snapshots discarding mode/type** — canonical listings now bind Git mode and type alongside the raw-aware content identity (`sourceListingEntry`); your `chmod +x` reproduction is a pinned regression test, and executable-bit/symlink/gitlink changes invalidate the owning receipt.

3. **Ignored source invisible to attribution** — explicitly claimed ignored paths are now rejected deterministically rather than certifying no bytes (exact claims via `check-ignore --no-index`; directory claims probe ignored descendants under a HEAD-seeded temporary index so `git add -f` cannot hide them; the only exception is a HEAD-tracked ignored path, which is bindable). Both shapes are test-pinned.

4. **Greenfield misclassified as legacy** — every newly emitted boundary now carries explicit modern metadata (including the content-addressed empty state for a Git-less greenfield), and modernity is monotonic: once any modern source-binding evidence exists in the intent's history, a fieldless boundary reads `invalid` (fail closed), not `legacy`. Fieldless is reserved for genuinely pre-upgrade histories.

5. **Merge authority from stale/unrelated convergence rows** — `convergedSourceRecord` is now correlated to the selected intent and the exact current lifecycle (stage, run floor, batch, `BOLT_STARTED`/`WORKTREE_CREATED` provenance), canonicalizes filesystem aliases, and fails closed on modern-fieldless, stale-attempt, mismatched, or ambiguous rows. The movable-branch fallback survives only for verified pre-binding (pre-upgrade) worktrees; durable modern metadata or retained reviewed-source refs make it unreachable.

Both required CI suites are green on the current head (smoke+unit and integration+e2e deterministic), and the deterministic e2e slice also passes in full locally, including the construction-worktree suites. Since your review the branch also went through eleven adversarial review rounds with reproduction probes. A small set of residual hardening items that survived triage (documented trust-boundary edges, multi-repo provenance, hygiene notes) is written up and ready to file as follow-up issues.

I'd appreciate a re-review when you have time.

leandrodamascena (approved):
Approved. I re-reviewed the updated branch and found no remaining blockers. The source-manifest binding and loopback fixtures are now consistent, with the affected integration tests passing.

PR #815, opened 2026-08-20
test: gate hook registration matchers against captured payload fixtures

### LINKED ISSUE #777: Proposal: check each hook registration's matcher against the harness's captured payload fixture

apackeer:
Adds `tests/unit/t308-hook-matcher-fixture.test.ts`, a deterministic drift gate asserting every matcher-bearing hook registration in scope selects at least one tool name actually captured for that harness. Codex uses literal matcher semantics and carries an explicit burn-down allowlist for the two known capture gaps (`spawn_agent` and `request_user_input`); Kiro IDE matcher patterns are checked against the documented live-captured vocabulary. Cursor and Copilot matcher discovery is covered for future registrations. Kiro CLI remains covered by t148 and the #778 fix line because its agent-v1 matchers use glob and alias-family semantics; Claude remains excluded until a captured fixture exists. No runtime behavior change and no framework version bump. Closes #777


leandrodamascena:
This PR should not bump the framework version.

The functional change is test-only: it adds a deterministic matcher/fixture drift gate without changing runtime behavior, distributed harness content, commands, flags, or user-facing semantics. Per the repository's changelog policy, test-only changes live in the commit history and do not receive a release entry.

Please remove:

- the `2.6.26` `CHANGELOG.md` entry;
- the README version badge change;
- the `core/tools/aidlc-version.ts` bump;
- the generated `dist/*/aidlc-version.ts` changes; and
- `(2.6.26)` from the PR title.

The final diff should contain the new test and any genuinely necessary test documentation only.

leandrodamascena (approved):
Rebased onto the current v2 branch and verified the final test-only diff. The framework version bump was removed, the test moved to the available t308 slot, Codex literal matcher semantics are now evaluated exactly, and the new request_user_input registration is represented by an anti-rot fixture-gap entry until a live capture lands. Focused tests, coverage-registry checks, package parity, typecheck, and lint all pass locally.

PR #817, opened 2026-08-20
fix(plugins): emit Kiro plugin compose wiring per runtime generation (2.6.47)

### LINKED ISSUE #784: [Bug]: the Kiro plugin projection emits a legacy .kiro.hook, which is not the form either runtime it targets reads
### Version

### Summary

Both Kiro trees project a plugin with `kind: "kiro"`
(`harness/kiro/manifest.ts:95`, `harness/kiro-ide/manifest.ts:145`), and that kind
emits the compose manifest as a legacy `.kiro.hook` (`scripts/package.ts:972-974`,
`when: { type: "promptSubmit" }`). `.kiro.hook` is the IDE 0.x format. For
`dist/kiro`, whose documented runtime is Kiro CLI ≥ 2.6 (`README.md:59`), the
hook wiring the tree itself uses is the `hooks` block inside
`.kiro/agents/aidlc.json` — it ships no standalone manifest of any kind — so the
emitted file is a form that tree's own runtime does not use.

### Evidence

- Emission: `scripts/package.ts:972-984` writes
  `<projection>/hooks/aidlc-plugin-compose.kiro.hook` for `kind: "kiro"`, and both
  Kiro harnesses declare that kind (`harness/kiro/manifest.ts:95`,
  `harness/kiro-ide/manifest.ts:145`).
- What the trees themselves do: `dist/kiro/.kiro/hooks/` contains 18 files, all
  `.ts` scripts and no manifest; the wiring is `.kiro/agents/aidlc.json`'s `hooks`
  block. `dist/kiro-ide` registers in both formats — 8 `.kiro/hooks/aidlc-*.json`
  for IDE ≥ 1.0 and 9 `.kiro/hooks/aidlc-*.kiro.hook` for pre-1.0 builds
  (`README.md:138`); `session-end` is the one that ships in the legacy form only.
- Kiro's own documentation, read 2026-08-17: the IDE 0.x reference states "In IDE
  0.x, hooks were stored in `.kiro.hook` files. In IDE 1.0, hooks moved to
  `.kiro/hooks/*.json` files using a versioned JSON schema"; the CLI 2.x reference
  states "In CLI 2.x, hooks were embedded directly in the agent configuration file
  rather than in standalone files."
- So by the vendor's description the emitted form is read by pre-1.0 IDE builds
  only: not by IDE ≥ 1.0 (which reads `.kiro/hooks/*.json`) and not by CLI 2.x
  (which reads the agent config). I have **not** probed a CLI 2.x install with the
  file in place; this is a read of the documentation and of what the trees ship.

### Impact

On `dist/kiro` a plugin's "composes on first interaction" hook has no runtime that
would execute it even if it were installed in `.kiro/hooks/`, so the explicit
composer is the only mechanism there. On `dist/kiro-ide` it works on pre-1.0 IDE
builds and not on the 1.x builds the tree otherwise supports with its own dual
registration. Both are silent: nothing in the projection or the install reports a
form/runtime mismatch.

### Expected behavior

Emit per generation, the way the harness already does for its own hooks — the
`.json` v2 schema where the runtime reads it, and for `dist/kiro` either the agent
config's `hooks` block or no manifest at all with the explicit composer documented
as the mechanism.

### Reproduce

```bash
# at 905eb3e2
bun scripts/package.ts plugin build test-pro kiro /tmp/plug-kiro
ls /tmp/plug-kiro/hooks           # aidlc-plugin-compose.kiro.hook  compose.ts
ls dist/kiro/.kiro/hooks | wc -l  # 18 — all .ts, no manifest
rg -c 'preToolUse|postToolUse' dist/kiro/.kiro/agents/aidlc.json   # the actual wiring
```

apackeer:
The Kiro plugin projection emitted a legacy .kiro.hook that neither target runtime reads (Kiro CLI 2.x reads hooks only from the agent config; Kiro IDE >=1.0 reads .kiro/hooks/*.json v2 registrations). The Kiro CLI projection now ships no hook registration — the documented explicit composer is the mechanism — and the Kiro IDE projection ships a v2 .kiro/hooks/aidlc-<plugin>-compose.json SessionStart registration that runs compose from the workspace root with AIDLC_PLUGIN_ROOT set (the old command also depended on a ${PLUGIN_ROOT} variable no Kiro runtime populates). Live IDE firing was not probed (needs macOS + Kiro.app); the registration mirrors the schema of the tree's own v2 hooks that current IDEs execute. Closes #784


leandrodamascena (approved):
Approved.

Kiro IDE now invokes the generated compose wiring through the cross-platform Bun launcher, without relying on `sh`, `$PWD`, `command -v`, or POSIX environment assignments.

The regression test executes the folder-dropped wiring end to end with `aidlc` absent from `PATH` and confirms successful plugin composition.

Validation: `bun run check`, package drift check, coverage registry, all 80 `t188` tests, and all 7 `t68` tests passed.

PR #818, opened 2026-08-20
fix(kiro): verb-intercept parses raw /aidlc prompts, not only expanded skill bodies (2.6.46)

### LINKED ISSUE #776: Kiro CLI: verb-intercept's argv parser expects an expanded skill body, so the terminal-command dispatch silently no-ops
## What I observed

The Kiro CLI `verb-intercept` seam recovers its argv by matching an **expanded skill body** in the prompt:

```ts
// harness/kiro/hooks/aidlc-kiro-adapter.ts:162-172
const m = expandedPrompt.match(/aidlc-orchestrate\.ts next ([^`\n]*)`/);
if (!m) return { raw: "", args: [] };
```

It is called with the hook payload's prompt at `harness/kiro/hooks/aidlc-kiro-adapter.ts:199`, and the seam is registered on `userPromptSubmit` at `harness/kiro/agents/aidlc.json:70`.

I cannot find a payload shape that matches that anchor.

1. **This repository's own captured fixture does not carry it.** `tests/fixtures/kiro-hook-payloads/payloads.json` (`_provenance`: *"Live-captured on kiro-cli 2.6.1 … FIELD NAMES are verbatim"*) has `userPromptSubmit.prompt = "Say exactly HELLO and nothing else."` — plain text, no slash command, no code span, no `aidlc-orchestrate.ts next` anchor.
2. **On the two surfaces I measured, the payload is the raw slash text.** Kiro IDE 1.0.309 and `kiro-cli 2.18.1 --v3`: `userPromptSubmit` delivers exactly what the human typed (a 34-character probe arrived verbatim on both).

## Why it matters

When the anchor does not match, `args` is `[]`, `classifyTerminalCommand([])` returns `null`, and the terminal-command dispatch returns without doing anything. The same seam still bumps the turn counter (`:212`) and records the human turn, so the failure is silent: the hook fires, the health beat is written, and only the deterministic handling of `/aidlc --status`, `/aidlc space`, `/aidlc intent` and friends is missing. A `.aidlc-hooks-health/*.last` file proves the hook ran, never that this branch did anything.

## Scope of the claim

I have **not** measured Kiro CLI 2.x in its classic UI, which is the surface `dist/kiro` targets, so I am not asserting the anchor never matches anywhere — I am saying I cannot point at a generation where it does, and this repository contains no payload example of the expanded-body shape.

If the expanded-body shape is real on some generation, adding that payload to `tests/fixtures/kiro-hook-payloads/payloads.json` would settle it and pin the behaviour. If it is not, the parser wants the raw slash text instead (`/aidlc(?![\w-])\s*(.*)` plus a small shell-word split is enough, and it is what I ended up using for a Markdown-agent Kiro tree in #775).

## Environment

- Measured with Kiro IDE 1.0.309 and kiro-cli 2.18.1 (`--v3`), macOS.

apackeer:
The Kiro CLI verb-intercept seam recovered its argv only from an expanded-skill-body anchor (`aidlc-orchestrate.ts next ...`), but measured `userPromptSubmit` payloads (repo fixture kiro-cli 2.6.1; Kiro IDE 1.0.309 and kiro-cli 2.18.1 `--v3` per the report) carry the raw typed text, so deterministic terminal dispatch (`/aidlc --status`, `space`, `space-create`, `intent`, ...) silently no-oped. This adds an anchored raw `/aidlc` fallback (the anchor still wins where present), keeps mid-sentence mentions and `/aidlc-foo` inert, and pins both payload shapes in t180. Closes #776


leandrodamascena (approved):
Rebased onto the current v2 branch and reviewed the raw-prompt fallback against the existing expanded-body path. The forwarding-loop anchor retains precedence, raw leading /aidlc prompts preserve shell-normalized argv, and ordinary text, mid-sentence mentions, and /aidlc-foo remain inert. The release metadata is refreshed to 2.6.46. Focused tests, version sync, package parity, typecheck, and lint all pass locally.

PR #819, opened 2026-08-20
fix: drop the inert Kiro plugin compose manifest — the explicit composer is the arming step (2.6.28)

### LINKED ISSUE #783: [Bug]: Kiro's plugin folder-drop lands the compose manifest outside the runtime's read path, so a dropped plugin never arms
### Version

### Summary

The documented Kiro plugin install is a folder-drop followed by an explicit
composer run. The drop copies the projection's compose manifest to
`<project>/hooks/`, but Kiro reads hook manifests from `<project>/.kiro/hooks/`,
and `compose.ts` installs no hook files. The emitted manifest therefore sits on
disk where no runtime looks, on every Kiro tree, and the composer runs only
because the chapter tells the reader to run it by hand.

### Evidence

- The manifest is written into the projection's own `hooks/` directory:
  `scripts/package.ts:974` (`join(hooksDir, "aidlc-plugin-compose.kiro.hook")`,
  where `hooksDir = join(outDir, "hooks")` at `:944`).
- The documented drop copies that projection root into the project root:
  `docs/harness-engineering/10-authoring-a-plugin.md:307`
  (`cp -r dist/plugins/<name>/kiro/. <project>/`) and
  `docs/reference/18-plugin-mechanism.md:394` (`cp -r "$PLUGIN_ROOT"/. <project>/`).
- Measured on a stock `dist/kiro` install (repro below): `.kiro/hooks/` holds 18
  entries before the drop and 18 after; the manifest lands at
  `<project>/hooks/aidlc-plugin-compose.kiro.hook`; and `.kiro/agents/aidlc.json`
  — the file that carries this tree's own hook wiring — contains no reference to
  it either.
- The same drop deposits eight directories at the project root — `agents/`,
  `contributions/`, `hooks/`, `knowledge/`, `scopes/`, `sensors/`, `stages/`,
  `tools/` — none of them inside `.kiro/`. They are inputs `compose.ts` reads
  through `AIDLC_PLUGIN_ROOT`, so nothing is broken by their presence, but a
  project root that suddenly carries `agents/` and `tools/` reads like a botched
  install.

### Impact

A reader who follows the drop and then skips, mistypes, or lacks the environment
for the composer step gets no plugin composition and no diagnostic: the manifest
is present, and nothing reports that it is in a directory the runtime never
scans. Claude and Codex mediate the same plugin through their own install path,
so Kiro is the one documented route where "installed" and "armed" come apart
without a diagnostic.

### Expected behavior

Either the drop (or a sync step) installs the manifest into `.kiro/hooks/` in the
schema the target tree's runtime reads, or the manifest is documented as build
output that Kiro does not consume, with the explicit composer named as the
mechanism. #775 already reworded the two sentences that described the drop as
self-arming, so what is left here is the mechanism.

### Reproduce

```bash
# at 905eb3e2
bun scripts/package.ts plugin build test-pro kiro /tmp/plug-kiro
PROJ=$(mktemp -d); mkdir -p "$PROJ/.kiro"; cp -R dist/kiro/.kiro/. "$PROJ/.kiro"/
ls "$PROJ"/.kiro/hooks | wc -l            # 18
cp -r /tmp/plug-kiro/. "$PROJ"/           # the documented folder-drop
ls "$PROJ"/.kiro/hooks | wc -l            # 18 — unchanged
find "$PROJ" -name 'aidlc-plugin-compose*'
# <project>/hooks/aidlc-plugin-compose.kiro.hook
```

apackeer:
The Kiro plugin projections emitted a `hooks/aidlc-plugin-compose.kiro.hook` that the documented folder-drop landed at `<project>/hooks/`, a path no Kiro runtime reads (CLI hooks ride the agent JSON, IDE hooks live under `.kiro/hooks/`) and whose command depends on a `${PLUGIN_ROOT}` no Kiro host sets — so a dropped plugin looked installed but never armed. The kiro-kind projection now ships `hooks/compose.ts` only, the docs name the explicit `aidlc plugin sync` / `bun compose.ts` run as the arming step (and the security note no longer claims the hook fires on the next prompt), and t188 pins the kiro hook inventory to exactly `compose.ts`. Auto-fire remains the recorded deferred surface; this removes the file that made "installed" and "armed" silently come apart. Closes #783


leandrodamascena:
Closing as superseded by #817, now merged into `v2`.

#817 removes the inert Kiro CLI manifest while preserving Kiro IDE auto-composition through a valid `.kiro/hooks/` `SessionStart` registration and a cross-platform Bun launcher. Rebasing this PR would remove that IDE wiring and reintroduce a functional regression.

The underlying issue in #783 is resolved by #817.

PR #820, opened 2026-08-20
fix: seed opencode and Copilot persona memory pointers at default so a space switch never rewrites shipped files

### LINKED ISSUE #785: [Bug]: a space switch rewrites 22 shipped persona files on opencode and Copilot, replacing the <active-space> placeholder with a concrete space
### Version

### Summary

`repointOpencodeAgentMemory` (`core/tools/aidlc-includes.ts:135-141`) rewrites
every occurrence of `aidlc/spaces/<anything>/memory/` in a persona body. The
persona bodies shipped for opencode and Copilot carry that path as a documentation
placeholder — `aidlc/spaces/<active-space>/memory/{org,team,project}.md` — so the
first space repoint substitutes the placeholder with a concrete space name. On a
`default → default` repoint (the no-op case) it rewrites **22 files** on each of
those two trees, while the Cursor tree, which is seeded for exactly this reason,
rewrites zero.

### Evidence

- The pattern: `core/tools/aidlc-includes.ts:136-138` —
  `raw.replace(/aidlc\/spaces\/[^/]+\/memory\//g, …)`. It is applied to
  `.cursor/rules/*.mdc` and `.cursor/agents/*.md` (`:210`, `:227`) and to
  `.aidlc/agents/*.md`, `.opencode/agents/*.md`, `.github/agents/*.md` (`:348`).
- The invariant it breaks is stated in the tree itself, `scripts/package.ts:269-272`:
  *"Cursor persona bodies are mutable active-space pointers. Ship their memory
  references on the default seed so the first startup's
  `repointHarnessIncludes(project, "default")` is byte-identical."* That seed is
  applied only when `harness === "cursor"` (`:274-279`), so opencode and Copilot
  ship `<active-space>` and are rewritten on the first repoint.
- Measured on untouched dist copies at `905eb3e2` (repro below):
  opencode `repointed 22`, 22 files differ; Copilot `repointed 22`, 22 files
  differ; Cursor `repointed 0`, 0 files differ. The 22 are the 11 personas in
  `.aidlc/agents/` and their 11 twins in `.opencode/agents/` (`.github/agents/`
  on Copilot).
- The rewrite is inconsistent within a single list, because only the `memory/`
  segment matches:

  ```diff
  -1. `aidlc/spaces/<active-space>/memory/{org,team,project}.md` — active-space guardrails …
  +1. `aidlc/spaces/default/memory/{org,team,project}.md` — active-space guardrails …
   2. `.aidlc/knowledge/aidlc-shared/` — methodology principles
   3. `.aidlc/knowledge/aidlc-developer-agent/` — agent-specific methodology
   4. `aidlc/spaces/<active-space>/knowledge/aidlc-shared/` — team shared knowledge (if exists)
  ```

### Impact

Three consequences, in order of how likely they are to be noticed:

1. The first startup on opencode or Copilot dirties 22 shipped files, so
   `git status` after an install-and-launch shows 22 modified personas that the
   user did not touch — and the "repoint to default is byte-identical" property
   the Cursor seed exists to guarantee does not hold on those two harnesses.
2. The persona body ends up internally inconsistent: item 1 names a concrete space
   while item 4 still says `<active-space>`, in the same numbered list.
3. The placeholder is not recoverable by switching spaces again — each switch
   rewrites the concrete segment to the new space — so the shipped prose is only
   restorable by re-copying the dist tree.

### Expected behavior

Either anchor the rewriter on the concrete pointer form it is meant to move, or
seed the opencode and Copilot persona bodies at `default` the way `package.ts`
already seeds Cursor's. For the anchoring option there is a precedent in the same
file added by #775: `repointKiroAgentMarkdownResources` matches the full
`file://aidlc/spaces/<space>/memory/**/*.md` resource form specifically so a bare
path fragment in prose is not a candidate.

### Reproduce

```bash
# at 905eb3e2
for h in opencode copilot cursor; do
  F=$(mktemp -d); cp -R dist/$h/. "$F"/
  leaf=$([ $h = cursor ] && echo .cursor || echo .aidlc)
  (cd "$F" && bun $leaf/tools/aidlc-utility.ts space default)
  echo "$h: $(diff -rq dist/$h "$F" | wc -l) files differ"
done
# opencode: repointed 22 → 22 files differ
# copilot:  repointed 22 → 22 files differ
# cursor:   (no repoint line) → 0 files differ
```

apackeer:
A space switch rewrote 22 shipped persona files on the opencode and Copilot dists because their bodies carried the `aidlc/spaces/<active-space>/memory/` documentation placeholder, which the repoint regex matches. This seeds those persona memory references at the concrete default space at package time — exactly the existing Cursor seed — across the packager transform and both emit paths, with byte-identity regression tests, so repoint-to-default is a no-op on all three harnesses. Closes #785


leandrodamascena (approved):
Rebased onto the current v2 branch and reviewed the default-space projection across both emission layers. The shared package transform covers the inline .aidlc persona copies, while the opencode and Copilot emitters cover their native .opencode and .github twins. Default-space repointing is byte-identical and non-default switches still rewrite all managed personas. Release metadata is refreshed to 2.6.48. Focused tests, package parity, typecheck, and lint pass locally.

PR #821, opened 2026-08-20
fix(kiro): project disallowedTools out of the Kiro-native agent surface (2.6.60)

### LINKED ISSUE #774: [Bug]: `disallowedTools: Task` reaches the Kiro-native agent surface unprojected — inert on both kiro dists (copilot/opencode project it and guard the leak)
### Version

2.6.2 (`core/tools/aidlc-version.ts`), commit `4569754`

### Summary

In `dist/kiro/` and `dist/kiro-ide/`, the file that serves as the **Kiro-native
agent surface** — `.kiro/agents/*-agent.md` — still carries the core frontmatter
line `disallowedTools: Task` verbatim. `disallowedTools` is a **Claude Code**
subagent field; it is not part of Kiro's agent schema in any generation, and
`Task` is not a Kiro tool name. On both Kiro surfaces the line is inert: a
declared denial that nothing enforces.

Copilot and OpenCode avoid this by separating the engine copy from the native
twin — the twin carries a projected, platform-supported form. Kiro has no such
separation: one file plays both roles, and the Claude-only key rides along into
the surface config.

Same class as #533 (`modelOverride:` vs `model:`) — a frontmatter key that reads
as a guarantee and is a no-op — except here the repo already recognizes the
hazard elsewhere and guards against it.

### Evidence

**The field is Claude-only, per this repo's own docs.**
`docs/reference/14-claude-features.md:326` records the spike result: "`disallowedTools`
is a real, working field on the denylist side" — a statement about Claude Code
subagents. `docs/reference/05-agent-system.md:23` and `docs/reference/03-orchestrator.md:414`
frame it the same way.

**It is not in Kiro's agent schema at any generation.**

| Kiro generation | Tool-restriction fields |
|---|---|
| CLI 2.x (agent-v1 JSON) | `tools`, `allowedTools`, `toolsSettings` |
| IDE 1.x / CLI v3 (unified agent harness) | `tools` (tag-based), `excludedTools`, `permissions` (capability rules, incl. the `subagent` capability); `toolsSettings` removed |

Sources: [Custom agents](https://kiro.dev/docs/custom-agents),
[Configuration reference](https://kiro.dev/docs/custom-agents/configuration-reference),
[CLI v3 agent config changes](https://kiro.dev/docs/cli/v3/agent-config). None
of the three lists `disallowedTools`.

**Where it lands, and what actually blocks delegation today.**

- `dist/kiro-ide/.kiro/agents/*-agent.md:11` — here the frontmatter *is* surface
  config. `harness/kiro-ide/manifest.ts:107-113` states it: "the IDE resolves a
  delegated subagent's tools from the agent .md frontmatter, not from the
  agent-v1 JSONs" (the finding behind #458, fixed via `frontmatterAdditions`).
  `disallowedTools` sits in that same block as an unrecognized key, one line
  above the injected `tools: ["read", "write", "shell"]` that does the real work.
- `dist/kiro/.kiro/agents/*-agent.md:11` — the CLI JSON references the persona as
  `"prompt": "file://aidlc-<role>-agent.md"`, so the frontmatter block is ingested
  as prompt text. The line is a string in the prompt, not a config. Nested
  delegation is prevented by the JSON `tools` array simply not listing a
  delegation tool.

**The other native-surface emissions project it instead of copying it.**

- copilot: `.github/agents/*.md` gets the projected form —
  `harness/copilot/emit.ts:66-99` turns the denial into a supported `tools:`
  allowlist, and its comment states the principle outright: "An unknown
  `disallowedTools` value fails the build instead of silently shipping an
  unenforced denial."
- opencode: `.opencode/agents/*.md` gets `permission:` / `task: deny` —
  `harness/opencode/emit.ts:38-57`, which drops the source line.
- Guards assert the twin stays clean: `tests/unit/t248-copilot-packaging.test.ts:134`,
  `tests/unit/t240-opencode-packaging.test.ts:126` ("no inert `disallowedTools`
  leak"), `tests/integration/t188-plugin-compose.test.ts:527,563`.

In both those dists the core line does survive in the engine copy under
`.aidlc/agents/` — which is correct, that file is not a platform surface. The
Kiro dists have no such two-file split: `.kiro/agents/*-agent.md` is both the engine
copy and the platform surface, and no projection or guard applies to it
(`tests/unit/t148-kiro-file-structure.test.ts`,
`tests/unit/t252-kiro-allowlist-semantics.test.ts` do not check for it).

### Impact

No runtime regression: on both Kiro surfaces the delegation tools are withheld
by allowlist, so nested delegation is in fact blocked. The cost is correctness of
the shipped surface:

- The one line in the file that *declares* a tool restriction is the one with no
  effect on the platform reading it, while the actual enforcement is implicit —
  an omission from an allowlist.
- On kiro-ide it is unrecognized config, and the unified agent format has a real
  way to express this (`excludedTools`, or a `permissions` deny on the `subagent`
  capability) that is going unused.
- On kiro (CLI) it is prompt noise: a Claude-Code-shaped directive naming a tool
  (`Task`) that does not exist on the surface reading it.

### Expected behavior

`disallowedTools` should not reach a Kiro-native agent file unprojected, on the
same terms the copilot and opencode emissions already hold themselves to —
either dropped, or projected into the Kiro-native form (`excludedTools`, or a
`permissions` deny on `subagent` for the unified format), with a t148/t252-style
guard asserting it stays out.

Two things make this cheap:

- The seam exists. `scripts/package.ts:152-193` (`projectTierFrontmatter`) already
  rewrites `tier:` into harness-native keys per harness during packaging.
- The engine does not read the field — `grep -rn disallowedTools core/tools/`
  returns nothing; `loadAgents()` parses `name`, `display_name`, `examples`,
  `description`. And the persona body already carries the prose backstop
  ("**IMPORTANT: Do NOT use the Task tool...**") on the line immediately after
  the frontmatter.

### Reproduce

```bash
# Kiro-native surfaces carry the Claude-only key (14 personas x 2 dists)
grep -c '^disallowedTools:' dist/kiro/.kiro/agents/*-agent.md dist/kiro-ide/.kiro/agents/*-agent.md | grep -c ':1'
# 28

# The other native twins carry a projected form instead
grep -n '^disallowedTools:' dist/copilot/.github/agents/*.md dist/opencode/.opencode/agents/*.md
# no hits
grep -n 'permission:\|task: deny' dist/opencode/.opencode/agents/aidlc-developer-agent.md
# the opencode-native denial

# The engine never reads the field
grep -rn 'disallowedTools' core/tools/
# no hits
```

apackeer:
The Claude-only `disallowedTools: Task` frontmatter key shipped verbatim in all 28 Kiro CLI/IDE agent .md files, where it is an inert, unenforced denial. The packager now validates that the authored denial is exactly `Task` and drops the line for the kiro tier flavor (nested delegation stays blocked by Kiro's native tool allowlists, which omit `subagent` for delegates), mirroring the copilot/opencode precedent. Plugin compose applies the same projection: new personas are projected with drop-logging for unprojectable values, and re-running compose migrates an installed persona when it is an exact, unchanged, same-plugin pre-projection copy; edited or foreign files keep no-clobber collision behavior, and an already-composed unsupported value gets a specific remove-and-recompose diagnostic. Central architecture and agent references now distinguish the authored/Claude `disallowedTools: Task` enforcement from Kiro's native delegate allowlists. Guarded by t148 (projected surface) and t188 (compose projection and migration paths). Ships as 2.6.56. Closes #774

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.57] - 2026-08-23
+
+Kiro CLI and IDE agent Markdown no longer carries the unsupported Claude-only `disallowedTools` key; nested delegation remains blocked by Kiro's native agent tool configuration. **Upgrade:** refresh your `dist/kiro/` or `dist/kiro-ide/` shell, and re-run compose for installed Kiro plugins so their agent personas receive the same projection.
+
+* Kiro packaging validates that the authored denial is exactly `disallowedTools: Task`, then omits that inert line from all 14 projected agent personas on both Kiro distributions.
+* Kiro plugin composition strips the supported `Task` denial from composed personas and drop-logs any agent that declares a different, unenforceable `disallowedTools` value.
+
 ## [2.6.56] - 2026-08-23
 
 Kiro CLI agent-v1 hooks retain the literal selectors required by the host's canonical-name/alias glob semantics, while the adapter now normalizes selected payloads before shared guards and observers run. **Upgrade:** re-copy `dist/kiro/` into the project.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [2.6.57] - 2026-08-23
+## [2.6.60] - 2026-08-23
 
-Kiro CLI and IDE agent Markdown no longer carries the unsupported Claude-only `disallowedTools` key; nested delegation remains blocked by Kiro's native agent tool configuration. **Upgrade:** refresh your `dist/kiro/` or `dist/kiro-ide/` shell, and re-run compose for installed Kiro plugins so their agent personas receive the same projection.
+Kiro CLI and IDE agent Markdown no longer carries the unsupported Claude-only `disallowedTools` key; nested delegation remains blocked by Kiro's native agent tool configuration. **Upgrade:** refresh your `dist/kiro/` or `dist/kiro-ide/` shell, and re-run compose for installed Kiro plugins. An unchanged same-plugin persona is migrated automatically; if doctor reports an already-composed unsupported value, fix the plugin source, remove the named installed persona, and re-run compose.
 
-* Kiro packaging validates that the authored denial is exactly `disallowedTools: Task`, then omits that inert line from all 14 projected agent personas on both Kiro distributions.
-* Kiro plugin composition strips the supported `Task` denial from composed personas and drop-logs any agent that declares a different, unenforceable `disallowedTools` value.
+* Kiro packaging requires exactly one authored `disallowedTools: Task` denial, then omits that inert line from all 14 projected agent personas on both Kiro distributions.
+* Kiro plugin composition strips the supported `Task` denial from new personas, migrates exact unchanged same-plugin legacy copies without clobbering edits, and drop-logs duplicate, unsupported, or already-composed invalid values with recovery guidance.
 
 ## [2.6.56] - 2026-08-23
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.57-blue)
+![version](https://img.shields.io/badge/version-2.6.60-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.56-blue)
+![version](https://img.shields.io/badge/version-2.6.57-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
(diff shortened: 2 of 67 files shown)


leandrodamascena (changes requested):
Found one upgrade blocker and one documentation inconsistency. The fresh core projections themselves look correct and required CI is green.

1. **P1: the documented plugin upgrade path never applies the projection to an already-composed agent** (`scripts/plugin-hooks-template/compose.ts:1113-1168`, especially 1137-1153). The changelog tells users to “re-run compose” so installed Kiro plugin personas receive this projection, but `copyTreeNoClobber()` never overwrites an existing destination and skips the new precheck. It transforms only for comparison; an old installed persona still containing `disallowedTools: Task` differs from the projected source, so compose records a generic collision and leaves the inert line in place. An old unsupported value is likewise retained and never reaches the specific drop check. Thus fresh installs are fixed, but the upgrade instruction does not repair the population that needs migration. Add a same-plugin, narrowly verified migration for this one projection, or change the recovery contract to require removing the owned installed persona before re-compose and test that exact upgrade path.

2. **P2: central references still describe `disallowedTools` as the universal runtime enforcement** (`docs/reference/01-architecture.md:58`, `docs/reference/03-orchestrator.md:423`, `docs/reference/05-agent-system.md:15-34,50-71`; similar claims remain in `docs/reference/06-hooks-and-tools.md` and `docs/reference/agents/README.md`). After this change, Kiro projections intentionally do not have that key; their boundary is the native JSON/frontmatter tool allowlist. The two updated documents explain the new model, but these primary architecture/agent contracts still say all agent files carry and enforce `disallowedTools: Task`. Qualify authored/Claude surfaces and document Kiro’s native enforcement consistently so users do not inspect a correct Kiro dist and conclude the safety control is missing.

Validation: all GitHub checks pass on `35203bd24`; `bun scripts/package.ts --check` is green locally. The focused tests could not be rerun in the isolated worktree because dependencies are not installed there, but CI ran both affected suites successfully.

apackeer:
@leandrodamascena both findings from your review are addressed, and the branch has been rebased through the recent v2 movement (now slotted as 2.6.57 on top of #788's merge).

1. **Upgrade path (P1):** re-running compose now migrates an installed Kiro plugin persona when it is an exact, unchanged, same-plugin pre-projection copy, with the rewrite going through the compose write transaction. Edited or foreign files keep the no-clobber collision behavior, and an installed persona with an unsupported `disallowedTools` value gets a specific diagnostic naming the file to remove before re-compose, instead of the generic collision. t188 covers all three paths: exact migration, user-edited preservation, and the unsupported-value recovery message.

2. **Docs (P2):** 01-architecture, 03-orchestrator, 05-agent-system, 06-hooks-and-tools, and agents/README now scope `disallowedTools: Task` to the authored core and Claude surfaces and document Kiro's native enforcement (delegate JSON/frontmatter allowlists that omit `subagent`). The conductor module and the guide/authoring chapters got the same treatment.

CI is green on the current head, including smoke+unit (an earlier push tripped t239's documentation-parity phrase pin; the pinned wording is restored). Ready for another look when you have time.

leandrodamascena:
@apackeer I addressed the two remaining blockers and updated the PR at `058b8bd3a`.

- Kiro core packaging now requires exactly one `disallowedTools: Task` declaration. Duplicate declarations fail packaging instead of allowing the first value to authorize removal of every denial.
- Kiro plugin compose now rejects duplicate `disallowedTools` declarations for new personas. Legacy exact-copy migration also refuses duplicates, preserves the installed file, and emits the specific remove-and-recompose recovery guidance.
- Added regressions for the authored core contract, fresh plugin composition, and legacy migration.
- Rebased against the current `v2` tip. The branch was already up to date.
- Re-slotted the release to `2.6.60` because `2.6.57`, `2.6.58`, and `2.6.59` are claimed by #813, #797, and #789 respectively. Version metadata, changelog, generated distributions, and the PR title are synchronized.

Local validation passed:

- `t04`: 6/6
- `t148`: 18/18
- `t188` plugin compose: 81/81
- `t68`: 7/7
- `bun run check`
- package parity
- coverage-registry freshness
- `git diff --check`
- real packager adversarial probe confirms `Task + WebSearch` now exits with an error

GitHub CI is currently running on the updated head.

leandrodamascena (approved):
Approved!

PR #822, opened 2026-08-20
feat: engine-owned atomic single-use continuation cursor on every harness (2.6.51)

### LINKED ISSUE #762: Enforce atomic single-use continuation replay across sessionless AI-DLC harnesses
## Problem

PR #749 makes continuation tokens single-use only when the active-directive marker is an exact-context, non-sessionless Copilot marker. Claude, Codex, Cursor, Kiro, Kiro IDE, opencode, and other sessionless paths intentionally retain the existing stateless behavior. The same valid continuation token can therefore be presented more than once and produce the same successor directive.

That compatibility decision is pinned by a real same-token repeat regression in PR #749. It must not become an accidental permanent contract.

## Goal

Define and implement one shared, crash-safe, atomic continuation cursor for every supported harness without weakening native token validation or moving host-specific Stop, Resume, ownership, compaction, and delivery semantics into the engine.

## Acceptance Criteria

1. Specify the authoritative cursor identity for every supported harness, including currently sessionless markers.
2. Compare the SHA-256 of the complete presented token with the complete current token under the canonical active-directive transaction.
3. Atomically replace the current token with the exact prepared successor, or a tokenless `run-stage`, before stdout.
4. Guarantee exactly one successful winner for concurrent uses of the same current token on every supported harness.
5. Define crash and retry behavior before commit, after commit, before stdout, after stdout, and across restart.
6. Define fresh `next` reset ordering against concurrent `continue` in both lock orders.
7. Preserve missing, malformed, legacy, and migrated-marker recovery through an explicit compatibility and rollout plan.
8. Add production-path tests for Claude, Codex, Copilot, Cursor, Kiro, Kiro IDE, and opencode. Mock-only adapter tests are not sufficient.
9. Add real same-token, mixed-process, stale-token, final-run-stage, and crash-boundary regressions with exact result cardinality.
10. Keep Post and host hooks cursor-read-only. Host correlation may enrich delivery evidence but may not own replay.
11. Document migration, release boundaries, rollback, unsupported filesystems, and harness-specific residual limits.
12. Preserve existing hook-owned Stop retention, Resume, ownership, compaction, conversation, and count semantics unless separately accepted criteria require a change.

## Non-Goals

- Do not close the remaining #740 Cancel, prompt, compaction, or TUI matrix as a side effect.
- Do not introduce a database, daemon, queue, or second marker lock hierarchy without a separate design review.
- Do not claim exactly-once `report` or `park` execution.
- Do not infer replay authority from PID, age, liveness, audit prose, model output, or caller-supplied identity.

## Relationship to PR #749

This is the required follow-up accepted during PR #749 review. PR #749 deliberately preserves sessionless same-token replay until this issue lands. A future change must replace that control explicitly and must not be folded into PR #749.

apackeer:
Closes #762 — the required follow-up accepted during PR #749 review.

## What

Continuation tokens are now **single-use on every harness** — Claude, Codex, Copilot, Cursor, Kiro CLI, Kiro IDE, and opencode share one engine-owned, crash-safe continuation cursor stored in the existing `.aidlc-active-directive.json` marker and serialized by the existing `.aidlc-active-directive.lock/`. This closes the deliberate sessionless same-token replay carve-out from PR #749.

- `continue` keeps all native token validation (decode, MAC, state, route), then — inside the active-directive transaction — compares the SHA-256 of the complete presented token with the persisted current token and atomically publishes the exact successor (or a tokenless `run-stage`) **before stdout**. Replaying a consumed token returns an error directive; concurrent uses of one token have exactly one winner.
- Fresh `next` is an explicit cursor reset and must commit its first directive before emitting on every harness; the previous non-Copilot fail-open publication path is removed. Publication failure/contention yields an error directive and no work directive.
- Cursor identity = canonical project, active record, intent UUID, complete state digest + presence, and the installed harness name (`tools/data/harness.json`), recorded as a new optional `cursor_harness` field on marker v2. Harness *directories* are not identities (Kiro/Kiro IDE share `.kiro`; Copilot/opencode share `.aidlc`).
- Missing, malformed, oversized, v1, and cross-harness markers permit exactly one natively validated recovery continuation under the same transaction; a mismatched token is stale, not a bootstrap opportunity. Pre-upgrade markers migrate on the first matching `continue` or fresh `next`.
- Copilot claim/settlement/Resume/ownership/Stop-count semantics remain host-owned marker enrichment and do not own replay (issue AC 10/12).

## Behavior change

Presenting the same continuation token twice now errors with "This continuation token is no longer current for this workflow. Run a fresh `next`" instead of repeating the successor directive. Recovery is always a fresh `next`.

**Upgrade:** replace the whole `dist/<harness>/` shell in one quiescent swap, then run a fresh `next` (a matching in-flight token migrates atomically). **Rollback** restores the older release's sessionless replay behavior until re-upgraded. Full crash/retry matrix, migration, rollback, filesystem requirements, and per-harness residual limits are documented in the new "Atomic continuation cursor" section of `docs/reference/06-hooks-and-tools.md`.

## Tests

`t283` was rewritten from a Copilot-only cursor test into a production-path matrix that drives the real packaged engine as separate processes against all seven installed harness trees (real `harness.json` identities), with **no test seams in production code**:

- same-token exactly-once consumption per harness; full walk to the final tokenless `run-stage`
- two real processes racing one token — exactly one winner, per harness (+ Copilot session-owned variant)
- missing/malformed/oversized/v1/legacy-v2/migrated-harness recovery — exactly one winner
- both fresh-`next`/`continue` commit orders (deterministic reset token bytes pinned) + a real waiter surviving a live lock hold
- crash boundaries: fabricated dead-owner lock (production reaper reclaim) and committed-but-stdout-failed (read-only fd → EBADF after the marker rename)
- `t248` inverts the old regression that pinned sessionless replay, now pinning single-use + `cursor_harness`

Evidence: full deterministic tiers (smoke+unit+integration) green — 340 files, 8292 assertions, 0 failures — plus `bun scripts/package.ts --check`, `bun run typecheck`, `bun run check`, and the t68 version/changelog pin. Validated independently twice (author run + conductor re-run). Deterministic e2e + live-harness slices to run as the pre-merge gate on this PR.

## Non-goals (per issue)

No change to Stop retention, Resume, ownership, compaction, conversation, or count semantics; no exactly-once `run-stage`/`report`/`park` execution claim; the pre-existing no-state-file markerless-continuation branch is documented as a residual limit, not changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.51] - 2026-08-21
+
+Steering continuation tokens are now single-use on every harness: Claude, Codex, Copilot, Cursor, Kiro CLI, Kiro IDE, and opencode share one engine-owned, crash-safe continuation cursor in the active-directive marker, closing the sessionless same-token replay carve-out accepted in PR #749 (issue #762). **Upgrade:** replace the whole `dist/<harness>/` shell in one quiescent swap (no AI-DLC command or hook running) - mixed old/new tool files are unsupported - then run a fresh `next`; a matching in-flight token instead migrates atomically. Rolling back to an older release restores its sessionless replay behavior until re-upgraded.
+
+* **Behavior change:** presenting the same continuation token twice now returns an error directive ("This continuation token is no longer current for this workflow. Run a fresh `next`; do not reuse an earlier token.") instead of repeating the successor. Concurrent uses of one token have exactly one winner on every harness; recovery is always a fresh `next`.
+* `continue` keeps all native token validation, then compares the SHA-256 of the complete presented token with the persisted current token inside the active-directive transaction and atomically publishes the exact successor - or a tokenless `run-stage` - before stdout. New marker publications record the installed harness as `cursor_harness` (marker v2, optional on read); pre-upgrade markers migrate on the first matching `continue` or fresh `next`.
+* Fresh `next` must now commit its first directive before emitting on every harness. Publication failure or contention yields "The directive could not be published, so no work directive was issued. Retry the command; if coordination remains busy, run `/aidlc --doctor`." - replacing the previous Copilot-only wording and the non-Copilot fail-open delivery.
+* Missing, malformed, oversized, v1, cross-harness, and pre-state bare-space markers permit one natively validated recovery continuation under the same transaction; a mismatched token is stale, not a bootstrap opportunity.
+* The exactly-one-winner guarantee requires the marker and lock paths on one local filesystem (atomic rename, exclusive directory creation, coherent cross-process visibility); NFS/SMB/FUSE-style shares are unsupported, and sudden power-loss durability is outside the claim. See the new "Atomic continuation cursor" section in the Developer Reference for the crash/retry matrix, migration, rollback, and per-harness residual limits.
+
 ## [2.6.50] - 2026-08-21
 
 Approval checkpoints now reject replies that do not exactly match a currently offered decision, while change-request feedback travels separately from the selected choice. **Upgrade:** refresh your `dist/<harness>/` shell so the gate protocol, deterministic transition guards, and harness instructions stay aligned.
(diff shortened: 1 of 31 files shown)


leandrodamascena (approved):
The remaining continuation-token replay path is resolved. Pre-state directives now use the bare-space cursor, preserving the exactly-one-winner guarantee before an intent or state exists.

### Verification

- Added a real concurrent pre-state regression: one successor and one stale-token result.
- Cursor, steering, and version tests: 51 passed.
- Package parity, typechecks, lint, coverage-registry validation, and `git diff --check`: passed.
- Rebased onto current `v2` and bumped the release to `2.6.51`.

Approved.

PR #814, opened 2026-08-20
docs: Kiro IDE hook payloads — correct the "no tool arguments" statements (2.6.53)

apackeer:
Kiro IDE tool-argument delivery varies across supported IDE generations: captured 0.12 and 1.0.165 payloads can carry empty inputs, while later 1.x builds may populate PreToolUse, delegation, shell/write, and PostToolUse inputs. This corrects the blanket "no tool arguments" wording in reviewer-scope, stage-rule-delivery, Kiro IDE orchestration guidance, and the hooks reference. Wiring and enforcement behavior are unchanged.

The empirical reference retains the measured captures and records issue #763’s reported Kiro IDE 1.0.309 behavior explicitly as reported rather than repository-measured evidence. Version 2.6.53. Upgrade by refreshing `dist/<harness>/` for the corrected generated comments and guidance. Closes #763.

PR #823, opened 2026-08-20
fix: require the reviewer receipt before presenting the approval gate (2.6.52)

### LINKED ISSUE #551: [Bug]: conductor agent misbehaving (version 2)
### Description

this could be a bug that it autocorrected:

> Before continuing, I need to flag something: I made a sequencing error on rough-mockups. Per protocol, the reviewer (Product Lead) is supposed to run before the approval gate, but I presented the gate and called report --result approved first — which the audit log confirms already completed the stage and advanced the phase to Inception (practices-discovery) — and only then ran the reviewer and fixed the findings it raised. The good news: the actual content fixes are legitimate and already applied to wireframes.md and user-flow.md. The gate itself already closed and the workflow has moved forward, so rewinding would mean jumping backward through Inception's start — a disruptive move I don't think is warranted for what were documentation clarity fixes, not scope or decision changes.

### Steps to Reproduce

I was running through the process on a greenfield app (explicit MVP)

### Expected Behavior

follow the regular process and do not need to autocorrect

### Actual Behavior

described above

### AI-DLC Version

v2 (alpha)

### Release / Commit

_No response_

### AI-DLC Phase

### Platform / IDE

Kiro IDE

### AI Model

Sonet 5

### Environment

macos

### Additional Context

_No response_

apackeer:
Reviewer-bearing approval gates now prove that the configured reviewer inspected the current artifact before the gate is presented or re-presented. Gate opening and revision re-entry enforce fresh reviewer receipts; rejection from an active stage records Active to Revising without fabricating an approval gate.

The receipt lifecycle is bound end to end: REVIEW_REQUESTED fingerprints the artifact bytes, REVIEW_COMPLETED requires the same fingerprint, unmatched iterations serialize, and only --retry-pending may repeat the exact outstanding ordinal. Stage-level and per-unit receipts remain separate; per-unit reviews require the authoritative Unit DAG or an exact active Bolt match for legacy no-DAG swarms. Existing open gates are revalidated through artifact, summary, pipeline, and reviewer guards.

The branch is integrated with current v2 behavior, including the engine-owned single-use continuation cursor, exact checkpoint choices, pipeline receipts, workspace source freshness, and loop-back replay. Version 2.6.52. Upgrade by refreshing `dist/<harness>/`; existing workflow state requires no migration. Closes #551.


leandrodamascena (changes requested):
The direction is correct, but the new gate precondition is not yet complete. I found multiple paths that still present or re-present a reviewer-bearing gate without proving that the configured reviewer inspected the current artifact.

1. **P1: rejection recovery still records a pre-review gate** (`core/tools/aidlc-orchestrate.ts:5642-5659`, `core/tools/aidlc-state.ts:2808-2812,2843-2847,2905-2915`).

   `report --result rejected` accepts an `in-progress` stage. `handleReject` then backfills `STAGE_AWAITING_APPROVAL` before `GATE_REJECTED` without calling `verifyReviewerPrecondition()`.

   A conductor can therefore skip the required `awaiting-approval` report, present the human gate, receive “Request Changes,” and record that gate before any review. This contradicts the changelog claim that reviewer-bearing stages can no longer present a gate before review.

   Reject recovery from `in-progress` must verify the reviewer receipt before emitting the recovered gate, or it must record rejection without claiming that an approval gate had been validly opened.

2. **P1: the approve-time revision backstop reopens the gate before re-review** (`core/tools/aidlc-state.ts:2663-2719`, `tests/unit/t205-gate-revision-backstop.test.ts:329-357`).

   When an artifact is revised at an open gate without a recorded rejection, the backstop emits `GATE_REJECTED`, `STAGE_REVISING`, and a recovered `STAGE_AWAITING_APPROVAL`, then persists that state before checking whether the revision invalidated the receipt.

   The receipt check correctly refuses completion afterward, but the stage is already persisted as `[?]`. The changed test explicitly records the recovery review while the gate remains open.

   Persist `[R]` after the recovered rejection, require the new review, and only then emit `STAGE_AWAITING_APPROVAL`.

3. **P1: a fake per-unit receipt can satisfy the no-DAG stage-level fallback** (`core/tools/aidlc-log.ts:831-843`, `core/tools/aidlc-lib.ts:4870-4906,5198-5207`, `core/tools/aidlc-state.ts:1833-1866`).

   `aidlc-log review --unit ghost` is accepted whenever the stage is per-unit, even when no DAG exists. `freshReviewReceipts()` assigns every terminal unit receipt to both `stageVerdict` and `unitVerdicts`. When `resolveBoltDag()` returns `none`, the gate guard accepts `stageVerdict` as the stage-level fallback.

   Keep stage and unit receipts separate. Only a receipt without `Unit` may satisfy the stage-level fallback, and reject `--unit` when the unit is not in the authoritative DAG.

4. **P1: writes between review dispatch and verdict are certified as reviewed** (`core/tools/aidlc-log.ts:1019-1035`, `core/tools/aidlc-lib.ts:5073-5197`).

   `REVIEW_REQUESTED` records no artifact fingerprint. The fingerprint is taken only when `REVIEW_COMPLETED` is logged. If the artifact changes while the reviewer is running, the old reviewer result is recorded with a fingerprint of the new bytes, and `gate-start` accepts content the reviewer never saw.

   Bind the request to an artifact fingerprint and require the same fingerprint when recording the verdict, or reject any completion after a matching write.

5. **P1: multiple normal review iterations may be outstanding concurrently** (`core/tools/aidlc-log.ts:875-988`).

   Iteration sequencing uses total request count but does not reject a new normal request while `pendingIterations` is non-empty. Iterations 1 and 2 can be requested before either completes; iteration 1 can then return NOT-READY and the already-dispatched iteration 2 can return READY without observing the findings or repair from iteration 1.

   Reject a normal request whenever an unmatched request exists. Only `--retry-pending` should be able to repeat that exact ordinal.

6. **P2: an already-open gate is not revalidated** (`core/tools/aidlc-orchestrate.ts:5630-5633`).

   `report --result awaiting-approval` returns success immediately when the checkbox is already `[?]`. Legacy gates and the recovered gate described above can therefore be re-presented without reviewer evidence. Revalidate the receipt before treating an existing gate as valid, or explicitly migrate it to a non-open state.

7. **Required checks are not green.**

   Integration/e2e fails because `tests/e2e/t113.test.ts:104-140` calls `gate-start` before recording the review. The fixture comment describes the new ordering but implements the old ordering. The branch is also conflicting with `v2`, so current CI does not represent a mergeable result.

8. **P2: version collision.**

   This PR and open PR #616 both claim `2.6.20`. Resolve the conflict-trap after rebasing, before merge.

`bun scripts/package.ts --check` is green and the focused reviewer-ordering tests pass, but those tests cover only the direct `gate-start`/`revise` happy paths. The recovery and receipt-lifecycle paths above still bypass the intended guarantee.

apackeer:
@leandrodamascena thanks for the deep pass. All eight findings were valid and are addressed on the rebased head ba41dd890 (2.6.23):

1. **Rejection recovery**: `report --result rejected` from an active stage no longer backfills a `STAGE_AWAITING_APPROVAL` row. The direct path records `GATE_REJECTED` + `STAGE_REVISING` and lands `[R]`; `STAGE_AWAITING_APPROVAL` is now only ever emitted by a guard-passed gate opening (`gate-start`, `revise`, or the backstop after its guards). Pinned by t263 ("rejecting directly from active records no approval gate") and t17 test 39.

2. **Approve-time revision backstop**: it persists `[R]` (with the recovered `GATE_REJECTED`/`STAGE_REVISING` pair) before running the summary and reviewer guards; the recovered `[?]` is emitted only when they pass, and re-entry otherwise goes through the guarded `revise` after a fresh review. t205 test 3b models the full ordering (revision -> `[R]` -> refused approve -> fresh review -> `revise` -> `[?]` -> approve).

3. **Receipt separation**: `freshReviewReceipts` keeps Unit-bearing receipts out of `stageVerdict`; the no-DAG fallback accepts only a Unit-less receipt; `review --unit` refuses a unit outside the authoritative DAG and refuses outright when no DAG exists. Pinned by t115 R19 (a forged ghost-unit receipt no longer opens the gate) and t271, plus a new t208 pin for the reverse direction (a stage-level receipt cannot satisfy per-unit requirements when a DAG exists).

4. **Fingerprint binding**: `REVIEW_REQUESTED` records the artifact fingerprint at dispatch; `REVIEW_COMPLETED` requires request == recorded == current bytes, so a mid-review write refuses the verdict and directs to `--retry-pending`, which re-stamps the same ordinal against current bytes as a fresh dispatch. The gate scan applies the same three-way match, so legacy fingerprint-less receipts fail closed and need a fresh review. Pinned by t271 and t115 R2/R12/R15.

5. **Iteration serialization**: a normal request is refused while any request in the same (stage, reviewer, unit, workflow) scope is unmatched; only `--retry-pending` repeats that exact ordinal, consuming no budget. The old t271 case that asserted concurrent requests succeed now asserts the refusal.

6. **Open-gate revalidation**: `report --result awaiting-approval` on an already-`[?]` stage falls through to `gate-start`, which re-runs the artifact, summary, and reviewer guards against current bytes without emitting a duplicate event; a legacy `[?]` with no receipt is refused with guidance to run the reviewer first, and a fresh review at `[?]` recovers it. Pinned by t263.

7. **CI**: the branch is rebased onto v2 (mergeable again) and the t113 fixture now records the review before `gate-start`. The full touched-file slice is green locally (24 files, 522 assertions, 0 failed), plus `package.ts --check` and typecheck.

8. **Version**: re-slotted to 2.6.23, the lowest patch slot no open v2 PR claims in its version file (2.6.20 is held by #616 and #813, 2.6.21 by #747 and #824, 2.6.22 by #814, 2.6.24 by #821).

Known residuals recorded during the review round, deliberately left for follow-up issues rather than this PR: `--retry-pending` rebinds a pending iteration to new bytes without tool-level proof the reviewer re-ran (the once-only rerun rule stays protocol prose); `AIDLC_SKIP_REVIEWER_GATE_GUARD=1` bypasses the gate-opening half at runtime (test-only by convention; completion paths stay unconditional); `next`/`resume` do not revalidate an existing `[?]` at presentation time (enforcement sits on `report` and the commit paths).

Title and body are updated for the expanded scope. Ready for another look.

leandrodamascena:
Fixed in `bb3b775d8`.

The regression came from requiring authoritative DAG membership for every per-unit review, which broke legacy no-DAG swarm execution.

The fix now:

- Keeps DAG membership strict when a DAG exists.
- Allows a no-DAG per-unit review only with an exact, active Bolt match.
- Rejects absent, mismatched, completed, or failed Bolts.
- Preserves the active Bolt across stage/rejection accounting resets.
- Adds regression coverage and regenerates all harness distributions.

The previously failing integration/e2e CI job is now passing.

leandrodamascena (approved):

leandrodamascena (approved):
Re-reviewed after integrating current v2. Reviewer request/verdict fingerprint binding, iteration serialization, stage/unit receipt separation, active-Bolt fallback, gate revalidation, and revision recovery remain intact. The merge also preserves exact checkpoint choices, pipeline receipts, workspace source freshness, and loop-back replay. The focused PR slice, cross-PR regression suites, package parity, typecheck, lint, and coverage registry checks pass locally. Release metadata is updated to 2.6.51.

leandrodamascena (approved):
Re-reviewed after integrating v2 2.6.51 and its engine-owned continuation cursor. Reviewer receipt lifecycle behavior remains intact, version metadata is now 2.6.52, and the continuation, reviewer-gate, source-freshness, swarm, and loop-back regression suites pass together. Package parity, typecheck, lint, and coverage registry checks are green locally.

PR #824, opened 2026-08-20
fix(kiro-ide): ship only IDE-native agent and settings surfaces (2.6.64) (#555)

### LINKED ISSUE #555: [Bug]: The Kiro IDE harness (`harness/kiro-ide/`) ships CLI-format surfaces — JSON agents, `settings/cli.json`, and legacy `.kiro.hook` files — that Kiro IDE 1.0 does not read (agents/settings) or flags for migration (hooks)

apackeer:
Drops the Kiro CLI agent-v1 JSONs and settings/cli.json from the Kiro IDE distribution, ships the conductor as an IDE-native agents/aidlc.md plus tools/permissions.rules grants on all 14 delegation personas, and teaches doctor, loadAgents, and the plugin composer the IDE agent shape (with strict dispatch-grant validation per the PR #653 review). Ports the reviewed #653 direction onto current v2, preserving the #658 steering include and the #614 hook coexistence; the permissions.rules shape follows the schema live-validated on IDE 1.0.165/0.12.333 in the issue thread. Closes #555


leandrodamascena (changes requested):
I reviewed commit `e7f5dd16f703357e60d414d36ab271ae68e4438a` against the current `v2` branch and ran the focused Kiro packaging, plugin-compose, smoke, and unit tests. The existing head is internally green, but the following issues remain:

P1: Preserve the Kiro IDE-specific plugin wiring from #817. `harness/kiro-ide/manifest.ts` still declares `kind: "kiro"`, which emits the legacy `.kiro.hook` registration that Kiro IDE 1.x no longer executes. Rebase onto current `v2` and retain `kind: "kiro-ide"`, the v1 `SessionStart` registration, and the cross-platform launcher. Otherwise installed plugins will not compose automatically on the supported IDE runtime.

P1: Remove the PR-specific `frontmatterRemovals` mechanism when rebasing. Current `v2` already removes `disallowedTools: Task` from every Kiro Markdown projection in `projectTierFrontmatter` through #821. If this PR second removal pass is retained, packaging will fail for all 14 personas because `applyFrontmatterRemovals` treats the already-removed key as an error. Use the shared Kiro projection and avoid maintaining two competing removal paths.

P2: Do not require `capability` to be the first field in a permission rule. `validIdePermissionRule()` only recognizes rules beginning with `- capability:`, so a semantically equivalent YAML mapping beginning with `- effect: allow` is rejected even when `capability` and `match` follow. YAML mapping order is not significant. Parse the rule structurally or accept the fields in any order, and add a reordered-rule test.

P2: Update the Kiro IDE topology protocol. The Kiro IDE section of `core/aidlc-common/protocols/stage-protocol-ensemble.md` still says every delegation target requires an agent config and a `trustedAgents` entry in the conductor, and that all personas ship both. This directly contradicts the Markdown-only surface introduced here and leaves the runtime-loaded protocol instructing the conductor to expect files that no longer exist.

The rebase must also preserve the other current-base changes and update the changelog, framework version, README badge, and PR title from `2.6.21` to the next available version after `2.6.61`.

apackeer:
@leandrodamascena thanks for the review - all five items are addressed on the new head `3af1fd76c`, a single commit rebased onto current `v2` (`cd107e09a`, 2.6.62).

**P1 - Kiro IDE plugin wiring from #817**: preserved through the rebase. `harness/kiro-ide/manifest.ts` declares `plugin: { manifestDir: ".kiro-plugin", kind: "kiro-ide" }`, and the emitted kiro-ide plugin projection (v1 `SessionStart` registration plus the cross-platform launcher) is byte-identical to the current base output; the only change under `dist/plugins/test-pro/kiro-ide/` is the shared compose hook change below.

**P1 - `frontmatterRemovals` removed**: the PR-specific contract and implementation are gone entirely (zero references remain). Both Kiro distributions rely on the shared `projectTierFrontmatter()` projection from #821, so there is no second removal pass and packaging succeeds for all 14 personas (`bun scripts/package.ts --check` green).

**P2 - permission-rule field order**: `validIdePermissionRule()` in `scripts/plugin-hooks-template/compose.ts` now parses each rule as a mapping and accepts `capability`/`effect`/`match` in any order. t188 gains an accepted "reordered-rule" variant that leads with `effect:` and puts `capability:` last, alongside the fail-closed cases (empty tools, empty permissions, empty rules, partial entry).

**P2 - ensemble protocol**: the Kiro IDE statement in `core/aidlc-common/protocols/stage-protocol-ensemble.md` now says Kiro IDE resolves all 14 personas directly from `agents/aidlc-*-agent.md`, each carrying a non-empty `tools:` grant and capability-scoped `permissions.rules`, with no agent-v1 JSON or conductor `trustedAgents` entry.

**Version**: re-slotted to **2.6.64** rather than 2.6.62; #716 merged as 2.6.62 while this fix round was in flight, and 2.6.63 is spoken for by another in-flight PR. `aidlc-version.ts`, the CHANGELOG heading, the README badge, and the regenerated dists all agree (t68 green), and the PR title now matches.

Tests on the new head: the smoke+unit+integration slice covering t68, t148, t188, t220, t223, and t252 passes clean (t188 83/83 including the reordered-rule case); `bun scripts/package.ts --check` and `bun run check` are green. Re-requesting your review.

leandrodamascena (approved):
I re-reviewed commit `3af1fd76c` against the current `v2` branch. All findings from my previous review have been addressed:

- Kiro IDE-specific plugin wiring from #817 is preserved.
- The conflicting `frontmatterRemovals` mechanism has been removed.
- Permission rules now accept YAML fields in any order, with regression coverage.
- The ensemble protocol correctly describes the Markdown-only Kiro IDE agent surface.
- Release metadata is consistently updated to `2.6.64`.

The focused suite passed with 196 assertions. `bun scripts/package.ts --check`, typecheck, and lint also passed. I found no remaining blocker.

Approved.

PR #825, opened 2026-08-20
docs: explain the available workflow profiles

leandrodamascena:
## Summary

Adds a customer-facing Workflow Profiles guide that explains Classic, Express, and every other shipped workflow choice, while clarifying that the engine models a workflow profile as a scope.

## Changes

- add a quick chooser and practical guidance for all 11 core workflow profiles
- explain when to use and avoid Classic, Express, Feature, Enterprise, MVP, PoC, Bugfix, Refactor, Infrastructure, Security Patch, and Workshop
- make the page visible from the documentation home, Getting Started navigation, Introduction, First Workflow, and Scopes reference
- complete the explicit CLI scope list with Classic, Workshop, and Express
- add a glossary entry for workflow profile
- derive a documentation parity test from `core/scopes/` so every shipped profile remains documented and invocable

## Validation

- `bun test tests/unit/t239-documentation-parity.test.ts` (10 tests, 169 assertions)
- `bun scripts/docs-rewrite-links.ts && uv run zensical build --strict`
- `bun run check` (typecheck clean; only pre-existing Biome informational suggestions in `t267`)
- `git diff --check`

This is documentation and test coverage only. It does not change runtime behavior, version metadata, the changelog, or generated distributions.


apackeer:
Reviewed at head 074d0854 against v2, and re-verified after v2 advanced to 840ba653: everything below still holds on the current base, and the merge is clean.

Most of the page checks out against the shipped engine, so first the good news: all 11 profile names match `core/scopes/`; every stage count in the quick chooser matches the compiled `scope-grid.json` (bugfix 7/33 through enterprise 33/33, recomputed from the grid rather than taken from other docs); the Depth and Test-strategy columns match scope frontmatter, including workshop's `testStrategy: Minimal` override; and the classic/express prose matches the authored scope files. The CLI scope-list completion is a genuine fix (the section said "11 named scopes" but listed 8) and the new t239 block pins it. `bun scripts/package.ts --check`, typecheck, and `git diff --check` are clean, and t239 passes with 10 tests / 169 assertions, matching the PR body.

Three asks before merge, two of them one-line wording fixes:

1. **P2** - `docs/guide/workflow-profiles.md:16-17`: "You always see the proposed route before the workflow starts" is not true for explicit invocation. An explicitly named scope births directly with no confirm ask, by design (`createPrintDirective` in `core/tools/aidlc-orchestrate.ts`: "an explicitly named scope births directly (no confirm ask by design), so the stage/gate counts ride here"). Only the keyword-match and compose flows confirm before starting. Suggested wording: "A keyword match or compose offer confirms the route before starting; an explicitly named profile starts immediately and prints its stage and gate counts."

2. **P2** - `docs/guide/workflow-profiles.md:181` (Related controls): "**Review class** with `--review adversarial|advisory|none`" reads as a symmetric knob. The engine's `resolveReviewClass` (`core/tools/aidlc-lib.ts`) is min() everywhere: an override or cap can only lower a stage's declared class, never raise it, so `--review adversarial` on Classic (cap advisory) or Express (cap none) does nothing. `docs/guide/12-cli-commands.md` already words this correctly as "Cap stage reviews for this run". Suggest renaming the bullet to "Review cap" with a lower-only note.

3. **P2** - the quick-chooser table hand-copies stage counts, depth, and test strategy, making it the third unguarded copy of these numbers in the User Guide (only the 05 matrix cells are guarded, by t244). The new t239 test checks profile names and `/aidlc <scope>` strings only, so a future grid change updates the guarded matrix but silently strands this table. Since this PR already adds the parity test, extending it to derive the counts from the compiled `scope-grid.json` and the two columns from `core/scopes/aidlc-<name>.md` frontmatter would close the gap (t244 has the loader pattern). The table survived the last 23 commits on v2 unchanged, but that was luck: nothing changed the grid, and nothing guards the table.

Two non-blocking notes:

4. **NOTE** - "workflow profile" appears nowhere in the product surface (zero grep hits in `core/` and `harness/`); every runtime message, the `--scope` flag, and the `Scope` state field say "scope". The glossary entry calls it "the user-facing name for a scope", but the only surface using the term is this page, so readers will meet "profile" in the guide and "scope" everywhere in the tool. Either plan to adopt the term in product output or soften the glossary to "the User Guide's name for a scope".

5. **NOTE** - the page overlaps the per-scope "Use when" sections in `docs/guide/05-scopes-and-depth.md` nearly 1:1. The differentiators (the do-not-choose-when guidance, the profile/scope bridge, the Getting Started placement) are real, but per-scope guidance now lives in three prose places plus the matrix. Consider trimming 05's per-scope prose to defer here, so each scope keeps one narrative home.

leandrodamascena:
@apackeer, I rebased the PR onto the latest `v2` and addressed the feedback in `0e0320f73`:

1. Corrected the startup guidance: keyword matches and compose offers require confirmation, while an explicitly named profile starts immediately and prints its stage and gate counts.
2. Renamed “Review class” to “Review cap” and documented the lower-only resolution across the stage declaration, profile cap, and per-run cap.
3. Extended `t239` so the Quick chooser is checked against the compiled `scope-grid.json` for stage counts and scope frontmatter for depth and effective test strategy.
4. Clarified that “workflow profile” is the User Guide’s term for the runtime’s scope concept.

I left the broader consolidation with `05-scopes-and-depth.md` as a separate follow-up because it was non-blocking and would expand this PR’s scope.

Validation completed:
- `t239-documentation-parity.test.ts`: 10/10
- strict Zensical documentation build
- `bun run check`
- `git diff --check`

Could you please re-review the updated head?

apackeer (approved):
**Approved.** Reviewed the new head 0e0320f73 against v2: this round rebased the PR and resolved all four areas raised in [the prior review](https://github.com/awslabs/aidlc-workflows/pull/825#issuecomment-5384579714). Two independent verification passes agree on the result. I verified the updated diff itself rather than relying on the reply:

- Explicit profile invocation is now correctly distinguished from keyword-match and compose confirmation (`docs/guide/workflow-profiles.md:15`); this matches the direct-birth implementation (`core/tools/aidlc-orchestrate.ts:898`).
- `--review` is now documented as a lower-only cap, matching the engine's low-wins resolution (`docs/guide/workflow-profiles.md:185`; `resolveReviewClass` in `core/tools/aidlc-lib.ts:8715`).
- `t239` now derives the quick-chooser stage counts from the compiled scope grid and the Depth/Test-strategy columns from scope frontmatter (`tests/unit/t239-documentation-parity.test.ts:441`), and its `testStrategy ?? depth` fallback mirrors the runtime's own default. A grid or frontmatter change now fails the test instead of stranding the table.
- The glossary now makes clear that "Workflow Profile" is the User Guide's name for the runtime's scope concept (`docs/guide/glossary.md:71`), avoiding any claim that the runtime surface uses that term.

Targeted verification passed twice: two independent runs of the t239 unit slice each report 10 tests, 237 expect() calls, 0 failures.

I also checked the current `v2` head: scope frontmatter and the compiled scope grid have not changed since this PR's merge base, v2's own edits to the three shared files sit in disjoint hunks, and GitHub reports the PR mergeable, so the page and the new test stay valid post-merge.

The consolidation note from the prior review remains valid but non-blocking: `05-scopes-and-depth.md` still duplicates some per-scope selection prose, but the docs now establish a defensible split between the new chooser-oriented page and the detailed routing reference, and the chooser values are machine-guarded. That cleanup can reasonably stay a follow-up rather than expanding this PR.

No new findings. Thanks for the thorough turnaround, Leandro.

PR #826, opened 2026-08-20
fix: parse flow-style YAML sequences in listField so inline scope keywords match (2.6.49)

### LINKED ISSUE #805: [Bug]: A flow-style YAML sequence in listField returns an empty list, so a scope with inline keywords is never matched
## What happens

`listField` (`core/tools/aidlc-lib.ts:8406`) only matches block sequences. A flow sequence is valid YAML, parses without error, and returns an empty list:

```yaml
keywords: [design review, architecture, ddd]   # → []
```

```yaml
keywords:                                      # → the three items
  - design review
  - architecture
  - ddd
```

The regex requires a newline after the key and lines beginning with `-`, so a flow sequence simply never matches and the function returns `[]` from the no-match branch.

## Expected vs actual

**Expected:** either both YAML sequence styles are read, or a flow sequence is rejected with a message pointing at the file and key.

**Actual:** silently empty. Nothing fails.

## Why it matters

For scope manifests the consequence is invisible in every way a person would check. A scope declaring `keywords` in flow style **composes, compiles, and appears in the graph** — and is then never matched by keyword. The author sees a healthy install and a scope that "doesn't work sometimes", with no diagnostic anywhere.

We lost time to exactly this: our plugin's scope was written with flow-style keywords (it reads naturally, and it is valid YAML), and the scope was simply never selected. There is no failure to search for, which is what makes it expensive.

## Reproduce

1. Author a scope manifest with `keywords: [a, b, c]` on one line.
2. Compose and compile — both succeed, no warning.
3. Describe work matching one of those keywords — the scope is never detected.
4. Rewrite the same keywords as a block sequence — detection works.

## Suggested direction

Reading both styles would be the friendliest fix. If flow style is deliberately out of scope, a rejection at manifest-validation time would be enough — the cost here is not the restriction, it is that the restriction is silent.

## Environment

macOS 15, Claude Code harness, bun 1.3.14, engine v2 at `3d02c500`, re-verified at the tip of `v2` (`fbb1460c`).

apackeer:
`listField` previously matched only block sequences, so valid flow-style fields such as `keywords: [a, b, c]` silently parsed as empty and prevented keyword-based scope detection. The parser now falls back to a single-line flow-sequence scanner while keeping block style preferred. The scanner respects quoted commas and brackets, accepts trailing YAML comments, and fails closed on malformed quoting. Unit coverage lives in t309, with an end-to-end dropped-scope `detect-scope` regression in t125. Closes #805.


leandrodamascena (approved):
Rebased onto the current v2 branch and reviewed the parser at its shared frontmatter boundary. Block sequences retain precedence; flow sequences now handle quoted commas and brackets, trailing YAML comments, empty lists, and malformed quote rejection without adding a YAML dependency. The dropped-scope regression verifies keyword routing end to end, and the direct coverage moved to t309. Release metadata is refreshed to 2.6.49. Focused tests, coverage registry, package parity, typecheck, and lint pass locally.

PR #827, opened 2026-08-20
fix: plugin sync fails loudly when configured plugin roots carry no compose hook (2.6.78)

### LINKED ISSUE #806: [Bug]: plugin-sync exits 0 with "nothing to sync" when installed plugins carry no compose hook
## What happens

`plugin-sync` reports success when it did nothing. If no installed plugin has a `hooks/compose.ts`, the handler prints and returns with exit code 0 (`core/tools/aidlc-utility.ts:981`):

```
no installed plugins; nothing to sync
```

## Expected vs actual

**Expected:** a caller can distinguish "synced everything" from "found nothing to sync". Today both are exit 0 with a friendly message.

**Actual:** a no-op announces itself as success, and any script gating on the exit code proceeds as if the plugins had been synced.

## Why it matters

The message is accurate for the case it was written for — no plugins installed at all. But it also covers the case where plugins **are** installed and none carries a compose hook, which is a setup problem rather than an empty set. In that state the command is the only thing that would tell you, and it says everything is fine.

We hit this while wiring plugin installation into an automated upgrade path: the step passed, nothing had been composed, and the failure only surfaced later when the plugin's stages were absent from the graph.

## Suggested direction

Whichever fits the CLI's conventions:

- distinct exit code when the plugin set is non-empty but no compose hook was found; or
- keep exit 0 and make the message name which installed plugins were skipped and why.

The second alone would have been enough for us — the problem was not knowing which of the two situations we were in.

## Reproduce

1. Install a plugin whose tree has no `hooks/compose.ts`.
2. Run `plugin-sync`.
3. It prints `no installed plugins; nothing to sync` and exits 0, while a plugin is in fact installed.

## Environment

macOS 15, Claude Code harness, bun 1.3.14, engine v2 at `3d02c500`, re-verified at the tip of `v2` (`fbb1460c`).

apackeer:
aidlc plugin sync previously printed "no installed plugins; nothing to sync" and exited 0 even when plugin-root env vars were configured but no root carried hooks/compose.ts, so automation could not tell a clean no-op from a broken plugin install. Now the no-roots case keeps the documented exit-0 no-op, an all-unusable root set exits 1 naming each root and reason (missing hooks/compose.ts, or a root directory that does not exist), and mixed sets warn per skipped root on stderr while composing the valid ones. Docs, unit tests, and the 2.6.78 changelog entry (flagged as a behavior change) are included. Closes #806


leandrodamascena (approved):
Thanks, Arden. I reviewed the current head `b78c858bd` against `v2` at `2dad181b6`.

This change aligns with the project direction and the contract confirmed in #806: preserve the exit-0 no-op when no plugin roots are configured, fail loudly when every configured root is unusable, and warn while continuing when valid and invalid roots are mixed.

I rebased the PR, updated the release to `2.6.78`, regenerated all harness distributions, and added regression coverage ensuring every unusable root is named and correctly classified.

Validation covered all no-root, invalid-root, mixed-root, and valid-root paths, plus version synchronization, packaging parity, coverage registry, typecheck, lint, smoke, unit, integration, and deterministic e2e checks. The final commit only corrects commit authorship and is byte-identical to the fully green validated tree.

PR #829, opened 2026-08-20
fix: compile rejects duplicate producers for a consumed artifact (2.6.65)

### LINKED ISSUE #804: [Bug]: Two stages declaring the same artifact compile clean, and consumers resolve to an arbitrary producer
## What happens

Two stages can declare the same artifact in `produces:`. The graph compiles clean, and every consumer of that artifact silently resolves to whichever producer comes first in graph load order.

`producersOf` returns a list (`core/tools/aidlc-graph.ts:817`), and four call sites take `[0]`:

- `core/tools/aidlc-orchestrate.ts:1539` and `:1589`
- `core/tools/aidlc-sensor.ts:297` and `:312`

The comment above the first one states the assumption directly (`core/tools/aidlc-orchestrate.ts:1527`):

> `the verified graph invariant is exactly one producer per artifact (a clean 1:1 map), so producersOf(name)[0] is the owner`

Nothing verifies that invariant.

## Expected vs actual

**Expected:** compile rejects two stages declaring the same artifact, the way it already rejects two stages with the same slug (`aidlc-graph.ts:1753`) and two sensor manifests with the same id (`:747`) — both with errors naming the two files.

**Actual:** compiles clean. Directive resolution and sensor producer lookup pick an arbitrary producer, and nothing is reported.

## Why this one is worth a guard

The two existing duplicate checks show the project already treats duplicate declarations as errors worth naming precisely. This is the same class, but the failure mode is worse: a duplicate slug is a loud refusal at compile time, while a duplicate artifact is a wrong path chosen quietly at runtime. A stage reads the artifact from the wrong producer's record directory and behaves as if it read the right one.

## Reproduce

1. In any stage graph, add a second stage whose `produces:` includes an artifact already produced by another stage.
2. Compile the graph — it succeeds.
3. Have a third stage `consumes:` that artifact.
4. The resolved path follows load order, not declaration intent, with no diagnostic.

## Context

We consume v2 as a pinned engine (currently `3d02c500`) and hit this while authoring a plugin: two of our stages ended up declaring the same artifact. We expected a compile error; we got a clean build and had to add our own test to reject it. Verified still present at the tip of `v2` (`fbb1460c`).

## A note on the target branch

`CONTRIBUTING.md` says to work against the latest `main`, but `core/tools/aidlc-graph.ts` does not exist on `main`, and every recent feature and fix PR merges into `v2`. If you'd like a PR for this, could you confirm which branch it should target?

We have a patch following the shape of the two existing duplicate checks. Happy to open a PR once you've decided whether this should be a compile error or a warning — that call is yours.

## Environment

macOS 15, Claude Code harness, bun 1.3.14, engine v2 at `3d02c500` and re-verified at `fbb1460c`.

apackeer:
Two stages could declare the same artifact in produces/optional_produces and the graph compiled clean, leaving every consumer to resolve to whichever producer came first in load order via producersOf(name)[0]. aidlc-graph compile now fails loud — naming every producing stage file and one consumer — whenever an artifact with two or more producers is consumed by any stage, while unconsumed shared names (the shipped 8-producer traceability pattern) remain valid and are pinned by a new test. Ships as 2.6.65 with regenerated dist trees, E10a/E10b compile tests, and updated collision-policy docs. Closes #804

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.65] - 2026-08-23
+
+Stage graph compilation now rejects ambiguous producers before consumers can resolve an artifact by file order. **Upgrade:** refresh your `dist/<harness>/` shell; existing custom stages or plugins that declare the same consumed artifact in `produces` or `optional_produces` must rename one output or update the consumer.
+
+* `aidlc-graph compile` now reports `Duplicate producers for consumed artifact`, naming every producing stage file and slug plus one consuming stage.
+* Shared artifact names remain valid when no stage consumes them, including the shipped multi-stage `traceability` output pattern.
+
 ## [2.6.64] - 2026-08-23
 
 The Kiro IDE distribution now ships only IDE-native agent and permission surfaces. **Upgrade:** re-copy `dist/kiro-ide/` into your project, then remove any `.kiro/agents/aidlc.json`, `.kiro/agents/aidlc-*-agent.json`, and `.kiro/settings/cli.json` files left by an older overlay copy.
@@ -8,13 +15,6 @@ The Kiro IDE distribution now ships only IDE-native agent and permission surface
 * Kiro IDE now ships the conductor as `.kiro/agents/aidlc.md` and all 14 personas as Markdown with native `tools:` grants and `permissions.rules`; CLI agent-v1 JSON and `settings/cli.json` files are no longer included.
 * `/aidlc --doctor` accepts either the Kiro CLI JSON conductor or Kiro IDE Markdown conductor and checks `settings/cli.json` only for CLI-shaped installs.
 * Plugin composition validates Kiro IDE dispatch targets by their installed Markdown capability grants, rejecting empty or malformed permissions with actionable remediation while leaving Kiro CLI trust checks unchanged.
-
-## [2.6.62] - 2026-08-23
-
-Stage graph compilation now rejects ambiguous producers before consumers can resolve an artifact by file order. **Upgrade:** refresh your `dist/<harness>/` shell; existing custom stages or plugins that declare the same consumed artifact in `produces` or `optional_produces` must rename one output or update the consumer.
-
-* `aidlc-graph compile` now reports `Duplicate producers for consumed artifact`, naming every producing stage file and slug plus one consuming stage.
-* Shared artifact names remain valid when no stage consumes them, including the shipped multi-stage `traceability` output pattern.
 
 ## [2.6.62] - 2026-08-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ The Kiro IDE distribution now ships only IDE-native agent and permission surface
 
 ## [2.6.62] - 2026-08-23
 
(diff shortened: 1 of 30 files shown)


apackeer:
@leandrodamascena could you review this one?

Freshly rebased onto `v2` @ 2.6.61 (re-bumped to 2.6.62 per the conflict-trap policy) — exactly 2 commits on top. Review focus:

- `core/tools/aidlc-graph.ts` — the new compile guard: hard-error when an artifact has ≥2 producers (across `produces`/`optional_produces`) **and** at least one consumer; unconsumed duplicates stay valid (the shipped multi-producer traceability pattern depends on that).
- `core/tools/aidlc-orchestrate.ts` (~line 1551) — the one non-trivial rebase conflict: upstream's multi-repo CodeKB path expansion was preserved alongside the guard's uniqueness comment. Worth a close look.
- `tests/integration/t-custom-harness-compile.test.ts` — 18 cases incl. consumed-duplicate rejection and unconsumed-duplicate acceptance (all green deterministically).

Everything else is version artifacts + packager-regenerated `dist/`. Fixes #804 — the reporter re-verified the defect still reproduces on current v2, so this is the awaited fix.

leandrodamascena (changes requested):
The implementation and rebase conflict resolution look sound, but two items remain before approval:

1. The new regression coverage exercises only a `produces`/`produces` collision. Please add a consumed collision where at least one producer declares the artifact through `optional_produces`. The implementation explicitly unions both namespaces, and the changelog/documentation promise that behavior, but removing the `optional_produces` branch would currently leave every new test green.

2. Please update the PR title and description from `2.6.31` to the rebased release version, `2.6.62`.

I verified the duplicate-producer guard, unconsumed shared-name exception, path resolution, version synchronization, and the multi-repo CodeKB conflict resolution. CI is green, and 153 targeted local tests passed.

apackeer:
@leandrodamascena both items are addressed, and the branch is freshly rebased onto the current v2 tip:

1. New E10b integration test: a consumed collision where the duplicate producer declares the artifact only through `optional_produces` (`produces: []`). Removing the `optional_produces` branch from the guard's producer union turns E10b red. The unconsumed-duplicate acceptance case you verified is retained as E10c.
2. On the version: the slot moved twice while this round was in flight (#716 merged as 2.6.62, then #824 as 2.6.64), so this PR now ships as **2.6.65** rather than the 2.6.62 you reviewed. `core/tools/aidlc-version.ts`, the README badge, the CHANGELOG heading, and the generated dist copies all agree (t68 green); I'll refresh the title and description to (2.6.65) to match.

Rebased onto v2 @ 62999bc39: this PR's CHANGELOG entry sits above [2.6.64], dist regeneration is byte-identical (`bun scripts/package.ts --check` passes), and the targeted post-rebase run is green (t68 7/7, t-custom-harness-compile 19/19 including the new E10b). Ready for another look.

leandrodamascena (approved):
I re-reviewed the updated head and applied two small follow-up corrections in commit `ca657a709`:

- Preserved the current Kiro IDE Markdown-only documentation instead of reverting it to the removed agent JSON and `settings/cli.json` surfaces.
- Aligned the PR title and description with the shipped `2.6.65` version.

The requested `optional_produces` regression coverage is present and passes. The focused graph suite passed with 118 assertions, the follow-up documentation/version suite passed with 16 assertions, and `bun scripts/package.ts --check`, typecheck, and lint completed successfully.

PR #832, opened 2026-08-20
fix: gate SUBAGENT_COMPLETED logging on running workflow state (2.6.54)

### LINKED ISSUE #710: [Bug]: `aidlc-log-subagent` gates on the audit shard existing rather than on workflow state — once a clone's shard is committed, every subagent completion in every session appends to the ledger, framework or not
# AI-DLC Version
v2 (`v2` branch). Observed on 2.3.3 (`83ed7a81`), verified still present at the current `v2` tip (`8b20390b`): `core/hooks/aidlc-log-subagent.ts` — harness-neutral, so all harnesses.

# AI-DLC Phase
None — that is the bug. The hook fires in sessions that are not running any workflow at all.

---

## Summary

Every audit-emitting hook self-gates on **workflow state existing** — `aidlc-mint-presence.ts`, `aidlc-session-start.ts`, `aidlc-session-end.ts`, `aidlc-validate-state.ts` all begin with some form of `if (!existsSync(stateFilePath(projectDir))) return 0`. The rationale is written into `aidlc-mint-presence.ts` itself:

> No workflow state on disk means nothing to gate, so the hook exits without writing — otherwise every prompt in a project that carries the harness shell but never ran the framework would scaffold and grow audit shards.

`aidlc-log-subagent.ts` is the one exception. Its gate is whether **this clone's audit shard file already exists**:

```ts
const auditFile = auditFilePath(projectDir);
if (!existsSync(auditFile)) return 0;
```

The audit trail is committed by design, and the shard name embeds the per-clone token from `.aidlc-clone-id`, which is stable within a clone. So on any checkout where a workflow has ever run, the shard exists — and from that day on, **every SubagentStop in every session in that checkout appends a `SUBAGENT_COMPLETED` entry to the version-controlled ledger, framework or not, forever**. Completing or tearing down the workflow doesn't stop it: the state file goes away, the shard stays.

## Evidence (one real workspace, 2.3.3, Claude Code harness on macOS)

Our workspace ran `practices-discovery` and `reverse-engineering` in July; the runs committed this clone's shard with the real events. Three weeks later the shard held **52 entries, of which 37 were `SUBAGENT_COMPLETED` noise** from ordinary interactive sessions — PR chores, doc edits — none of them framework work. The entries look like this:

```
## Subagent Completed
**Timestamp**: 2026-08-03T08:01:51Z
**Event**: SUBAGENT_COMPLETED
**Agent Type**:
**Agent ID**: a389061a2b8d7bb64
**Message**: commit it
```

Note the payload: `Agent Type` is empty (these aren't AI-DLC personas), and `Message` carries the session's chat text. Two consequences beyond the git noise:

1. **The ledger's signal drowns.** 71% of this shard was non-framework entries; anything reading `audit/*.md` to reconstruct a run (replay, session-cost, outcomes pack) wades through them.
2. **Chat from sessions that never opted into the framework gets committed.** Interactive messages land in a version-controlled file as a side effect of having once run a workflow in that checkout.

## Proposed fix

Give the hook the same self-gate as its siblings — state, not shard:

```diff
 import {
-  auditFilePath,
   type ClaudeCodeHookInput,
   errorMessage,
   hooksHealthDir,
   isClaudeCodeHookInput,
   isoTimestamp,
   recordHookDrop,
   resolveProjectDirFromHook,
+  stateFilePath,
 } from "../tools/aidlc-lib.ts";
@@
-const auditFile = auditFilePath(projectDir);
-if (!existsSync(auditFile)) return 0;
+// Gate on workflow STATE, same self-gate as every other hook: no
+// aidlc-state.md on disk means no workflow is running, so a subagent
+// completing belongs to ordinary harness use, not to the ledger.
+if (!existsSync(stateFilePath(projectDir))) return 0;
```

Behaviour change is confined to the broken case: during a workflow, state exists before subagents are dispatched, so legitimate `SUBAGENT_COMPLETED` logging is unchanged (and `appendAuditEntry` creates the shard if missing, so the old gate wasn't needed for file-existence either). Sessions with no workflow state stop appending.

We've applied exactly this as a downstream patch on 2.3.3 and verified both paths by piping a synthetic SubagentStop payload through the hook: with no state file the shard is byte-identical afterwards; with a state file present the event appends as before.

Happy to send the PR (edit in `core/hooks/`, regenerate `dist/`, changelog + version bump per CONTRIBUTING) if the approach looks right to you.

apackeer:
aidlc-log-subagent.ts was the one audit-emitting hook gating on its clone shard existing instead of on workflow state, so once a shard was committed every SubagentStop in every session (framework or not) appended chat-bearing SUBAGENT_COMPLETED entries to the ledger forever. It now self-gates on aidlc-state.md like its siblings (gate before heartbeat, matching session-start), with t09 regression coverage for the no-state/shard-present, no-state/no-shard, and state/no-shard paths, doc updates, and version 2.6.20. Closes #710

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.54] - 2026-08-21
+
+Subagent completion auditing now runs only while the active workflow state is explicitly `Running`, instead of writing whenever a state file or audit shard remains from an earlier workflow. **Upgrade:** refresh your `dist/<harness>/` shell so project-wide SubagentStop hooks no longer write audit entries or health heartbeats during non-AI-DLC sessions or after workflow completion.
+
+* SubagentStop exits when active intent state is absent, unreadable, or does not declare `Status: Running`; a running workflow still records `SUBAGENT_COMPLETED` and creates the clone's audit shard if needed.
+
 ## [2.6.53] - 2026-08-21
 
 Kiro IDE hook guidance now reflects that tool-argument delivery varies across supported IDE generations instead of describing all payloads as empty. Hook registration and enforcement behavior are unchanged. **Upgrade:** refresh your `dist/<harness>/` shell to receive the corrected generated comments, Kiro IDE orchestrator guidance, and version metadata.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.53-blue)
+![version](https://img.shields.io/badge/version-2.6.54-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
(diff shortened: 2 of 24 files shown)


leandrodamascena (approved):

leandrodamascena (approved):
Approved.

The hook now emits `SUBAGENT_COMPLETED` only when the active workflow has `Status: Running`. Completed or invalid state no longer writes audit entries or heartbeats, while running workflows still create the audit shard when needed.

Regression coverage and all relevant checks pass. No remaining blockers.

PR #835, opened 2026-08-20
fix: swarm finalize merges converged unit worktrees back to the base branch

### LINKED ISSUE #513: [Bug]: autonomous swarm never merges per-unit artifacts back to the main checkout (complete --merge is metadata-only)
## Description

The autonomous Construction swarm flow never integrates a converged unit's ARTIFACTS into the main checkout. `aidlc-bolt complete --merge` merges metadata only: it delegates to `aidlc-state merge`, `aidlc-audit audit-merge`, and `aidlc-runtime fragment-merge` (state file, audit delta, runtime-graph fragment). No record-dir `.md` artifacts and no generated code are copied out of the unit's Bolt worktree, and the swarm's finalize path never dispatches `aidlc-worktree merge` (the only tool that performs a real git branch merge) - grep of `core/tools/aidlc-swarm.ts` shows only `aidlc-worktree create`. The conductor-prose git-merge step exists only for the NON-swarm Bolt path.

Net effect: after an autonomous swarm run completes a stage, the per-unit artifacts (code-generation-plan, code-summary, the generated code itself) live only in the per-unit worktrees under `.aidlc-worktrees/`. The main checkout's record tree has the audit events (SWARM_UNIT_CONVERGED etc.) but not the work.

This was discovered while fixing the batch-advance bug (#439, PR #497): the fix had to use the audit ledger as the completion signal precisely because artifact coverage in the main tree can never observe swarm work. Two engine sites already encode awareness of the gap: the report-side coverage guard exempts the autonomous swarm, and the per-unit engine treats referee-converged units as covered without artifacts.

A related sharp edge found in the same review: finalize computes convergence verdicts BEFORE its merge loop, so a unit that converges but then fails its metadata merge keeps its SWARM_UNIT_CONVERGED row forever. Nothing un-converges a unit, so after a human remediates via halt-and-ask, the swarm can never re-fan that unit.

## Expected Behavior

Either the swarm flow merges each converged unit's worktree back (artifacts + code) as part of finalize or a post-batch step, or the docs/prose state explicitly that swarm output stays in worktrees and define whose job the merge is, with a deterministic verb for it. The converged-then-merge-failed recovery path should have a way to re-open a unit.

## Environment

- v2 (2.2.x), all harnesses
- Discovered during PR #497 (issue #439); test-coverage note in that PR body: every existing e2e/integration swarm fixture is single-batch

apackeer:
The autonomous swarm's finalize was metadata-only: converged units' generated code and worker artifacts stayed in their Bolt worktrees forever while only the AIDLC metadata (state/audit/runtime-graph) reached the main checkout. Finalize now commits each genuine unit's worktree work (excluding fork-managed AIDLC metadata), squash-merges it to the prepare-time base branch via `aidlc-worktree merge` (`WORKTREE_MERGED` lands; worktree + Bolt branch removed on success, preserved on conflict), skips the already-landed metadata merge on retry, and gains `finalize --repo` for multi-repo intents. Covered by extended t134 (main-tree landing, wrong-branch failure + idempotent retry) and a new t166 multi-repo finalize case. Closes #513


leandrodamascena (changes requested):
Thanks, Arden. I reviewed the current head `20843d4c` against `v2` at `4f944c007`.

The fix is aligned with the project direction and #513: a successful autonomous swarm must not leave reviewed source and artifacts stranded in disposable worktrees. However, I found two blocking correctness issues.

P1 — Commit hooks can replace source after convergence and review. `commitSwarmWorktree` runs a normal `git add -A` and `git commit` after the final check. I reproduced a `pre-commit` hook replacing and staging `impl.txt`: the original bytes passed `--check-cmd`, but finalize returned 0, merged the hook-modified bytes, and emitted `SWARM_UNIT_CONVERGED`. The merged source therefore was neither tested nor reviewed. This needs to use the immutable reviewed-source mechanism now present on `v2`: snapshot the validated bytes without ambient hooks, merge with hooks bypassed, and verify the landed tree and source fingerprint.

P1 — A merge that lands but fails cleanup cannot follow the documented retry path. `aidlc-worktree merge` can return `[merge-succeeded:<sha>]` after the target commit has landed when subsequent worktree or branch cleanup fails. Finalize currently treats every non-zero result as an unmerged, retryable failure and suppresses `SWARM_UNIT_CONVERGED`. I reproduced the source landing, the worktree being removed, and finalize returning exit 2 without convergence authority. Retrying finalize then returns exit 2 with `no worktree on re-verify`, so the unit cannot recover through the documented retry. Post-merge cleanup failures must be distinguished from pre-merge failures and conflicts; already-landed source must never be merged again, and cleanup reconciliation must be idempotent.

The branch also conflicts semantically with the source-authority model merged in #813. A mechanical rebase would bypass its immutable `Source Commit`, per-unit manifest, repository provenance, aggregate fingerprint, and `SWARM_SOURCE_MERGED` guarantees. The correction should rebase onto current `v2`, integrate through those mechanisms, and separately ensure the reviewed AIDLC record artifacts required by #513 are included rather than excluded with framework metadata.

Validation on the original head passed the 69 focused swarm, multi-repo, retry, and batch-advance tests, package parity, coverage registry, and typecheck. I added temporary adversarial reproductions for both blockers; both reproduced deterministically. The existing CI is green for the old head, but the branch is currently conflicting and has not been validated against current `v2`.

No fixes or commits were pushed as part of this internal review. Requesting changes.

apackeer:
Thanks, Leandro, for the detailed reproductions and for directing this through #813’s source-authority model.

I reviewed current head `d0674139c` against `v2` at `42164eee`. The branch now:

- Lands application source only from the immutable reviewed `Source Commit`, with merge and commit hooks disabled.
- Preserves reviewed record artifacts and the bound source manifest transactionally.
- Requires `SWARM_SOURCE_MERGED` before modern Units advance.
- Distinguishes pre-merge failures from landed-source cleanup failures and supports idempotent cleanup-only retries.
- Retains repository provenance, aggregate fingerprints, legacy compatibility, and multi-repo behavior.

Validation passed package parity, typecheck, lint, focused authority and recovery suites, and all GitHub CI checks, including smoke/unit and deterministic integration/e2e.

The corrective implementation also completed two independent Codex closure-review cycles, with the final verdict `MERGEABLE`.

leandrodamascena (approved):

leandrodamascena:
Thanks, Arden. I reviewed head `d0674139c` against current `v2` at `e4d17257b`.

The change remains aligned with #513, and I found no new blocking correctness issues. The previous findings are resolved through immutable reviewed-source binding, hook isolation, transactional record-artifact preservation, durable `SWARM_SOURCE_MERGED` authority, and idempotent cleanup-only recovery.

I validated package parity, coverage, typecheck, lint, record rollback and symlink guards, hook isolation, audit-append failure, and post-authority cleanup recovery. I also simulated integration with current `v2`; the authored conflict is limited to preserving both the new `reviewArtifactEntries` export and this PR's snapshot types.

Approved with the required follow-up to rebase onto current `v2`, re-slot the release version, regenerate dist and coverage metadata, and rerun required CI before merge.

PR #836, opened 2026-08-20
feat: gate-bound sensor firing with blocking severity (2.6.72)

### LINKED ISSUE #529: [Feature]: debounce sensor firing - the Write/Edit hook re-fires the same sensor + stage + path tuple on every save, so doc-heavy stages accumulate hundreds of advisory fires per run

apackeer:
Document-shape sensors (claim-sources, required-sections, upstream-coverage) now fire once per existing deliverable at gate presentation instead of on every Write/Edit, via a new fire_on: write|gate manifest field — eliminating the per-save fan-out that accumulated hundreds of advisory fires in doc-heavy stages. Sensors may also declare default_severity: blocking: a failed blocking gate sensor refuses gate-start and revise re-entry until fixed or explicitly overridden with the audited --override-blocking-sensors flag (groundwork delivered for #431). Closes #529


leandrodamascena (changes requested):
**Request changes**

1. **P1: Blocking sensors fail open on execution failures.**
   `fireGateSensors()` ignores non-zero dispatcher exits and only blocks on a valid `failed` verdict with a detail path. I reproduced malformed output, exit 2, and timeout cases: all opened the gate with status 0. A blocking sensor must distinguish a verified pass from an unavailable or invalid evaluation.

2. **P1: The approve-time revision backstop bypasses gate sensors.**
   Recovered revision re-entry proceeds toward approval without calling `fireGateSensors()`. Revised bytes can therefore be approved without rerunning the blocking checks. Normal entry, `revise`, and recovered re-entry need one shared enforcement path.

3. **P1: The model can authorize its own override.**
   `--override-blocking-sensors` requires no human turn, offered override choice, or authorization receipt, and is accepted in autonomous mode. This conflicts with #431’s requirement that unattended runs halt loudly.

4. **P2: `--artifacts` can send files outside the project to sensors.**
   Candidates are accepted using basename plus `isRegularFile()` only. I reproduced an external `/private/tmp/.../one.md` being passed to the sensor. Paths must be canonically confined to the stage’s resolved produce directories.

The branch also needs a substantial rebase. It is still versioned as `2.6.21`, while current `v2` is `2.6.61`, and conflicts in state and orchestration code. The resolution must preserve reviewer-before-gate checks and already-open gate revalidation.

The existing head passes 85 targeted tests, package parity, typecheck, and Biome. These failures come from additional adversarial coverage absent from the current suite.

leandrodamascena (approved):
I reviewed the rebased implementation at commit `4c62ff1da` and addressed the remaining gate-sensor enforcement gaps.

- Gate dispatch now applies each sensor's `matches` filter before invocation.
- Blocking verdicts are bound to the evaluated artifact bytes using SHA-256 checks before dispatch, after evaluation, and inside the state transaction.
- Overrides on already-open gates are now explicitly audited with `Revalidated: true` and consume their human authorization receipt.
- The Kiro adapter regression test now correctly verifies that gate-fired document sensors do not run during intermediate writes.
- Added adversarial coverage for mixed outputs, artifact mutation during evaluation, and override reuse.

The focused validation suite passes with 109 tests, along with TypeScript, Biome, packaging drift, and diff checks. Approved.

PR #837, opened 2026-08-20
fix: cost previews match the effective grid; approval gates say how long they have waited (2.6.73)

### LINKED ISSUE #530: [Feature]: cost-preview refinements (per-unit clause, greenfield-effective counts) and a first unattended-gate story

apackeer:
Scope cost previews now describe the workflow the engine will actually run: the per-unit clause is omitted when units-generation is skipped (those stages run a single pass), and cold-start confirms apply the same greenfield reverse-engineering adjustment intent creation writes into state. For the first unattended-gate story, /aidlc --status shows the organic gate-open timestamp and pending duration, and /aidlc --doctor adds a passing advisory for gates open longer than 24 hours (statusline/notification seams and the #766 presence-mint design intentionally deferred). Closes #530


leandrodamascena (changes requested):
P1: Bound the pending-gate timestamp to the current stage attempt

`pendingOrganicGate()` considers only gate-open and gate-resolution events for the matching stage. It ignores attempt boundaries such as `WORKFLOW_STARTED`, `STAGE_JUMPED`, and `STAGE_STARTED`, allowing an unresolved gate from an earlier execution of the same stage to supply the timestamp for the current gate.

I reproduced this with an organic gate-open on August 1, followed by a redo jump and a new `STAGE_STARTED` on August 20, then a recovered-only gate-open for the current attempt. `--status` incorrectly reported the gate as waiting since August 1, and `--doctor` emitted the corresponding stale-gate advisory. The current gate intentionally has no organic timestamp, so both surfaces should omit its age.

Please constrain the scan to the current workflow/stage attempt, using the latest relevant workflow, jump, or stage-start boundary, and add regression coverage for both `--status` and `--doctor`.

The branch also needs to be rebased onto the current `v2` and have its version metadata re-bumped before merge.

leandrodamascena (approved):
I re-reviewed this against commit `1ed3fe0cb` after rebasing onto the current `v2`.

The requested attempt scoping is now correctly enforced. Pending-gate age resets at workflow, jump, and matching stage-start boundaries. Recovered and revalidated rows cannot open or replace the organic timer.

I also verified the cross-shard behavior adversarially: same-second boundary/gate events no longer depend on shard filename order and fail closed when causally ambiguous, while same-shard append order remains authoritative.

The generated distributions are synchronized, and the focused tests, typecheck, lint, coverage registry, version synchronization, and package drift checks pass.

No remaining blocker from my review. Approved.

PR #838, opened 2026-08-20
fix: Stop hook no longer fires the pending-step nudge while a background subagent is in flight

### LINKED ISSUE #548: [Bug]: Stop hook fires 'pending step' error while conductor awaits background Agent() subagent
## Description

The `aidlc-stop.ts` hook fires a blocking error ("The AIDLC workflow has a pending step") when the conductor has legitimately dispatched a background subagent via the `Agent()` tool and is waiting for its completion notification. The conductor CANNOT report the stage result until the subagent returns — but the stop hook doesn't recognize this in-flight async state.

This creates a confusing UX: the user sees a blocking error telling them to "run next" or "park", but neither action is appropriate — the conductor is correctly waiting for work to complete.

## Steps to Reproduce

1. Start an AI-DLC workflow on Claude Code
2. Reach a `mode: subagent` stage (e.g., `reverse-engineering` or `code-generation`) OR any stage where the conductor dispatches `Agent()` with `run_in_background: true`
3. The conductor launches the background agent and outputs "waiting for completion"
4. Before the background agent returns, the stop hook fires
5. Observe: blocking error "The AIDLC workflow has a pending step (a run-stage directive for X). You haven't finished the forwarding loop yet."

## Expected Behavior

The stop hook should recognize that a background agent is in-flight and NOT block. Possible approaches:
- Check for a `.aidlc-agent-inflight` marker file (set by conductor before Agent dispatch, cleared on notification)
- Simply suppress the hook when the conductor's last output mentions "waiting for" or "background agent"
- Add a timeout: only fire if the pending step has been idle for >5 minutes with no background work

## Actual Behavior

The stop hook fires immediately, producing a confusing error that tells the user to `report` or `park` — neither of which is valid while a subagent is running. The conductor correctly continues waiting anyway (the hook is blocking-advisory, not truly blocking execution), but the error message clutters the UX and confuses users.

## AI-DLC Version

v2 (alpha)

## Release / Commit

Latest (installed via `.claude/` workspace shell, July 2026)

## AI-DLC Phase

Not phase-specific — happens on any stage that dispatches a background Agent().

Observed specifically during:
- Stage 2.1 Reverse Engineering (subagent mode, aidlc-developer-agent)
- Stage 2.3 Requirements Analysis (inline mode but with background Agent for artifact generation)

## Platform / IDE

Claude Code

## AI Model

global.anthropic.claude-opus-4-6-v1[1m] (Opus 4.6, 1M context)

## Environment

- OS: macOS Sequoia (Darwin 25.5.0)
- Shell: zsh
- Terminal: Warp
- Claude Code: latest

## Additional Context

The stop hook at `.claude/hooks/aidlc-stop.ts` checks `aidlc-state.md` for an in-progress stage and fires if it finds one. It has no awareness of whether the conductor is actively waiting for a subagent or is genuinely idle/stuck.

The hook's guard condition needs to include: "AND no background agent is in-flight" — which requires either a marker file protocol or a query to the Claude Code task system.

apackeer:
The Stop hook had no signal for an in-flight background Agent/Task dispatch, so a conductor that legitimately ended its turn to await the result was hit with the confusing "pending step" forwarding-loop nudge. This adds a deterministic marker protocol (written by the dispatch hook on run_in_background:true, cleared on SubagentStop) honoured by a new autonomy-guarded, 2h-freshness-bounded Stop-hook carve-out, plus a doctor probe, t304/t204 coverage, and docs. Closes #548


leandrodamascena (changes requested):
Thank you for addressing the pending-step Stop-hook behavior in #548. The intent to let the conductor stop cleanly while a background subagent is genuinely still running is useful.

I reviewed the current head `1ff065e86f11b6a5f401d25c6376c8d75e08bdb6` against the current `v2` head `1790d41c69cf20403775b116cdb89d92feaf91fd`. Project alignment is established by this being maintainer-owned work implementing the accepted #548 scope. I found two blocking correctness issues:

### P2: The workspace-wide Boolean marker cannot represent concurrent workers or sessions

`subagentInflightMarkerPath()` defines one marker for the entire workspace (`core/tools/aidlc-lib.ts:5683`). Every accepted background dispatch writes that same marker (`core/hooks/aidlc-deliver-stage-rules.ts:288`), and every `SubagentStop` unconditionally deletes it (`core/hooks/aidlc-log-subagent.ts:46`).

When two background dispatches overlap, the first completion removes the only marker while the second worker is still active. The next Stop event therefore loses the in-flight signal and resumes forwarding-loop enforcement even though the condition this change is intended to represent is still true. I reproduced this with two accepted background dispatches: the first `SubagentStop` removed the marker while the second remained active. The current base's per-session bindings make the shared marker additionally unsafe across sessions, because one session can authorize or clear another session's state.

Please represent accepted in-flight work per session/dispatch (or with equivalent reference-counted identity), remove only the completing worker's state, and cover overlapping workers both within one session and across sessions.

### P2: A rejected dispatch leaves a fresh in-flight marker even though no worker started

The marker is written before `dispatchHookOutput()` validates and accepts the dispatch (`core/hooks/aidlc-deliver-stage-rules.ts:281-298`). Error and size-limit rejection happen afterward. I reproduced this with an oversized rules bundle: the hook returned exit code 2 and reported that no worker started, but a fresh marker remained. That marker can suppress Stop enforcement for up to two hours without any in-flight subagent.

Please create the marker only after the dispatch has passed the hook's acceptance path, and add regression coverage proving rejected dispatches leave no marker.

I did not modify or push the maintainer-owned branch. Validation on the reviewed head:

- `t121` + `t304`: 81 tests, 208 expectations passed
- `t09` + `t204`: 14 tests, 43 expectations passed
- `t68`: 7 tests passed
- Typecheck passed
- Lint passed with three existing Biome informational suggestions in `t267`
- Coverage registry check passed
- Focused diagnostic tests failed exactly on the two paths above and were removed afterward

The head CI is green, but it predates the current `v2` base. The PR is currently conflicting; the head carries version `2.6.19` while the base is at `2.6.82`, so the resolved result still needs fresh validation after rebasing.

apackeer:
Addressed the blocking review findings on head `e47cdb7a1` and rebased the branch onto current `v2` (`62b868c57`).

Changes:
- Replaced the workspace-wide Boolean marker with a workspace-locked, session-scoped in-flight ledger. Accepted background dispatches add one entry; `SubagentStop` removes one matching entry, preserving overlapping workers in the same session and isolating concurrent sessions.
- Stop now honors only a fresh entry for its own session, prunes stale entries, and fails closed for malformed or foreign-session evidence. Autonomous Construction remains enforced.
- Moved ledger creation after dispatch rule validation and output-size checks. Missing-rule and oversized rejected dispatches now leave no in-flight state; the accepted preload-fallback path still records the worker.
- Forwarded the same host session identity through Kiro CLI/IDE, Copilot, and opencode dispatch/completion adapter paths.
- Updated doctor reporting, docs, generated distributions, coverage metadata, and re-slotted the release metadata to `2.6.94`.

Post-rebase validation:
- `bun run check`: packaging parity for all seven harnesses, all TypeScript configs, and lint passed (three existing informational `t267` suggestions only).
- `t09` + `t121` + `t327`: 105 assertions passed.
- `t204` + `t248` + `t319` + `t324` + `t68` + coverage registry: 110 assertions passed.
- Kiro/Kiro IDE adapters: 91 assertions passed.
- opencode adapter: 16 assertions passed.
- Copilot adapter: 54 assertions passed with a freshly rebuilt `2.6.94` native binary.

Please re-review when ready.

leandrodamascena (changes requested):
Thank you for addressing the previous review findings. The session-scoped ledger, reference counting, and accepted-dispatch ordering resolve the original blockers.

I reviewed the current head `e47cdb7a11f7db4910281be8c297e891f7e2abf5` against base `62b868c575bbbac8904051a9255fad10931d47f6`. Project alignment remains established by this maintainer-owned implementation of the accepted #548 scope.

### P2: A foreground completion can consume a background worker's ledger entry

Only dispatches with `run_in_background === true` create entries (`core/hooks/aidlc-deliver-stage-rules.ts:271-298`). However, every `SubagentStop` calls `completeSubagentInflight()` (`core/hooks/aidlc-log-subagent.ts:47-50`), which removes the first entry matching only the session (`core/tools/aidlc-lib.ts:9662-9667`).

When a background worker is active and a foreground worker in the same session completes first, the foreground completion removes the background worker's entry. Stop enforcement then resumes while that background worker is still running.

I reproduced this through the real completion hook: the ledger changed from `freshCount: 1` to absent after a foreground `SubagentStop`.

Please correlate completion with the registered background dispatch identity, or otherwise prevent foreground completions from decrementing background work. Add coverage for overlapping foreground and background workers in one session.

### P2: A failed ledger deletion silently leaves completed work active

When the final entry is removed, `writeSubagentInflightLedger()` catches and ignores every `unlinkSync()` failure (`core/tools/aidlc-lib.ts:9605-9611`). `completeSubagentInflight()` still returns success (`core/tools/aidlc-lib.ts:9667-9668`), while subsequent matching rereads and authorizes the unchanged entry (`core/tools/aidlc-lib.ts:9703-9709`).

I reproduced this with a non-writable ledger directory: completion returned `true`, but `matchSubagentInflight()` still returned `active: true`. The Stop carve-out can therefore remain active for up to two hours after completion.

Please propagate or durably record deletion failure so stale readable evidence cannot continue authorizing Stop. Add regression coverage for permission or partial-write failures.

I did not modify or push the maintainer-owned branch.

Validation on this head:

- Both failure paths reproduced against the real hooks/helpers
- 99 focused tests passed across `t327`, `t248`, `t204`, `t147`, and `t241`
- Typecheck, focused Biome, and `git diff --check` passed
- All GitHub checks for this exact head are green
- The PR is mergeable but remains blocked by the existing requested-changes review

The remaining risk is incorrect Stop enforcement whenever foreground and background workers overlap, and stale authorization after ledger cleanup failures.

leandrodamascena (approved):
Thank you for reworking the background-subagent marker into a locked, session-scoped ledger and for moving marker creation after dispatch validation.

I reviewed the current head `e47cdb7a11f7db4910281be8c297e891f7e2abf5` against `62b868c575bbbac8904051a9255fad10931d47f6`. Project alignment is established by issue #548, its `v2-launch` scope, and the prior maintainer review.

The two previous blockers are addressed: overlapping background workers are reference-counted and isolated by session, while rejected dispatches no longer create in-flight evidence.

**Non-blocking follow-up:** ledger entries are not yet bound to individual workers. A foreground `SubagentStop` in a session with an active background worker can consume that session’s background entry. Please track this as a follow-up to correlate completions with actual background dispatches and add mixed foreground/background coverage.

Validation completed across the focused ledger, Stop-hook, dispatch, doctor, adapter, version, coverage, packaging, typecheck, and lint suites. Remote CI is fully green for the exact head.

No branch changes were made.

Proposed review event: **APPROVE**

PR #842, opened 2026-08-20
fix: add quality-target verification gate to Construction (never relax defined NFR targets) (2.6.74)

### LINKED ISSUE #273: [Bug]: No quality-target verification gate — measurable NFR targets (e.g. coverage thresholds) can be silently relaxed in Code Generation / Build & Test

apackeer:
Measurable NFR targets (e.g. coverage thresholds) could be silently relaxed downstream: Code Generation had no rule forbidding a developer agent from lowering test-config thresholds to pass, and Build and Test never compared actual results against the defined targets. This ports the reporter's v1 PR #274 direction to the v2 architecture: a never-relax Critical Rule + developer-subagent instruction in code-generation, Coverage Target / Target Met verdicts and a per-target actuals comparison plus a no-weakening on-failure rule in build-and-test, an overconfidence red flag in the stage protocol, and an org Testing Posture invariant - pinned by tests/unit/t317-quality-target-gate.test.ts (v2.6.74). Closes #273

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.74] - 2026-08-24
+
+Defined quality targets now remain binding from Code Generation through Build and Test: every measurable target is inventoried from NFR Requirements, NFR Design, and the approved Testing Contract, then finalized with evidence before the stage can succeed. **Upgrade:** refresh your `dist/<harness>/` shell so the updated Construction stage and protocol guidance are installed.
+
+* Code Generation treats measurable quality targets as required inputs and forbids lowering, relaxing, or disabling a target to manufacture a pass.
+* Build and Test executes every applicable generated check, permits deployed-environment deferral only to a named owning validation stage, and keeps deferred or missing evidence visibly `Unverified`.
+* The target matrix records target ID and source, expected and actual values, evidence, owning stage, and a final `Met`, `Not Met`, or `Unverified` verdict on every exit path; `N/A` is valid only when no target applies.
+* `Not Met` and `Unverified` targets now enter the same bounded Build-and-Test failure-escalation ladder as command failures, preserving the Code Generation loop-back added in 2.6.20.
+
 ## [2.6.73] - 2026-08-24
 
 Scope cost previews now describe the workflow the engine will actually run, and approval gates expose how long the current gate attempt has been waiting for a person. **Upgrade:** refresh your `dist/<harness>/` shell to install the corrected preview, status, and doctor behavior; existing workflow records need no migration.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.73-blue)
+![version](https://img.shields.io/badge/version-2.6.74-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/aidlc-common/protocols/stage-protocol.md
+++ b/core/aidlc-common/protocols/stage-protocol.md
@@ -491,6 +491,8 @@ When contradictions are detected:
   - "Whatever you think is best" or "up to you" — ask what outcome they care about most
   - Contradictory signals between different answers
   - Answers that dodge the question or change the subject
+  - Relaxing, lowering, or disabling a previously defined quality target (e.g.
+    a test coverage threshold) instead of meeting it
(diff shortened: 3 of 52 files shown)


leandrodamascena (changes requested):
I reviewed commit `2d1ec8d` against the current `v2` head, the requirements in #273, and the Build-and-Test failure-loop behavior introduced by #616. I found three remaining blockers:

P1: Treat unmet or unverified quality targets as Build-and-Test failures

The new rule surfaces an unmet target at the approval gate, while failure handling still runs only when build or test commands fail. A command can exit successfully while reporting coverage below an external NFR target. This would bypass the failure ladder and could be auto-approved under autonomous Construction.

A `Not Met` or `Unverified` target must enter the same bounded loop-back or halt-and-ask flow as other Build-and-Test failures.

P1: Execute the checks required to produce every claimed actual

The PR claims to compare actuals against every measurable NFR target, including performance budgets. However, Step 10 executes only build, unit, and integration commands; the generated performance and security instructions are never executed.

Please execute all applicable checks or explicitly defer targets requiring a deployed environment to their owning validation stage. Deferred targets must remain visibly unverified and cannot receive a successful verdict.

P2: Build and finalize a source-complete target matrix

Step 9 requires `Target Met: Yes/No/N/A` before Step 10 produces actual results. The contract also does not define when `N/A` is valid, require failure paths to update the summary, or inventory targets from every named source: NFR Requirements, NFR Design, and the approved Testing Contract.

Please require a per-target matrix containing target ID/source, expected value, actual value, and verdict. It may begin as `Pending`, but must be finalized on every exit path. `N/A` is valid only when no target applies; an applicable target without an actual is `Unverified` and must enter failure handling.

The branch also needs to be rebased onto the current `v2` and have its version metadata re-bumped. The rebase must preserve and integrate #616’s Build-and-Test failure loop.

leandrodamascena (approved):
I re-reviewed this against commit `8ce11d53e` after rebasing onto the current `v2`.

The previously requested changes are now addressed:

- Unmet or unverified quality targets enter the bounded Build-and-Test failure ladder.
- Applicable generated checks must run or be explicitly deferred to a named validation stage.
- Every exit path produces a complete target matrix with source, expected and actual values, evidence, ownership, and final verdict.
- The rebase preserves the current `source-manifest.json` contract and Build-and-Test loop-back behavior.

Focused unit and integration tests, type checking, linting, coverage-registry validation, version synchronization, and package drift checks pass.

No remaining blockers. Approved.

leandrodamascena:
Integration+e2e tests are failing because the runner, not because the tests.

PR #843, opened 2026-08-20
fix: bugfix and refactor scopes now include the deployment stages (2.6.70)

### LINKED ISSUE #800: [Feature]: v2 bugfix and refactor scopes should include deployment
### Description

The v2 scopes for bug-fix and refactor scopes do not include the operations phase for deployment-pipeline and deploument-exec stages.  However the security-patch scope does include these stages.  I don't see why you would want fix a bug or implement a refactor without deploying it.  These stages should be added to the scopes.

### Use Case

User wants to fix a bug and have it deployed.  User wants to refactor a portion of the application and fix a bug.

### Version

v2 (current)

### Area

Operations phase rules

apackeer:
Adds deployment-pipeline and deployment-execution to the bugfix and refactor scope grids, matching the deployment tail security-patch already runs, so a verified fix or refactor actually ships (environment-provisioning and the rest of Operation stay skipped). Bugfix now runs 9 of 33 stages with 6 approval gates and refactor 10 of 33 with 7; all dist trees, fixtures, pinned tests, and docs are regenerated/updated in the same change, with the version slot at 2.6.66. Closes #800


leandrodamascena (changes requested):
I reviewed this PR against commit `6402759b` with an adversarial pass and found one functional blocker.

**P1: Preserve deployment fallback when `deployment-pipeline` is skipped**

When an existing pipeline is already adequate, `deployment-pipeline` correctly reports `skipped`. However, `deployment-execution` still requires `cd-config.md` and `deployment-strategy.md`. Because their producer is on the active scope path, the engine emits both as `consumes_absent` with `expected: false`, which the protocol treats as a real gap requiring recovery.

I reproduced this on the `bugfix` path. Instead of deploying through the existing pipeline, the workflow asks to rerun the skipped stage or provide its artifacts manually. This breaks a primary scenario introduced by this PR.

Please treat outputs from a conditionally skipped producer as an expected absence with an explicit workspace fallback, or model these inputs as conditional/optional. Add an end-to-end test that skips `deployment-pipeline` because the existing pipeline is adequate and verifies that `deployment-execution` can proceed for both `bugfix` and `refactor`.

**P2: Rebase and refresh release metadata**

The branch is currently conflicting and 27 commits behind `v2`. Please rebase it, resolve the overlapping stage, test, and generated-tree changes, then update the title, version, README badge, and changelog from `2.6.19` to the next available version after `2.6.61`.

The targeted suite passed with 137 assertions, and `bun scripts/package.ts --check` passed, but the current tests fabricate deployment artifacts and therefore do not exercise the P1 path.

apackeer:
@leandrodamascena thanks for the thorough review - this is ready for another look.

**P1 - fixed in the engine.** `splitConsumesByPresence` now consults the recorded state: a required input whose on-path producers are all recorded `[S]` after a conditional runtime skip is emitted in `consumes_absent` with `expected: true`, at both the single-stage and per-unit wave call sites (previously `expected: false`, which routed Deployment Execution into missing-artifact recovery). The deployment-execution stage body now directs the run to inspect and use the workspace's existing pipeline configuration when Deployment Pipeline skipped itself as already adequate, and the stage-definition protocol plus all seven conductor SKILLs carry the matching "absent by scope design or a recorded conditional skip" wording.

**Tests.** New deterministic e2e `t302-deployment-pipeline-skip-fallback.test.ts` walks both `bugfix` and `refactor`: Deployment Pipeline reports skipped because the existing workspace pipeline is adequate, Deployment Execution then arrives with its three absent inputs all `expected: true`, proceeds without recovery against the workspace config, and the intent completes - no fabricated pipeline artifacts in the new path. t116 gained the unit-level contrast (skipped on-path producer -> `expected: true`, un-skipped on-path producer -> `expected: false`), and the coverage registry is regenerated for the new test.

**P2 - rebased and re-slotted.** The branch now sits directly on the current `v2` tip (`62999bc39`) and the version source, README badge, and changelog are at **2.6.65**: 2.6.62 (#716) and 2.6.64 (#824) merged while this was in flight, so the slot moved above the merged head rather than the 2.6.63 originally staged.

**Verification.** Full deterministic suite green (423 files, 8727 assertions, 0 failures, `--no-llm --all`); after the final rebase the version/coverage/fallback slice (t68, gen-coverage-registry, t134, t116, t302), `bun scripts/package.ts --check`, typecheck, and biome all pass.

leandrodamascena (changes requested):
I re-reviewed this PR against commit `fb3d04a7` with an adversarial pass. The original deployment-pipeline fallback blocker is fixed, but the implementation introduces a broader functional blocker.

**P1: Distinguish conditional runtime skips from explicit jumps**

`splitConsumesByPresence` now treats every on-path producer recorded as `[S]` as a conditionally skipped producer. However, `[S]` is also written by explicit stage and phase jumps.

I reproduced this by jumping a `bugfix` workflow directly to `deployment-execution`. The jump marked both `deployment-pipeline` and `build-and-test` as skipped, and the resulting directive emitted every missing input as `expected: true`, including `construction/build-and-test/test-results.md`.

This allows Deployment Execution to bypass missing-artifact recovery even though the bugfix scope explicitly retains Build and Test and the stage has no documented fallback for missing test results. A user jump can therefore silently turn a required verification artifact into an expected absence.

Please use skip provenance, such as the latest `STAGE_SKIPPED` audit event, to classify only conditional runtime skips as expected absences. Inputs skipped by an explicit jump should remain `expected: false` unless the consuming stage explicitly supports that fallback. Add a regression test covering both cases:

- `report --result skipped` for an adequate existing pipeline produces `expected: true`.
- An explicit jump over an on-path required producer does not suppress missing-artifact recovery.

**P2: Restore the Kiro IDE README corrections from the current base**

The rebase reintroduced two statements removed by #824:

- The Kiro IDE instructions claim that `.kiro/settings/cli.json` is bundled.
- The repository layout claims that Kiro IDE ships agent JSONs and settings.

The generated Kiro IDE distribution contains only the Markdown conductor at `.kiro/agents/aidlc.md`; it does not contain `settings/cli.json` or agent JSON files. Please restore the current `v2` wording in both README locations.

**P2: Refresh the PR metadata and release slot**

The PR title and body still identify this change as `2.6.19`, while the branch is currently versioned as `2.6.65`. PR #829 is also open, approved, and using `2.6.65`, so whichever PR merges second must rebase and move to the next available version. Please update the title and body after the final release slot is known.

The new `t116` and `t302` coverage passes, as do package parity, typecheck, and lint. The integration/e2e CI job is green.

leandrodamascena (approved):
Addressed the requested changes and rebased the PR onto the latest `v2`.

- Extended the bugfix and refactor scopes through Deployment Pipeline and Deployment Execution.
- Added explicit skip provenance so conditional runtime skips are distinguished from forward jumps.
- Fixed missing consumed artifacts after a jump to remain `expected: false`.
- Added regression coverage for conditional pipeline fallback and skipped `build-and-test`.
- Updated the release metadata and generated distributions to `2.6.70`.

Validation completed successfully: package parity, typecheck, lint, targeted unit/integration/E2E tests, and 207 runner assertions. Ready for re-review.

PR #844, opened 2026-08-20
fix: focused reverse-engineering rescans merge into the shared CodeKB

### LINKED ISSUE #632: [Bug]: reverse-engineering is not stored in the current intent

apackeer:
Focused Reverse Engineering rescans now merge their newly analyzed area into the shared per-repo CodeKB instead of wholesale-replacing it: verified-current coverage accumulates as a union, stale/unverified prior coverage is preserved as prose and honestly demoted to shallow, and full rescans still replace all nine artifacts. This implements the cumulative behavior the issue thread converged on after #670 shipped the honest-overwrite guard, using that guard's existing deterministic machinery (COVERS is now the merge backstop) with no tool changes. Closes #632


leandrodamascena (changes requested):
Thanks for addressing #632. The cumulative CodeKB direction is aligned with the issue and with the overwrite/fingerprint work from #670.

I reviewed head `49ca9c6f` against merge base `4d0968fc` and tested its integration with current `v2` at `b1a95632`.

## Blocking: the shared CodeKB merge is not generation-safe

In `core/aidlc-common/stages/inception/reverse-engineering.md`, the workflow:

- reads the current shared artifacts and coverage;
- unions the requested paths;
- creates a fresh source fingerprint;
- performs a scope-only comparison;
- writes all nine shared artifacts directly.

There is no generation token, compare-and-swap, repository-scoped lock, or atomic publication step.

This allows two data-integrity failures:

1. A previously covered source can change after the initial status check. A focused scan of another area can preserve stale prose for that source while minting a fingerprint over its new contents. The scope comparison still returns `COVERS`, and the resulting stale CodeKB is reported as `CURRENT`.

2. Two concurrent same-space intents, now supported following #858, can read the same baseline and publish independent merged results. The last writer silently drops the first focused scan, recreating the data loss this change is intended to prevent.

I reproduced the first case with a temporary diagnostic regression: seed a CURRENT store for `src/payments/`, mutate that source, perform a focused merge for `src/catalog/`, preserve the old payments prose, and mint the union fingerprint. The next status incorrectly returned `CURRENT`.

Please bind the merge to both the source and CodeKB generations and publish it atomically. A suitable approach would be to stage the merged artifacts, commit them through a repository-scoped CAS utility using the expected store generation and source fingerprint, and re-read/re-merge if either changed. Please also add deterministic interleaving tests for source mutation and concurrent focused merges.

## Non-blocking follow-up: utility messaging still describes replacement

`core/tools/aidlc-utility.ts` still tells users that narrower or unknown-scope reruns replace the store and discard knowledge (`UNKNOWN_SCOPE` around line 4659 and `NARROWER` around line 4701). That contradicts this PR’s cumulative behavior, where existing prose is retained and only its verification status changes.

Please update the user-facing text to say that paths/components are no longer claimed as verified coverage. The existing JSON field names can remain for compatibility if needed.

## Validation

- New `t304-codekb-cumulative-merge`: 5/5 passed.
- `t68-version-changelog-sync`: 7/7 passed.
- Coverage registry is fresh.
- `bun run check`: package parity, typecheck, and lint passed.
- Two `t248` failures reproduced unchanged on the merge base, so I am not attributing them to this PR.
- Existing CI on `49ca9c6f` is green.

A diagnostic rebase onto current `v2` auto-merged the functional changes. Only changelog/version metadata and generated version copies conflicted. Since current `v2` is `2.6.110`, this PR will need to re-slot to `2.6.111` and regenerate `dist`.

I made no commits or pushes to the contributor branch.

apackeer:
Addressed the review findings on head `698e27069` (rebased onto current `v2` at `d1d2b931d`, re-slotted to `2.6.115`).

## What changed

- Added `codekb-snapshot`, which captures the exact shared-store generation and source fingerprint under a space+repo lock immediately before scanning.
- Added `codekb-publish`, the sole shared-store writer. It validates an exact nine-artifact candidate, rechecks the expected source and store generations plus the candidate scope fingerprint, and publishes the complete directory with rollback/crash recovery.
- A changed store now returns `CODEKB_STORE_CHANGED` and requires re-reading/re-merging the winner's store. Changed source returns `CODEKB_SOURCE_CHANGED` and requires a fresh scan. A stale candidate fingerprint is also refused.
- Staged candidate paths are confined to the project, including real-path checks against symlinked ancestor escapes; recovery refuses symlinked transaction state.
- Reverse Engineering now writes to a temporary complete candidate directory and never writes cumulative artifacts directly into the shared CodeKB.
- Updated `UNKNOWN_SCOPE` and `NARROWER` human output to describe paths/components that are no longer claimed as verified coverage. Existing JSON field names remain unchanged.

## Regression coverage

- Source mutation after the snapshot, followed by minting a new union fingerprint over stale preserved prose, is refused; the old store remains `STALE`, never falsely `CURRENT`.
- Two focused candidates built from the same baseline contend through real subprocesses: exactly one publishes, the other receives `CODEKB_STORE_CHANGED`, and a fresh re-merge preserves both additions.
- An interrupted directory swap is recovered before a later snapshot reads either generation.

- `t304-codekb-cumulative-merge`: 5/5 passed.
- Expanded unit slice: 9 files, 203 assertions, 0 failures.
- Deterministic integration slice (`t44`, `t55`): 50 assertions, 0 failures.
- `bun run check`, package parity, typecheck, lint, and coverage-registry freshness passed.
- GitHub contract, markdown, changelog, gitleaks, Bun audit, and Semgrep checks are already green; the remaining test jobs are still running.

@leandrodamascena, please re-review head `698e27069`.

leandrodamascena (changes requested):
Thanks, Arden. The generation-safe publication direction is aligned with #632, #670, and the previous review.

I reviewed the current head `80d3d85ce7a6ab017919505d4b8274e49c376676` against `v2` at `d1d2b931d5a703bf8c2a506744403ca632796f69`, and checked integration with current `v2` at `7f42a9081ebf4e60bbf82325d9fe037293a34960`.

**P1: Transaction recovery can delete data outside the project**

`recoverCodekbTransactions()` checks the final transaction root and transaction directories with `lstat`, but it does not check their parent components (`core/tools/aidlc-utility.ts:6321-6344`).

I reproduced this by making `.aidlc-codekb-transactions` a symlink to an external directory and placing a file under the expected `<repo>/<transaction>/` path. Running `codekb-snapshot` returned success, but recovery followed the symlink and deleted the external file and its parent directories through `rmSync`.

Please validate every component from a trusted project anchor before recovery performs any rename or recursive removal. The check must reject symlinked ancestors, not only a symlink at the final path. Please add a regression that points `.aidlc-codekb-transactions` outside the project and verifies that recovery refuses without modifying the target.

The branch also conflicts with current `v2`, including `core/tools/aidlc-utility.ts`. After the fix, it needs a rebase, release re-slot from `2.6.115` to the next available version, and regenerated distributions.

Validation performed:

- Focused suite: 4 files, 88 assertions, all passing.
- The external-deletion path was reproduced on the exact head.
- `git diff --check` passes.
- All GitHub checks are green for this head, but the PR is currently conflicting with `v2`.

No changes or commits were pushed. The remaining risk is destructive recovery outside the project boundary.

apackeer:
Addressed the transaction-recovery P1 on head `58334b355`, rebased onto current `v2` at `3ac57def5`, and re-slotted the release to `2.6.117`.

## What changed

- `recoverCodekbTransactions()` now anchors recovery at the real project root and validates every component of the transaction and CodeKB store paths before any rename or recursive removal.
- The guard is rechecked immediately before restoring a backup or deleting a transaction directory, so a redirected `.aidlc-codekb-transactions` ancestor is refused instead of traversed.
- Added a regression that points `.aidlc-codekb-transactions` at an external directory and verifies `codekb-snapshot` fails while the external sentinel and transaction tree remain untouched.
- Rebased through the latest `v2` changes, resolved the utility/documentation integration, regenerated every harness distribution, and refreshed the coverage registry.

## Validation

- Final-base focused slice: `t304-codekb-cumulative-merge` and `t68-version-changelog-sync`, 13 assertions, 0 failures.
- `bun run check`, package parity, typecheck, lint, and coverage-registry freshness passed.
- Final-head GitHub CI: all 11 checks passed; smoke+unit passed in 11m05s and deterministic integration/e2e passed in 3m14s.
- The PR is currently mergeable.

@leandrodamascena, please re-review head `58334b355`.

leandrodamascena (approved):
Thanks, Arden. The cumulative CodeKB design is aligned with the resolution of #632, #670, and the previous maintainer reviews.

I reviewed the current head `cdf49453400d23e5ec82a823074d82785ef7d032` against `v2` at `a09e55cc97c52199a49f87bdd639b028da5e3b02`.

The previous blockers are resolved:

- CodeKB publication is generation-safe and atomic.
- Transaction recovery rejects symlinked ancestors before destructive operations.
- Snapshot and publication resolution now use the invoking session's bound space and intent instead of the shared active-space cursor.
- A regression confirms that a session bound to `team-b` cannot snapshot or publish into `default`.

I pushed the session-bound fix in `164c8308909324cdd591001e004475eaa357d277` and integrated the current `v2` in `cdf49453400d23e5ec82a823074d82785ef7d032`.

Local validation passed: 134 focused tests across seven files, package parity for all harnesses, typecheck, lint, coverage-registry freshness, version synchronization, and `git diff --check`.

GitHub CI is green for the exact head, including smoke/unit, deterministic integration/e2e, contract checks, security scans, changelog, documentation, and build. No unresolved review threads or known blockers remain.

PR #845, opened 2026-08-20
fix: bootstrap stage diaries at directive emission, ending the Kiro ACP read-probe tool-call fumbles (2.6.55)

### LINKED ISSUE #695: [Bug]: Kiro CLI (ACP) reviewer intermittently fumbles a tool argument (orphan 'Directory not found') and silently recovers - no audit signal for recovered tool-call failures
## What was observed

During the PR #658 pre-merge gate (Kiro ACP e2e slice, and reproduced on a clean v2 baseline - so pre-existing, not introduced by that PR), the live Kiro CLI reviewer run intermittently emits an **orphan failed tool call**: a validation failure such as `Directory not found` on one tool call's arguments, which the agent recovers from on its next call. The stage still completes, the reviewer verdict lands, and the on-disk artifacts are correct.

`t-acp-kiro-reviewer.serial.test.ts` originally pinned `toolCallIssues == []`, so this transient fumble failed an otherwise-green run. The pin was loosened in PR #658 to reject only SYSTEMATIC issues (hook blocks, or repeated failures of the same `toolCallId`) - see the comment block around the `systematic` filter in that test. That keeps the gate honest but leaves the underlying behavior unexamined.

## Why it deserves an issue rather than staying a test comment

- It is live-agent argument fumbling on a shipped harness (Kiro CLI over ACP): the agent composes a path argument that does not exist, eats a failed tool call, and retries. Each occurrence costs a turn and a tool round-trip, and on latency-sensitive live runs it widens the window for real timeouts.
- The failure is invisible outside the test harness: nothing audits or surfaces recovered fumbles, so we have no signal on how often real users' runs pay this cost.
- If the fumble rate rises (model drift, prompt changes), the only detector today is a flaky e2e slice.

## What a fix could look like (open)

- Cheapest: audit visibility - count recovered tool-call failures per stage in the audit trail so `--doctor`/traces expose the rate.
- Better: identify why the reviewer composes a non-existent directory argument in this flow (observed on the requirements-review path) and steer it - e.g. the brief could name the concrete artifact paths it already knows instead of leaving the reviewer to derive them.
- The test-side handling (systematic-only rejection) is already in place and should stay regardless.

## Repro pointers

- `AIDLC_KIRO_ACP_LIVE=1 bash tests/run-tests.sh --debug --e2e --filter "t-acp-kiro-reviewer"` - the fumble is intermittent; the run stays green under the current pin but the ndjson trace records the orphan issue when it occurs.
- Gate evidence from the #658 effort: reviewer slice traces showed the orphan `Directory not found` failure mid-run with the stage completing normally, on both the PR branch and clean v2.

apackeer:
The intermittent orphan "Directory not found" failures on the live Kiro ACP slices trace to the conductor probing for the maybe-absent stage diary with fs_read (live trace: two validation-failed calls per fresh stage entry, plus a wrong-path guess for question-rendering.md). The engine now creates memory.md from the shipped template when it emits a run-stage directive (idempotent, never overwrites, advisory), stage guidance forbids read-tool probes of maybe-absent paths, and a live spike proved kiro-cli fires no hook for validation-failed calls — so no deterministic audit counter is possible for that class and removing the root cause is the fix. Deterministic unit coverage in t304; the e2e systematic-only pin stays. Closes #695


leandrodamascena (approved):
Approved.

Stage diaries are now created only when real run-stage directives are emitted. Internal Stop-hook probes remain write-free, and bounded waves create diaries only for the Units included in the emitted directive.

Regression coverage validates both paths, and all relevant packaging, typecheck, lint, coverage, orchestration, memory, and Stop-hook checks pass. No remaining blockers.

PR #846, opened 2026-08-20
fix: /aidlc --resume resumes directly instead of re-asking the resume menu (2.6.43)

### LINKED ISSUE #794: `--resume` lands on the Resume/Redo/Jump/Start-Fresh menu instead of resuming directly
**What happens:** Running `/aidlc --resume` on a project with an existing workflow returns the four-choice re-entry menu (Resume / Redo / Jump / Start Fresh). The user must then select "Resume" a second time to actually continue from the last checkpoint.

**Expected:** `--resume` is an explicit, unambiguous intent to continue. It should short-circuit the menu and route straight to the last checkpoint. The menu is appropriate for a bare `/aidlc` (ambiguous re-entry) but redundant when the user already typed `--resume`.

**Impact:** Minor friction, but it makes the `--resume` flag feel like a no-op — the user answers "resume" twice.

**Suggested fix:** In the engine's resume handling (`aidlc-orchestrate.ts next`), when the `--resume` flag is present and an existing workflow is found, emit the resume directive directly rather than the `ask` menu. Bare `/aidlc` keeps the menu.

**Environment:** Kiro CLI harness.

apackeer:
Explicit `--resume` now short-circuits the Resume/Redo/Jump/Start-Fresh menu and routes straight to the saved checkpoint (the same continuation directive bare `next` emits); bare `/aidlc` session re-entry keeps the conductor-offered menu and its `report --result resumed` round-trip. Riding along: `--resume --stage <slug>` now honors the explicit jump target, Copilot treats resumed reports as plain reports (the engine-armed resume gate can no longer arm; legacy pre-2.6.19 markers are superseded by the next explicit resume), and the off-tier SDK calibration was re-anchored from the resume gate to the Branch-8 compose offer (not live-run). Closes #794


leandrodamascena (approved):
The requested changes are resolved.

- Explicit `/aidlc --resume` now continues directly through normal steering and stage routing.
- `/aidlc --resume --stage <slug>` preserves the explicit jump target.
- Bare re-entry retains the Resume / Redo / Jump / Start Fresh menu, including OpenCode.
- Copilot no longer arms new Resume markers, while explicit Resume supersedes legacy markers.
- The branch is rebased onto current `v2`, versioned as `2.6.43`, with generated distributions synchronized.

All CI checks pass. I found no remaining code or protocol blockers.

Approved.

PR #847, opened 2026-08-20
fix: preserve lossless compose birth descriptions (2.6.42)

### LINKED ISSUE #807: [Bug]: compose→birth drops the user's task description — composed-scope workflows get the `[Project description]` placeholder, breaking `[desc]` grounding
## Summary

On the **compose** path, the user's task description is dropped before `intent-birth`, so every composed-scope workflow is created with the literal `[Project description]` placeholder as its state **Project** field. There is no supported way to set it afterward (`aidlc-state.ts set` is lifecycle-blocked), so the placeholder is permanent for the life of the intent. This breaks `[desc]` grounding in Intent Capture and trips the `claim-sources` sensor.

## Affected area

- **Engine / routing:** `core/tools/aidlc-orchestrate.ts` — `composeDispatchDirective()` (front / no-state mode) and its interaction with `birthPrintDirective()` / the `intent-birth` handler.
- **Stage:** `intent-capture` (Ideation) — its Step 3/Step 5 grounding contract expects a resolvable `[desc]` source.
- **Sensor:** `claim-sources` — validates the `[desc]` register value against `aidlc-state.md`'s **Project** field, which is the placeholder.
- **Scope:** harness-agnostic (the engine emits the directive for every harness). Observed on Kiro IDE.

## Expected vs. actual

**Expected:** When a workflow is composed via `/aidlc compose "<task text>"` (or a cold-start compose offer answered with `compose`) and then born, the task text the user provided to the compose step should flow into `intent-birth --arguments`, so the state **Project** field and the record-dir slug carry the real description — exactly as a direct `/aidlc "<task text>"` start does.

**Actual:** After the composer proposes and the human approves, the birth is run as `intent-birth --scope <name>` with no `--arguments`. `birthPrintDirective()` only threads `--arguments` when a description is passed, so the description provided to `next compose "<task>"` is lost and the **Project** field is written as `[Project description]`. `intent-capture` then cannot ground its `[desc]` source, and `claim-sources` flags a `[desc]` mismatch.

## Reproduction

1. On a fresh workspace, run a cold-start compose: `/aidlc "<a real multi-sentence task description>"`.
2. Answer the "no ready-made plan fits" offer with `compose`; let the composer propose a custom scope; approve it.
3. The conductor writes the scope files and births via `next --scope <name>` (no `--arguments`).
4. Inspect the intent's `aidlc-state.md`: **Project** is `[Project description]`, not the task text.
5. In Intent Capture, the `[desc]` source cannot be verified against state; `claim-sources` emits an advisory `[desc]` finding.

Contrast: a direct `/aidlc "<task text>"` start (no compose) threads `--arguments` correctly and records the real description — so the defect is specific to the compose→birth handoff, not to `--scope`-only starts with genuinely no description (those legitimately have none).

## Root cause

- `birthPrintDirective(scope, flags, description?)` pushes `--arguments ${JSON.stringify(description)}` only when `description` is non-empty.
- `composeDispatchDirective()` (front branch) instructs the conductor to render the proposal and, on approval, birth the workflow — but never says to carry the original `flags.intent` task text into that birth. The engine is stateless between calls, so the description given to `next compose "<task>"` exists only in the conductor's context at approval time; nothing tells the conductor to re-supply it, and nothing persists it for the birth to read.

## Suggested fix

Minimal, single-file, harness-agnostic: extend `composeDispatchDirective()`'s front-mode message so the post-approval birth instruction threads the original task text, e.g. `intent-birth --scope <name> --arguments "<original task text>" --label "<essence>"`. Since every harness's conductor follows the engine's directive message, this fixes all harnesses without editing the five per-harness `SKILL.md` copies.

A more robust alternative (no reliance on the conductor re-supplying text): have the compose step persist the task text (e.g. a small sidecar written when the composer runs) and have `intent-birth` read it when `--arguments` is absent on a composed-scope birth, then consume it. This survives context loss between the compose turn and the birth turn.

Either way, a unit test asserting the composed-birth directive/flow carries the description would prevent regression.

## Environment

- **Platform:** macOS (Apple Silicon / darwin)
- **Harness:** Kiro IDE distribution (`dist/kiro-ide`)
- **Model:** Kiro "Auto" model selection (Anthropic Claude)
- **Framework version:** `awslabs/aidlc-workflows` `v2` branch, HEAD `1806d44` at time of report (reproduced against the installed Kiro IDE distribution in a separate project)

apackeer:
On the compose path the post-approval birth ran as a bare `next --scope <name>`, so composed-scope intents were born with the literal `[Project description]` placeholder, breaking `[desc]` grounding and the claim-sources sensor. The front compose dispatch now directs the conductor to thread the original task text verbatim into the birth, and every harness SKILL.md approve step names the same carry; pinned by a t198 unit test and a t192 live-journey assertion. Closes #807


leandrodamascena:
Maintainer follow-up pushed at `2b8f56ff0`.

Compose-to-birth description handling is now lossless and shell-safe:

- Every front/report proposal requires a nonblank `birthDescription`.
- Task-backed proposals copy the dispatch task exactly.
- Report-only and task-less proposals derive a grounded description from the approved report or plan.
- Scope-only compose births are forbidden.
- Birth descriptions are passed after the literal `--` boundary as one shell-safe argv value.
- `intent-create` uses `--arguments=<value>`, allowing descriptions that begin with `--`.
- `$()`, backticks, variables, embedded quotes, newlines, and flag-like text remain literal instead of being expanded or discarded.
- The top-level dispatcher preserves literal task arguments after `--` and inserts global `--project-dir` before that boundary.
- The composer agent schema, all seven harness conductor surfaces, documentation, and generated distributions now share the same contract.
- The branch was rebased onto current `v2` and rebumped to `2.6.42`.

Validation:

- Focused parser/dispatcher/documentation slice: 129 tests, 1,959 assertions, 0 failures.
- Compose and dispatcher suites: 79 tests, 512 assertions, 0 failures.
- `t118` under the CI harness: 16 tests, 78 assertions, 0 failures.
- The integration sweep retained only environment-sensitive timeouts; the affected `t52` and `t99` files passed in isolation.

The PR is ready for CI completion and final approval.

leandrodamascena (approved):
All blocking findings are resolved at `2b8f56ff0`. Compose birth descriptions are required, exact, flag-safe, and shell-safe across the dispatcher, engine, composer schema, and all harness surfaces. Approved for merge once CI is green.

PR #848, opened 2026-08-20
fix: detect workspaces whose source is nested up to three container levels (#438)

### LINKED ISSUE #438: [Bug]: project type defaults to Greenfield in the bugfix workflow

apackeer:
Workspace detection classified projects Greenfield when their source sat more than one container level down (e.g. services/api/src/), so incremental scopes like bugfix skipped Reverse Engineering. The nested-project fallback is now a deterministic bounded recursive walk (cap 3 levels) that aggregates hits without double counting and reports slash-joined nested roots; scope routing and the empty-workspace advisory are unchanged. Ships as 2.6.19 with regenerated dist trees and t203 regression coverage. Closes #438


leandrodamascena (approved):
The branch now addresses the adversarial boundary cases and is current with `v2`.

The nested scanner evaluates each visited directory independently. It no longer looks through an additional arbitrary container, so the documented three-level limit is exact and sibling projects retain their own language, framework, build-system, and `nestedRoot` signals.

### Verification

- Nested workspace tests: 21 passed.
- Version/changelog synchronization: 7 passed.
- Package parity, typecheck, and lint: passed.
- Coverage registry check: passed.
- Added regressions for direct source and manifests in a fourth container, plus sibling Python and React/TypeScript projects.

I found no remaining code or protocol blocker. Approved.

PR #839, opened 2026-08-20
fix: durable ordered pipeline-link receipts and per-repo codekb completion matrix

apackeer:
Pipeline stages now mint durable, ordered PIPELINE_LINK_COMPLETED receipts per link (aidlc-log.ts link), resume from the first missing link via directive.pipeline, and cannot enter or complete approval from conductor-written artifacts alone; repeated and out-of-order links are refused. Reverse Engineering requires the complete declared artifact set for every registered repository with repo-scoped ARTIFACT_REUSED reuse exemptions, and multi-repo codekb directive paths resolve from the intent registry; isolated --single receipts never satisfy the main workflow. Closes #741

PR #850, opened 2026-08-20
fix: require complete current-attempt CodeKB evidence per repo (2.6.39)

### LINKED ISSUE #514: [Bug]: multi-repo codekb writes drift on Kiro ACP / Codex exec (split from the gate-verb issue)
## Description

Splitting this out of #442, whose fix (PR #510) deliberately covers only the Request Changes half. The original report described a second symptom class observed in live runs: on Kiro ACP and Codex exec, multi-repo codekb writes drifted - reverse-engineering output for a repo landed somewhere other than the canonical `aidlc/spaces/<space>/codekb/<repo>/` location, or a bare `next` was run where the per-repo codekb pass was expected.

The community comment on #442 warned from experience that this class LOOKS identical to the skipped-verb class but has a different mechanism, and folding them into one fix hides the second bug. The gate-revision backstop shipped in PR #510 cannot catch it: codekb stages are explicitly excluded from its predicate (their artifact paths carry no stage-slug segment), and the drift is about WHERE artifacts land, not whether a gate verb was recorded.

What an investigation needs to establish:

1. Reproduce on Kiro ACP and Codex exec: drive a multi-repo intent through reverse-engineering and compare where each repo's codekb artifacts land vs the canonical per-repo path.
2. Determine the mechanism: conductor forwarding drift (the ACP/Kiro conductor forwarding fragility is a known weak seam), a missing --repo thread on one harness's dispatch, or prose ambiguity in the RE stage for the multi-repo case.
3. Decide the deterministic guard: the artifact guard and per-unit coverage guard precedents suggest a path-level check at the approve/report chokepoint for codekb stages (right repo dir per recorded repo set), rather than more prose.

## Environment

- v2, primarily Kiro ACP and Codex exec harnesses per the original #442 report
- Related prior finding: a t50-era codekb placement gap was caught by trace-reading a green test, so ndjson/trace inspection (not just pass/fail) is the right verification method here

apackeer:
On multi-repo intents the stage-completion artifact guard accepted a single codekb artifact anywhere under the space codekb root, so a Kiro ACP / Codex exec conductor that drifted (skipped a repo's reverse-engineering pass or wrote to a hand-composed dir) could still gate and approve. The guard now requires at least one declared artifact in every recorded repository's canonical `aidlc/spaces/<space>/codekb/<repo>/` dir, refusing with the missing repos and the `codekb-path` resolver command; intents with no recorded repo set keep the legacy any-repo fallback. Deterministic, tested in t185; version 2.6.20.

**Overlap note (#839):** open PR #839 (issue #741) touches the same seam — it rewrites the `producesDirsForStage` codekb arm in `core/tools/aidlc-state.ts` with a stricter per-repo completion matrix (complete declared artifact set per registered repo, repo-scoped `ARTIFACT_REUSED` reuse exemptions) and claims the same 2.6.20 version slot. The mechanisms overlap and the diffs will conflict; this PR is the narrower guard (>=1 declared artifact per recorded repo). If #839 lands first, its matrix subsumes this check and this PR should be rebased down to its remaining delta (or closed in favor of #839); whichever merges second re-bumps the version per the changelog conflict policy.

Closes #514


leandrodamascena:
Addressed the remaining blockers and updated the branch to the current `v2`.

- Reverse Engineering now requires all nine declared CodeKB artifacts for every registered repository.
- Pipeline completion requires ordered, current-attempt receipts for each scanned repository.
- Rejections, stage jumps, and later stage starts invalidate stale receipts.
- Repo-scoped reuse only grants an exemption for `Decision=keep`; `modify` and `redo` do not.
- Main and isolated runs remain separated, and interrupted pipelines resume from the first missing link.
- Consolidated the overlapping pipeline-receipt work from #839 and bumped the release to `2.6.39`.

Validation:
- `bun run check`
- affected unit tests: 391 assertions passed
- pipeline receipt regressions: 10/10 passed
- artifact guard integration: 28/28 passed

Pushed in `1414a5f40`.

leandrodamascena (approved):

PR #853, opened 2026-08-21
fix: make Testing Posture HTML-comment-aware (2.6.38)

### LINKED ISSUE #841: [Bug]: aidlc-testing-posture resolver treats a COMMENTED-OUT ## Testing Posture example as an affirmed methodology — fresh projects silently get BDD
### Summary

`aidlc-testing-posture.ts render` resolves `"methodology": "bdd"` with `"source": "team"` on a project where **no posture was ever affirmed** — the only BDD text in `team.md` is the shipped template's commented-out example. The resolver appears to read the raw `## Testing Posture` section text without stripping HTML comments, so the example line counts as a real affirmation.

Impact: every fresh project that keeps the shipped `team.md` gets a **BDD scenario-first** code-generation plan instead of the documented org default (`test-after`), silently, with no affirmation and (in a custom scope) without practices-discovery having run. It also propagates into the plan's embedded Testing Contract + `contract_sha256`, so the wrong methodology is fingerprinted into the Plan Approval guard.

### Reproduce

On a project with the **unmodified** shipped `team.md` `## Testing Posture` section (comments only):

```
## Testing Posture
<!-- Affirmed during practices-discovery. Example: -->
<!-- We use BDD. Specifications drive scenarios; scenarios drive code. -->
<!-- Each Unit ships with feature files in /features/. -->
```

Run:

```
bun .kiro/tools/aidlc-testing-posture.ts render
```

Observed (abridged):

```json
{
  "methodology": "bdd",
  "source": "team",
  "ordering": "Define executable behavior scenarios before implementing each observable feature slice.",
  "applicable_notes": [
    { "layer": "team",
      "text": "<!-- Affirmed during practices-discovery. Example: -->\n<!-- We use BDD. Specifications drive scenarios; scenarios drive code. -->\n<!-- Each Unit ships with feature files in /features/. -->" }
  ]
}
```

The `team` layer `text` the resolver selected from is 100% HTML comments, yet `methodology` resolves to `bdd` and `source` to `team`.

### Expected

An HTML-commented example is not an affirmation. With no real `Methodology`/`Ordering` fields in any layer, the resolver should fall through to the org default (`methodology: "test-after"`, `source: "org"`). Comment content under `## Testing Posture` must be stripped before methodology detection.

### Likely origin

This looks like a regression/over-correction from **#724** ("code-generation plan ignores the affirmed `## Testing Posture` — TDD teams get an implementation-first plan"). #724 correctly made the resolver *read* `team.md`'s Testing Posture; this bug is that it now reads *too much* — it does not exclude the commented example lines the template ships with. The two together suggest the detector should: (a) strip `<!-- ... -->` (and ideally require the explicit `Methodology:`/`Ordering:` field shape described in org.md) before matching a methodology keyword, and (b) keep #724's affirmed-posture behavior for real, uncommented fields.

### Suggested fix

- In the posture resolver, strip HTML comments from each layer's `## Testing Posture` body before keyword/field detection.
- Prefer matching the documented explicit shape (`**Methodology**: <tdd|bdd|atdd|test-after>` / `**Ordering**: ...`) rather than a bare keyword anywhere in the section, so prose that merely mentions "BDD" doesn't flip the methodology.
- Add a regression test: unmodified shipped `team.md` (comments only) → `methodology: test-after`, `source: org`.

### Environment

Harness: Kiro IDE, v2 base. Surfaced live during a Construction code-generation Plan step (the wrong `bdd` contract would have been fingerprinted into Plan Approval). Related: #724 (closed).

apackeer:
The posture resolver classified methodology from raw section text, so the shipped team.md commented-out BDD example silently affirmed BDD for every fresh project and fingerprinted the wrong Testing Contract into Plan Approval. Comment content is now stripped (via the existing multi-line-aware visibleMarkdownLines) before structured-field/keyword detection and applicable_notes, while raw sections still feed input_sha256 and visible prose/structured affirmations keep their #724 behavior; regression tests pin the unmodified shipped templates to methodology=test-after, source=org. Closes #841

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.38] - 2026-08-21
+
+Testing Posture resolution now ignores HTML-commented examples while preserving visible prose, structured fields, and fenced notes. **Upgrade:** refresh your `dist/<harness>/` shell; contracts whose Testing Posture sections contain comments will re-resolve, so re-approve affected Code Generation plans.
+
+* Commented headings and comment-only `Testing Posture` sections no longer select, truncate, or affirm a methodology; the resolver falls through to the real visible section or next visible memory layer.
+* Visible `Methodology` and `Ordering` fields remain authoritative beside comments, and visible prose and fenced content remain in `applicable_notes`.
+* Testing Contract input fingerprints retain each raw resolved section, including comments and fenced content, so comment-only changes still invalidate stale approvals.
+
 ## [2.6.37] - 2026-08-21
 
 Code-generation review receipts are now bound to the workspace source state the reviewer inspected, and every completion route refuses when that state is stale. The same binding follows autonomous swarm work from finalize through source merge. **Upgrade:** refresh your `dist/<harness>/` shell so the source-fingerprint tools, guards, audit schema, and generated harness files are installed.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.37-blue)
+![version](https://img.shields.io/badge/version-2.6.38-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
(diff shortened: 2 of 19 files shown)


leandrodamascena:
Maintainer follow-up pushed at `8b1dd0bfd`.

The Testing Posture parser is now HTML-comment-aware without dropping visible Markdown:

- Commented `## Testing Posture` headings and nested headings can no longer select or truncate the real section.
- Commented methodology examples and fences cannot affect classification.
- Visible fenced notes remain in `applicable_notes`, while methodology examples inside fences do not affirm a posture.
- Same-line inline-code literals such as `<!--` remain visible and do not start comments.
- Escaped backticks, mixed fence markers, invalid backtick fence info strings, trailing heading comments, and multiline-comment boundaries are handled fail-closed.
- The raw selected sections, including comments and fenced content, remain in `input_sha256`, so comment-only changes still invalidate stale approvals.
- The branch was rebased onto current `v2`, regenerated, and rebumped to `2.6.38`.

Validation:

- Focused regression suite: 88 tests, 1,643 assertions, 0 failures.
- Testing Posture plus version synchronization: 45 tests, 221 assertions, 0 failures.
- Deterministic integration/E2E completed with one unrelated `t99` timeout; `t99` passed in isolation with 14 tests and 42 assertions.
- The smoke/unit sweep retained only the existing unrelated failures in `t248`, `t255`, and `t298`.

The PR is ready for CI completion and final approval.

leandrodamascena (approved):

PR #855, opened 2026-08-21
fix: emit ARTIFACT_* audit rows for relative-path native writes (Kiro CLI conductor deadlock) (2.6.19)

apackeer:
## Problem

On Kiro CLI, the main session's `write`/`fs_write` tool passes a **model-composed** `tool_input.path` that can be project-relative. The write-audit hook's path-gate compared it verbatim (`startsWith`) against absolute record/codekb roots, so every conductor-authored `produces[]` artifact write fell out at "not under record or codekb root" and emitted **no** `ARTIFACT_CREATED`/`ARTIFACT_UPDATED` row.

Downstream this deadlocks the workflow: `aidlc-state gate-start`'s completion precondition finds no native-tool write after the consolidated summary confirmation and refuses the approval gate, while every workaround is blocked by design — re-saving trips the review-freeze (which *does* normalize relative paths), and manual append of `ARTIFACT_*` is CLI-reserved. Stages whose artifacts are subagent-authored were unaffected only because those sessions happened to pass absolute paths.

From a user-filed report with `AIDLC_HOOK_DEBUG` evidence: `path-gate tool="Write" file="aidlc/spaces/default/intents/…" recordRoot="/Users/…/intents/…" underRecord=false` → `exit: not under record or codekb root`; the identical write with an absolute path emits correctly.

## Fix

- **`core/hooks/aidlc-write-audit-log.ts`** — absolutize `file_path` against `projectDir` before the path-gate; the absolutized form flows into the audit `File:` field, context breadcrumb, and CREATE-vs-UPDATE stat probe. Fixes every harness at once.
- **`harness/kiro/hooks/aidlc-kiro-adapter.ts`** (`audit-and-sensors`) — absolutize before forwarding, matching the codex adapter idiom (`isAbsolute(p) ? p : join(projectDir, p)`); kiro-ide already did this.
- **`core/hooks/aidlc-run-sensors.ts`** — same normalization before the `.aidlc-sensors/` recursion guard (which a relative path silently bypassed) and glob matching; corrected the falsified "always an absolute path" comment.

Deliberately **not** taken from the report: a gate-start mtime fallback — it would weaken the audit-backed invariant; fixing the row producer removes the deadlock.

## Why tests missed it, and the coverage added

The kiro payload fixture was live-captured once with an absolute path (the shape is model-chosen, not harness-fixed); every unit test built `file_path` via `join(proj, …)`; and the one test on this seam asserted only exit 0 against an always-exit-0 advisory hook — structurally blind to a silently dropped row. Added:

- `t07`: relative `file_path` under the record root emits an `ARTIFACT_*` row with an absolute `File:` field (**revert-checked**: fails without the core fix).
- `t147` test 7: upgraded from exit-code-only to asserting the audit row lands from a relative Kiro payload.
- `t94`: relative-path write into the sensors detail dir is caught by the recursion guard.

## Release

`2.6.19` — version, CHANGELOG entry (upgrade: refresh `dist/<harness>/` shell copies), README badge; all dists regenerated, `bun scripts/package.ts --check` green.

## Evidence

- smoke+unit: 236 files / 6847 assertions, 0 failed (`tests/logs/2026-08-21T07-11-01Z-p3`)
- integration: 104 files / 1447 assertions, 1 environmental red (`t145`, missing worktree `node_modules` for its scratch symlink) re-run green-alone after `bun install` (`tests/logs/2026-08-21T07-27-11Z-p3994994`, rerun `2026-08-21T07-46-08Z-p4157211`)
- typecheck + biome clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.19] - 2026-08-21
+
+Kiro CLI conductor writes now preserve the audit-backed artifact lifecycle when the native write tool reports a project-relative path, and sensor dispatch applies the same normalization before its recursion guard and glob matching. **Upgrade:** refresh your `dist/<harness>/` shell copies so the updated adapters and core hooks are installed.
+
+* Kiro CLI `write` / `fs_write` events now emit `ARTIFACT_CREATED` or `ARTIFACT_UPDATED` for relative paths under the active record, fixing the "no recorded native-tool write after the human's consolidated summary confirmation" gate refusal and resulting workflow deadlock.
+* The run-sensors hook now resolves relative write paths against the project before checking the `.aidlc-sensors/` recursion guard and applicable sensor globs.
+
 ## [2.6.18] - 2026-08-19
 
 Classic and Express are new scope options, and the implicit default scope is now Classic — a **declared behavior change**: invocations that name no scope and match no keyword now run the v1-style lifecycle without Ideation instead of the full-lifecycle Feature scope. Exactly two things control the implicit default: the `AWS_AIDLC_DEFAULT_SCOPE` env var (which overrides) and the framework's hard-coded `classic` fallback. Express has a deterministic requirements-to-conditional-deploy path, and conditional protocol modules reduce fixed context without dropping reviewer recovery behavior. **Upgrade:** refresh your `dist/<harness>/` shell; in-flight workflows keep their persisted scope and need no migration; set `AWS_AIDLC_DEFAULT_SCOPE=feature` (Claude: the `.claude/settings.json` `env` block, which now ships `classic`) to keep the previous full-lifecycle default.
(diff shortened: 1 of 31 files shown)


leandrodamascena (changes requested):
**Request changes**

This PR has been superseded by #788 (`2.6.56`), which fixes the original Kiro failure with broader coverage:

- Normalizes `path`, `file_path`, `paths[]`, and `operations[].path`.
- Resolves project-relative paths before invoking audit and sensor hooks.
- Supports Write/Edit aliases, replacements, appends, and batched writes.
- Verifies relative batches emit both `ARTIFACT_*` and `SENSOR_FIRED` rows.

The remaining core-hook normalization is reasonable defense in depth, but no current harness relies on it: adapters normalize paths before forwarding, and Claude supplies absolute paths directly.

This branch is also based on `2.6.19` and conflicts with current `v2` at `2.6.61`, including the Kiro adapter, `t147`, changelog, version, and generated distributions.

I recommend closing this PR as superseded. If the core-hook hardening is still desired, it should be submitted separately against current `v2`, without duplicating the merged Kiro implementation and tests.

The existing head passes 66 targeted tests, package parity, typecheck, and Biome; the blocker is duplication and stale integration, not a separate defect in the original patch.

apackeer:
Closing as superseded, per your review - #788 (2.6.56) covers the Kiro adapter path normalization and this branch duplicated it on a stale 2.6.19 base.

The core-hook hardening you said would be reasonable separately is now #880, submitted against current v2 with no adapter or t147 changes.

PR #858, opened 2026-08-21
feat: per-session workflow bindings for concurrent same-space intents (2.6.80)

apackeer:
## Problem

Two live sessions in one checkout previously shared `aidlc/active-space` and
`aidlc/spaces/<space>/intents/active-intent`. A switch or intent creation in one
session could silently redirect another session's workflow selection and
attribution.

## Solution

- Each session records a machine-local `{space, intent}` binding under
  `aidlc/.aidlc-sessions/`. Explicit selectors win, then the session binding,
  then the existing shared-cursor and lone-intent fallbacks.
- Cold sessions persist an explicit null-intent binding before any workflow
  exists. Intent creation and intent/space switching move the binding, UUID
  attribution, usage ownership, and lifecycle ownership together.
- `AIDLC_SESSION_OVERRIDE` and validated PID ancestry provide inherited session
  identity for spawned tools. Conflicting identities refuse before workflow
  paths are derived.
- Resume/rebind offers and fresh-intent handoffs are receipted per session.

## Review Follow-Up

All findings from the latest review are addressed:

- `aidlc-learnings surface` resolves one session-aware workflow selection before
  reading state, runtime graph, or stage memory.
- Null-intent usage attribution remains
  `record:<bound-space>/legacy` after another session moves the shared cursor.
- The Stop hook validates fresh-intent handoffs against the owning session's
  binding and UUID stamp, not the shared cursor, and retains a non-expired
  receipt when ownership is temporarily unresolved.
- On POSIX Codex, the adapter pins the validated hook payload session into every
  Bash command and core-hook child. Sandboxed macOS therefore does not depend on
  `ps`; a denied-`ps` regression seam covers this path.
- The rebase preserves the newer token-first load-steering recovery order while
  retaining env-inherited session identity.
- Rebase overlaps with current source-freshness behavior preserve omitted
  intent selection rather than converting it to explicit null.

The branch is rebased onto `v2` at `1c0188a2` (`2.6.79`) and this change is
re-slotted to `2.6.80`.

## Known Limitation

Harness-native method includes remain one mutable checkout-wide surface.
Concurrent intents within one space are isolated, but simultaneous sessions in
different spaces can still race on ambient delivered rules. Windows and
shared-process harness sessions retain shared-cursor behavior where no distinct
identity is available.

## Verification

- `bun run typecheck`
- `bun run lint`
- Post-rebase regression slice: 20 files, 473 assertions, PASS.

Codex users upgrading this change must refresh `dist/codex/` and re-run:

`bun scripts/package.ts codex trust --project <absolute-project-path>`


leandrodamascena (changes requested):
I reviewed this PR against commit `0abb3470` with an adversarial pass. The core binding helpers work in the covered scenarios, but I found three isolation gaps that still allow cross-session state or context contamination.

**P1: Persist cold-session bindings before the no-workflow return**

`SessionStart` resolves a cold session as `{ space: "default", intent: null }`, but returns when no state file exists before calling `writeSessionBinding`.

I reproduced the resulting race:

1. Session A starts in an empty workspace.
2. Its binding remains absent.
3. Session B creates an intent and moves the shared cursor.
4. Session A resolves through the fallback cursor and silently selects Session B's intent.

This recreates the active-intent stealing that the PR is intended to prevent. Persist the resolved null-intent binding before the early return and add an integration test where two sessions start before either creates an intent.

**P1: Do not load another session's space method through the global include**

Session workflow selection is binding-aware, but `SessionStart` still calls `repointHarnessIncludes(projectDir, activeSpace(projectDir))` using the shared cursor.

I reproduced Session A resolving its state from `default` while every Claude `@` import pointed to `team-b` after Session B moved the cursor. Session A therefore executes the correct workflow with another team's ambient method.

Changing the argument to `selection.space` fixes the immediate mismatch but does not eliminate the underlying race: the harness include is one mutable checkout-wide file, so two concurrent sessions in different spaces can still overwrite each other's ambient method. Please either provide a session-local method-delivery mechanism or explicitly limit this increment to concurrent intents within one space and retain the existing multi-space limitation. Add a two-session, two-space regression test that verifies both workflow selection and delivered rules.

**P1: Move usage and lifecycle attribution when intent creation rebinds a session**

Creating a second intent updates the session binding but preserves the previous UUID stamp. The PostToolUse binding path explicitly returns without replacing an existing stamp.

I reproduced a session moving from intent A to a newly created intent B:

- The shared cursor and session binding pointed to B.
- `readSessionIntentUuid` still returned A.
- `intentUsageKey` continued charging usage to A.

`SessionEnd` also uses this stamp, so its lifecycle event can be written to the previous workflow after the session has moved to the new one. The same stale-stamp condition can occur when switching to an empty space because the binding becomes `intent: null` without clearing the previous stamp.

Please make binding, usage attribution, and lifecycle ownership transition together, while preserving any handoff receipt needed by continuation logic. Add coverage for creating a second intent from an already stamped session and switching from an intent-bearing space to an empty space.

**P2: Rebase and refresh release metadata**

The branch is 31 commits behind `v2`, currently conflicts, has no current CI result, and still claims version `2.6.36`, which is already present in the current history. Please rebase onto the current `v2`, reconcile the overlapping session, continuation, engine, and generated-tree changes, then move the title, body, version source, README badge, and changelog to the next available slot after `2.6.65`.

The new `t310`, `t311`, and `t312` suites pass, as do package parity, typecheck, and lint. The blockers above come from concurrent lifecycle transitions that those tests do not currently cover.

leandrodamascena (changes requested):
I reviewed this PR against commit `f62a0c330` with an adversarial concurrency pass. The previous findings are addressed, but I found four remaining isolation gaps.

**P1: Resolve the learnings ritual through the session binding**

`aidlc-learnings surface` still resolves its workflow with `activeSpace()` and `activeIntent()`, bypassing `resolveWorkflowSelection()`.

I reproduced Session A being bound to intent A while the shared cursor pointed to B. Running `surface` from A returned B’s intent and learning candidates. This can surface or persist learnings from the wrong workflow.

Please resolve `{space, intent}` once through the session-aware selection path and add a two-session regression test where the binding and shared cursor disagree.

**P1: Keep null-intent usage attribution bound to its session**

`intentUsageKey()` consults the session UUID stamp, but when that stamp is absent it falls back directly to the shared space and intent cursors.

I reproduced a session bound to `{space: "team-empty", intent: null}` receiving the UUID usage key of another session’s active intent after the shared cursor moved. Usage from the empty-space session is therefore charged to an unrelated workflow.

Please derive the fallback key from the session binding, including `record:<bound-space>/legacy` for a null intent, and test after another session moves the cursor.

**P1: Validate intent handoff against session ownership, not the shared cursor**

The Stop hook accepts a fresh intent-creation handoff only when `activeIntentUuid()` still matches the receipt.

I reproduced Session A creating intent A2 and receiving a valid handoff receipt, followed by Session B moving the shared cursor before A’s Stop hook ran. A’s Stop was incorrectly blocked, and the valid receipt was cleared unconditionally, preventing a retry.

Please validate the handoff against A’s binding and UUID stamp. Add a regression test that moves the global cursor between intent creation and the owning session’s Stop event.

**P1: Do not depend on sandbox-blocked `ps` for Codex/macOS isolation**

Darwin ancestry resolution depends exclusively on spawning `ps`. In a sandboxed Codex session, `ps` is denied, so ancestry resolution silently returns null and every unpinned tool falls back to the shared cursor.

The ancestry suites pass outside the sandbox but fail inside the actual Codex sandbox, including workflow selection, rebinding, and divergent-identity checks. This means the PR’s central isolation guarantee does not hold for a declared harness/platform combination.

Please provide a session-identity path that works without sandbox-forbidden process inspection, or explicitly limit and document the unsupported environment rather than claiming Codex session isolation there. Add a denied-`ps` regression seam.

**P2: Rebase and re-slot the release version**

The branch is four commits behind `v2`, is currently conflicting, and overlaps recent changes in the core selection and orchestration paths. Both the current base and this PR claim `2.6.69`.

Please rebase onto the current `v2`, reconcile the functional overlaps, regenerate `dist`, and move the title, version source, README badge, and changelog to the next available version.

The existing learnings and usage suites pass, but they do not exercise these interleavings. The failures above were reproduced directly against the current PR head.

apackeer:
Review update for `1226bfd8c`:

### Latest findings addressed

- **Learnings ownership:** `aidlc-learnings surface` now resolves one session-aware `{space, intent}` selection before reading state, runtime graph, or stage memory. A binding-vs-cursor regression proves the bound session's candidates win.
- **Null-intent usage:** usage fallback now derives from the session selection, preserving `record:<bound-space>/legacy` after another session moves the shared cursor.
- **Fresh-intent handoff:** Stop validates the receipt against the owning session's binding and UUID stamp rather than the shared cursor. Non-expired receipts are no longer consumed on a transient ownership mismatch.
- **Codex/macOS identity:** POSIX Codex pins the validated hook payload session into every Bash command and core-hook child. A denied-`ps` seam proves sandboxed macOS no longer depends on ancestry inspection.

### Rebase reconciliation

- Preserved omitted intent selection through the source-freshness paths instead of coercing it to explicit null, restoring recorded multi-repo source ownership.
- Removed the stale intermediate `report_command` conductor wording; the engine and all seven harness skills retain the current v2 directive contract.
- Re-slotted release metadata, changelog, badge, and generated trees to **2.6.76**.

### Verification

- `bun run typecheck`
- `bun run check`
- Post-final-rebase regression slice: **15 files, 374 assertions, PASS**
- GitHub CI, deterministic tests, documentation, Semgrep, Gitleaks, and Bun audit: **green**

The documented limitation remains: workflow selection is isolated for concurrent intents within one space, while ambient method delivery is still checkout-global across spaces.

Note: `v2` advanced again after this push (currently `e18e4cbe`), so GitHub now reports a base conflict. This comment describes the verified `1226bfd8c` head; a further rebase has not been force-pushed.

leandrodamascena (approved):

PR #880, opened 2026-08-23
fix: normalize relative paths in core write hooks (2.6.75)

apackeer:
## Summary

- Resolve project-relative `tool_input.file_path` values in the shared audit and sensor hooks before any path-sensitive behavior.
- Preserve absolute-path behavior, gate-only sensor isolation, audit path redaction, and regenerate all harness distributions.
- Add focused coverage for audit logging, sensor dispatch, and the sensor recursion guard.

## Why this is separate from #788

#788 normalized Kiro adapter payloads. This PR does not change any adapter or `t147`.

On the current `v2` base, the shared hooks still consumed the raw payload in `core/hooks/aidlc-write-audit-log.ts` and `core/hooks/aidlc-run-sensors.ts`; the sensor hook also documented absolute paths as an assumption. Shipped adapters already normalize paths, so this is defense in depth: the shared hooks now enforce the invariant they previously only assumed.

## Tests

- `t07`: project-relative artifact paths pass the absolute path gates and retain `<project-dir>` redaction in the recorded `File` value.
- `t94`: project-relative paths reach write-fired sensor globs and `--output-path` as absolute paths, the recursion guard catches relative sensor-detail writes, and gate-fired sensors remain excluded from per-write dispatch.
- `t305`: the Build-and-Test prose pin recognizes the quality-target-aware failure ladder from `v2`.
- 86 focused `t07`, `t68`, `t94`, and `t305` tests pass.
- `bun run typecheck`
- `bun run lint`

Version slot: `2.6.75`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.75] - 2026-08-24
+
+Shared write hooks now normalize project-relative file payloads before applying audit and sensor path logic, enforcing the hooks' absolute-path invariant even if an adapter boundary is bypassed. **Upgrade:** refresh your `dist/<harness>/` shell; existing workflows require no migration.
+
+* Artifact audit logging resolves relative record paths before its path gates while preserving project-prefix redaction in the recorded `File` value.
+* Sensor dispatch resolves relative paths before its recursion guard, glob match, and `--output-path` argument while preserving gate-only sensor isolation.
+
 ## [2.6.74] - 2026-08-24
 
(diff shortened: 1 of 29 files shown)


leandrodamascena (approved):
I re-reviewed this against commit `7017bb499` after rebasing onto the current `v2`.

The shared hooks now normalize project-relative paths before audit and sensor path-sensitive behavior. The rebase preserves both gate-only sensor isolation and project-prefix redaction in audit records.

Tests now explicitly verify write-fired dispatch, exclusion of gate-fired sensors, recursion protection, redacted audit paths, and version synchronization. All generated distributions are synchronized at `2.6.75`.

The 44 focused tests, package drift check, type checks, lint, and coverage-registry validation pass.

No remaining blockers from my review. Approved.

PR #881, opened 2026-08-23
fix: scope-aware validity resolution and zero-Unit lifecycle consistency (2.6.77)

apackeer:
## Summary

Three follow-ups recorded during the stage-validity review (#716, merged as 2.6.62). Version metadata: 2.6.66.

- **Scope-aware artifact resolution.** The validity resolver treated every per-unit stage as Unit-expanded, so express workflows (no Unit DAG, stage-level construction paths) never had their Code Generation / Build and Test outputs fingerprinted, and the no-DAG directory fallback could pick up stale per-unit directories. The resolver now mirrors the engine's existing effective-plan decision (shared helpers in `aidlc-lib.ts`, also used by the orchestrator and the approval artifact guard): when the approved plan skips Units Generation, a per-unit stage resolves one stage-level instance before any DAG or directory scan. Ambiguity remains fail-open (Validation Warning at capture, inspection warning at `next`/`--status`). The known-limitation paragraph in the state-machine reference is replaced by the new behavior.
- **Seeded codekb parent at intent birth.** `t-acp-kiro-journey-workspace` was red on v2: the live agent list-probes `aidlc/spaces/<space>/codekb`, which the default space never created (named spaces already ship the parent), producing an orphaned failed tool call. Birth now ensures the empty `codekb/` parent; per-repo stores stay lazy per the existing contract.
- **Explicit t304 timeouts.** `t304-source-freshness-receipts` tripped bun's 5s default per-test timeout under `-P 8` load (5.18s, green alone). The slow CLI-chain tests now carry explicit 60s timeouts, following the t188/t224/t300 precedent from #852.

## Validation

- Focused slice (unit+integration+e2e): 15 files / 410 assertions, 0 failed - includes the three new express-resolution cases in t310, the extended birth assertion in t27, t304-source-freshness 57/57, the deterministic express e2e journey (t301), and the version/CHANGELOG sync pin (t68).
- Live Kiro ACP journey (`AIDLC_KIRO_ACP_LIVE=1 t-acp-kiro-journey-workspace`): PASS 42/42 assertions - the previously red base test, green with the birth fix.
- `bun scripts/package.ts --check`, typecheck, and lint all green; coverage registry regenerated (function coverage 171 -> 173, no class decreased).


leandrodamascena (approved):
I reviewed this again against commit `c9782b7d5` after rebasing onto the current `v2`.

The adversarial findings are resolved:

- The effective zero-Unit plan is now authoritative across routing, completion evidence, summary/reviewer cardinality, swarm eligibility, and Unit lifecycle commands, even when a stale valid DAG remains.
- Stage-level review fingerprints now bind the actual artifacts and reject mutations between review request and verdict.
- Validation Basis schema 3 makes existing schema-2 receipts fail open as untracked instead of falsely reporting unchanged workflows as stale.
- Upstream coverage now resolves consumes from stage-level producer directories in zero-Unit workflows.

The release was re-slotted to `2.6.76`, generated distributions and coverage metadata were refreshed, and the focused suites, full `t304`, parity, typecheck, lint, version synchronization, and coverage checks pass.

Approved.

PR #882, opened 2026-08-24
fix(copilot): deny native question pickers during active workflows (2.6.91)

### LINKED ISSUE #840: [Bug]: VS Code AI-DLC asks workflow questions twice when native question tools are used
## Bug report

**Title**  
VS Code AI-DLC asks workflow questions twice when native question tools are used

**Summary**  
In VS Code agent mode, if AI-DLC asks a workflow question via a native picker/tool first, the selected answer comes back as a tool result rather than a real chat turn. AI-DLC’s human-presence guard then rejects the answer, so the agent asks the same question again as plain text.

**Repro**  
1. Start an AI-DLC stage that asks workflow questions.
2. The agent uses a native question tool first.
3. Select an answer in the picker.
4. AI-DLC tries to record the answer with `aidlc-log.ts`.
5. The log step refuses because no trusted human turn was recorded.
6. The agent falls back to asking the same question again in chat.

**Expected**  
One question, one answer, no duplicate prompt.

**Actual**  
The same question is asked twice:
- first via picker/tool
- then again as plain text in chat

## Root cause

This is an enforcement gap in the Copilot harness:

1. The harness docs explicitly say **not** to use native picker tools for AI-DLC questions.  
   See `question-rendering.md` and `SKILL.md`.

2. Why: picker answers do **not** fire the trusted `UserPromptSubmit` hook that records `HUMAN_TURN`.  
   See `question-rendering.md`, `aidlc.json`, and `aidlc-copilot-adapter.ts`.

3. AI-DLC answer logging **requires** that `HUMAN_TURN`.  
   See the refusal in `aidlc-log.ts`.

So the flow is:

`picker answer` -> **no `HUMAN_TURN`** -> `aidlc-log.ts answer` rejects -> agent re-asks in plain text

## Why you may only notice it in VS Code

The framework already assumes both Copilot surfaces are unsafe for picker-based workflow questions, but in practice the bug only appears when the model actually uses the picker path. VS Code seems to hit that path more readily. The CLI likely “works” because it followed the required numbered-prose path in your successful runs.

## Severity

**Medium-high**  
- breaks workflow UX
- causes duplicate questioning
- can stall approval/question flows
- undermines trust in the harness

## Recommended fix

**Best fix: block native question tools for AI-DLC workflow questions in Copilot harness mode.**

Specifically:
1. Add a `PreToolUse` deny for question-picker tools during active AI-DLC workflow steps.
2. Force workflow questions through the numbered-prose path defined in `question-rendering.md`.
3. Do **not** “fix” this by treating picker tool results as trusted `HUMAN_TURN` events unless GitHub provides a deterministic verified event for them.

## One-line diagnosis

AI-DLC relies on `UserPromptSubmit` for trusted human answers, but VS Code picker answers bypass that hook; because the harness does not technically block picker usage, the agent can ask once with a picker and then has to ask again in chat.

If you want, I can turn this into a ready-to-file GitHub issue or patch the harness to prevent it.

### Steps to Reproduce

Image

### Expected Behavior

Looks like the harness is supposed to not use askquestion due to hook concerns, which is a shame.

### Actual Behavior

Image

### AI-DLC Version

v2 (alpha)

### Release / Commit

4d0968fc1891f98a06a4404d76f526f0e9901db2

### AI-DLC Phase

### Platform / IDE

GitHub Copilot

### AI Model

_No response_

### Environment

macOS: 26.6 (build 25G72)
VS Code: 1.134.0
GitHub Copilot CLI: 1.0.80
Architecture: arm64

### Additional Context

_No response_

apackeer:
Copilot surfaces (VS Code agent mode especially) could ask an AI-DLC workflow question twice: the model would call a native picker (`ask_user` / `vscode/askQuestions`), whose answer returns as a tool result that never fires `UserPromptSubmit`, so no `HUMAN_TURN` is recorded, `aidlc-log.ts` refuses the answer, and the question is re-asked in chat. The Copilot adapter's matcher-free PreToolUse guard now denies the native picker execution IDs while workflow state exists, directing the model to render numbered prose and end the turn; without workflow state the guard fails open so non-AI-DLC projects keep native pickers, and picker results are still never treated as trusted human turns.

Note: this branch carries version slot 2.6.24 as committed; origin/v2's CHANGELOG head has since moved past it, so the slot re-bumps at landing per the changelog conflict policy (rebase + re-bump by the second-to-merge).

Closes #840


leandrodamascena (changes requested):
Thank you for addressing #840. Blocking untrusted picker results from bypassing the human-turn contract is the right direction.

I reviewed the current head `41d024f148bcd0b552535c2c593cc25c43786390` against `48f106b9e75d14e35f94546645005f1dbf81b498`. Project alignment is established by the linked issue’s explicit recommendation and this maintainer-owned implementation.

### P2: Completed workflows keep native Copilot pickers disabled indefinitely

The new guard checks only whether `aidlc-state.md` exists (`harness/copilot/hooks/aidlc-copilot-adapter.ts:939-942`). Workflow completion retains that file and sets `Status: Completed`, so every later `ask_user` or `vscode/askQuestions` call in the workspace is still denied, even though no AI-DLC workflow is active.

I reproduced this using the repository’s completed-state fixture: `ask_user` returned `permissionDecision: deny` with `Status: Completed`. This contradicts the documented “during an active workflow” boundary and prevents ordinary Copilot picker use after AI-DLC finishes.

Please gate the denial on the selected session-bound workflow actually being nonterminal, fail open for completed or unusable state, and add regression coverage for a completed workflow.

I did not modify or push the maintainer-owned branch. Validation performed:

- Original `t249` tests: 49 passed, 2 compiled-binary cases skipped
- Diagnostic completed-state test failed exactly as described and was removed afterward
- `t248`, `t250`, and `t68`: 39 tests, 518 expectations passed
- Typecheck passed
- Lint passed with three pre-existing informational suggestions in `t267`
- Coverage registry check passed

GitHub reports no CI checks for this branch. The PR is also conflicting: its version is `2.6.24`, while the current base is `2.6.83`, so it needs a rebase, a fresh version slot, and validation of the resolved tree. The latest base change does not alter this finding.

leandrodamascena (approved):

PR #883, opened 2026-08-24
feat: doctor advisory for duplicate producers in the loaded stage graph

### LINKED ISSUE #831: [Feature]: doctor advisory for duplicate producers in the loaded stage graph (defense-in-depth behind the compile guard)
## Context

#829 (closes #804) makes `aidlc-graph compile` reject an artifact with two or more producers whenever any stage consumes it, because runtime consume resolution takes `producersOf(name)[0]` (`core/tools/aidlc-orchestrate.ts:1539`, `:1589`; `core/tools/aidlc-sensor.ts:297`, `:312`) and would otherwise pick an arbitrary producer by load order.

That guard is compile-time only. Nothing validates producer uniqueness on the **loaded** graph:

- a hand-edited or stale `stage-graph.json` bypasses it (the file is generated, but nothing at load time re-checks the invariant)
- the `AIDLC_STAGE_GRAPH` env override (`core/tools/aidlc-graph.ts:259`) points the engine at an arbitrary JSON that never went through compile
- `validateGrid`/`validateScope` check only zero-producer orphans and off-path producers (`core/tools/aidlc-graph.ts:1178-1198`) — a duplicate-producer consumed artifact in the loaded graph still resolves silently to `[0]`

## Proposal

Teach `/aidlc --doctor` (or the grid validation it delegates to) to surface "consumed artifact has N producers" on the loaded graph — advisory tier is enough, since a graph produced by compile can no longer contain the defect and the check exists to catch out-of-band edits.

Optional fold-in while touching the area: the three duplicate guards (stage slug, sensor id, consumed-artifact producer) each throw on the first collision found; collecting all collisions before throwing would save plugin authors one compile round-trip per defect.

Low urgency: normal installs regenerate `stage-graph.json` through compile, so this is defense-in-depth for the out-of-band paths, filed so the gap is on record.

apackeer:
Adds a `Duplicate producers` advisory row to `/aidlc --doctor`: every consumed artifact in the LOADED stage graph is checked for a single producer, catching hand-edited/stale stage-graph.json and AIDLC_STAGE_GRAPH overrides that bypass compile, where runtime silently resolves producersOf(artifact)[0] by load order. Advisory tier (never fails doctor), with the ambiguous artifacts and their producer slugs named in the row and classified as a warning in the doctor bundle. Note: the version slot is committed as 2.6.23; origin/v2 CHANGELOG head has since moved past it, so the slot re-bumps at landing per the changelog conflict policy (rebase + re-bump by the second-to-merge). Closes #831


leandrodamascena (changes requested):
Thanks for adding the loaded-graph defense behind the compile-time guard from #829. This addresses the out-of-band graph paths described in #831 while keeping duplicate producers advisory-only.

I reviewed the current head e2c07aef712cc8444a0a7d471024198b02280e1b against b44b9c9c1ad1afeb351f1543b384936cd3d16839. The change is project-aligned: it is maintainer-owned, directly implements #831, and remains scoped to doctor diagnostics, documentation, and regression coverage.

[P2] Preserve graph load order when reporting the selected producer
core/tools/aidlc-graph.ts:851

Runtime selects `producersOf(artifact)[0]`, and `producersOf` preserves graph load order. This helper sorts the producer slugs alphabetically before the doctor renders them. With a loaded graph ordered as `[stage-z, stage-a, consumer]`, runtime selects `stage-z`, but the diagnostic reports `[stage-a, stage-z]`. Because the message says runtime resolves the first producer by load order, the first displayed producer can be wrong.

Please preserve producer load order or return and render an explicit `selectedProducer`. Add a regression case where graph order differs from alphabetical order. That case should also confirm the duplicate-producer advisory does not change an otherwise healthy doctor's exit status.

[P2] Rebase and re-slot the release version
CHANGELOG.md:4

The PR still declares 2.6.23, while the current `v2` base is at 2.6.88. Please rebase and use the next available version, currently 2.6.89, updating the changelog, README badge, authored version file, generated distributions, documentation conflict, and coverage registry together. GitHub currently reports the PR as conflicting.

No fixes or commits were made because this is an internal contribution.

Validation performed at head e2c07aef712cc8444a0a7d471024198b02280e1b:

- 64 focused tests passed across t37, t68, and t243
- `bun scripts/package.ts --check` passed for all seven harnesses
- `bun run typecheck` passed
- `bun run lint` passed with three informational diagnostics in unrelated t267
- coverage registry and `git diff --check` passed
- A diagnostic reproduction confirmed runtime selected `stage-z` while the helper reported `[stage-a, stage-z]`
- GitHub reports no CI checks for this head

Remaining risk: the current regression test uses the same alphabetical and load order, so it cannot detect the misleading producer ordering or prove the advisory exit-status contract.

Proposed review event: REQUEST_CHANGES

apackeer:
Addressed both review findings in `af2ba31ea`:

- Preserved producer graph load order in `consumedArtifactProducerCollisions()`, so the first rendered producer is the one runtime actually selects.
- Added a full-scaffold regression where load order differs from alphabetical order, asserting both the rendered order and an unchanged doctor exit status of 0.
- Rebased onto current `v2`, resolved the documentation and generated-file conflicts, re-slotted the release to `2.6.93`, and regenerated all harness distributions and coverage metadata.

Validation:

- 69 focused assertions passed across `t37`, `t68`, and `t243`
- `bun scripts/package.ts --check` passed for all seven harnesses
- `bun run typecheck` passed
- `bun run lint` passed with the three existing informational diagnostics in unrelated `t267`
- Coverage registry and `git diff --check` passed

@leandrodamascena, please re-review when ready.

leandrodamascena (approved):
Thanks, Arden. I reviewed the current head `af2ba31e` against the current `v2` at `62b868c57`. The change remains aligned with the maintainer-owned scope established in #831.

The previous findings are resolved: producer order now matches runtime graph order, and the regression confirms the advisory preserves doctor’s exit status. I also validated triple producers, optional producers, repeated outputs, unconsumed collisions, generated parity, typecheck, lint, coverage, and the combined behavior with the latest doctor changes.

The current CI is green. Approved with one follow-up action before merge: please rebase onto the latest `v2`, preserve the #890 doctor documentation, re-slot this change to `2.6.94`, regenerate the distributions and coverage metadata, and let CI rerun on the rebased head.

Issue #793, opened 2026-08-18
[Feature]: Plugin authoring and test toolchain outside the framework checkout (aidlc plugin build/validate/test)


apackeer:
### Description

Follow-up from #790 (reusable test framework for plugins, implemented in #792).

Plugins are install-time composed, and the distribution model is already fully external: a plugin lives in its own repository, is published to its own marketplace, trust is host-native, and the compose hook ships inside every emitted plugin, so a user's install never needs this repository. First-party and third-party plugins are mechanically identical; the `plugins/` tree here is just where first-party plugins live, and customers can run their own plugin stores.

What is still bound to a framework checkout is the authoring toolchain:

- The builder only exists as `scripts/package.ts plugin build <plugin> <harness> <outDir>`, and plugin discovery scans this repository's `plugins/` directory.
- The test kit from #792 (`tests/harness/plugin-kit.ts`: `validatePluginContent`, `composePluginFixture`, `invokeHarness`) uses relative imports into the checkout, and the package is private with no exports, so nothing is installable from outside.

So a third-party author's build step today is effectively "clone this repository, drop your plugin into `plugins/`, run bun from there". That works, but the build tool is a whole framework checkout at whatever commit was cloned.

Proposal: ship the toolchain as a versioned, installable tool. After the Distribution RFC's reshape (#722, PR #756), the existing `plugin select|sync|list` verbs sit in the hidden `engine` namespace pending the Plugins RFC (#723), which makes plugins first-class for users under a top-level `aidlc plugin` command. The authoring verbs ride that same command (#723 § 2e):

- `aidlc plugin validate` - wrap `validatePluginContent` (content tier, runs anywhere, offline). Ships first: it is the highest-value piece in the field evidence below.
- `aidlc plugin build <harness> [outDir]` - wrap the emitter so an author projects their plugin from their own repository. One implementation, two callers: the packager and this verb share the emitter, so the in-repo byte-parity guard makes external builds identical to first-party builds by construction.
- `aidlc plugin test` - compose the built projection into a disposable candidate of a real install, on the transaction engine's mkdtemp candidate-tree machinery that plugin sync already uses (never mutating the author's pinned install).
- Later: `aidlc plugin create` - scaffold `.aidlc-plugin/plugin.json` plus the core-shaped subtrees (the gh/helm authoring on-ramp).

Settled (2026-08-23, maintainer decision):

[...]

PR #887, opened 2026-08-24
feat: aidlc plugin validate/build — plugin authoring toolchain outside the framework checkout (2.6.66)

apackeer:
First cut of the authoring toolchain from #793, per its settled decisions: the authoring verbs land alongside the consumption verbs, and the whole toolchain ships inside every distribution so pinning `aidlc` pins the toolchain with no framework checkout in the loop.

- **`aidlc plugin validate [path]`** — content tier, offline, runs from the plugin's own repository. Defaults to the current directory, supports `--json`, exits 1 on any finding.
- **`aidlc plugin build <harness> [outDir]`** — the emitter, relocated into the shipped `core/tools/aidlc-plugin-author.ts`. One implementation, two callers: the repository packager delegates to the same library, and the new `t301` integration test proves an external-repository build is byte-identical to the packager's projection. Supports all seven harness targets (including the new per-runtime Kiro wiring from #817), refuses to erase a non-empty directory that is not a prior AIDLC projection, and defaults output to `dist-plugin/<harness>`.
- Every dist now carries `tools/data/plugin-targets.json` and a reference copy of the compose-hook template under `tools/data/plugin-hooks-template/` — the ship-the-source-of-truth half of field requirement 1.
- The #792 test kit (`tests/harness/plugin-kit.ts`) re-exports the moved validator, so existing plugin tests and doc snippets are unchanged.

Deferred per the issue's phasing (it stays open): `validate` hardening (vendored-hook match, the #804/#805/#806 silent-failure refusals, `tools/` hygiene) and `aidlc plugin test` on the candidate-tree machinery. Version 2.6.66.

Part of #793.


apackeer:
Plan: PR #892 now contains the consolidated plugin authoring toolchain, including the functionality ported from this PR. If PR #892 is approved, this PR (#887) will be closed as superseded. Until then, this PR will remain open.

leandrodamascena:
Thanks for the work on this implementation. I reviewed the current head `49f9100998` against `v2` and the consolidated implementation in #892.

The retained VALIDATE and BUILD functionality from this PR has been ported to #892, which now owns the complete plugin authoring toolchain and is under active review. There is no remaining unique scope here, and keeping both implementations open would only create unnecessary duplication and conflict.

Closing this PR as superseded by #892.

PR #889, opened 2026-08-24
fix: make audit locks generation-safe across platforms

apackeer:
## Summary

- bind every audit, active-directive, and usage-ledger lock acquisition to a unique owner generation and stable canonical identity
- serialize canonical gate publication, inspection, retirement, recovery, and doctor clearing with `flock` on POSIX and `LockFileEx` on Windows
- distinguish missing, malformed, unreadable, dead, live, and PID-reused owners so unsafe recovery fails closed while provably abandoned generations remain recoverable
- retain failed-retirement receipts and recovery handlers, bound retry budgets, and support glibc, macOS libSystem, and common/discovered musl libraries
- add deterministic race/fault coverage plus repeated Windows state and usage stress

## Verification

- local `t161-per-intent-lock-reaper`: passed
- local `t145-state-lock-concurrency|t163-reaper-steal-race|t267-usage`, 20 rounds: passed
- Windows `t161-per-intent-lock-reaper`: passed
- Windows `t145-state-lock-concurrency|t163-reaper-steal-race|t267-usage`, 50 rounds: passed
- TypeScript configurations and focused lint
- `git diff --check`

## Notes

- no full-suite run was performed; verification was limited to lock ownership, recovery, callers, packaging, and Windows stress
- final adversarial closure review reported no findings and `MERGEABLE`


leandrodamascena (changes requested):
Thanks, Arden. I reviewed the current head `ed947bed` against `v2@f73472aa`. The bug fix aligns with the existing lock-safety contract and internal project direction, but I found two recovery blockers.

**P2 — Reject malformed generation tokens before deriving recovery paths.**

`inspectOwnerStamp` accepts any non-empty token (`core/tools/aidlc-lib.ts:7261`), while `reapGateReleasablePath` uses it as a path component (`:7522`). I reproduced a live gate stamped with `token: "../external-dir"` and an external `releasable` file. Doctor classified that live gate as released and removed it, allowing another owner to enter while the original generation was still active.

Please validate tokens as canonical generated identifiers, verify containment and the expected token directory, and treat any mismatch as an invalid owner that remains fail-closed. This needs regression coverage for traversal, missing token directories, and external/symlinked releasable markers.

**P2 — Preserve lock acquisition when the optional process-generation probe is unavailable.**

`writeOwnerStamp` returns `null` whenever `/proc`, `GetProcessTimes`, or `proc_pidinfo` cannot provide a generation on a nominally supported platform (`core/tools/aidlc-lib.ts:7214`). I reproduced the unavailable-probe path: even an uncontended `acquireAuditLock(..., 0, 1)` returns `false`. Consequently, every state, audit, active-directive, and usage-ledger mutation fails in restricted Linux environments or when the native probe is temporarily unavailable.

The protocol already handles missing generations safely: a live owner with unknown generation remains fail-closed, while a provably dead PID can be recovered. Please retain the PID/token stamp when the optional OS generation cannot be read and add coverage that forces the self-generation probe unavailable.

I made no branch changes or pushes. The diagnostic rebase showed that the implementation applies cleanly to the current `v2`; only release metadata conflicts and the version must be rebumped to `2.6.90`.

Validation performed:

- `t161`: 102/102 assertions passed across three repeated runs
- `t163` recovery races and `t267` usage concurrency passed
- `t76` audit-lock behavior passed 15/15 when isolated
- package parity, all TypeScript configurations, lint, and diff checks passed
- the native binary compiled and executed its lock-using paths
- unrelated `v2` test failures were reproduced unchanged on the base branch

The current GitHub head has no reported CI checks, and Windows should be rerun after these corrections.

**Review event: REQUEST_CHANGES**

apackeer:
Addressed both requested changes:

- Owner tokens must be canonical UUIDv4s with non-symlinked, contained token directories; releasable markers must be regular in-tree files. Invalid gates remain fail-closed and doctor reports `invalid-owner` without clearing.
- An unavailable optional process-generation probe no longer blocks acquisition. PID/token stamps are retained, while live unknown-generation owners remain fail-closed.

Added regression coverage for traversal, missing token directories, redirected token directories/markers, and unavailable self-generation probes.

Validation passed on Linux and native Windows: `t161`, `t145+t163`, `t76`, `t267`, package parity, typecheck, and lint.

@leandrodamascena, please re-review.

leandrodamascena (approved):

Issue #876, opened 2026-08-23
[Bug]: Compose copies the plugin tools/ tree recursively with no filter, so co-located tests and fixtures land in the engine tree


apackeer:
### Description

Field report on #793 (point 3): `copyTreeNoClobber(join(PLUGIN_ROOT, "tools"), join(HARNESS_DIR, "tools"), "tool")` (`scripts/plugin-hooks-template/compose.ts:1434`) walks the plugin's `tools/` recursively with no precheck or filter — unlike stages/scopes/sensors, which each have prechecks. Tests and fixtures authored next to the programs are composed into the install (56 files instead of 10 in the field case, 42 of them fixtures).

The intended convention exists but is invisible: plugin tests live in top-level `tests/`, which the emitter's `contentDirs` (`scripts/package.ts:914`) excludes from the projection — but authors composing a raw plugin root (no emitter available outside the checkout until #793) never pass through that filter, and the authoring guide never states the convention.

### Proposed

- Compose-side: a `tools/` precheck (or filter) that drops obvious non-tool payloads — at minimum nested `tests/`, `fixtures/`, `*.test.ts` — with a recorded drop, matching the existing precheck pattern.
- Docs: state the convention in `docs/harness-engineering/10-authoring-a-plugin.md`.
- The author-side warning belongs in `aidlc plugin validate` and is folded into #793.

### Version

v2 (alpha)

### Area

Plugin mechanism / compose hook

### Additional Context

Split out of the field report on #793: the compose hook ships inside every emitted plugin, so this fix reaches users through the engine, not the authoring toolchain.

apackeer:
Fix in #909. Compose now skips test and fixture files co-located inside a plugin's `tools/` tree (nested `tests/`, `__tests__/`, `fixtures/`, and `*.test.ts`/`*.spec.ts`) instead of installing them, and records an advisory drop for each — including for payloads an older compose already landed, so `/aidlc --doctor` tells you what to remove. The authoring guide now states the top-level `tests/` convention.

leandrodamascena:
Fixed by https://github.com/awslabs/aidlc-workflows/pull/909

PR #756, opened 2026-08-14
feat: add native distribution, config policy, and hardened release lifecycle

apackeer:
Implements the locally buildable surface of the distribution RFC (#722) - originally opened as the milestone 1 reshape, this branch now carries milestones 1 through 3 plus the milestone 2 hardening: the six-command CLI surface, the full config policy lane (six sections, three presets, the aidlc.settings.json hierarchy, the first-run setup walk, per-command help, and the /aidlc --config in-session alias), the supply-chain hardening with its reviewable design chapter, and the end-to-end Windows proof. Each push carries a comment describing that round's delta and evidence; this body summarizes the whole.

## The surface (milestone 1)

**Public surface is six commands:** `config`, `doctor`, `version`, `update`, `use`, `uninstall`. `init` became `config`, `upgrade` became `update`, and the public `rollback`, `versions`, `harness`, `package`, `plugin`, and `completions` verbs are retired. Failed updates recover automatically through the transaction engine, so there is no public rollback verb. Machine install asks nothing: one runtime bundle carries every harness payload, and harness selection moved into `aidlc config`, which scaffolds offline from the installed runtime.

**Two hidden namespaces, each with its own help, neither in default help:** `aidlc engine <noun> <verb>` is the trusted machinery generated harness surfaces invoke - every route under it is project-scoped or read-only with network forbidden, and it is the prefix the shipped host trust seeds pre-allow. `aidlc system <...>` holds machine and network internals (lifecycle passthrough, retained versions, machine config, completions); no trust seed grants it. The project pin lives on `config --pin`/`--unpin`; `use` selects only the machine-active version, so no public command carries project-and-machine mutation scope.

### The part most worth reviewing

The RFC asserted that an agent trusted to run the engine is never thereby trusted to run machine mutations. **The first implementation did not satisfy that**: hidden lifecycle routes stayed reachable under the trusted prefix, and a permissive routing fallback forwarded unmatched nouns into the lifecycle module, so an agent pre-allowed `aidlc engine` could uninstall the framework or repoint the release mirror unattended. Five independent adversarial reviews proved it by execution, pre-ship. It is enforced now rather than asserted: every route declares its namespace; `t230` walks the route table and rejects any engine route that mutates beyond the project or touches the network; unmatched nouns are usage errors; and the trust seeds derive from the same table (`t243` pins the bytes), so a seed cannot silently disagree with it.

## The config policy lane (milestone 3, sections 3d + 1f)

[...]

PR #892, opened 2026-08-24
feat: plugin authoring toolchain - create, validate, build, and test a plugin with no framework checkout

### LINKED ISSUE #793: [Feature]: Plugin authoring and test toolchain outside the framework checkout (aidlc plugin build/validate/test)

apackeer:
Implements the authoring-toolchain phases of #793, per its Settled block and RFC #723 § 2e: four shipped tools that let a plugin author work entirely from their own repository against a pinned, copied distribution — no framework checkout anywhere in the loop.

## The four tiers (one commit each)

1. **`aidlc-plugin-validate.ts`** — offline content-tier validation, refuse-ambiguity posture. Rules from the #793 field report: manifest schema, stage frontmatter (via the real stage schema), scope naming + keywords-parse-non-empty (the #805 silent-failure shape), agent naming, duplicate artifact producers within the plugin across `produces`/`optional_produces` consumed or not (the stricter local rule the #804 reporter keeps), non-tool payloads under `tools/` (#876's validate half), vendored `hooks/compose.ts` byte-check against the now-bundled template (closing the field report's missing-reference-copy gap).
2. **`aidlc-plugin-build.ts`** — projects one plugin for one harness. The emitter moved to `core/tools/aidlc-plugin-emit.ts` with **one implementation, two callers** (packager + shipped tool); per-harness targets ship as packager-derived `tools/data/plugin-targets.json`. t315 proves byte-parity: from a mkdtemp with no checkout, the shipped tool rebuilds `test-pro` **byte-identical to `dist/plugins/test-pro/` on all seven harnesses** (independently reproduced during review).
3. **`aidlc-plugin-test.ts`** — composes the built projection into a disposable mkdtemp candidate of a real install, runs the real compose hook, reports composed files/drops/post-compose graph, includes a second-pass idempotency check, and **never mutates the live install** (before/after hash-verified in t316 and in review). `--dist` reserved until the release channel exists (RFC #722 milestone 2).
4. **`aidlc-plugin-create.ts`** — deterministic scaffold that validates green on arrival; t317 proves create → validate → build → test end-to-end from copied tools.

## Deliberate boundaries
- `core/tools/aidlc.ts` untouched (empty diff): the `aidlc plugin validate|build|test|create` verb wiring lands on PR #756's route table per RFC #723 § 2e.
- `tests/harness/plugin-kit.ts` delegates shared validation + compose execution to the shipped tools (one implementation); checkout-aware policies stay test-side.
- Harness manifests gain optional `plugin.installRoots` (opencode's `.opencode`) for the test tier's candidate copy.

## Evidence
Every phase ran smoke+unit (245→248 files, 0 failures) + the deterministic plugin integration slice (3 files, 0 failures); `package.ts --check` green throughout (the build-tier refactor's byte-parity guard); final gate post-rebase: t68+t314-t317 33/0. One unrelated t276 red in phase 4's first run passed green-alone and in the full rerun.

One version bump for the whole PR: **2.6.67** (2.6.66 is reserved by open #827).

Refs #793


apackeer:
Rebased this PR onto the current `v2` head and completed the review-fix pass.

What changed:

- Moved the release bookkeeping from `2.6.67` to `2.6.69` to avoid the version claimed by changes already merged into `v2`.
- Hardened plugin BUILD output handling. Symlinks inside the caller-owned output boundary, including a linked `<plugin-root>/dist`, are rejected before cleanup or writes. Symlinked environmental ancestors remain supported.
- Made plugin content symlinks explicit validation failures, including linked files, directories, and `.aidlc-plugin/`. VALIDATE returns a path-specific `content-symlink` finding; BUILD and TEST stop before creating output or changing the install under test.
- Added regression coverage for linked authored content, linked metadata, linked default output parents, custom output beneath an environmental alias, and TEST install isolation.
- Regenerated every harness distribution from authored sources.

Post-rebase verification:

- `bun scripts/package.ts --check`: pass
- `bun run typecheck`: pass
- `bun run check`: pass, with only the three existing informational `t267` suggestions
- smoke + unit: 248 files passed, 7,115 assertions, 0 failures
- plugin integration slice (`t188|t300|test-pro`): 3 files passed, 89 assertions, 0 failures
- symlinked-`TMPDIR` package probe: pass
- linked-`.aidlc-plugin` validation probe: rejected as expected with `content-symlink`

The branch is now based on current `v2`; GitHub reports it mergeable and the refreshed CI/security checks are running.

leandrodamascena (changes requested):
I reviewed commit `8a52b229` against the current `v2` direction. Thanks for the comprehensive implementation and test coverage. I found two blockers before this can merge:

**P1 – BUILD can delete another plugin’s projection**

`assertPluginBuildOutput()` treats any output containing a host manifest whose name starts with `aidlc-` as an owned prior projection. `buildPluginProjection()` then recursively deletes that directory.

I reproduced this by building `test-pro`, adding a sentinel file, and building a valid `other-plugin` into the same output. The second BUILD succeeded and deleted the previous projection and sentinel. Please bind the output marker to the plugin being built, harness, and preferably an explicit tool-owned marker before allowing replacement.

**P2 – `aidlc.contributes` paths validate successfully but are ignored**

I created a valid plugin with:

```json
"contributes": {
  "stages": "custom-stages/"
}
```

VALIDATE returned valid and BUILD completed successfully, but the stage was silently omitted because validation and emission still read the hardcoded `stages/`, `scopes/`, and other canonical directories.

Please either honor these manifest paths consistently across validation, symlink scanning, BUILD, and TEST, or reject non-canonical values until configurable paths are supported. A successful build must not silently omit declared content.

The branch also needs a rebase onto current `v2`; GitHub reports it as conflicting. The version and title must be updated from the currently claimed `2.6.67`/`2.6.69` to the next available version. PR #887 independently implements the VALIDATE/BUILD surface, so please also explicitly supersede and close it to avoid merging parallel implementations.

Validation performed on the reviewed head:

- focused `t314`–`t317`: 32 tests passed
- plugin integration slice: 3 files, 89 assertions passed
- package parity, typecheck, and lint passed

Thank you for the substantial work here. Once the two false-safe cases and branch ownership/rebase are resolved, I can review the updated head.

apackeer:
Addressed the requested changes and rebased this branch onto current `v2` (`c36559c33`).

What changed:

- Fixed P1 by adding an explicit projection ownership marker bound to the logical plugin and exact harness. BUILD now holds a canonical output-scoped lock across marker verification, cleanup, and emission; missing, malformed, or mismatched markers fail without mutation, and there is no force bypass.
- Fixed P2 by requiring exact canonical `aidlc.contributes` paths until configurable routing ships. Unsupported `memory` declarations, custom paths, absolute/traversal paths, and malformed/unknown declarations now fail VALIDATE, BUILD, TEST, and direct emission before output changes.
- Ported the retained PR #887 functionality into this implementation:
  - `/aidlc plugin validate [path] [--json]`
  - `/aidlc plugin build <harness> [outDir] [--plugin-root <path>] [--json]`
  - core/plugin agent-reference validation, artifact namespace checks, contribution-target validation, and non-empty stage-body validation.
- Kept one implementation: the top-level routes delegate to the standalone validator/builder and the repository plugin-kit delegates shared content rules to the shipped validator.
- Regenerated all harness distributions and re-slotted release bookkeeping to `2.6.96`.

Verification:

- `bun run check`: pass; only the three existing informational `t267` suggestions.
- `bun scripts/package.ts --check`: pass.
- coverage registry check: pass.
- focused routing/authoring/plugin slice: 8 files, 220 assertions, 0 failures.
- Kiro CLI adapter: 33 assertions, 0 failures.
- Kiro IDE adapter: 58 assertions, 0 failures.
- test-suite drift: 9 assertions, 0 failures.

@leandrodamascena could you please re-review the updated head?

leandrodamascena (changes requested):
Thanks for the thorough follow-up. I reviewed the current head `b37560e54e8ed7b071e52c948f4e5cb3c7734989` against current `v2` at `1c6e03a1264c5376b854033adfae4f1e7ccef6eb`. Alignment is established by the settled maintainer decision in #793 and the prior substantive review.

The previous P1 output-ownership issue and P2 contribution-path issue are correctly resolved. Two P2 blockers remain:

**P2 - A crashed BUILD permanently locks that output**

[`withPluginBuildOutputLock()`](https://github.com/awslabs/aidlc-workflows/blob/b37560e54e8ed7b071e52c948f4e5cb3c7734989/core/tools/aidlc-plugin-emit.ts#L120) creates a lock directory and only removes it in `finally`. If the process is killed or crashes, subsequent builds only treat `EEXIST` as live contention and eventually fail after 30 seconds. Although `owner.json` records the PID, it is never read for stale-owner recovery.

This permanently disables BUILD for the same output, including the default `<plugin-root>/dist/<harness>`, until the temporary lock is manually removed. Please implement ownership-safe dead/unstamped lock recovery and add regression coverage for an orphaned lock. After rebasing, the existing owner-stamped lock/reaper in `aidlc-lib.ts` may be reusable.

**P2 - The new public routes remain classified as uncovered**

The committed coverage registry explicitly marks both `aidlc-utility plugin-build` and `aidlc-utility plugin-validate` as [`UNDER-MECHANISM`](https://github.com/awslabs/aidlc-workflows/blob/b37560e54e8ed7b071e52c948f4e5cb3c7734989/tests/.coverage-registry.json#L7692). `t327` does spawn the dispatcher, but the mechanism detector classifies it as `none`, so its claims do not meet the required CLI mechanism and do not increase the coverage ratchet.

Please make the spawned dispatcher recognizable as a CLI mechanism, regenerate the registry, and pin both new subcommands as `covered`.

The branch is also 7 commits behind `v2` and currently conflicting. It needs a rebase, regeneration, and release re-slot from `2.6.96` to the next available version, currently `2.6.102`; the PR title still references `2.6.67`.

No fixes or commits were pushed during this internal review.

Validation on the reviewed head:

- 258 focused tests passed, 0 failures
- package parity passed
- typecheck passed
- lint passed with three pre-existing informational findings
- coverage registry check passed
- `git diff --check` passed
- all GitHub checks on the exact head are green

apackeer:
Addressed the latest requested changes and rebased the branch onto current `v2` (`4b0fd9c37`).

What changed:

- BUILD now uses the shared generation-bound owner-stamped lock and reaper from `aidlc-lib.ts`. Dead owners and aged unstamped crash remnants are recovered; live, malformed, unreadable, and generation-unknown owners fail closed.
- Added BUILD regression coverage for dead owner recovery, live owner preservation, fresh unstamped contention, and aged unstamped recovery.
- Updated the coverage mechanism detector to recognize a real runtime spawn of the public `aidlc.ts` dispatcher, with a calibrated synthetic test and an explicit `t327` migration pin.
- Regenerated the coverage registry and ratchet. `aidlc-utility plugin-build` and `aidlc-utility plugin-validate` are now `covered` at CLI mechanism, and subcommand coverage increased to 99/111.
- Re-slotted release bookkeeping to `2.6.102` and regenerated every harness distribution.

Verification on pushed head `82b81b26e`:

- `bun run check`: pass; only the three existing informational `t267` suggestions.
- `bun tests/gen-coverage-registry.ts --check`: pass.
- `git diff --check`: pass.
- Post-rebase BUILD/public-route slice: 2 files, 20 assertions, 0 failures.
- Consolidated plugin-authoring slice: 8 files, 139 assertions, 0 failures.

@leandrodamascena could you please re-review the updated head?

leandrodamascena (approved):
Thanks for the thorough follow-up. I reviewed head `82b81b26e` against the current `v2` direction established in #793.

The previous lock-recovery and coverage blockers are resolved. BUILD safely handles stale locks, both public routes are covered, and #887 is closed as superseded. I found no remaining implementation defects.

Validation passed: 139 plugin-authoring tests, focused lock and coverage tests, package parity, typecheck, lint, coverage registry, and all GitHub checks.

Follow-up before merge: rebase onto current `v2`, move release `2.6.102` to the next available version, resolve the changelog conflict, regenerate distributions, and rerun CI. GitHub currently reports the branch as conflicting.

PR #901, opened 2026-08-24
test: harden Windows TUI process cleanup

apackeer:
## Summary

- Identify Windows TUI daemons and descendants by session token and process creation identity, preventing stale or reused PIDs from authorizing unrelated `taskkill` targets.
- Await the complete ConPTY process tree with bounded graceful and forced teardown, preserving diagnostics for surviving or ambiguous processes.
- Retry transient Windows workspace and session-directory locks while keeping genuine teardown leaks visible.
- Route every TUI e2e cleanup through helpers that propagate kill failures instead of reporting a false green or cleanup-only failure.

This is test infrastructure only, so it does not bump the framework version.

## Testing

- Focused Linux tests: 59 pass, 0 fail.
- Focused native Windows tests: 59 pass, 0 fail.
- Native Windows orientation journey repeated twice: 2 pass, no residual test-owned processes.
- Focused typecheck, TUI lint, and package drift checks passed.
- Codex implementer/reviewer loop reached `MERGEABLE` on review cycle 4.

No full TUI or repository test suite was run.

### DIFF
--- a/tests/harness/tui-drive.ts
+++ b/tests/harness/tui-drive.ts
@@ -155,6 +155,9 @@ export const WIN_KILL_TIMEOUT_MS = 8_000;
 const WIN_PROCESS_QUERY_TIMEOUT_MS = 750;
 const WIN_TASKKILL_TIMEOUT_MS = 750;
 const WIN_CONSOLE_LIST_TIMEOUT_MS = 5_000;
+const WINDOWS_SESSION_CLEANUP_ATTEMPTS = 20;
+const WINDOWS_SESSION_CLEANUP_WAIT_MS = 100;
+const RETRYABLE_WINDOWS_RM_CODES = new Set(["EBUSY", "ENOTEMPTY", "EPERM"]);
 
 type Args = {
   positionals: string[];
@@ -410,6 +413,41 @@ function sleepSync(ms: number): void {
   Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, ms);
 }
 
+export interface WindowsSessionCleanupOptions {
+  attempts?: number;
+  waitMs?: number;
+  remove?: (path: string) => void;
+  sleep?: (ms: number) => void;
+}
+
+export function removeWindowsSessionDirWithRetry(
+  path: string,
+  options: WindowsSessionCleanupOptions = {},
+): void {
+  const attempts = options.attempts ?? WINDOWS_SESSION_CLEANUP_ATTEMPTS;
+  const waitMs = options.waitMs ?? WINDOWS_SESSION_CLEANUP_WAIT_MS;
+  const remove = options.remove ??
+    ((target: string) => rmSync(target, { recursive: true, force: true }));
+  const wait = options.sleep ?? sleepSync;
+
+  for (let attempt = 1; ; attempt++) {
+    try {
+      remove(path);
+      return;
+    } catch (error) {
+      const code = (error as NodeJS.ErrnoException).code;
+      if (
+        typeof code !== "string" ||
+        !RETRYABLE_WINDOWS_RM_CODES.has(code) ||
+        attempt >= attempts
+      ) {
+        throw error;
+      }
+      wait(waitMs);
+    }
+  }
+}
+
 function processIsAlive(pid: number): boolean {
   try {
     process.kill(pid, 0);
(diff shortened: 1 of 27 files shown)


tests/harness/tui-drive.ts
```diff
@@ -325,6 +333,628 @@ function sleep(ms: number): Promise<void> {
...
+  const forceDeadline = ops.now() + Math.max(0, forcedMs);
+  while (state.alive.length > 0 && ops.now() < forceDeadline) {
+    const root = state.alive.find((processInfo) => processInfo.pid === rootPid);
+    const targets = root
+      ? [root]
+      : state.alive.filter((processInfo) => !forcedPids.has(processInfo.pid));
+    for (const processInfo of targets) {
+      if (forcedPids.has(processInfo.pid)) continue;
+      forcedPids.add(processInfo.pid);
+      forced.push(ops.terminateTree(processInfo.pid));
```
leandrodamascena: **[P2] Enforce one deadline across all forced terminations**

If the daemon exits but multiple descendants remain, `targets` contains every survivor and this loop checks `forceDeadline` only before processing the batch. Each `terminateTree` call can independently block for 5 seconds. A deterministic three-descendant case with `forcedMs: 5000` completed after 15000 ms and attempted all three PIDs.

Please pass the remaining global budget into each termination call and stop issuing calls when it is exhausted. The current `v2` implementation from #905 already follows this pattern, so this older loop should preferably not survive the rebase. Please add a multi-descendant timeout regression.

leandrodamascena (changes requested):
Thanks for hardening the Windows TUI cleanup path and, in particular, for making journey cleanup propagate failed kills instead of silently reporting success.

I reviewed the current head `5e730ffedcfc9d9036eaedaec1e60910301878e7` against `v2` at `f73472aa388590763e3d3f8685647e9a488ff5fd`. The test-infrastructure direction is aligned, and the later merged Windows work in #905, #926, and #927 confirms it. Those PRs now overlap substantially with this head, however.

**[P2] The forced teardown is not globally bounded.** When the root has exited and several descendants remain, every synchronous `taskkill` receives its own 5-second timeout. I reproduced a 15-second elapsed teardown with a configured 5-second forced budget and three surviving descendants.

This PR is also currently conflicting in `tui-drive.ts`, the orientation and preflight journeys, and `t142`. Please rebase and consolidate rather than restoring this head's older process-management implementation. Keep `v2`'s current authenticated ownership, stable wrapper, spawn/exit authority, portable-path, Unicode, and startup-cleanup behavior from #905/#926/#927. Reapply the still-useful cleanup-result propagation, workspace/session cleanup retries, and caller updates on top.

During that integration, preserve a reliable project-to-session association: this head's pending-session protection depends on `meta.cwd`, while current `v2`'s metadata schema does not record `cwd`. Please cover the integrated schema with a test proving a failed kill leaves the workspace intact. Also update the Windows user-settings journey added since this branch so its kill result is not ignored.

Validation performed:
- focused `t142` and `t152`: 56 passed, 0 failed, 234 assertions
- `bun scripts/package.ts --check`: passed for all seven harnesses
- `bun run typecheck`: passed
- focused Biome check: passed
- `git diff --check`: passed

No fixes were committed or pushed. Current CI is green for this stale head, but the PR remains `CONFLICTING/DIRTY`; native Windows and broad TUI validation still need to run after consolidation.

apackeer:
Addressed the requested changes on head `7393cae2c`.

What changed:
- rebased and consolidated onto current `v2`, preserving the authenticated ownership, stable wrapper, spawn/exit authority, portable-path, Unicode, and startup-cleanup behavior
- enforced one shared forced-termination deadline and added the multi-process regression
- added `cwd` session association, bounded session/workspace cleanup retries, failed-kill workspace preservation, and kill-result propagation across TUI callers
- updated the Windows user-settings journey
- native Windows validation exposed two masked `v2` regressions, so process metadata now uses base64 UTF-8 JSON, the attached ConPTY console is switched to UTF-8, and successful teardown clears stale target authority

Validation:
- all GitHub CI, security, documentation, contract, smoke/unit, and deterministic integration/e2e checks pass
- local focused `t142`/`t152`: 61 assertions, 0 failures
- `aidlc-dev-2` native Windows:
  - focused `t142`/`t152`: PASS
  - full `t-tui-preflight`: PASS, 22 driver traces
  - orientation journey: PASS, 3 native launches
  - Windows user-settings isolation: PASS through Bedrock
- `bun scripts/package.ts --check`, typecheck, focused Biome, and `git diff --check`: PASS

leandrodamascena (approved):
Thank you for the comprehensive follow-up and native Windows validation.

I reviewed the current head `7393cae2c0948bfc65e3051baebde97cc42c2118` against current `v2` at `cb30d6422e716247f2bb9d91a928ea86078dec9d` (merge base `f390fe3ae4289de7f0be8f9c7caf10ba69493db5`).

The test-infrastructure direction remains aligned with the previously reviewed scope and preserves the Windows ownership and startup contracts from #905, #926, and #927.

The previous blocker is addressed: forced termination now shares one deadline across the process batch, with a deterministic multi-process regression. The updated cleanup paths also preserve workspaces after failed kills, associate sessions through `cwd`, retry transient Windows locks, propagate kill failures, and cover the Windows user-settings journey.

Validation performed:

- Focused `t142` and `t152`: 61 assertions passed.
- Package parity, typecheck, focused Biome, coverage-registry freshness, and `git diff --check`: passed.
- A diagnostic merge with current `v2` completed without conflicts; the same 61 assertions and contract checks passed.
- All GitHub checks are green for this exact head.
- Local TUI preflight was skipped because `tmux` is unavailable. The author reports passing native Windows focused tests, full preflight, orientation, and user-settings journeys.

No fixes or commits were pushed because this is an internal contribution. The remaining unresolved inline thread is outdated and refers to the replaced implementation.

I found no remaining P1/P2 correctness issues.

**Proposed review event: `APPROVE`.**

PR #903, opened 2026-08-24
fix: close the review-freeze / summary-confirmation deadlock and make refusals name an executable step (2.6.76)

apackeer:
## Summary

A stage could reach a state where two integrity checks each demanded to be the last event over the same document: the write-freeze refused to let a reviewed artifact change, while the completion check demanded that artifact be saved again after the human's confirmation. The freeze's message pointed at the approval gate, which the completion check was itself refusing to open. A field session spent 1h37m looping between the two before a human found the only exit.

Repeating an identical, fully scoped summary confirmation now keeps the earlier confirmation's authorization for an unchanged document, so that inversion cannot form. Reviews additionally cannot begin over answers that were never confirmed or over missing required output documents, so a final review cannot certify work that was never admissible. Stale-source recovery gets a write window bound to a validated pending recovery request that also requires the staleness to still hold and ends at a session boundary. Every refusal touched here names a next step that is executable in the state that triggered it.

**Upgrade:** refresh the whole `dist/<harness>/` shell so the logger, state checks, reviewer protocol, cursor adapter, and freeze hook move together.

## What changes for users

* Re-confirming the same answers no longer invalidates an already-reviewed document; changed answers still require a fresh confirmation.
* `aidlc-log.ts review` refuses a request until the stage's answers are confirmed and every required output document exists, and rechecks required outputs when the verdict is recorded.
* Recording a verdict for an already-accepted request is no longer gated on Unit-set resolvability, so an in-flight review cannot become unclosable.
* A validated pending stale-receipt recovery request opens a write window for its own stage or Unit only while the stale condition still holds in the current session; restoring the reviewed workspace source, recording the verdict, or restarting the session re-arms the freeze.
* Refusals now route by stage state: active stages to a human Request Changes decision, revising stages to `/aidlc --stage <slug>`, completed stages to restoring the reviewed source or redoing the stage, and off-plan `SKIP` stages to a scope change.
* A blocked write caused by a reviewer *suggestion* now says to quote it at the gate instead of applying it, matching the reviewer protocol.
* A stage-level review request on a per-Unit stage requires every applicable Unit's required outputs. `AIDLC_SKIP_ARTIFACT_GUARD=1` also bypasses that new review-time existence check.
* The reviewer protocol records the review request *before* replacing an existing `## Review` section, and uses `--retry-pending` to rebind the fingerprint afterwards.

## Verification

The deadlock was reproduced against v2 before fixing it, and the reproduction is now a regression test (`t320`). On an unmodified checkout of v2 the sequence still refuses; on this branch it passes, the freeze still blocks the artifact, and the gate opens.

Full `smoke+unit+integration` tier on the rebased tree: 358 files, 8819 assertions, 0 failures. `package.ts --check`, `typecheck`, `biome`, `t68`, and the coverage registry are clean. New tests: `t320` (deadlock prevention and guidance routing), `t321` (request-bound recovery suspension scope), `t322` (Bolt attempt floors), `t323` (verdict closure after topology change).

## Deliberately not included

Two mechanisms were built during review and then removed, because each caused a worse failure than the one it fixed:

* **Prepared-unit-set preservation.** Persisting a swarm-supplied Unit set through `runtime-graph.json` recompiles could truncate an open Unit set irrecoverably, and made guards that document authored-DAG authority trust conductor input. Removed in full; `bolt_dag` authority again comes only from an authored `unit-of-work-dependency.md`, matching base v2.
* **Plan Approval as a pre-review check.** The Plan Approval guard already enforces at the dispatch point, which is the documented boundary; a second enforcement point in the review logger added no protection and conflicted with in-flight work.

Known inherited limitation (present on v2 today, not introduced here): a per-Unit stage with no authored `unit-of-work-dependency.md` requires a stage-level review receipt once its Bolts close. The workaround is to author that document before Construction. First-class no-DAG per-Unit continuity is deferred.

## Notes for reviewers

* Worth a close look: the duplicate-confirmation branch in `checkSummaryConfirmationEvidence`. It reuses the same receipt scoping as the primary check and resolves each candidate receipt under **its own** `Hash Scope`, so a legacy unscoped receipt is compared against the raw file hash and a scoped one against the confirmed-content hash. Ordering uses the shard-aware comparison introduced in #789; cross-shard timestamp ties still fail closed.
* Two review budget notes for the record: `BOLT_STARTED` clears a Unit's verdict, staleness, and spent-recovery state, which aligns `freshReviewReceipts` with the request-side floor that already behaved that way; and the `state === "none"` arm returns a single stage-level artifact path where base enumerated per-Unit directories, which affects in-flight upgrades on no-DAG per-Unit stages.
* A plain-language pass over the remaining error-recovery refusal strings is a separate follow-up branch, so this PR keeps its wording changes to the messages that named an action the engine would refuse.


leandrodamascena (changes requested):
I reviewed this against commit `77c1adf24`, including the review lifecycle, recovery freeze, upgrade behavior, and cross-shard ordering.

**P1 — An accepted review can still become impossible to close after a resolved Unit topology change**

`--retry-pending` re-applies Unit membership admission. I reproduced this sequence: a review request is accepted for Unit `alpha`; the authoritative DAG is rebuilt with only `beta`; the reviewer updates `## Review`; the verdict correctly requires `--retry-pending`, but that retry is rejected because `alpha` is no longer in the DAG. The pending request can then never receive a verdict.

`t323` only covers the DAG becoming unavailable. Please also cover a still-resolvable DAG that no longer contains the accepted Unit and allow an existing pending request to close without reapplying new-request membership admission.

**P1 — New session and Bolt boundaries do not fail closed on cross-shard timestamp ties**

`freshReviewReceipts` processes `SESSION_STARTED`/`SESSION_RESUMED` and `BOLT_STARTED` directly in the stream ordered by `shardIndex`. When either boundary shares its second-precision timestamp with a recovery request or receipt in another shard, filename order decides authority.

This can leave a recovery write suspension active after a session boundary or let a prior-attempt receipt survive a new Bolt boundary. Other authority-sensitive boundaries in this scanner explicitly fail closed on these ties. These new boundaries need equivalent ambiguity handling and tests for both shard orderings.

The main deadlock fix and request-bound recovery design otherwise address the reported failure, but the two cases above remain blockers.

Requesting changes.

apackeer:
@leandrodamascena Thanks for the review. The two blockers are addressed in `1a5ab14fe`:

- `--retry-pending` now re-dispatches an already accepted request without reapplying new-request Unit membership admission. `t323` now rebuilds the authoritative DAG from `alpha` to `beta` and verifies that the pending `alpha` request can still be retried and closed.
- Cross-shard same-second session and Bolt boundaries now apply at the timestamp-group boundary instead of relying on shard filename order. A session-tied recovery request remains retryable but cannot keep its write suspension active, and a tied `BOLT_STARTED` clears prior-attempt pending/receipt state in either shard ordering.
- `t321` covers both `SESSION_STARTED` and `SESSION_RESUMED` in both filename orders; `t322` covers both filename orders for the Bolt/receipt tie.

The branch is rebased onto current `v2`, is mergeable, and all CI checks are green. Could you please re-review?

leandrodamascena (approved):

PR #904, opened 2026-08-24
fix: bind source freshness without Git

apackeer:
## Summary

- Replace Git-dependent Code Generation source receipts with one bounded, content-addressed filesystem identity that works with Git, without Git, and when the Git executable becomes unavailable.
- Bind ordinary and ignored application source while excluding framework state, VCS metadata, dependencies, caches, and unregistered generated-output boundaries.
- Add `.aidlc-source-paths.json` for real binary, extensionless, or generated-directory source that must participate in freshness checks and swarm snapshots.
- Preserve exact multi-repo partitions and harden swarm Source Commit handling for ignored source, symlinks, embedded repositories, and recursively initialized submodules.
- Require advertised submodule history to be fetchable through a bounded, non-interactive recovery proof.

## Testing

- Final targeted local slice: 4 files, 110 assertions, 0 failures.
- Native Windows targeted coverage: source-freshness unit 100 assertions and loop-back coverage 3 assertions, all passing.
- Package parity passed for all seven harness distributions.
- Codex implementer/reviewer loop reached `MERGEABLE` after 15 review cycles.

No full repository test suite was run.


leandrodamascena (changes requested):
I reviewed commit `deecedd7` adversarially against the current `v2` head. Thank you for addressing the non-Git source-freshness gap; the direction is correct, and I verified that the merge-base returns `null` for non-Git source while this head produces a stable content identity. I found the following blockers:

**P1: Clean-filter rewriting can reintroduce excluded files into the immutable Source Commit**

`shapeSourceSnapshotIndex()` restores generated-output and framework boundaries to `HEAD`, but `filteredRawIndexEntries()` subsequently enumerates every regular index entry and hashes its current worktree bytes. `bindReviewedSource()` writes those hashes back, while its final restore omits conditional boundaries such as `dist/`, `build/`, and the active harness directory.

I reproduced this with a tracked, clean-filtered `dist/out.js`. After shaping, the temporary index correctly contained `BASELINE`; after the raw-filter rewrite, it contained `UNREVIEWED_GENERATED`. The resulting Source Commit can therefore carry bytes deliberately excluded from the reviewed identity.

Please restrict raw-filter rewriting to exact paths included by the filesystem source boundary, and add regressions for clean-filtered tracked files under generated-output and framework boundaries.

**P1: The branch does not typecheck**

`bun run typecheck` fails with four `TS2345` errors at `core/tools/aidlc-lib.ts:6347`, `:6374`, `:6717`, and `:6750`. The sets inferred from the `as const` arrays accept only their literal union, while the walker passes arbitrary strings. These sets need an explicit `Set<string>` type or equivalent correction.

**P1: The rebase must preserve the per-unit attribution contract from #813**

This branch predates #813 and currently implements only a global `workspaceSourceFingerprint()`. Current `v2` requires `workspaceSourceState()` to produce the fingerprint and canonical per-path listing in the same pass; that listing drives Unit Source Fingerprints, manifests, stage baselines, newest-claimant shielding, and unclaimed-path rejection.

The branch tests still assert that three laundering cases are accepted as documented limitations, while current `v2` correctly refuses them. The diagnostic rebase conflicts in `aidlc-lib.ts`, `aidlc-swarm.ts`, t304, and the related documentation. Please integrate the filesystem walker into the current state/listing model rather than selecting the old global-only side, preserve the inverted t304 cases and t305 attribution coverage, then re-bump from current `2.6.81`.

**P2: Submodule recovery has no aggregate time bound**

Each new submodule can independently spend 15 seconds in `ls-remote` and another 15 seconds fetching advertised history. This runs inside the submodule loop, which permits thousands of materialized checkouts, so the operation described as bounded can multiply into an extremely long finalize. Please enforce one cumulative deadline and a practical cap on new recovery proofs.

Validation at this head:

- Focused official tests: 3 files, 103 assertions, all passing.
- Package parity: all seven harnesses passing.
- Lint: passing, with three unrelated informational suggestions.
- Typecheck: failing as described above.
- GitHub reports no CI checks, and the PR currently conflicts with `v2`.

Thank you for the substantial hardening work here. The non-Git identity should land, but the Source Commit bypass, current compile failure, and #813 integration need to be resolved first.

apackeer:
Addressed the review findings in `04b069040` and rebased onto the current `v2` head.

* Restricted clean-filter raw-byte rewriting to exact filesystem-included regular paths. Excluded generated and active-harness files now remain at `HEAD`; registered generated source still retains the reviewed raw bytes.
* Corrected the inferred literal-set type errors with explicit string sets.
* Integrated the Git-independent walker into `workspaceSourceState()` so the fingerprint and canonical per-path listing are produced together, preserving Unit Source Fingerprints, newest-claimant shielding, and unclaimed-path rejection from #813.
* Added one cumulative 30-second new-submodule recovery deadline and a 32-proof cap shared across the full `finalize` call.
* Re-slotted the release metadata to `2.6.86`.

Validation:

* Focused official slice: 5 files, 165 assertions, all passing (`t304`, `t305`, version sync, and both loop-back integrations).
* `bun run check`: passing; only the three existing informational Biome suggestions in `t267` remain.
* `bun scripts/package.ts --check`: all seven harnesses passing.
* Merge-tree against the current `v2`: conflict-free; GitHub now reports the PR mergeable.

@leandrodamascena please re-review when ready.

leandrodamascena (changes requested):
Thanks for the thorough follow-up. I reviewed current head `04b069040eb3cfb700e82ecaa6087f86b6d3def2` against current `v2` at `4cae87cf27364cb70a8e2777db61e297e5d5cffe`. The direction remains aligned with the previous maintainer review, and all four previously reported implementation blockers are addressed.

**P2: The validity-capture integration fixture now creates application source after review**

`tests/integration/t166-multi-repo-construction.test.ts:475` writes `composition-stage-graph.json` at the workspace roof after recording the review. The new contract correctly binds workspace-roof files, so the expected successful completion fails with `source-fingerprint mismatch`.

I reproduced the CI failure. Moving this test-only graph under framework-owned `.aidlc/` made the complete `t166` file pass: 42 tests and 179 assertions.

**P2: The coverage registry was not regenerated**

`shapeSourceSnapshotIndex`, `workspaceSourceSnapshotPaths`, and `workspaceSourceEmbeddedGitPaths` are newly enumerated functions but are absent from `tests/.coverage-registry.json`. Running `bun tests/gen-coverage-registry.ts` generated the expected entries, and the registry test then passed all 34 tests. Please commit the regenerated registry.

**Rebase required**

The PR currently conflicts with `v2` in release metadata and `t305`. Current `v2` is already at `2.6.102`, so please rebase, preserve both the current intent-creation terminology and this PR's no-Git expectations, retain the source-freshness test renumbering to `t314`, and re-slot to the next available patch version.

Validation performed:

- Confirmed the previous clean-filter, typecheck, per-unit attribution, and aggregate submodule-budget findings are fixed.
- `t305`: 51 tests passed.
- Version synchronization: 7 tests passed.
- Coverage registry after local regeneration: 34 tests passed.
- `t304`: 103 tests passed; the large argv stress case and its cleanup hook timed out locally, although CI did not identify that case as a failure.
- Contract, documentation, security, and changelog checks are green.
- Smoke/unit and integration/e2e CI remain red for the two issues above.

No commits were pushed because this is maintainer-owned work. Remaining risk is the required post-rebase validation against the latest `v2`.

apackeer:
Addressed both review findings in `a2cac564a` and rebased onto the current `v2` head (`2.6.108`).

* Moved the t166 validity-capture test graph under framework-owned `.aidlc/`, so the test-only fixture no longer invalidates the recorded review's bound workspace-roof source. The complete t166 file passes: 42 tests.
* Regenerated and committed `tests/.coverage-registry.json` with the three newly enumerated functions; the registry suite passes: 36 tests.
* Rebase resolutions follow your guidance: kept the current intent-creation terminology with this PR's no-Git expectations in t305, retained the t314 renumbering, and re-slotted to `2.6.109`.
* The rebase also surfaced a semantic conflict with the plain-language refusal strings (#908, which postdates your rebase base): five t314 expectations updated to the current refusal contract, and the recorded-repo two-unit case now stales the unit's claimed scope before spending its bounded recovery — under accurate per-unit claims, re-reviewing an unchanged scope correctly no longer consumes the recovery.

Validation: t314-source-freshness 104 tests, t305 51, t166 42, t304/t305/t307 loop-backs, version sync, and coverage registry all passing; `bun run typecheck`, `bun run check`, and `bun scripts/package.ts --check` clean; merge-tree against `v2` conflict-free.

@leandrodamascena please re-review when ready.

leandrodamascena (changes requested):
Thanks for the thorough follow-up and for resolving the previous fixture and coverage-registry findings. I reviewed the current head `a2cac564a3fbd885eab75b5bd09458f1a1f5591d` against current `v2` at `33e5d01907a556360342104d35752c7b51deb198`. Project alignment remains established by the previous maintainer reviews, and the scope remains within the accepted non-Git source-freshness work.

**P1: External symlink targets make the supposedly immutable Source Commit validation depend on live filesystem bytes**

`filesystemSourceIdentity()` records an external symlink target under `<link>@target` (`core/tools/aidlc-lib.ts:8757`), while `gitCommitSourceListing()` reconstructs a commit by checking it out and walking that live target (`core/tools/aidlc-lib.ts:7997`). The commit itself contains only the symlink text.

I reproduced this on the current head: without changing the commit, editing its external target changed the reconstructed entry from `e8762c...` to `2190a3...`. The post-merge validator then uses these mutable listings as the committed delta (`core/tools/aidlc-worktree.ts:2147`). A target changed after the pre-merge check at `aidlc-worktree.ts:1571` can therefore be accepted as committed reviewed content and receive `SWARM_SOURCE_MERGED` authority.

Please make commit reconstruction depend only on bytes represented by the tree. Bound swarm snapshots should reject external targets that cannot be represented immutably, or materialize them through an explicitly immutable representation. Add a regression covering target mutation during source landing.

**P2: An initialized submodule crashes source fingerprinting when the Git executable is unavailable**

The filesystem walker still invokes `git rev-parse` for nested repositories (`core/tools/aidlc-lib.ts:8840`) and reads `nestedHead.stdout` before checking whether spawning Git succeeded (`:8845`).

I reproduced this with one initialized submodule and a child process whose `PATH` contained no Git. `workspaceSourceFingerprint()` threw `TypeError: null is not an object`, rather than preserving the fingerprint or failing closed. This contradicts the advertised Git-independent identity.

Please remove the Git-executable dependency from this identity path, guard spawn failures, and add a cross-process regression with an initialized submodule and no Git on `PATH`.

**Rebase required**

Current `v2` added #835 after this branch's parent and conflicts in `aidlc-swarm.ts`, release metadata, generated distributions, and the coverage registry. A diagnostic rebase preserved both #835's reviewed-record snapshot transaction and this PR's `sourceFingerprint`/shared recovery-budget binding, then re-slotted the release to `2.6.110`.

Validation performed:

- Exact-head GitHub CI: all reported checks green.
- Diagnostic rebase: `bun run check`, typecheck, lint, package parity, and `git diff --check` passed.
- Focused cross-contract suite: 347 tests passed across source freshness, per-unit attribution, swarm routing/referee, multi-repo construction, loop-backs, version sync, and coverage registry.
- The Windows argv stress test timed out locally again at 214 seconds; it passes on exact-head CI and remains a local validation gap.

No commits were pushed because this is maintainer-owned work. The remaining blockers are the two findings above and the required rebase.

apackeer:
Rebased once more: #835 (swarm finalize merge-back) landed on `v2` right after the previous push and took `2.6.109`, so this is re-slotted to `2.6.110`. The finalize convergence path now composes #835's reviewed-record snapshot capture with this PR's bounded source identity and submodule-recovery budget. Re-validated: t134-swarm-referee (23), t135, t201, t314-source-freshness (104), t305 (53), t166 (42), loop-backs, version sync, and coverage registry all passing; typecheck/lint/packager parity clean; merge-tree against `v2` conflict-free.

leandrodamascena (changes requested):
Thanks for rebasing this and integrating #835. I reviewed the current head `d5bd756e0c3b9577264e12c8fc56494de14358dd` against `33e5d01907a556360342104d35752c7b51deb198`.

Project alignment remains established by the previous maintainer reviews, and the scope has not expanded. The two blockers from my previous review are still present.

**P1: Source Commit reconstruction still depends on a live external symlink target**

`gitCommitSourceListing()` checks out the commit and then calls `filesystemSourceIdentity()` (`core/tools/aidlc-lib.ts:8043-8071`). The symlink handling follows an external target and records its current bytes under `<link>@target` (`core/tools/aidlc-lib.ts:8765-8841`).

I reproduced this again on the current head. Without changing the commit, changing only the external target changed the reconstructed entry from `5d8f65...` to `f4918c...`.

This listing is later used to reconstruct the committed source delta (`core/tools/aidlc-worktree.ts:2473-2501`). A target changed during landing can therefore be treated as content from the immutable commit after the pre-merge check at `:1572`.

Please make Source Commit reconstruction depend only on bytes represented by the Git tree. For bound swarm snapshots, external targets need to be rejected or represented immutably. This also needs a regression where the target changes during landing.

**P2: An initialized submodule still crashes when Git is unavailable**

The filesystem walker runs `git rev-parse` for a nested repository and calls `nestedHead.stdout.trim()` before checking the spawn result (`core/tools/aidlc-lib.ts:8906-8924`).

I reproduced this with an initialized submodule and no `git` executable on `PATH`. `workspaceSourceFingerprint()` exits with:

```
TypeError: null is not an object (evaluating 'nestedHead.stdout.trim')
```

Please remove this Git executable dependency from the identity path, or handle the spawn failure and fail closed with a proper error. Please also add a cross-process regression for an initialized submodule with no Git on `PATH`.

Validation performed:

- Reproduced both findings on the current head.
- Focused official tests: 3 files, 61 assertions, all passing.
- Typecheck and package parity passed.
- Version synchronization: 7 tests passed.
- Coverage registry: 36 tests passed.
- `git diff --check` passed.
- The Windows command-line stress test timed out locally again. It passes on exact-head CI.
- All GitHub checks are green for `d5bd756e`.

No fixes or commits were pushed because this is maintainer-owned work. The remaining risk is that mutable external bytes can receive source authority and the no-Git submodule path can crash.

apackeer:
Addressed the latest review blockers in `38a38125a` and rebased onto the current `v2` head, re-slotting the release to `2.6.115`.

* Source Commit reconstruction is now tree-derived: symlinks contribute only committed link text, so immutable commit listings never depend on live external target bytes. Bound swarm snapshots reject external target chains before Source Commit authority is minted, while valid internal registered aliases remain supported.
* Source-control metadata is fail-closed: symlinked source registries and worktree control metadata cannot influence immutable reconstruction, including dangling and out-and-back link chains.
* Initialized nested repositories and submodules no longer require the Git executable for source identity. HEAD is resolved from bounded `.git`/`commondir`/ref metadata and verified as a real commit in loose or packed object storage, including linked worktrees and v1/v2 pack indexes.
* Added regressions for landing-time external-target mutation, external and out-and-back links, symlinked control files, packed detached submodules without Git on `PATH`, linked-worktree shared refs, malformed OIDs/config, and missing commit objects.
* Preserved the current per-Unit attribution, no-DAG continuity, reviewed-record snapshot, and bounded new-submodule recovery contracts from current `v2`.

Validation:

* `t314-source-freshness-receipts`: 110 assertions passed.
* Version sync, coverage registry, and `t305`: 96 assertions passed.
* `t166-multi-repo-construction`: 42 assertions passed.
* `t134-swarm-referee`: 23 tests passed.
* `bun run check`, package parity across all seven harnesses, and `git diff --check` passed.
* All completed GitHub checks are green; smoke/unit CI is still running.

@leandrodamascena please re-review when ready.

leandrodamascena (changes requested):
Thanks for addressing the external-symlink and no-Git submodule blockers. I reviewed current head `38a38125aea861dd21134ab392cb706c3127f882` against current `v2` at `7f42a9081ebf4e60bbf82325d9fe037293a34960`.

Project alignment remains established by the previous maintainer reviews, and the scope has not expanded.

**P1: Mutable checkout filters can still receive Source Commit authority**

`gitCommitSourceListing()` describes its result as tree-derived, but reconstructs the commit through `git checkout-index` (`core/tools/aidlc-lib.ts:8318-8350`). That command applies repository-local `smudge`/`process` filters, so the resulting listing can depend on mutable bytes outside the commit.

I reproduced this in two ways:

- The same immutable commit SHA produced different source-listing hashes after changing only `filter.<name>.smudge`.
- In the complete swarm flow, I changed an external filter payload after review and convergence. The squash merge succeeded, the main checkout contained `UNREVIEWED` instead of the reviewed bytes, and `SWARM_SOURCE_MERGED` was still emitted.

The post-merge validation at `core/tools/aidlc-worktree.ts:2476-2500` uses the same filtered checkout reconstruction, so it validates the transformed bytes rather than bytes represented by the commit.

Please derive immutable listings directly from tree/blob contents or otherwise guarantee that checkout transformations cannot participate. Add a regression where mutable smudge output changes after convergence and verify that no source authority is emitted.

No fixes or commits were pushed because this is maintainer-owned work.

Validation:

- Exact-head GitHub checks are green.
- `t314-source-freshness-receipts`: 109 passed; the known Windows command-line stress test timed out locally and passes on exact-head CI.
- The additional end-to-end smudge-filter reproduction passed, confirming the authority bypass.
- The PR now conflicts with current `v2` after #828 and requires a rebase and version re-slot.

The remaining risk is that bytes not represented by the reviewed Source Commit can become active source and receive merge authority.

leandrodamascena (changes requested):
Thanks for addressing the previous symlink and no-Git submodule blockers. I reviewed current head `38a38125aea861dd21134ab392cb706c3127f882` against current `v2` at `7f42a9081ebf4e60bbf82325d9fe037293a34960`.

Project alignment remains established by the previous maintainer reviews, and the scope remains within the accepted source-freshness work.

**P1: Local smudge filters can authorize bytes the reviewer never saw**

`gitCommitSourceListing()` materializes an immutable commit through `checkout-index`, then hashes that checkout (`core/tools/aidlc-lib.ts:8337-8350`). Git applies repository-local `filter.<name>.smudge` or process filters during that checkout, so the result is not tree-derived.

The post-merge validator uses these transformed listings as the expected committed delta (`core/tools/aidlc-worktree.ts:2476-2500`).

I reproduced this with a non-idempotent smudge filter:

- Review and finalize bound `export const reviewed = true;`.
- The source merge returned success.
- The main checkout contained `SMUDGED:export const reviewed = true;`.
- `SWARM_SOURCE_MERGED` was emitted for those transformed bytes.
- Changing only the smudge command also changed the listing reconstructed for the same immutable commit.

A smudge-only variant, with no clean filter configured, also returned a successful merge. This gives source authority to bytes outside the reviewed fingerprint.

Please reconstruct commit listings directly from tree modes and blob bytes, without checkout conversions, and prevent or reject smudge transformations before mutating the target checkout. Add an end-to-end regression using a smudge-only, non-idempotent filter.

**Rebase required**

`v2` advanced to `7f42a908` while this review was running. GitHub now reports the PR as conflicting. `CHANGELOG.md` and `tests/.coverage-registry.json` conflict, and `2.6.115` is already occupied, so this needs rebasing and re-slotting to `2.6.116` or the next available version.

The previously reported external-symlink and no-Git initialized-submodule blockers are fixed.

Validation performed:

- Reproduced the P1 both at listing level and end to end.
- Typecheck and seven-harness package parity passed.
- Version synchronization and coverage registry: 43 tests passed.
- Changed `t166`/`t304`/`t307`/`t305` slice: 97 passed; one macOS reference-transaction hook injection did not fire locally.
- `t314`: 109 passed; the known Windows argv stress case timed out locally and caused its cleanup hook to time out.
- All exact-head GitHub checks are green, but they predate the new `v2` base.

No fixes or commits were pushed because this is maintainer-owned work. Remaining risk is the smudge-filter authority bypass plus post-rebase integration.

apackeer:
Addressed the smudge-filter review blocker in `7dc4c96ca`.

* `gitCommitSourceListing()` now enumerates the commit with `ls-tree` and reads raw object bytes through `cat-file --batch`; it no longer uses `checkout-index`, so smudge/process, EOL, ident, and working-tree-encoding conversions cannot change the listing reconstructed for an immutable commit.
* Bound source merges now inspect effective repository configuration before the audit/mutation boundary and refuse any configured `filter.<name>.smudge` or `filter.<name>.process` command. The explicit `AIDLC_SKIP_SOURCE_FRESHNESS=1` override retains its existing behavior.
* Added a listing-level regression proving that changing only mutable smudge output leaves the immutable commit listing unchanged.
* Added a smudge-only end-to-end regression proving the merge is refused before target HEAD, index, or worktree mutation, no `[merge-succeeded:...]` state is reached, and no `SWARM_SOURCE_MERGED` authority is emitted.
* Preserved the existing initialized/new-submodule recovery behavior and all prior source-boundary regressions.

Validation:

* `t314-source-freshness-receipts`: 112 tests passed.
* Version synchronization and coverage registry: 43 tests passed.
* `bun run check`, `git diff --check`, and package parity across all seven harnesses passed.
* All current GitHub checks on `7dc4c96ca` are green.

The branch was rebased onto `v2` at `7f42a9081` and re-slotted to `2.6.116` as requested. Since that push, `v2` advanced again to `3ac57def5`; GitHub currently reports the PR as conflicting in release metadata, so one more rebase/re-slot is still required before merge.

@leandrodamascena the smudge/process authority finding itself is addressed in this head.

leandrodamascena (approved):
Thanks for addressing the smudge/process-filter authority blocker.

I reviewed the current head `7dc4c96caecaceab89a83d38c85b02ccef8585f7` against current `v2` at `2399542dd53ea598a6674c93bdd15f3118ecf6d8` (merge base `7f42a9081ebf4e60bbf82325d9fe037293a34960`).

Project alignment remains established by the previous maintainer reviews, and the scope remains within the accepted source-freshness work. I found no new implementation blockers in this head.

The previous finding is addressed:

- Immutable commit listings are reconstructed from raw tree modes and blob bytes through `ls-tree` and `cat-file`, without checkout filters or other worktree conversions.
- Bound merges reject effective smudge/process filter configuration before target mutation or source-merge authority.
- The listing-level and end-to-end regressions both pass locally: 2 tests, 12 assertions, 0 failures.
- All exact-head GitHub checks are green.

Follow-up before merge: rebase onto current `v2`, re-slot from `2.6.116` to the next available version after `2.6.121`, reconcile the authored `aidlc-lib.ts`, audit documentation, and `t314` conflicts semantically, regenerate `dist/`, and rerun the full suite. The current green checks predate that reconciliation.

No fixes or commits were pushed because this is maintainer-owned work. Remaining risk is limited to the required integration with the newer base.

PR #906, opened 2026-08-24
fix: reduce minimal workflow context overhead

apackeer:
## Summary

- Restrict Minimal Intent Capture and Requirements Analysis to stage-relevant shipped knowledge while preserving active plugin, rule, persona, consumes, and user/team context.
- Stop repeating artifact and scan bodies in subagent returns; use durable files plus path, decision, and concern summaries.
- Move the Reverse Engineering developer-to-architect handoff to a digest-bound record file and remove the duplicate conductor source scan.
- Harden isolated pipeline lifecycle, one-repository identity, reuse freshness, required CodeKB artifact validation, and recursive plugin-knowledge provenance.
- Preserve both Reverse Engineering agents, all nine CodeKB artifacts, sensors, reviews, and approval gates.

## Measured Result

- Representative native Windows Reverse Engineering run: 670s to 574s.
- Output tokens: 51,267 to 47,863.
- Conductor source-discovery calls: 1 to 0.
- Agent dispatches and nine-artifact output remained unchanged.
- Recorded cost increased from $4.92 to $5.30; this demonstrates elapsed-time and context improvement, not a cost reduction.

## Testing

- Final exact-commit local slice: 7 files, 132 assertions, 0 failures.
- Native Windows targeted slice during implementation: 7 files, 132 assertions, 0 failures.
- Recorded native Windows benchmark preserves pipeline, artifact, sensor, digest, approval, and validation invariants.
- Package parity passed for all seven harness distributions.
- Codex implementer/reviewer loop reached `MERGEABLE` after the final r4 closure review.

No full repository test suite was run.


leandrodamascena (changes requested):
Thank you for the substantial work reducing Minimal workflow context while strengthening pipeline and reuse evidence.

I reviewed the current head `607e16fa2ab4349860f6d753f8789508ce31e489` against base `c06482d267c62afc6f87cbd23ba0414e8fad5d0b` (merge base `75d8af80ac46890ccd863acb9464aa42ee8ac5a8`).

The direction is aligned with the project: this is maintainer-owned performance and internal-integrity work intended to preserve the existing workflow and artifact contracts.

**[P1] Reject symlinked Reverse Engineering handoffs**

`core/tools/aidlc-log.ts:741-776` validates and hashes the developer handoff with `statSync` and `readFileSync`, both of which follow symlinks. The receipt consumer repeats this in `core/tools/aidlc-lib.ts:8614-8629`.

I reproduced this by making the exact expected `<record>/inception/reverse-engineering/developer-scan.md` path a symlink to a file outside the intent record. `aidlc-log link` returned success and minted the authoritative receipt. This bypasses the new durable handoff boundary and can make the architect consume content outside the record.

Please reject symlinks throughout the path and read only a regular file without following links in both receipt minting and verification. The existing `assertNoSymlinkInChainOrThrow` and `readRegularFileNoFollowOrThrow` primitives provide the intended pattern. Add regression coverage for a symlinked handoff.

**[P2] Preserve plugin provenance when knowledge is removed**

`scripts/plugin-hooks-template/compose.ts:1766-1789` rebuilds the ownership sidecar only from files still present in the plugin source. Because composition is no-clobber, a knowledge file removed in a plugin upgrade remains installed but disappears from the sidecar.

I reproduced composing recursive plugin knowledge, removing it from the plugin, recomposing, and explicitly deselecting the plugin. `core/tools/aidlc-orchestrate.ts:2094-2105` treated the stale file as unowned custom knowledge and still included it in Minimal `inline_context_paths`.

Please retain provenance for previously owned installed files or safely remove/tombstone them, so removed or disabled plugin knowledge cannot become unowned active context. Add an upgrade/removal plus plugin-deselection regression test.

**[P2] Regenerate and claim the coverage registry**

`bun tests/gen-coverage-registry.ts --check` fails at this head: the enumerated universe changes from 611 to 615. These new exports are currently `UNCOVERED`:

- `codekbStoreIsCurrent`
- `pipelineAttemptStartedAt`
- `singleStageAttemptIsOpen`
- `latestPipelineLinkArtifactMtime`

Please add honest `covers:` claims where exercised, add missing direct coverage where necessary, and regenerate the committed registry and ratchet.

Validation performed:

- Passed the six changed test files plus `t68-version-changelog-sync`.
- Passed packaging parity for all harnesses, typecheck, and lint.
- The two temporary adversarial regressions above failed as described and were removed afterward.
- `t248-codekb-scope-diff` has two unrelated failures that reproduce identically at the merge base, so I am not attributing them to this PR.

No fixes or commits were pushed because this is an internal contribution.

GitHub currently reports no CI checks, no prior reviews, and a conflicting merge state. The head still carries version `2.6.69`, while the current base is `2.6.85`; after rebasing, please resolve the changelog/version/badge conflict using the next available version and rerun the full guards.

**Proposed review event: `REQUEST_CHANGES`.**

apackeer:
Addressed the requested changes in `fbc4c32d4` and rebased the branch onto current `v2` (`f390fe3ae`), re-slotting the release metadata to `2.6.97`.

- Reverse Engineering handoffs now reject symlinks at the leaf or any path component during both receipt minting and verification. Reads are regular-file, no-follow, descriptor-bound, checked against the prevalidated real path before and after reading, and use the descriptor's mtime. Regression coverage includes leaf replacement, post-receipt replacement, and a symlinked parent directory.
- Plugin composition now retains prior provenance for still-installed knowledge files removed from a later plugin source. The regression composes recursive knowledge, removes it from the plugin, recomposes, deselects the plugin, and verifies the stale installed file is excluded from Minimal inline context.
- Added executable coverage for `codekbStoreIsCurrent`, `pipelineAttemptStartedAt`, `singleStageAttemptIsOpen`, and `latestPipelineLinkArtifactMtime`; regenerated the coverage registry and ratchet.
- Updated the older stage-artifact fixture and native binary gate to follow the new isolated-stage `STAGE_STARTED` lifecycle contract.

Validation on the pushed tree:

- `bun run typecheck`
- `bun run check`
- Focused post-rebase runs: 73 unit assertions, 110 `t185` assertions, 84 `t188` assertions, and the native binary build/gates all pass.
- The broad smoke+unit+integration run exercised 368 files; every red was either corrected or passed when rerun green-alone. The live Reverse Engineering SDK journey also passed.

core/tools/aidlc-lib.ts
```diff
@@ -11393,6 +11470,37 @@ function pipelineAttemptFloor(
   return floor;
 }
 
+export function pipelineAttemptStartedAt(
+  projectDir: string,
+  stageSlug: string,
+  options: { singleRun?: boolean } = {},
+): string {
+  const events = orderedPipelineEvidenceEvents(projectDir);
```
leandrodamascena, line 14541: **[P1] Preserve the shard-aware attempt ordering from #839**

These new helpers still rely on the pre-#839 `orderedPipelineEvidenceEvents()`, which flattens audit shards and resolves equal timestamps using concatenation/file-name order. I reproduced this with a same-second `GATE_REJECTED` in `aaaa-boundary.md` and both receipts in `zzzz-evidence.md`: `pipelineLinkEvidence()` accepts the entire stale chain. Current `v2` correctly treats that ordering as causally ambiguous and fails closed.

During the rebase, please integrate `pipelineAttemptStartedAt`, `singleStageAttemptIsOpen`, `latestPipelineLinkArtifactMtime`, and isolated reuse with #839's `AuditShardEvent` / `PipelineAttemptFloor` / `pipelineEventAfterFloor` model instead of restoring the flattened implementation. Please retain regression coverage for both shard filename orderings.

leandrodamascena (changes requested):
The previous handoff-symlink, plugin-provenance, and coverage findings are addressed. The focused pipeline, Minimal-context, single-stage, artifact-guard, and plugin-compose suites pass, as do package parity, typecheck, lint, and coverage-registry checks.

One ordering blocker remains above. The branch also needs rebasing onto current `v2` (`cb30d6422`), resolving the overlap with #839 without losing its fail-closed receipt semantics, and re-slotting `2.6.97` to the next available version. The PR is currently `CONFLICTING`.

apackeer:
Addressed the remaining shard-ordering blocker and rebased the branch onto current `v2` (`1c6e03a12`). The current head is `6bff98dc8`, with release metadata re-slotted to `2.6.102`.

- Preserved #839's shard-aware `AuditShardEvent` / `PipelineAttemptFloor` / `pipelineEventAfterFloor` model instead of restoring the flattened audit implementation.
- `pipelineAttemptStartedAt` now reads the shard-aware attempt floor.
- `singleStageAttemptIsOpen` fails closed when same-second isolated lifecycle events are causally ambiguous across shards.
- Current pipeline receipts and main/isolated reuse evidence are filtered through the shard-aware floor.
- Historical developer-handoff mtimes use the maximum matching value across shards, preserving the cross-attempt rewrite backstop without filename-order dependence.
- The original same-second `GATE_REJECTED` reproduction now asserts raw current receipts, reuse evidence, and the completed chain are rejected in both shard filename orderings.
- Added direct regressions for both filename orderings of ambiguous isolated start/completion events and cross-shard prior artifact mtimes.

The previously confirmed handoff-symlink, plugin-provenance, and coverage-registry fixes remain intact.

Validation:

- `bun run typecheck`
- `bun run check`
- `t315-pipeline-link-receipts`: 24 tests passed
- `t205` unit slice: 29 tests passed
- `t185-stage-artifact-guard`: 110 tests passed
- `t188-plugin-compose`: 84 tests passed

After the final test-only base advance from #901, I reran the package, coverage, typecheck, lint, and `t315` guards. CI is running on the pushed head.

@leandrodamascena, please re-review the current head when ready.

scripts/plugin-hooks-template/compose.ts
```diff
@@ -1717,7 +1752,48 @@ try {
...
+  const knowledgeSource = join(PLUGIN_ROOT, "knowledge");
+  const knowledgeTarget = join(HARNESS_DIR, "knowledge");
+  changed = copyTreeNoClobber(
+    knowledgeSource,
+    knowledgeTarget,
+    "knowledge",
+    undefined,
+    undefined,
+    undefined,
+    knowledgeWritten,
```
leandrodamascena: **[P2] Backfill ownership when upgrading pre-sidecar installs**

`knowledgeWritten` only receives paths written during the current compose. On an installation produced by the previous hook, an identical plugin knowledge file already exists, so `copyTreeNoClobber` skips it and no prior sidecar exists to preserve its ownership.

I reproduced this by pre-seeding an identical plugin knowledge file, running the current compose hook, then deselecting the plugin. The sidecar omitted the path and `next` still included it in the Minimal `inline_context_paths`.

Please cover the upgrade path where the installed identical file exists but the ownership sidecar does not, and add that case to the regression test.

core/tools/aidlc-lib.ts
```diff
@@ -11939,7 +11976,24 @@ export function readRegularFileNoFollowOrThrow(
   path: string,
   what: string,
   maxBytes?: number,
-): Buffer {
+  expectedRealPath?: string,
+): Buffer;
+
+export function readRegularFileNoFollowOrThrow(
```
leandrodamascena, line 13767: **[P2] Keep overload declarations from inflating the coverage ratchet**

`enumerateExportedFunctions()` matches every top-level `export function` declaration. These overloads therefore generate three identical `function:readRegularFileNoFollowOrThrow` rows. The function count rises from 410 to 416 although only four distinct exported function identities were added.

Please deduplicate enumerated units by class and ID, or otherwise ensure overload declarations count once, add a uniqueness assertion, and regenerate the registry.

leandrodamascena (changes requested):
Arden, thanks for rebasing this and preserving #839's shard-aware attempt ordering.

I reviewed the current head `6bff98dc84418909548eb090a3d54f16771fb41f` against `v2` base `1c6e03a1264c5376b854033adfae4f1e7ccef6eb`. This remains aligned with the maintainer-reviewed goal of reducing Minimal workflow context and Reverse Engineering overhead while preserving the existing artifact, pipeline, and audit contracts.

The shard-aware ordering from #839 is now preserved, including fail-closed handling of same-second cross-shard evidence. The previous handoff-symlink fixes also remain intact.

Two P2 issues remain: plugin provenance is not backfilled when upgrading an installation created before ownership sidecars existed, and the new overload declarations inflate the coverage registry.

Validation: all current GitHub checks are green; `t315` passes 24/24; the focused gate, Minimal-context, plugin-compose, package-parity, typecheck, lint, and coverage checks pass. The two timeout-only failures observed under concurrent local execution passed when rerun serially.

Requesting changes for the two inline findings.

apackeer:
Addressed the two remaining P2 findings in `08765713e`.

- Plugin composition now treats a byte-identical installed knowledge file as composed for ownership-sidecar migration. This backfills ownership for installations created before sidecars existed, while different-content collisions remain unowned and drop-logged. The regression pre-seeds identical knowledge without a sidecar, verifies ownership backfill and retention after source removal, deselects the plugin, and confirms Minimal context excludes the stale installed file.
- Coverage enumeration now deduplicates exported units by class and ID, so TypeScript overload declarations count once. A uniqueness regression pins `readRegularFileNoFollowOrThrow` to one unit, and the regenerated registry/ratchet now report 414 function units and 223 covered functions instead of the inflated 416/225.

Validation on the pushed head:

- `bun run typecheck`
- `bun run check`
- `gen-coverage-registry`: 35 tests passed
- `t188-plugin-compose`: 84 tests passed

@leandrodamascena, please re-review the current head when ready.

leandrodamascena (approved):
Arden, thanks for addressing the remaining findings.

I reviewed head `08765713e5c8ac4fa42abe75ed1bdffada9a2006` against current `v2` `7f42a9081ebf4e60bbf82325d9fe037293a34960`. The implementation remains aligned with the previously reviewed direction.

The plugin-provenance and coverage-registry findings are resolved. Focused validation passed: plugin composition 84/84, coverage registry 35/35 plus freshness check, and pipeline receipts 24/24. All current GitHub checks are green. No fixes or commits were pushed.

Approved for implementation correctness at this head.

Follow-up before merge: rebase onto current `v2`, preserve the overlapping isolated-stage and plugin-composition behavior, re-slot the release metadata to `2.6.116`, regenerate derived artifacts, and rerun CI. The rebased head should receive a final delta check because the PR is currently conflicting.

**Review event: `APPROVE`.**

PR #833, opened 2026-08-20
fix: actionable Review brief at every reviewer-backed human gate

apackeer:
Reviewer-backed approval gates now open with a compact Review brief: the stage, a plain-language outcome, why the decision is needed now, the findings table with stable R-NN IDs, exact artifact path + section, required action and status, and what Approve vs Request Changes each does. Re-review dispatches extract the prior findings table before replacing `## Review`, so finding IDs and dispositions (New/Unresolved/Resolved/Rejected/Accepted risk) survive revisions and stale-artifact recovery instead of being silently replaced. Pinned by the new t304 unit test across core and all seven dist projections. Closes #765

PR #908, opened 2026-08-24
fix: plain-language error recovery narration and refusal strings

apackeer:
## Summary

The voice work in #707 (issue #682) made the conductor speak in the project's terms on clean paths, and named one residual it did not close: error recovery. When a framework check refuses an action, the conductor relays or paraphrases that refusal, and its register collapses back into internals: output-declaration syntax, protocol section cites, audit event names. A recent field session confirmed it, spending over an hour narrating engine vocabulary at a user while retrying a blocked action.

This adds a refusal-narration contract to every harness orchestrator and rewrites the refusal strings a user is most likely to hit. When a check declines an action, the conductor states in one plain sentence what was declined and why in project terms, gives the next step, stops retrying after two consecutive refusals of the same action and escalates to the human, and diagnoses from the refusal itself rather than by reading framework source.

**Upgrade:** refresh the whole `dist/<harness>/` shell. Existing workspaces keep their own memory tree, so no record migration is needed.

## What changes for users

* The refusal-narration clause ships byte-identical across Claude, Codex, Copilot, Cursor, Kiro CLI, Kiro IDE, and opencode, pinned by `t325`.
* Roughly forty gate, question, review, collaboration, reviewer-scope, plan-approval, and state-transition refusals now name the project action that was declined and a next step, instead of internal vocabulary. For example, a per-Unit review gap now reads "Changed after review: alpha. Not yet reviewed: none" rather than "Invalidated receipts: alpha".
* After two consecutive refusals of the same action, the conductor is required to stop and present the state and options to the human rather than continuing to retry.
* Exact commands, paths, stage slugs, and machine-readable audit fields are preserved wherever recovery or wire compatibility depends on them; only prose changed.

## Verification

Full `smoke+unit+integration` tier: 355 files, 8810 assertions, 0 failures. `package.ts --check`, `typecheck`, `biome`, and `t68` clean.

Every reworded string had its assertions re-pointed at the shipped text in the same change, per the repository's documentation policy. No assertion was deleted or weakened to accommodate a rewrite, and total assertion count increased.

## Scope notes

* The stale-recovery refusals are deliberately untouched here. PR #903 owns those and makes them state-aware, because their previous advice to present the approval gate is unreachable in some of the states that trigger them. Merge #903 first; this branch was written to rebase cleanly on top of it.
* Approve-path refusals are reworded; the reject-path equivalents still carry the older phrasing, so the register is not yet uniform across that pair. Left as a follow-up rather than widening this change.
* Lower-frequency parsing, knowledge-index, audit-protection, and malformed-state errors are not in this pass. They are a later inventory sweep.


leandrodamascena (changes requested):
Thanks for extending the project-language voice into recovery paths. I reviewed the current head `0b082fcbb` against current `v2` (`f73472aa`). The direction is aligned with the merged #707 work and the accepted recovery behavior from #903.

P1 — Preserve #903’s state-aware recovery routing during the rebase

This head predates #903 and still routes exhausted recovery to a fixed approval-gate response in `core/tools/aidlc-log.ts` and `core/tools/aidlc-state.ts`. `t325` explicitly pins that superseded wording.

Current `v2` replaced it with state-aware guidance because the approval gate is unreachable from some active, revising, completed, and off-plan states. I reproduced semantic rebase conflicts in exactly these files. Resolving them in favor of this PR would restore the recovery loop that #903 fixed.

Please rebase onto current `v2`, preserve #903’s `recoveryGuidance` behavior and signatures, and update `t325` to assert the combined plain-language and state-aware contract.

P2 — Resolve the contradictory refusal instructions

Every harness now says that a tool, hook, or workflow-check refusal must be translated and may be retried up to twice. A few lines later, the `directive.kind === "error"` arm still requires printing `directive.message` verbatim, stopping immediately, and never retrying.

Report and transition failures are emitted as error directives, so the same result currently has two mutually exclusive instructions. Please define the boundary explicitly: either error directives are already user-facing and terminal and therefore exempt from refusal translation, or the error arm must follow the new translation contract. Add coverage that prevents these rules from diverging again.

P2 — Define refusal identity and reset semantics

“Two consecutive refusals of the same action” does not define what identifies an action or what resets the count. A corrected retry can use different arguments for the same project action, while an unrelated successful call may occur between refusals. Depending on interpretation, the conductor can either escalate prematurely or continue the loop this change is intended to stop.

Please define the identity and reset rules and cover representative sequences. `t325` currently checks only that the sentence exists; it does not catch contradictory or incomplete retry semantics.

Validation on the reviewed head:
- Official focused unit run: 7 files, 150 assertions, 0 failures
- Typecheck and Biome
- Version/changelog synchronization
- Existing GitHub CI is green for this head

The PR is currently conflicting with `v2`, and the affected recovery files require semantic resolution rather than only a version bump.

Requesting changes.

apackeer:
Addressed the review on head `20927e49d`, rebased onto current `v2` (`cb822bae`).

Changes:
- Preserved #903 state-aware `recoveryGuidance` routing in `aidlc-log.ts` and `aidlc-state.ts` for active, revising, completed, and off-plan stages.
- Made terminal `directive.kind === "error"` messages explicitly user-facing, verbatim, terminal, and exempt from refusal translation/retry.
- Defined refusal identity as project operation plus target. Corrected incidental arguments retain identity; unrelated successful actions do not reset it; success, explicit abandonment, or a workflow transition to another operation/target starts a fresh count.
- Expanded deterministic coverage for representative retry sequences, terminal error parity, state-aware recovery routing, and byte-identical behavior across all seven harnesses (`t327`).
- Re-slotted the user-visible change to `2.6.95` dated 2026-08-26 and regenerated every `dist/<harness>/` tree.

Validation:
- Post-rebase focused unit slice: 4 files, 124 assertions, 0 failures.
- Final version/changelog sync slice: 7 assertions, 0 failures.
- `bun run check`: package parity, all three TypeScript configs, and Biome passed.
- Pre-push outgoing scan: clean.

A broader affected-unit run exposed timing instability in `t249-copilot-adapter`; an isolated rerun also timed out across several unrelated long adapter cases. This PR changes only expected refusal text in that file, not those timed execution paths.

@leandrodamascena could you please re-review the current head?

leandrodamascena (approved):
Thanks for the thorough follow-up. I reviewed head `20927e49d` against current `v2` (`1c6e03a`) and validated that the previous protocol and recovery blockers are resolved.

The implementation preserves #903’s state-aware recovery, clearly separates terminal error directives from retryable refusals, and defines consistent refusal identity and reset semantics across all harnesses.

Validation included the focused recovery-language tests and package parity across all seven harnesses. `t327` passes completely.

Follow-up before merge:

- Update the stale expectations in `t321` and `t323` to match the new plain-language verdict message.
- Rebase onto current `v2`, preserving #839’s isolated pipeline receipts and #833’s review-brief behavior.
- Re-slot the release version and regenerate the distribution trees.

These are integration follow-ups and do not change my assessment of the implementation.

Approved.

PR #909, opened 2026-08-24
fix: filter co-located test/fixture payloads out of plugin tools/ composition

### LINKED ISSUE #876: [Bug]: Compose copies the plugin tools/ tree recursively with no filter, so co-located tests and fixtures land in the engine tree

apackeer:
Compose copied a plugin tools/ tree recursively with no content filter, so co-located tests and fixtures (nested tests/, __tests__/, fixtures/, and *.test.ts / *.spec.ts files) landed silently in the installed engine tools tree. Adds a toolsTestPayloadPrecheck to the plugin compose hook that skips those payloads with a recorded advisory drop (plus an up-front audit of already-landed payloads, which are reported but never deleted), documents the top-level tests/ convention in the plugin authoring guide, and ships as 2.6.75. Closes #876


leandrodamascena (changes requested):
Thanks for addressing #876 at the compose boundary and documenting the top-level plugin test convention. I reviewed the current head `5e8c34f829c51f618b40915ebb45ce48ec1da745` against `v2` at `ff012fd94474fe9e4ad07aa9d1d54a6fb79400f8` (merge base `3f3df4dda4b30ae8e585b5fa162394d5984b1307`).

**[P2] Audit stale installed payloads even after they disappear from the new plugin projection**

The migration audit at `scripts/plugin-hooks-template/compose.ts:652` walks only the current `PLUGIN_ROOT/tools` tree and then checks whether each matching source-relative path exists under the installed tools tree. This misses the normal remediation path: a plugin author follows the new guidance, moves `tools/tests/run.test.ts` to top-level `tests/`, and publishes a corrected projection. The old compose-installed `.claude/tools/tests/run.test.ts` remains on disk, but it is no longer present under the new plugin’s `tools/`, so the loop never sees it and no advisory is recorded.

I reproduced this by preinstalling `.claude/tools/tests/run.test.ts` and composing a new `alpha` projection with no corresponding source file. The stale file remained installed and the compose drops output was empty. This breaks the changelog and upgrade contract that previously composed test payloads are audited and reported with removal guidance; `/aidlc --doctor` cannot tell the user what to remove in exactly the upgrade where the author has corrected the plugin layout.

Please make the installed-payload audit independent of whether the same file still exists in the current source projection, while retaining enough ownership information to avoid attributing another plugin’s file to this plugin. Add a regression where the destination contains the old test payload but the new plugin source no longer does.

**Project alignment**

This is aligned as the compose-side bug fix specified by internal issue #876: prevent non-tool test payloads from entering the installed engine tree and surface actionable migration evidence for files landed by older compose versions. The related #892 covers the author-side validation/toolchain half and does not replace this runtime compose fix.

**Validation**

- The PR head’s official `t188-plugin-compose.test.ts` integration run passed all 85 tests and 553 expectations.
- A synthetic integration on current `v2` at `ff012fd9`, with the authored changes applied and distributions regenerated, passed all 85 `t188` tests and 588 expectations.
- Packaging parity passed for all seven harnesses on both the PR head and the current-base integration tree.
- TypeScript typechecking passed.
- Biome completed with no errors or warnings; it reported three informational suggestions in the existing `t267-usage.test.ts`.
- Version/changelog synchronization passed all 7 `t68` tests on the PR head.
- The focused stale-payload diagnostic failed as described above: the installed file survived and the advisory output was empty.

I made no fixes, commits, or pushes because this is an internal contribution.

**Integration status**

The implementation and `t188` changes combine with the current base, but the PR is presently `CONFLICTING/DIRTY` in the changelog, README badge, authored version, and generated version copies. `v2` is now at `2.6.81`, while this branch still carries `2.6.75`. After fixing the migration audit, please rebase, re-slot the user-visible bug-fix version/changelog/README updates against the then-current base, regenerate all distributions, and rerun the relevant checks.

GitHub currently reports no CI results for this head, so remote CI remains unverified.

leandrodamascena (changes requested):
Thanks, Arden. I reviewed the current head `12366761` against its merge base `f30161fa` and the current `v2` head `b9fe852f`.

The previous blocker is resolved for a single-harness installation: the destination audit now finds stale payloads even after they disappear from the corrected plugin projection, without attributing legacy files to the currently composing plugin.

I found two remaining blockers:

**P2 – A clean second harness erases another harness’s advisory.**

`plugin-compose-installed-tool-payloads.drops` is shared across the project, but each compose scans only its own `HARNESS_DIR/tools`. I reproduced this with a stale payload under `.claude/tools`: Claude compose created the advisory, then a clean Codex compose deleted it while the stale Claude payload remained installed. Multi-harness installations are explicitly supported, so the record must either be harness-scoped or aggregate every installed harness without allowing the last clean harness to win. Please add a cross-harness regression.

**P2 – The installed-tools audit follows directory symlinks and can abort composition.**

The new destination scan uses `walk()`, whose `statSync()` follows symlinked directories. With a circular symlink under `.claude/tools`, compose hit `ELOOP`, returned externally with status 0, recorded a degraded drop, and failed to install the plugin’s tools. Please use bounded, symlink-safe traversal that does not escape or recursively revisit the installed tools root, with regressions for circular and external directory symlinks.

Validation on the current head:

- `t188-plugin-compose`: 85 passed, 0 failed, 593 expectations
- Version/changelog synchronization: 7 passed
- All GitHub CI checks are green

The PR remains `CONFLICTING/DIRTY` against current `v2`. After addressing these findings, please rebase, re-slot the version beyond current `2.6.108`, regenerate distributions, and rerun the affected checks.

Thank you for addressing the original migration gap. Requesting changes for the two reproduced destination-audit regressions.

apackeer:
Thanks for the thorough round-2 review — both findings reproduced exactly as described and are fixed in the current head `c8ce0f5ce`.

**Cross-harness erasure:** the installed-payload record is now keyed by the composing harness leaf (`plugin-compose-installed-tool-payloads-<harness>.drops`), so each compose overwrites or removes only its own harness's record; `--doctor` scans any `*.drops` name so the scoped records stay visible. Regression: a stale payload under `.claude/tools` keeps its advisory across a subsequent clean `.codex` compose in the same project.

**Symlink traversal:** the destination audit now uses a bounded lstat-based walk that never follows links — a circular, self-referencing, or external directory link can no longer ELOOP, abort the tools install, or pull unrelated trees into the audit, and a payload-named link is still flagged as a leaf. A failed scan degrades to keeping the previous harness record rather than erasing it from a partial result. Regressions cover circular and external directory symlinks (junction-based so they run on Windows too).

Both new tests fail on the previous behavior and pass with the fix. Rebased onto current `v2`, re-slotted to 2.6.110 (2.6.109 was taken by #835 mid-review), distributions regenerated with byte parity, and t188 (89/89), t68, typecheck, and packaging parity are green locally; all four GitHub workflows are green on this head.

leandrodamascena (approved):
Thanks, Arden. I reviewed head `c8ce0f5c` against `v2` at `33e5d019`.

The previous blockers are resolved: installed-payload records are harness-scoped, and the installed-tools audit no longer follows circular or external directory symlinks. The new regressions cover both cases, alongside incoming filtering and stale installed payload detection.

Non-blocking follow-up: consider adding an entry/depth budget to `walkInstalledNoFollow()`. The installed tree is user-writable, and very large trees can add noticeable SessionStart latency because each plugin runs the audit. This can be addressed separately.

Validation:

- `t188-plugin-compose`: 89 passed, 0 failed
- Version/changelog synchronization: 7 passed
- Packaging parity, typechecking, lint, and `git diff --check` passed
- All GitHub checks are green
- Head is mergeable against the current `v2`

I made no fixes, commits, or pushes.

Review event: **APPROVE**

PR #910, opened 2026-08-24
fix: doctor detects plugin composition erased by an engine reinstall

### LINKED ISSUE #875: [Bug]: Reinstalling dist/<harness> erases composed plugin state, and no doc or check says to re-sync
### Description

Field report on #793 (point 2): composition is copy-into-the-engine-tree — the compose hook writes plugin stages/scopes/agents/knowledge/sensors/tools directly into the installed harness dir (`scripts/plugin-hooks-template/compose.ts:1390-1434`). An engine upgrade performed the documented way (copy the new `dist/<harness>/` over the install) removes everything composition put there.

Nothing surfaces this:

- No user-facing doc says "after upgrading the engine, re-run plugin sync" (`docs/guide/` has no hit for it; the authoring guide only discusses re-compose idempotency, `docs/harness-engineering/10-authoring-a-plugin.md:217`).
- `plugin-sync` exit status doesn't distinguish "recomposed N files" from "nothing to do", so an upgrade script can't verify recomposition happened (adjacent to #806).
- `/aidlc --doctor` has no check for "harness.json enables plugin X but its composed files are missing".

The field workaround was an explicit recomposition step verified by checking the files landed, plus a hash-baseline branch in their integrity check.

### Proposed

- Document the upgrade procedure: reinstall dist, then re-run plugin sync, in the User Guide and the authoring guide.
- Add a doctor check: enabled plugin whose composed surface is absent → flagged with the re-sync command as the fix. (Fits the plugin-extensible doctor-check seam from #796/#797.)

### Version

v2 (alpha)

### Area

Plugin mechanism / upgrade lifecycle

### Additional Context

Split out of the field report on #793 because it affects installed users regardless of the external authoring toolchain; #793's CLI cannot fix it.

apackeer:
Reinstalling dist/<harness> overwrites the compiled stage graph and core stage sources, silently erasing composed plugin graph entries and contribution merges while doctor stayed green. Adds a failing "Composed plugin surface" doctor check (missing plugin-owned graph entries + stale contribution sidecars, fix text naming /aidlc plugin sync), routes the uncompiled-stage advisory for plugin-owned files to plugin sync, and documents re-composition as an explicit engine-upgrade step across the User Guide, authoring guide, mechanism reference, and README. Guarded by the new tests/integration/t314-plugin-reinstall-doctor.test.ts (reinstall→fail→re-compose→heal, plus a stale-sidecar-only case).

Note: origin/v2 has since landed its own 2.6.74, so this branch's version slot is stale; per the changelog conflict policy the slot re-bumps (with the matching CHANGELOG heading rename) at landing.

Closes #875


leandrodamascena (changes requested):
Thanks for addressing #875. The upgrade guidance and the general doctor-check direction are aligned with the maintainer-authored issue.

I reviewed the current head 35458f0adcee1aa9895c5181348bc3786f3f6a14 against 4c666fbc8badf7999503a2ec5620eaadef18e016.

[P2] `core/tools/aidlc-utility.ts:2255-2267` fails open when an enabled plugin's contribution sidecar is unreadable, malformed, or references a stage that no longer exists. These paths are skipped or emitted as passing advisories, leaving `missingComposition` empty and allowing doctor to exit 0. I reproduced this with corrupt JSON and a valid sidecar targeting a removed stage. These cases need to fail the Composed plugin surface check and name `plugin sync` as recovery, with regression coverage.

[P2] The check cannot detect erased prose-only contributions. The compose hook persists a sidecar only when structural additions were made; fragment sentinels are not recorded. A plugin containing only a valid prose fragment therefore leaves no sidecar or plugin stage evidence. After reinstalling the engine, the fragment disappears and doctor still exits 0. I reproduced this with a prose-only plugin projection. Persist and verify fragment provenance, or provide an equivalent fail-closed signal, and add a reinstall regression test.

The existing two t314 scenarios pass. A diagnostic integration with current `v2` also passed those tests plus all 15 `plugin-sync` handler tests. The three adversarial probes above fail on the current head as described.

The branch is currently conflicting and has no CI results. It also needs rebasing onto current `v2` and re-slotting from `2.6.74` to `2.6.88`.

apackeer:
Addressed the requested changes and rebased onto current `v2` at `7573cd5b9` (`2.6.95`).

- The **Composed plugin surface** check now fails closed for unreadable, malformed, empty, or invalid enabled-plugin sidecars, and for sidecars whose target stage no longer exists.
- Compose now persists each successfully applied prose fragment's anchor, order, and content hash in the existing contribution sidecar, including for prose-only plugins. Doctor verifies the exact sentinels and enclosed prose hash.
- A corrupt sidecar is no longer replaced with incomplete fragment-only provenance; recovery preserves the failure until the stock engine is refreshed, the invalid sidecar is removed, and `plugin sync` can rebuild complete provenance.
- Disable/re-enable cleanup remains byte-stable, and all generated harness/plugin distributions were regenerated.

Validation:

- `bun run typecheck`
- `bun run lint`
- focused integration: `t188-plugin-compose`, `t224-plugin-selection`, and `t314-plugin-reinstall-doctor` - 101 assertions, 0 failures
- version/changelog sync: `t68-version-changelog-sync` - 7 assertions, 0 failures

@leandrodamascena, could you please re-review the updated head?

leandrodamascena (changes requested):
Thank you for addressing the earlier review findings. I reviewed the current head `7573cd5b97fc8844509ae0ecfa47a8afcf5c9fc9` against the current `v2` head `1c6e03a1264c5376b854033adfae4f1e7ccef6eb`. Alignment is established by issue #875 and the prior substantive maintainer review. The previous fail-closed and prose-provenance blockers are resolved.

**[P2] Preserve and verify the complete `consumes` contract**

Compose currently records only each consumed artifact name in the sidecar ([compose.ts#L1475-L1503](https://github.com/awslabs/aidlc-workflows/blob/7573cd5b97fc8844509ae0ecfa47a8afcf5c9fc9/scripts/plugin-hooks-template/compose.ts#L1475-L1503)), and doctor consequently compares only artifact membership ([aidlc-utility.ts#L770-L779](https://github.com/awslabs/aidlc-workflows/blob/7573cd5b97fc8844509ae0ecfa47a8afcf5c9fc9/core/tools/aidlc-utility.ts#L770-L779)).

This misses changes to `required` and `conditional_on`. I reproduced this by composing `test-pro`, changing its recorded consume from `required: false` to `required: true`, and running doctor. Doctor exited `0` instead of `1`.

This is reachable after an engine upgrade when a stock stage introduces the same artifact with different semantics: the artifact remains present, doctor reports the contribution as unchanged, and `plugin sync` deduplicates by artifact name so it cannot restore the plugin contract.

Please persist and compare complete consume entries (`artifact`, `required`, and optional `conditional_on`), retain compatibility with existing string-only sidecars, and add regression coverage for changed consume semantics.

The PR is also 8 commits behind `v2` and currently conflicting. After addressing the finding, please rebase, move the release metadata from `2.6.95` to the current next slot (`2.6.102`), regenerate distributions, and obtain fresh CI.

Validation on this head: the focused suites passed with 108 tests; `t314` passed all 5 tests/41 assertions; packaging parity, typecheck, lint, coverage registry, and `git diff --check` passed. The diagnostic regression failed as described and was removed afterward. All GitHub checks for this exact head are green.

No fixes, commits, or pushes were made because this is an internal contribution.

apackeer:
Addressed the latest review finding and rebased onto current `v2` at `edeca9c76` (`2.6.103`).

- Compose now persists the complete contract for each consume it actually adds: `artifact`, `required`, and optional `conditional_on`.
- The **Composed plugin surface** doctor check compares those fields exactly, so changes to requiredness or conditions fail instead of passing on artifact membership alone.
- Existing string-only sidecars remain valid and retain their legacy artifact-membership behavior.
- Disable/re-enable cleanup and Cursor's merge-aware installer accept both sidecar formats. Cursor transfers consume ownership to a newly adopted core entry only when the complete object-form contract matches.
- Regression coverage verifies `required` drift, `conditional_on` drift, object-form persistence, and legacy string compatibility.
- Release metadata was re-slotted to `2.6.103`, and all generated harness/plugin distributions were regenerated.

Validation:

- `bun run typecheck`
- `bun run lint`
- focused integration: `t188-plugin-compose`, `t224-plugin-selection`, and `t314-plugin-reinstall-doctor` - 102 tests, 0 failures
- focused unit: `t275-cursor-packaging` and `t68-version-changelog-sync` - 27 tests, 0 failures

@leandrodamascena, could you please re-review the updated head?

leandrodamascena (approved):
Thanks for addressing the previous finding. I reviewed head `edeca9c76e9d82ccf15b1aad9800276bcf1e9f70` against `v2` at `a329e8a4db7eb04dfa7f5cb065b8f8e3e9612c45`.

The complete `consumes` contract is now persisted and verified, including `required` and `conditional_on`, while legacy string sidecars remain compatible. The previous blocker is resolved.

Non-blocking follow-up: when upgraded core adopts the same artifact with different consume semantics, Cursor completes the reinstall but retains an incompatible sidecar; doctor correctly fails afterward, while plugin sync cannot reconcile it due to artifact deduplication. Please track this so the installer either rejects the conflict or defines explicit precedence.

Before merge, please rebase onto current `v2`, move the release metadata from the now-occupied `2.6.103` slot to `2.6.104`, regenerate distributions, and obtain fresh CI.

Validation: focused unit suites and package parity passed. `t224` and `t314` passed; the isolated lock-wait test from `t188` passed after a concurrent local run timed out. All GitHub checks for the exact reviewed head are green.

No commits or pushes were made because this is an internal contribution.

PR #911, opened 2026-08-24
feat: add rule lifecycle properties and stale drift triage

### LINKED ISSUE #874: [Feature]: Add OKF lifecycle properties (status, stale_after) to rule files so --doctor can distinguish stale from live rule drift
### Description

Add the OKF lifecycle properties `status` and `stale_after` as optional, file-level frontmatter fields on rule files (`org.md`, `team.md`, `project.md`, `phases/*.md`):

- `status` — one of `active` | `deprecated` | `draft`
- `stale_after` — ISO 8601 date (`YYYY-MM-DD`); a file whose date has passed is treated as stale even without an explicit `status: deprecated`

Both fields are validated at compile time (a bad enum value or malformed date fails the load naming the file) and are forward-compatible, additive, and optional — existing rule files never need rewriting.

The consumer is `/aidlc --doctor`'s rule-drift row: today every `##` heading that a team/project rule file shares with a populated `org.md` heading renders as a "review for contradiction" candidate. With these fields, overlaps on a file marked `status: deprecated` or carrying a past `stale_after` date instead render in a separate `Rule drift: N stale-suppressed` advisory row — still visible, still passing, never a silent drop. `status: active`, `status: draft`, future `stale_after` dates, and frontmatter-less files behave byte-identically to today.

### Use Case

The identified gap is **stale-vs-contradictory conflation** in post-write drift detection. A `team.md` heading that once matched `org.md` verbatim — and whose org guidance has since moved on — fires the same "review for contradiction" row on every `--doctor` run, forever. Doctor has no mechanical way to know the team rule is simply outdated rather than a live disagreement, so stale content keeps masquerading as unreviewed contradiction and drowns the real ones.

This closes that gap mechanically (a plain date/enum comparison, zero LLM involvement): a team whose whole `team.md` was superseded marks it `deprecated` (or stamps a `stale_after` date at affirmation time) and its overlaps triage into the stale-suppressed row instead of the live-candidate list.

Known limitation (deliberate, Tier 1 scope): the fields are file-level, so marking a file deprecated suppresses *every* heading's overlap in it, not just one stale heading. A heading-scoped acknowledgment marker (recording a resolved escalation on the specific `##` heading pair) is the natural follow-up.

### Version

v2 (alpha)

### Area

Other (rule system / `--doctor`)

### Additional Context

- This enhancement adds the OKF lifecycle properties for `status` and `stale_after`. OKF reference: https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#54-lifecycle-status — vocabulary only, not schema import: the memory layer keeps its filename-derived scope, zero-dep YAML frontmatter, and forward-compat unknown-key tolerance; only the two document-level lifecycle ideas map onto this subsystem's existing file-level frontmatter.
- `--doctor` remains fully deterministic and read-only: detection stays byte-reproducible heading/string work; the contradiction verdict stays with the orchestrator-LLM at observation time. These fields only change how a known-stale file's overlaps are triaged.
- Spec surface: `core/tools/aidlc-rule-schema.ts` (`RuleFrontmatter`), `core/tools/aidlc-utility.ts` (rule-drift block), `docs/reference/08-rule-system.md`.

apackeer:
Adds optional OKF lifecycle frontmatter (status: active|deprecated|draft, stale_after: YYYY-MM-DD) to rule files, validated at compile time with errors naming the file. /aidlc --doctor now triages overlaps from deprecated or date-expired team/project files into a separate passing "Rule drift: N stale-suppressed" row while live output stays byte-identical. Note: v2 CHANGELOG head has since moved to 2.6.74, so the committed 2.6.73 slot re-bumps at landing per the changelog conflict policy. Closes #874


leandrodamascena (changes requested):
Thanks for implementing the lifecycle metadata and stale-drift separation from #874. The overall direction is aligned with the accepted rule-system design.

I reviewed the current head `e65ffd2ec0e785fb9b7866a144d1d50673c0d79b` against `v2` at `1c0188a2f6ecf48152bb506bd06af747fc1c5203`.

**P2 - Reject lifecycle fields that are present without a supported scalar value**

In `core/tools/aidlc-rule-schema.ts`, `parseRuleFrontmatter()` only records `status` and `stale_after` when `scalarField()` returns a non-empty string. Blank values and unsupported block scalars therefore disappear before validation:

```yaml
status:
stale_after:
status: >
  deprecated
stale_after: |
  2026-01-01
```

All four cases currently parse as `{}` and pass `validateRuleFrontmatter()`. Consequently, compilation treats malformed lifecycle declarations as absent instead of failing with the promised file-specific diagnostic. Please preserve field presence through parsing, reject empty/unsupported scalar forms, and add compile-path regression coverage for these cases.

**P2 - Rebase and re-bump the release version**

The PR is currently conflicting with `v2`. The base is now `2.6.79` and already contains an unrelated `2.6.73` entry. Please rebase and bump this feature to `2.6.80`, updating the PR title, `CHANGELOG.md`, README badge, authored version source, and regenerated distributions. While resolving the troubleshooting conflict, preserve both the lifecycle stale-suppression text and the existing approval-gate advisory.

I validated the feature on a diagnostic rebase with 54 focused tests covering rule parsing, graph compilation, doctor behavior, and version synchronization. All seven harnesses passed `bun scripts/package.ts --check`, and `git diff --check` passed. No fixes or commits were pushed because this is an internal contribution. GitHub currently reports no CI checks for this head.

Once these two items are addressed, the remaining behavior appears additive and compatible: existing rule files remain valid, runtime rule delivery is unchanged, and only explicitly stale doctor output is reclassified.

**Review event: REQUEST_CHANGES**

apackeer:
Addressed both requested changes in `bd64daa25`:

- Lifecycle fields now preserve key presence when the scalar is blank or uses an unsupported block form, so validation rejects all four reported declarations with a file-specific diagnostic.
- Added regression coverage through both `loadRules()` and the full `compileStageGraph()` path for blank `status`, blank `stale_after`, folded `status: >`, and literal `stale_after: |`.
- Rebased onto current `v2` (`ff012fd94`) and re-slotted the release to `2.6.82`. The troubleshooting resolution retains both lifecycle stale-suppression and the long-wait approval-gate advisory.
- Regenerated all seven harness distributions.

Verification:

- Focused smoke/unit/integration slice: 5 files, 89 assertions, 0 failures.
- `bun scripts/package.ts --check`: all seven harnesses in sync.
- `bun run typecheck`: pass.
- `bun run lint`: pass.
- `git diff --check`: pass.

leandrodamascena (approved):
Thank you for addressing the previous review findings. I reviewed the current head `bd64daa25f4bfe1682f1d537eb54031abb85bd1e` against the current `v2` head `1c6e03a1264c5376b854033adfae4f1e7ccef6eb`.

Project alignment is established by issue #874 and the prior substantive maintainer review. The earlier lifecycle parsing and compile-path coverage blockers are resolved: blank and unsupported block-scalar declarations now fail with file-specific diagnostics.

I found no remaining implementation blockers. The lifecycle metadata remains additive, stale files continue participating in runtime rule delivery, and only their doctor overlap classification changes. Active, draft, future-dated, and frontmatter-less files preserve the existing live output.

Validation on the exact PR head:

- 49 focused rule-schema, graph-compilation, and doctor tests passed with 110 assertions.
- All GitHub checks are green, including smoke/unit, integration/e2e, parity/typecheck/lint, security, and changelog completeness.

I also performed a diagnostic rebase onto the current `v2`. After preserving the accumulated troubleshooting advisories and moving the release metadata to `2.6.102`, the same 49 focused tests passed. All seven harnesses passed packaging parity, typecheck passed, lint passed with only three unrelated existing informational suggestions, version/changelog synchronization passed, and the coverage registry remained fresh.

The PR is currently one commit ahead and 25 commits behind `v2`, with merge conflicts. Before merge, please rebase, move the release metadata from `2.6.82` to the current next slot (`2.6.102`), regenerate all distributions, preserve the current troubleshooting advisories, and obtain fresh CI.

No fixes, commits, or pushes were made to the author branch because this is an internal contribution. The only remaining risk is the actual rebase result and its fresh CI.

PR #913, opened 2026-08-24
fix: doctor detects a dead hook seam when the health dir is empty after a stage ran

### LINKED ISSUE #861: [Bug]: doctor reports hook heartbeats as "not yet fired" when the health dir is present and empty, so a dead hook seam passes
## What happens

The doctor's hook-heartbeat check documents three states and implements two. The state it documents as the failure — *"directory exists but no `.last` files → hooks registered but have never fired. Genuine drift; fail"* (`core/tools/aidlc-utility.ts:1874-1875`) — is unreachable, and that exact state reports **pass**.

`hasHookFiredContent` is set from the same predicate that populates the heartbeat list (`core/tools/aidlc-utility.ts:1891-1892`):

```ts
const files = readdirSync(healthDir).filter((f) => f.endsWith(".last"));
if (files.length > 0) hasHookFiredContent = true;
```

So with the health dir present and holding zero `.last` files, `hasHookFiredContent` is `false`, and control reaches the state-(a) branch (`:1912`) instead of the state-(b) branch (`:1922`):

```
✓ Hook heartbeats: not yet fired (first workflow stage will populate)
```

The failure branch is reachable only when `.last` files **exist** but every one of them is unreadable, so `heartbeatEntries` stays empty while `hasHookFiredContent` is true — a state the comment above the block does not describe.

The refinement that causes this is deliberate and its reasoning is sound: a dir holding only `hook-debug.log` or `.first-fired` should not be read as drift, or enabling `AIDLC_HOOK_DEBUG` on a fresh install would flip the check to a false failure for exactly the person trying to diagnose hooks (`:1880-1887`). The intent is right; the discriminator chosen for it also swallows the real case.

## Expected vs actual

**Expected:** health dir present, zero `.last` files, and a ledger that already contains engine events → fail, with the fix hint that is already written at `:1925` (*"health dir exists but no hooks have fired — verify hooks are registered in settings.json"*).

**Actual:** pass, with a label that predicts a future the ledger has already falsified — `STAGE_STARTED` events are on record, so "first workflow stage will populate" has already had its chance.

## Why this one is worth a guard

This is the only check that answers "is the hook seam alive?", and the seam is what the human gate rests on.

`humanActedSinceGate` fails closed by design (`core/tools/aidlc-lib.ts:3770-3772`): *"a workflow ledger without HUMAN_TURN fails closed"*. `HUMAN_TURN` can only come from the `UserPromptSubmit` hook. So an install whose hooks never fire cannot close **any** approval gate — not slowly, not degraded: never. The workflow parks and the reason is invisible.

We hit precisely that, and the doctor said the install was fine. Measured on our install: 700+ audit events across the shards, **zero** `HUMAN_TURN`, and both `.aidlc-hooks-health/` directories present and empty. Every event on record was written by a tool (`aidlc-orchestrate`, `aidlc-state`, `aidlc-log`, `aidlc-utility`); not one by a hook. The cause on our side was environmental — the harness session was opened in a directory above the one the engine is installed in, so every `$CLAUDE_PROJECT_DIR/...` hook command resolved to a path that does not exist. That part is ours to fix. What we are reporting is that the check written to catch it reported success.

The heartbeat mechanism itself is well designed for this: every hook writes `<hook>.last` as its **first** act, before reading stdin, which makes "did any hook ever run?" directly measurable rather than inferred from event taxonomy. The signal is there; only the branch that reads it is out of reach.

## Reproduce

1. Create `<workspace>/.aidlc-hooks-health/` and leave it empty (this is also the natural state after the engine ensures the directory and no hook has run).
2. Run any tool that appends to the audit ledger, so the ledger is non-empty — e.g. the doctor itself, which writes `HEALTH_CHECKED`.
3. Run `bun <harness>/tools/aidlc-utility.ts doctor`.
4. Section 6 reports `✓ Hook heartbeats: not yet fired (first workflow stage will populate)` and the doctor's exit code is unaffected, even though the ledger shows the engine has been running and no hook has ever written a heartbeat.

Keep the `AIDLC_HOOK_DEBUG` protection but invert the discriminator: instead of requiring `.last` files to be present in order to consider state (b), treat the dir as "fresh" only when it holds **nothing but** the known diagnostic artifacts (`hook-debug.log`, `.first-fired`), and as drift otherwise. That preserves the case the refinement was written for and restores the case the comment promises.

If it is worth being stricter, the ledger is available at that point and makes the distinction sharper: an empty health dir alongside a ledger that already contains non-DocumentKB events is drift regardless of what else the dir holds, because the engine demonstrably ran.

Happy to open a PR against `v2` following the shape of the existing three states — the call on whether this should fail the exit code or warn is yours.

## Environment

macOS 15, Claude Code harness, bun 1.3.14. Verified on the pinned engine we consume (`fbb1460c`, 2.6.18) and re-verified unchanged at `e744a050` (2.6.44) — same block, same line numbers.

apackeer:
Doctor documented a fail state for a health dir with no .last heartbeats but the branch was unreachable, so an install whose hooks never fire (which permanently parks every human gate) passed section 6. The check now fails with exit 1 when the same-scope ledger contains STAGE_STARTED and no hook ever wrote a heartbeat, while fresh installs, pre-created empty dirs, debug-only dirs, and fresh clones still pass; covered by new t37 cases 16b-16e and shipped as 2.6.71. Note: v2's CHANGELOG head has since moved past 2.6.71 (currently 2.6.74), so the version slot re-bumps at landing per the changelog conflict policy. Closes #861

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.90] - 2026-08-25
+
+`/aidlc --doctor` now distinguishes hook directories that have not had a chance to populate from dead hook seams after a workflow stage ran. **Upgrade:** refresh your `dist/<harness>/` shell so doctor can detect this condition.
+
+* When `.aidlc-hooks-health/` exists without any `.last` files and the active audit ledger contains `STAGE_STARTED`, doctor reports `Hook heartbeat data` and exits 1; CI jobs that gate on doctor will now fail for this broken hook state.
+* Missing, pre-created, and debug-only health directories still report the existing not-yet-fired pass row before any stage starts, while real heartbeat files continue to report their last-fired timestamps.
+
 ## [2.6.89] - 2026-08-25
 
 The conductor no longer parks a workflow because context *feels* heavy. The orchestrator skill previously licensed parking when context seemed low even though the conductor cannot measure its own usage, causing unnecessary park/resume interruptions while most of the window remained free. **Upgrade:** refresh your `dist/<harness>/` shell to install the bounded parking guidance.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.89-blue)
+![version](https://img.shields.io/badge/version-2.6.90-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-utility.ts
+++ b/core/tools/aidlc-utility.ts
@@ -2581,24 +2581,27 @@ function handleDoctor(projectDir: string, flags: Record<string, string> = {}): v
     }
   }
 
+  // Read across every per-clone audit shard (single shard in the common case).
+  // Both hook-health and state-drift checks use the same intent-scoped ledger.
+  const auditAllShards = readAllAuditShards(projectDir);
+  const auditShardEvents = readAuditShardEvents(projectDir);
+
   // 6. Hook heartbeats
-  // Three states:
-  //   (a) .aidlc-hooks-health/ missing entirely → fresh install, hooks haven't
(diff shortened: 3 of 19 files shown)


leandrodamascena (approved):
Thanks, Arden. I reviewed the current head `d1270b1cd6f1e974aba319ae8ba921a9114156e0` against `v2` at `f73472aa388590763e3d3f8685647e9a488ff5fd`. The direction is aligned with the accepted diagnosis in #861.

The fix correctly makes the previously unreachable doctor failure observable: an empty or debug-only `.aidlc-hooks-health/` directory now fails after `STAGE_STARTED`, while fresh workflows and valid heartbeats retain their expected behavior.

The branch was rebased, preserving the current pending-gate audit logic, rebumped to `2.6.90`, and all generated harness distributions were refreshed.

Validation completed:

- 53 focused tests passed
- Package parity passed for all harnesses
- Typecheck passed
- Lint passed
- All GitHub CI and security checks passed

The PR is mergeable, and I found no remaining code or protocol blockers.

Thank you for closing this diagnostic gap.

PR #914, opened 2026-08-24
docs: restore CodeKB fingerprint upgrade guidance

### LINKED ISSUE #896: [Bug]: in the lone-repo layout every codekb scope fingerprint is git's empty tree, so codekb-scope-diff cannot tell two scopes apart
## What happens

In the lone-repo layout, every codekb scope fingerprint is git's **empty tree**, so
`codekb-scope-diff` cannot distinguish one scope from another.

`handleCodekbScopeDiff` adds an exclusion when the repo root *is* the workspace root, so
that writing the framework's own `aidlc/` tree cannot stale the repo's fingerprint
(`core/tools/aidlc-utility.ts:5118`):

```ts
const fingerprintExcludes = repoDir === projectDir ? ["aidlc"] : [];
```

`codekbScopeFingerprint` (`core/tools/aidlc-lib.ts:1634`) turns that into a pathspec and
passes it to `git add -A` alongside the **narrow, user-supplied** scope paths:

```ts
const add = spawnSync("git", ["add", "-A", "--", ...paths, ...exclusions], …);
```

On git 2.52, `git add -A -- <narrow-path> :(exclude)<anything>` stages **nothing at all**.
`write-tree` then returns `4b825dc642cb6eb9a060e54bf8d69288fbee4904` — the well-known
empty-tree hash — for *every* scope.

**The exclusion's content is irrelevant.** Excluding a path that does not exist in the repo
(`:(exclude,literal)zzz`) produces the same empty tree. The mere presence of an exclude
pathspec next to a narrower positive pathspec is what zeroes the result.

Both call sites pass user-supplied narrow paths together with the exclusion —
`aidlc-utility.ts:5128` (`--mint`) and `:5208` (status/compare) — so in the lone-repo layout
both are affected.

## Expected vs actual

**Expected:** two different scopes fingerprint differently, and a stored fingerprint
compares equal only when that scope's files are unchanged.

**Actual:** every scope fingerprints identically, so the verdict the verb exists to give is
not trustworthy in this layout:

```console
  src/payments/    4b825dc642cb6eb9a060e54bf8d69288fbee4904
  src/auth/        4b825dc642cb6eb9a060e54bf8d69288fbee4904
```

- a **minted** fingerprint records the empty tree rather than the scope's contents;
- a stored fingerprint therefore compares equal for a scope that was **never analysed**,
  which can report `CURRENT` for unanalysed code — the false-negative direction, and the one
  that silently skips work;
- conversely, a scope minted in the multi-repo layout (where `fingerprintExcludes` is `[]`)
  compares **STALE** forever against a lone-repo checkout.

This degrades silently: `git add` exits 0, `write-tree` exits 0, and the hash matches
`/^[0-9a-f]{40,64}$/`, so every guard in `codekbScopeFingerprint` is satisfied. Nothing
returns `null`, so no caller can tell.

## Reproduce

No AI-DLC needed — plain git:

```console
$ R=$(mktemp -d); mkdir -p "$R/src/payments"; echo a > "$R/src/payments/gw.ts"
$ git init -q "$R"

# A — no exclusion: correct
$ I=$(mktemp -u); GIT_INDEX_FILE=$I git -C "$R" add -A -- src/payments/
$ GIT_INDEX_FILE=$I git -C "$R" write-tree
15aa89e9cb5d8cfe70a690d61f25cb02fde8034b

# B — with the exclusion the tool passes: EMPTY TREE, nothing staged
$ I=$(mktemp -u); GIT_INDEX_FILE=$I git -C "$R" add -A -- src/payments/ ':(exclude,literal)aidlc'
$ GIT_INDEX_FILE=$I git -C "$R" ls-files      # <- prints nothing
$ GIT_INDEX_FILE=$I git -C "$R" write-tree
4b825dc642cb6eb9a060e54bf8d69288fbee4904

# C — excluding a NONEXISTENT path does it too, so the exclusion content is not the issue
$ … add -A -- src/payments/ ':(exclude,literal)zzz'   ->  4b825dc6…

# D — the exclusion works as intended when the positive pathspec is the whole root
$ … add -A -- . ':(exclude,literal)aidlc'             ->  stages src/, omits aidlc/  ✓
```

## It is already caught, but reads as environmental

`tests/unit/t248-codekb-scope-diff.test.ts` fails two assertions on this today:

```
(fail) status mode > matching fingerprint → CURRENT; edit inside scope → STALE
       Expected: "CURRENT"   Received: "STALE"
(fail) mint mode > --mint prints the same fingerprint the lib computes
       Expected: "15aa89e9cb5d8cfe70a690d61f25cb02fde8034b"
       Received: "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
```

The second failure names the empty-tree hash outright. We had been carrying both as a
"known local red" — they are not environmental, they are this bug. (The in-process
`codekbScopeFingerprint` call in the test computes the *correct* hash because the test
invokes it without exclusions; the subprocess under test adds them.) Whether CI is green
depends only on whether the runner's git shows the same pathspec behaviour.

Pick one; the first is the smallest.

1. **Do not pass an exclusion when the positive pathspec is already narrower than it.** An
   exclusion of `aidlc` is only meaningful when the fingerprint covers the repo root; if
   every requested path is outside `aidlc/`, the exclusion is a no-op and can be dropped.
   Measured — this preserves the property the exclusion exists to provide:

   ```console
   # scope src/payments/, no exclusion, with an aidlc/ ledger present
   src/payments/           -> 49ef214c928b0034a9ebd7d77d5dd4fea0e3699e
   (staged: src/payments/p.ts only — aidlc/ does NOT leak in)
   after editing aidlc/    -> 49ef214c928b0034a9ebd7d77d5dd4fea0e3699e   # unchanged ✓
   after editing the scope -> adc34351d760931a4bba61005587c73db157f358   # changed  ✓
   ```

2. **Filter the paths in-process instead of via pathspec magic** — enumerate the scope
   paths, drop any under `aidlc/`, and pass only positive pathspecs.

3. **Refuse the empty tree.** `codekbScopeFingerprint` should treat it as `null` rather than
   a valid fingerprint — a scope that stages zero files is not a fingerprint, it is a
   failure. Worth doing regardless of 1 or 2, as defence in depth. Both object formats need
   covering (measured, not transcribed):

   | format | empty tree |
   |---|---|
   | sha1 | `4b825dc642cb6eb9a060e54bf8d69288fbee4904` |
   | sha256 | `6ef19b41225c5369f1c104d45d8d85efa9b057b53b14b4b9b939dd74decc5321` |

   Equivalently, and without hardcoding either constant: check that `git ls-files` staged a
   nonzero count before accepting the hash.

A regression test should assert two *different* scopes get *different* fingerprints — the
current test compares against a lib call that happens to be exclusion-free, so it cannot
catch a collision. Happy to open a PR against `v2` for whichever option you prefer.

## Environment

macOS 15, Claude Code harness, bun 1.3.14, **git 2.52.0**. Verified against `upstream/v2` at
`cc08c1612` (2.6.65) and re-verified unchanged at `75d8af80` (2.6.68).

_Found while auditing why a long-standing local red had never been explained; investigation
assisted by an AI agent, evidence verified with plain git and against the source._

apackeer:
In the lone-repo layout, the framework-owned aidlc exclusion was passed to git add -A next to narrow scope pathspecs - a combination git handles unreliably (deterministically staging nothing on e.g. macOS git 2.52), so every codekb scope fingerprinted as the empty tree and codekb-scope-diff could report false CURRENT for unanalysed scopes. codekbScopeFingerprint now drops exclusions no positive path covers, drops positives covered by an exclusion, and refuses zero-staged fingerprints (null -> UNVERIFIED, never a false verdict); regression tests pin distinct fingerprints for distinct scopes.

Rebased onto v2 at 350e0c3e8 and re-slotted to 2.6.81 per the changelog conflict policy. The CHANGELOG upgrade instruction telling users to re-mint stored scope timestamps whose fingerprint is the empty-tree hash (`4b825dc642cb6eb9a060e54bf8d69288fbee4904`) is adapted from #919 by @tejasavora, who independently diagnosed the same defect and provided the macOS git-version evidence.

Closes #896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,11 @@ Filesystem descendant checks now enforce one mixed-separator boundary on every h
 
 ## [2.6.91] - 2026-08-25
 
-Copilot no longer asks an AI-DLC workflow question twice when the model attempts to use a native question picker. Running workflows now refuse picker calls and direct the model to render numbered prose in chat, preserving the trusted human-turn check. **Upgrade:** refresh your `dist/copilot/` shell; in-flight workflows need no migration.
+Lone-repository CodeKB scope fingerprints now avoid Git's unreliable combination of positive and exclude pathspecs and reject fingerprints that contain no files, while Copilot no longer asks an AI-DLC workflow question twice when the model attempts to use a native question picker. **Upgrade:** refresh your `dist/<harness>/` shell before minting or checking CodeKB scope fingerprints, and re-mint (or re-run reverse engineering for) any stored scope timestamp whose fingerprint is `4b825dc642cb6eb9a060e54bf8d69288fbee4904`, the empty-tree value written by the pre-fix bug; in-flight Copilot workflows need no migration.
 
+* `codekbScopeFingerprint` omits exclusions outside analyzed roots, drops analyzed paths covered by exclusions, and preserves root and nested-project scope boundaries without silently rewriting malformed absolute paths.
+* Scopes that stage zero paths now mint `unknown` and report `UNVERIFIED`, never a false `CURRENT` or permanent `STALE` verdict.
+* Deterministic regression coverage verifies distinct narrow fingerprints, full-root exclusions, nested-project boundaries, malformed paths, and empty scopes.
 * The Copilot PreToolUse adapter denies native picker execution IDs such as `ask_user` and `vscode/askQuestions` only when the session-selected workflow has valid `Status: Running` state; projects with no workflow, completed workflows, and unusable state fail open.
 * Picker tool results still never count as `HUMAN_TURN`; the user's next chat message records the trusted turn used by answer and approval logging.
 


CHANGELOG.md
```diff
@@ -88,6 +96,7 @@ Non-English workflows now keep the agent's human-facing conversation and structu
 
 * Chat messages, status updates, progress reports, and transitional narration between tool calls now follow the resolved conversation language on every turn, for orchestrators and delegated agents alike.
 * Structured-question `prompt`, `header`, `options[].description`, and free-text follow-ups are localized, while protocol-defined `options[].label` literals such as `Approve` and `X. Other (please specify)` remain unchanged English tokens.
+||||||| parent of f266fec85 (fix: avoid empty codekb scope fingerprints (#896))
```
leandrodamascena: **[P2] Remove the committed conflict marker**

This literal `||||||| parent of ...` marker is part of the PR and makes `git diff --check` fail. Please remove it while rebasing and preserve the surrounding changelog entries.

core/tools/aidlc-version.ts
```diff
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.6.80";
+export const AIDLC_VERSION = "2.6.81";
```
leandrodamascena: **[P2] Rebase and re-slot the release version**

This PR still targets `2.6.81`, while `v2` is already at `2.6.89`, leaving the PR conflicting. Please rebase onto current `v2`, move this change to the next available version slot (currently `2.6.90`), update the changelog, README badge, and title, regenerate `dist`, and rerun the checks.

leandrodamascena (changes requested):
The CodeKB fingerprinting fix itself looks sound, including the narrow-scope exclusion handling, nested-project behavior, and zero-staged-entry guard.

I verified:
- `bun test tests/unit/t248-codekb-scope-diff.test.ts` - 39 passed
- `bun scripts/package.ts --check` - passed

Requesting changes for the committed conflict marker and stale release metadata/rebase.

apackeer:
Addressed the review and rebased this PR onto current `v2`.

What changed:

* #919 merged the same CodeKB fingerprint implementation and a superset of the deterministic regression coverage after the original review, so this PR no longer duplicates that code or those tests.
* The subsequent #882 merge replaced #919's `2.6.91` changelog text instead of preserving both changes. This PR now restores the CodeKB release summary and empty-tree re-mint instruction while retaining the Copilot notes that also shipped in `2.6.91`.
* The committed conflict marker and stale `2.6.81` version, badge, and generated-distribution changes are gone. The current PR changes only `CHANGELOG.md`; no new version bump or `dist` regeneration is needed for this release-note repair.

Local validation:

* `t68-version-changelog-sync`: 7 passed, 0 failed
* `bun scripts/package.ts --check`: all seven harness trees passed
* `git diff --check`: clean

@leandrodamascena, please re-review the current head `a27a423a`.

leandrodamascena (approved):

PR #915, opened 2026-08-24
fix(knowledge): make duplicate-record recovery fail closed (#895)

### LINKED ISSUE #895: [Bug]: knowledge sync index rebuild can keep a stale row over the digest-matching one (realpath asymmetry in rebuildIndex)
## What happens

`rebuildIndex` resolves the `documents/` anchor with `realpathSync` but compares it against
a path that is **not** resolved. When the project path traverses a symlink, the containment
check correctly refuses, a bare `catch` swallows the refusal, and the duplicate-record
tie-break silently degrades from *"prefer the row whose digest matches the file on disk"* to
*"keep the newest `indexed_at`"*.

The relevant lines (`core/tools/aidlc-knowledge.ts:3849-3856`, inside `rebuildIndex` which
begins at `:3788`):

```ts
const docsReal = existsSync(docsAbs) ? realpathSync(docsAbs) : null;   // /private/var/folders/...
// ...
const abs = join(knowledgeDir(projectDir, space), sourcePath.split("/").join(sep));  // /var/folders/...
currentDigest = sha256Hex(readCandidate(docsReal, abs));
} catch { /* absent/refused source gives no preferred digest */ }
```

`readCandidate` does `relative(documentsReal, absPath)`. With one side resolved and the
other not, that yields `../../../../../var/folders/...`; `resolveContainedPath` refuses it
(correctly — it looks like traversal), and `currentDigest` stays `null`. The first sort key
at `:3860-3862` can then never fire.

## Expected vs actual

**Expected:** index recovery keeps the row whose digest matches the file currently on disk.

**Actual:** on a symlinked project path it keeps whichever row has the newest `indexed_at`,
which can be a stale row — and the survivor is a **spliced record that validates**: the
stale row's `id` and `indexed_at` alongside the live file's `sha256` and
`extraction: "extracted"`, while the corresponding `content.md` never existed.

```console
$ cat .../knowledge/documents/policy.md
current policy

$ # index.json after the documented `sync` recovery
id=019fda80-8f8d-7bfa-b56c-983cf0353fab      # the planted stale row
  sha256=9c7b5c77676ceadab4509d8328b151c60b82b037825c8b62855fa3551d13ca6a   # the LIVE file's digest
  indexed_at=2099-01-01T00:00:00Z
  extraction=extracted                        # but no content.md was ever written
```

So `show <id>` on a recovered catalog can cite one identity while the digest belongs to
another. That is the failure mode the rebuild was specified to prevent: a
misclassification that looks successful is worse than a rebuild that refuses.

## Reproduce

Two runs of the same test, one variable — whether `TMPDIR` traverses a symlink:

```console
$ bun test tests/unit/t298-knowledge-transaction.test.ts
 53 pass
 1 fail          # rebuild keeps the duplicate live row whose digest matches the source

$ TMPDIR=/private/tmp bun test tests/unit/t298-knowledge-transaction.test.ts
 54 pass
 0 fail
```

And through the shipped CLI, as a controlled pair — the only variable is whether the project
path traverses a symlink. Onboard one document, plant a second `metadata.json` for the same
source path whose digest does **not** match the file but whose `indexed_at` is newer, delete
`index.json`, then run the documented recovery (`sync`):

```console
$ R=$(mktemp -d /tmp/repro-XXXX)              # /tmp is a symlink to /private/tmp on macOS
$ D="$R/aidlc/spaces/default/knowledge/documents"; mkdir -p "$D"
$ printf 'current policy\n' > "$D/policy.md"
$ bun <harness>/tools/aidlc-knowledge.ts onboard "$D/policy.md" --json --project-dir "$R"
#   -> id 01a032e0-f4de-71aa-92b2-25b376c13c51   (the row whose digest matches disk)

$ # plant a stale duplicate for the SAME source path, with a newer indexed_at:
$ #   id=019fda80-8f8d-7bfa-b56c-983cf0353fab
$ #   sha256=sha256("stale policy\n"), indexed_at=2099-01-01T00:00:00Z
$ #   extraction={state:"unsupported_type"}, no content/content_sha256
$ rm -f "$R/aidlc/spaces/default/knowledge/documentkb/index.json"
$ bun <harness>/tools/aidlc-knowledge.ts sync --json --project-dir "$R"

$ # survivors in the rebuilt index:
    id=019fda80-8f8d-7bfa-b56c-983cf0353fab          # <- the PLANTED STALE row won
      sha256=9c7b5c77676ceadab4509d8328b151c60b82b037825c8b62855fa3551d13ca6a   # the LIVE file's digest
      indexed_at=2099-01-01T00:00:00Z
      extraction=extracted                            # but its content.md never existed
```

The same script with `R=$(mktemp -d /private/tmp/repro-XXXX)` — an already-resolved path,
nothing else changed:

```console
  CONTROL — project path: /private/tmp/repro-80gT  (no symlink)
    survivor id=01a032e1-38d0-7f03-810c-0a7f7bd658e2  indexed_at=2026-08-24T08:26:56.078Z
    # the digest-matching row, as specified
```

| Project path | Survivor |
|---|---|
| `/tmp/...` (symlink to `/private/tmp/...`) | the **stale** row ✗ |
| `/private/tmp/...` (already resolved) | the digest-matching row ✓ |

Resolving the candidate the same way the anchor is already resolved turns the failing test
green with nothing else changed:

```diff
-        const abs = join(knowledgeDir(projectDir, space), sourcePath.split("/").join(sep));
+        const abs0 = join(knowledgeDir(projectDir, space), sourcePath.split("/").join(sep));
+        const abs = existsSync(abs0) ? realpathSync(abs0) : abs0;
         currentDigest = sha256Hex(readCandidate(docsReal, abs));
```

```console
$ bun test tests/unit/t298-knowledge-transaction.test.ts   # with the two lines above
 54 pass
 0 fail
```

(Applied to the shipped `dist/` copy the test actually imports, then reverted — diagnosis,
not a proposed patch; see below for why the real change is larger.)

## Why this is easy to miss

Scope is narrow: it needs a duplicate live record for one source path (crash residue or a
hand repair — exactly what the tie-break exists for) **plus** a project path that traverses
a symlink. macOS `mkdtemp`/`tmpdir()` and symlinked checkouts hit the second condition
routinely; a Linux CI runner does not, which is why the red reads as environmental. It is
not: `tests/unit/t298-knowledge-transaction.test.ts` has been catching this the whole time.

Two independent parts:

1. **Resolve both sides before comparing** — realpath the candidate path (or derive it from
   `docsReal` rather than from `knowledgeDir`) so `relative()` sees two paths in the same
   namespace. Worth grepping for other `realpathSync` results later compared against
   unresolved paths.
2. **Do not let a `catch` silently weaken selection.** A containment refusal here changes
   the ordering policy; it should be distinguishable from "the source is genuinely absent"
   rather than collapsing into the same `null`.

A regression test that exercises a symlinked project path would pin part 1 on Linux CI too.
Happy to open a PR against `v2` for whichever shape you prefer.

## Environment

macOS 15, Claude Code harness, bun 1.3.14, git 2.52.0. Verified against `upstream/v2` at
`cc08c1612` (2.6.65) and re-verified unchanged at `75d8af80` (2.6.68).

_Found while building DocumentKB summaries on top of #731; investigation assisted by an AI
agent, evidence verified against the source and by driving the shipped CLI._

apackeer:
rebuildIndex resolved the documents/ anchor with realpathSync but built the digest candidate from the unresolved project path, so on a symlinked project path the containment check refused the read and the duplicate-record tie-break silently degraded to newest indexed_at — letting a stale row survive index recovery as a spliced record. The candidate is now derived from the resolved documents/ root, an absent source keeps the documented indexed_at fallback, and a present-but-refused source now stops the rebuild with the source path and reason; three new t298 tests pin the symlinked-path regression on Linux CI plus the refused and absent branches.

Note on the version slot: this branch carries 2.6.77 (committed when v2 was at 2.6.69); v2 has since advanced to 2.6.74. Per the changelog conflict policy the slot re-bumps at landing if taken by then.

Closes #895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.100] - 2026-08-26
+
+Knowledge index recovery now selects duplicate records against the live managed source even when the project path traverses a symlink, and refuses present sources that cannot be safely read. **Upgrade:** refresh your `dist/<harness>/` shell to receive the corrected recovery behavior.
+
+* `aidlc-knowledge sync` rebuilds a missing `index.json` using digest matching through symlinked project paths, preventing a newer stale duplicate record from surviving recovery.
+* When duplicate records require a digest tie-break, a present-but-refused managed source, including a dangling symlink, now stops the rebuild with the source path and refusal reason; only a genuinely absent source retains the deterministic newest-`indexed_at` fallback.
+
 ## [2.6.99] - 2026-08-26
 
 Audit and usage-ledger read-modify-write transactions now remain serialized during Windows lock handoff instead of allowing retirement or stale recovery to displace a successor. **Upgrade:** re-copy `dist/<harness>/` so all state, audit, active-directive, and usage writers receive the generation-bound lock protocol.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.99-blue)
+![version](https://img.shields.io/badge/version-2.6.100-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-knowledge.ts
+++ b/core/tools/aidlc-knowledge.ts
@@ -3876,11 +3876,25 @@ export function rebuildIndex(projectDir: string, space: string): DocumentIndex {
   for (const [sourcePath, rows] of liveByPath) {
     if (rows.length < 2) continue;
     let currentDigest: string | null = null;
-    if (docsReal !== null && rows.some((row) => row.source.kind === "managed")) {
+    if (
+      docsReal !== null &&
+      rows.some((row) => row.source.kind === "managed") &&
+      sourcePath.startsWith("documents/")
+    ) {
+      const abs = join(
+        docsReal,
+        sourcePath.slice("documents/".length).split("/").join(sep),
+      );
       try {
-        const abs = join(knowledgeDir(projectDir, space), sourcePath.split("/").join(sep));
         currentDigest = sha256Hex(readCandidate(docsReal, abs));
-      } catch { /* absent/refused source gives no preferred digest */ }
+      } catch (e) {
+        if ((e as NodeJS.ErrnoException).code !== "ENOENT") {
+          throw new Error(
+            `cannot choose among duplicate records for ${sourcePath} while rebuilding the index: ` +
+              errorMessage(e),
+          );
+        }
+      }
     }
     rows.sort((a, b) => {
       const aMatches = currentDigest !== null && a.sha256 === currentDigest;
--- a/core/tools/aidlc-version.ts
+++ b/core/tools/aidlc-version.ts
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.6.99";
+export const AIDLC_VERSION = "2.6.100";
--- a/dist/claude/.claude/tools/aidlc-knowledge.ts
+++ b/dist/claude/.claude/tools/aidlc-knowledge.ts
@@ -3876,11 +3876,25 @@ export function rebuildIndex(projectDir: string, space: string): DocumentIndex {
   for (const [sourcePath, rows] of liveByPath) {
     if (rows.length < 2) continue;
     let currentDigest: string | null = null;
-    if (docsReal !== null && rows.some((row) => row.source.kind === "managed")) {
+    if (
+      docsReal !== null &&
+      rows.some((row) => row.source.kind === "managed") &&
+      sourcePath.startsWith("documents/")
+    ) {
+      const abs = join(
+        docsReal,
+        sourcePath.slice("documents/".length).split("/").join(sep),
+      );
       try {
-        const abs = join(knowledgeDir(projectDir, space), sourcePath.split("/").join(sep));
         currentDigest = sha256Hex(readCandidate(docsReal, abs));
-      } catch { /* absent/refused source gives no preferred digest */ }
+      } catch (e) {
+        if ((e as NodeJS.ErrnoException).code !== "ENOENT") {
+          throw new Error(
+            `cannot choose among duplicate records for ${sourcePath} while rebuilding the index: ` +
+              errorMessage(e),
+          );
+        }
+      }
     }
     rows.sort((a, b) => {
       const aMatches = currentDigest !== null && a.sha256 === currentDigest;
--- a/dist/claude/.claude/tools/aidlc-version.ts
+++ b/dist/claude/.claude/tools/aidlc-version.ts
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.6.99";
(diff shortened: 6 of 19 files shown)


leandrodamascena (changes requested):
Thanks for addressing the symlinked-project recovery bug reported in #895. The issue provides a concrete reproduction, and the maintainer-owned implementation is aligned with the documented fail-closed recovery contract.

I reviewed the current head `b1e4b26a38ac26b9dccc838763224ad790f4e632` against `b44b9c9c1ad1afeb351f1543b384936cd3d16839`.

**P2 - A broken source symlink is still treated as absent**

In `core/tools/aidlc-knowledge.ts:3608`, `existsSync(abs)` returns `false` for a dangling symlink. Therefore, when `documents/policy.md` is a broken symlink and duplicate metadata records exist, `readCandidate` is never called and the rebuild silently uses the newest-`indexed_at` fallback.

I reproduced this with two valid duplicate metadata records and a dangling `documents/policy.md` symlink. The rebuild succeeded and selected the stale record dated 2099 instead of refusing:

- selected id: `00000000-0000-7000-8000-000000000002`
- selected digest: the planted stale digest

This violates the PR's stated contract that a present-but-refused source stops recovery, and preserves the same spliced-record failure the change is intended to eliminate. A symlink entry is present and must be refused even when its target is missing.

Please use a no-follow existence check such as `lstatSync`, treat only `ENOENT` as genuinely absent, and pass every present entry through `readCandidate`. Add a regression case alongside `tests/unit/t298-knowledge-transaction.test.ts:1589` for a dangling source symlink.

I validated that this correction shape makes the reproduction fail closed with the existing symlink refusal and keeps all four rebuild-focused `t298` cases passing.

Validation performed on the unchanged PR head:

- `t298-knowledge-transaction`: 58/58 passed.
- `t284` and `t289`: 123/123 passed.
- Typecheck passed.
- Version/changelog synchronization passed, 7/7.
- Lint completed with only three non-blocking infos in unrelated `t267`.
- The targeted dangling-symlink regression fails on the PR head and passes with the diagnostic local correction.

No commits or pushes were made. The PR currently has no GitHub CI checks and is conflicting with `v2`; after fixing the blocker it also needs a rebase and version re-bump from `2.6.77` beyond the current `2.6.88`.

**Proposed review event: `REQUEST_CHANGES`**

apackeer:
Addressed the requested dangling-symlink recovery case and rebased onto the current `v2`.

Changes on head `e4336370e`:

* Removed the followed `existsSync` precheck from the duplicate-record digest tie-break. The rebuild now runs the full guarded `readCandidate` path and treats only `ENOENT` as a genuinely absent source.
* Dangling source symlinks and other present-but-refused paths now stop recovery instead of allowing the newest stale duplicate metadata row to win.
* Added a regression covering a dangling `documents/policy.md` symlink while preserving the genuinely absent-source fallback.
* Re-slotted the release metadata to `2.6.95` and regenerated all seven harness distributions.

Local validation:

* `t68`, `t284`, `t289`, and `t298`: 189/189 assertions passed.
* Packaging parity and typecheck passed.
* Lint completed with only the three existing non-blocking infos in unrelated `t267`.

The PR is now mergeable and CI is running. @leandrodamascena, please re-review the current head when ready.

apackeer:
Current-head update for re-review:

* Renamed the PR to describe the user-visible contract: duplicate-record recovery now fails closed.
* Rebuild resolves managed sources from the real `documents/` root and runs the full guarded `readCandidate` path; only `ENOENT` is treated as genuinely absent.
* Present-but-refused sources, including dangling symlinks, stop recovery instead of allowing a newer stale duplicate metadata row to win.
* Regression coverage pins the dangling-symlink refusal, symlinked-project digest match, unreadable-source refusal, and genuinely absent-source fallback.
* The branch is rebased onto `v2`, release metadata is slotted at `2.6.95`, and all seven harness distributions are regenerated.

Current head: `e4336370e`. All GitHub CI, security, documentation, and changelog checks are green.

@leandrodamascena, the formal review request remains active; please re-review this head when ready.

leandrodamascena (approved):
Thanks, Arden, for addressing the previous review and strengthening the recovery contract.

I reviewed the current head `e4336370eea72c293dc114f446984fe5f343b4af` against its submitted base `cb822bae43ef65c99d1d4deec200c8e804e2898d`, and checked its interaction with the current `v2` head `cb30d6422e716247f2bb9d91a928ea86078dec9d`. This remains aligned with #895 and the previous maintainer review.

The previous blocker is resolved. Duplicate-record recovery now sends managed sources through the guarded read boundary, treats only `ENOENT` as genuinely absent, and fails closed for dangling symlinks and other present-but-refused sources. The symlinked-project digest selection and deterministic absent-source fallback are both preserved.

Validation performed:

- `t298`: 59/59 passed
- `t284`: 31/31 passed
- `t289` boundary tests passed
- `t68` version synchronization passed
- packaging parity, typecheck, lint, coverage registry, and `git diff --check` passed
- GitHub CI is fully green for this exact head

`v2` advanced after this validation, so the branch now requires a rebase. The overlap is limited to release metadata: please re-slot from `2.6.95` to `2.6.100`, retain the new `2.6.96` through `2.6.99` changelog entries, regenerate all distributions, and let CI rerun.

I found no remaining correctness, compatibility, or protocol blocker in the implementation itself. Approving, with the rebase and release re-slot required before merge.

PR #916, opened 2026-08-24
fix: keep type-check sensor cache in the project record, not per-package tsconfig dirs

### LINKED ISSUE #900: [Bug]: type-check sensor writes `.tsbuildinfo` scratch into every package (monorepo with per-package `tsconfig.json`), escaping the root `.gitignore`
### Description

- **Framework version:** AI-DLC 2.6.65
- **Component:** `type-check` sensor (`.claude/tools/aidlc-sensor-type-check.ts`, `sensorsDir()` in `.claude/tools/aidlc-lib.ts`)
- **Severity:** Low (inert `tsc` cache; type-checking itself is correct) — but it clutters source packages and can leak into commits
- **Environment:** macOS (Darwin 24.6.0), Node v25.2.1, bun, npm workspaces monorepo (`packages/{shared,api,web,infra}`, each with its own `tsconfig.json`)

In a monorepo where each workspace package has its own `tsconfig.json`, the
`type-check` sensor drops a stray AI-DLC record tree inside **every** package:

```
packages/api/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/web/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/shared/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/infra/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
```

Two problems follow:

1. **Scattered scratch.** The sensor's `.tsbuildinfo` (tsc incremental build
   cache) is written under each package instead of once under the active
   intent's real record dir (`aidlc/spaces/<space>/intents/<intent>/.aidlc-sensors/`).
2. **Escapes the shipped `.gitignore`.** The sensor code comments assert the
   scratch "never pollutes commits" because `.aidlc-sensors/` is gitignored,
   but the shipped project `.gitignore` anchors `aidlc/spaces/*/intents/.aidlc-*`
   at the **repo root**. A path like `packages/api/aidlc/spaces/default/intents/.aidlc-sensors/`
   is therefore **not** ignored, so without a custom rule these files get
   committed into the user's source tree.

### Root cause

`aidlc-sensor-type-check.ts` anchors its scratch directory on the **nearest
`tsconfig.json` directory** rather than the active workspace record dir:

```ts
// aidlc-sensor-type-check.ts
const tsconfigPath = findTsconfig(args.filePath); // walks UP to nearest tsconfig.json
const tsconfigDir = dirname(tsconfigPath);        // = packages/<pkg> in a monorepo
const sensorsBaseDir = sensorsDir(tsconfigDir);   // -> <tsconfigDir>/aidlc/.../.aidlc-sensors
const tsBuildInfoFile = join(sensorsBaseDir, ".tsbuildinfo");
```

```ts
// aidlc-lib.ts
export function sensorsDir(baseDir: string, intent?, space?): string {
  if (intent === undefined && space === undefined) {
    return join(docsRoot(baseDir), ".aidlc-sensors"); // docsRoot(baseDir) re-roots at baseDir
  }
  return join(docsRoot(baseDir, intent, space), ".aidlc-sensors");
}
```

Because `findTsconfig()` resolves to each package's own `tsconfig.json`, the
"project anchor" becomes the package directory, and `sensorsDir()`/`docsRoot()`
build the `aidlc/spaces/default/intents/.aidlc-sensors` path **relative to that
package**. The behaviour assumes a single `tsconfig.json` at the repo root; a
per-package-tsconfig monorepo breaks that assumption.

Note also that the sensor's own manifest (`.claude/sensors/aidlc-type-check.md`)
documents the failure-detail location as
`aidlc/spaces/<active-space>/intents/<active-intent>/.aidlc-sensors/<stage-slug>/...`
— i.e. the true record dir — so the tool diverges from its documented output
location.

### Steps to Reproduce

1. Start an AI-DLC workflow that generates a TypeScript **monorepo** where each
   package carries its own `tsconfig.json` (e.g. `express` scope scaffolding a
   project with `packages/{shared,api,web,infra}`, each with a `tsconfig.json`
   extending a root `tsconfig.base.json`).
2. Let Code Generation / Build and Test write `.ts`/`.tsx` files into the
   packages. The `type-check` sensor fires on each write (`matches: "**/*.{ts,tsx}"`).
3. Inspect the package directories:
   ```bash
   find packages -type d -name aidlc
   find packages -iname "*.tsbuildinfo"
   ```

### Expected Behavior

- The sensor writes its `.tsbuildinfo` (and any failure detail) **once**, under
  the active intent's record dir:
  `aidlc/spaces/<active-space>/intents/<active-intent>/.aidlc-sensors/`
  — the location its own manifest documents.
- No AI-DLC record/scratch directories are created inside `packages/*` (or any
  application source directory).
- The scratch is reliably covered by the shipped `.gitignore` regardless of
  repository layout, so it can never be committed.

### Actual Behavior

- A separate `aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo` tree is
  created inside **each** package that has its own `tsconfig.json`.
- These paths are **not** matched by the shipped root `.gitignore`
  (`aidlc/spaces/*/intents/.aidlc-*` is root-anchored), so they are tracked by
  git unless the user adds a custom ignore rule.

### AI-DLC Version

v2 (alpha)

### Release / Commit

73164cad

### AI-DLC Phase

Not phase-specific

### Platform / IDE

Claude Code

### AI Model

Opus 4.8

### Environment

- macOS 15.7.9
- VSCode 1.134.0

### Additional Context

Observed tree after a scaffold run:

```
$ find packages -iname "*aidlc*"
packages/infra/aidlc
packages/web/aidlc
packages/shared/aidlc
packages/api/aidlc
packages/infra/aidlc/spaces/default/intents/.aidlc-sensors
packages/api/aidlc/spaces/default/intents/.aidlc-sensors
packages/shared/aidlc/spaces/default/intents/.aidlc-sensors
packages/web/aidlc/spaces/default/intents/.aidlc-sensors

$ find packages -iname "*.tsbuildinfo"
packages/web/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/shared/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/api/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
packages/infra/aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo
```

Each directory contains only a single `.tsbuildinfo` file (no source, config,
or user data).

### Suggested fix

- Resolve the sensor scratch dir from the **active intent record** (the engine
  already knows `aidlc/spaces/<space>/intents/<intent>/`) rather than from
  `findTsconfig()`'s directory. Keep using the discovered `tsconfig.json` as the
  `--project` anchor for `tsc` (that part is correct and needed for
  multi-package type resolution), but point `--tsBuildInfoFile` at the record
  dir — optionally namespaced per package (e.g.
  `.aidlc-sensors/type-check/<package-relative-path>.tsbuildinfo`) to avoid
  incremental-cache collisions across packages.
- Alternatively, place the `.tsbuildinfo` in the OS temp dir when the record
  dir cannot be resolved, matching the existing read-only-fs fallback.
- Harden the shipped `.gitignore` so `.aidlc-sensors/` scratch is ignored at any
  depth (e.g. `**/.aidlc-sensors/`), not just at the repo root, as defence in
  depth.

### Workaround (user side)

```bash
rm -rf packages/*/aidlc
# add to .gitignore:
#   **/.aidlc-sensors/
#   packages/*/aidlc/
#   *.tsbuildinfo
```

apackeer:
The type-check sensor anchored its incremental tsc cache on the nearest tsconfig.json directory, planting stray aidlc/spaces/default/intents/.aidlc-sensors/.tsbuildinfo trees inside every package of a per-package-tsconfig monorepo — paths the shipped root-anchored .gitignore does not cover, so they leak into commits. The cache now resolves through the project record (sensorsDir(resolveProjectDir()), matching the dispatcher and the sensor manifest) with a per-tsconfig .tsbuildinfo-<hash8> name to avoid cross-package cache thrash, and every harness gitignore gains a depth-tolerant **/aidlc/spaces/*/intents/**/.aidlc-sensors/ rule covering legacy strays. Adds a real-dispatcher monorepo regression test (t92 Group P) plus functional check-ignore coverage across all seven harnesses. Closes #900

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.96] - 2026-08-26
+
+Type-check sensor incremental state now stays in the project record and uses a distinct cache per tsconfig, preventing package-local `aidlc/` trees in monorepos. **Upgrade:** refresh your `dist/<harness>/` shell and drop any user-side workaround ignore rules; legacy package-local cache directories may be removed individually after confirming they contain only sensor cache data.
+
+* Type-check keeps `--project` and the tsconfig-directory working directory while writing `.tsbuildinfo-<sha256>` files under the project record's `.aidlc-sensors/` directory, keyed by the project-relative tsconfig path.
+* Every harness now ignores the full engine-shaped sensor-cache path at any depth, covering legacy package-local trees without ignoring unrelated `.aidlc-sensors` directories.
+
 ## [2.6.95] - 2026-08-25
 
 `/aidlc --doctor` now reports consumed artifacts with multiple loaded producers before runtime silently selects the first producer by graph load order. **Upgrade:** refresh your `dist/<harness>/` shell to install the new advisory check; no workflow migration is required.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.95-blue)
+![version](https://img.shields.io/badge/version-2.6.96-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/templates/onboarding.md
+++ b/core/templates/onboarding.md
@@ -62,6 +62,7 @@ Commit the `aidlc/` workspace tree — the record (state, the per-clone audit sh
 - `aidlc/active-space` and `aidlc/spaces/*/intents/active-intent` (per-user cursors)
 - `aidlc/.aidlc-clone-id` (per-clone audit-shard token) and `aidlc/.aidlc-sessions/`
 - `aidlc/spaces/*/intents/.aidlc-*` (pre-intent hooks-health scratch)
+- `**/aidlc/spaces/*/intents/**/.aidlc-sensors/` (engine-shaped sensor caches at any depth, including legacy package-local trees)
 - `aidlc/spaces/*/intents/*/runtime-graph.json` (also covers per-Bolt worktree fragments by relative-path glob)
 - `aidlc/spaces/*/intents/*/.aidlc-*` (recovery, hooks-health, sensors scratch)
 {{SLOT:gitignore_extra}}
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -9504,9 +9504,10 @@ export const COMPOSE_MARKER_TTL_MS = 24 * 60 * 60 * 1000;
 // `<baseDir>/.aidlc-sensors` — the sensor detail-output / tsbuildinfo directory.
 // `baseDir` is the project dir for current dispatcher and type-check callers;
 // callers append a stage slug as needed. Before 2.6.94, type-check passed a
-// tsconfig directory instead, creating legacy package-local record trees. A
-// bare baseDir still keeps the flat `.aidlc-sensors` leaf when no intent/space
-// context is supplied.
+// tsconfig directory instead, creating legacy package-local record trees. With
+// no explicit intent/space, `docsRoot` follows the active-intent cursor (or lone
+// record) when one resolves, so caches and failure details share the manifest's
+// per-intent location; only pre-intent does it fall back to the flat space root.
 export function sensorsDir(baseDir: string, intent?: string, space?: string): string {
   if (intent === undefined && space === undefined) {
     return join(docsRoot(baseDir), ".aidlc-sensors");
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -7421,11 +7421,12 @@ export function repoDir(projectDir: string, repoName: string): string {
 // pathspec a directory match, so a root FILE named `aidlc` stays in the walk.
 const AIDLC_SHELL_DIR_NAMES = ["aidlc/", ".aidlc/"];
 
-// The sensor cache is the one member of the family that is NOT root-anchored:
-// the type-check sensor anchors `.aidlc-sensors/.tsbuildinfo` at the tsconfig
-// dir (core/tools/aidlc-sensor-type-check.ts's `sensorsDir`), which a
-// monorepo's per-package tsconfig can put anywhere under repoDir. So this one
-// needs a depth-tolerant pathspec while the shell names stay root-anchored.
+// The sensor-cache exclusion remains depth-tolerant for legacy and worktree
+// paths. Before 2.6.94, the type-check sensor anchored
+// `.aidlc-sensors/.tsbuildinfo` at the nearest tsconfig directory, so monorepo
+// package caches could appear anywhere under repoDir. Those stray trees persist
+// in upgraded repositories, and Bolt worktree record mirrors can also sit below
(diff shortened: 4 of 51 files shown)


leandrodamascena (changes requested):
The cache relocation itself is aligned with #900 and works as intended. The focused integration test confirms that package-local AIDLC trees are no longer created and that each tsconfig receives a distinct cache under the project record.

[P2] Narrow the destructive upgrade cleanup
CHANGELOG.md:6

`rm -rf packages/*/aidlc` removes the entire directory, not only the legacy `.aidlc-sensors` cache. This repository explicitly treats nested `aidlc/` directories as potentially legitimate application source, so a package containing both the stale cache and real `aidlc` content would lose that content. Please target only the exact engine-shaped `.aidlc-sensors` directories and prune parent directories only when they are empty, or omit the cleanup command.

[P2] Rebase and re-slot the release version
CHANGELOG.md:4

The PR still declares 2.6.80, while the current `v2` base (`4c666fbc`) is already at 2.6.87 and 2.6.80 has been used. Please rebase and use the next available version, currently 2.6.88, updating the changelog, README badge, authored version file, and regenerated distributions together. The PR currently has merge conflicts in those files.

[P2] Keep the current bounded t294 timeout
tests/unit/t294-document-extractors-seam.test.ts:161

This unrelated change raises one package-check test timeout to 120 seconds. The current base uses 15 seconds, and the same test completed in 7.5 seconds during this review. Please retain the base timeout when resolving the conflict so a packaging hang does not delay failure detection by two minutes.

Validation performed at head `0ba08eed53bb04313e3753ea72c006722b6b3092`:

- 140 focused tests passed across t68, t92, t157, t294, and t304
- `bun scripts/package.ts --check` passed for all seven harnesses
- `bun run typecheck` passed
- `bun run lint` passed; only three informational diagnostics in unrelated t267
- coverage registry and `git diff --check` passed
- GitHub reports no CI checks for this head
- Current target base: `4c666fbc8badf7999503a2ec5620eaadef18e016`

apackeer:
Addressed the review feedback and rebased onto the current `v2` head.

Changes:

- Re-slotted the release metadata to `2.6.94` and regenerated every harness distribution.
- Removed the blanket `rm -rf packages/*/aidlc` upgrade command; the changelog now gives non-destructive guidance for individually verified legacy cache directories.
- Restored the bounded 15-second timeout in `t294`.
- Hardened the per-tsconfig cache key from an eight-character prefix to the full SHA-256 digest after confirming a real prefix collision, with regression coverage using the colliding paths.
- Preserved the active-intent sensor-cache documentation correction through the rebase.

Validation:

- Focused `t68`, `t92`, `t157`, `t294`, and `t314` slice: 5 files, 141 assertions, 0 failures.
- `bun run typecheck`
- `bun run lint` (only the three existing informational diagnostics in `t267`)
- `git diff --check`

@leandrodamascena, please re-review when convenient.

leandrodamascena (approved):
Thanks for addressing all of the previous findings.

I reviewed the current head `f360a79dbc0f621af4b44fa699cd634e720ceb96` against `cb822bae43ef65c99d1d4deec200c8e804e2898d`. The implementation remains aligned with issue #900, and all prior blockers have been resolved:

- Removed the destructive `rm -rf packages/*/aidlc` guidance.
- Restored the package-check timeout to 15 seconds.
- Replaced the truncated cache key with the full SHA-256 digest.
- Added regression coverage for real prefix collisions.

Local validation passed: 141 focused tests, packaging parity for all seven harnesses, typecheck, lint, coverage registry, and `git diff --check`. All GitHub checks for this exact head are also green.

The current base/version conflict is a normal integration race and is tracked separately as a non-blocking follow-up. I found no remaining code-quality blockers.

leandrodamascena:
`v2` advanced to `cb822bae43ef65c99d1d4deec200c8e804e2898d` while validation was running and now owns `2.6.94`. Please rebase onto the current `v2`, re-slot this change to `2.6.95`, update the CHANGELOG, README badge, authored version, and generated distributions, then rerun CI.

This is a follow-up integration task, not a blocker on the reviewed implementation.

PR #917, opened 2026-08-24
docs: describe the org-conflict admission check as orchestrator judgement, not a framework-enforced stop

### LINKED ISSUE #897: [Bug]: the guide promises a framework-enforced org-contradiction stop, but persist never reads org.md
## What happens

The user guide describes the org-contradiction check as something *the framework* performs
and that *stops* a write (`docs/guide/09-rules-and-the-learning-loop.md:90`):

> Before a kept learning lands on disk, **the framework runs a section-level check** against
> `memory/org.md`. If your proposed entry contradicts an org rule under the same heading,
> **the gate stops** and quotes the conflicting org sentence inline.

No tool performs that check. `aidlc-learnings.ts persist` never reads `org.md` — nothing
under `core/tools/` reads it for a conflict comparison at all — and the source says so
plainly (`core/tools/aidlc-learnings.ts:97-101`):

```ts
// The conflict COMPARISON is the orchestrator-LLM's job (the "single-line
// variant" of the §5 gate model); persist receives only conflict-clear or
// user-escalated selections and never judges.
```

That split is deliberate and the reference docs state it correctly
(`docs/reference/07-sensor-system.md:307-308`). The guide sentence is the mismatch: it reads
as a deterministic enforcement boundary when the comparison is a judgement the orchestrator
is *asked* to make, and `persist` accepts whatever selection it is handed.

## Expected vs actual

**Expected**, from the guide: a contradicting entry cannot land, because a framework check
refuses it.

**Actual:** whether it lands depends on the orchestrator having performed the comparison and
reported honestly. `persist` has no independent view of `org.md` to refuse against.

## Reproduce

Write an `org.md` rule, then hand `persist` a selection that directly contradicts it under
the same heading. No orchestrator, no conflict check — just the deterministic writer:

```console
$ R=$(mktemp -d); M="$R/aidlc/spaces/default/memory"; mkdir -p "$M"
$ cat > "$M/org.md" <<'EOF'
# Organization Rules

## Testing

- All database tests MUST run against a real database. Mocks are forbidden.
EOF
$ cat > "$R/sel.json" <<'EOF'
{
  "stage_slug": "code-generation",
  "space": "default",
  "intent": null,
  "selections": [
    {
      "candidate_id": "contradiction-probe-1",
      "type": "learning",
      "scope": "team",
      "heading": "Testing",
      "text": "Always mock the database in tests; never use a real one."
    }
  ]
}
EOF
$ bun <harness>/tools/aidlc-learnings.ts persist \
    --slug code-generation --selections-json "$R/sel.json" --project-dir "$R"
{"stage_slug":"code-generation","rule_learned":1,"sensor_proposed":0,"notes":[]}
$ echo $?
0

$ cat "$M/team.md"
# Team-Level Rules

## Testing
- Always mock the database in tests; never use a real one. (learned 2026-08-24) <!-- cid:... -->
```

`org.md` forbids mocks; the entry that lands mandates them, under the same heading. Exit 0,
`rule_learned: 1`, no note, nothing quoted back. This is the documented "gate stops" case.

## Why I am reporting this instead of sending a patch

I tried to close the gap by making the escalation genuinely enforceable — so the tool itself
could distinguish "reviewed against the org section as it actually stood" from "claims it
was." Every anchor available for that decision has to be read out of the **committed audit
shard**, which is an ordinary version-controlled file in the user's own repository, writable
by exactly the population such a gate would constrain.

It took eight rounds, and in rounds 5–8 **the fix itself opened the next hole**:

| Round | The fix | How that fix opened the next hole |
|---|---|---|
| 5 | MAC-authenticate the anchor row (HMAC + `timingSafeEqual`) | the MAC verified fields through a helper returning a field's **first** occurrence, so a row with a duplicated field authenticated under one value and was read under another |
| 6 | reject rows whose MAC-covered fields appear more than once | made **genuine** rows unverifiable: tainting a real row dropped its digest out of the deny set, erasing history a second check depended on |
| 7 | stop verifying during the history scan, so every well-formed row counts | made the deny set **injectable**: a forged row naming a not-yet-existing digest pre-emptively refuses a legitimate future escalation |
| 8 | — | I stopped. |

Round 7 also runs into a real asymmetry: **authorization sets may shrink under doubt; deny
sets may not.** That makes an unauthenticated deny set a denial-of-service surface rather
than a safety mechanism — so 6 → 7 → 8 is not three sloppy patches, it is one contradiction
surfacing three ways.

#710 already established in this codebase that the committed shard's *existence* is not a
trust signal. My conclusion is the stronger version: its *contents* cannot be one either,
for any property whose whole point is to constrain someone who can write the repo. A ninth
round on the same substrate would most likely produce a ninth self-inflicted hole.

## One finding worth separating out

The round-6 root cause is independent of this idea and live in current `v2`:
`auditBlockField` (`core/tools/aidlc-lib.ts:4949`) returns the **first** matching line for a
field name.

```ts
for (const raw of block.split("\n")) {
  const line = raw.startsWith("- ") ? raw.slice(2) : raw;
  if (line.startsWith(prefix)) return line.slice(prefix.length).trim();
}
```

It is shared and exported, so any authenticate-then-re-read pattern built on it is
decorative — the verifier and the consumer can disagree about which bytes they mean. Worth
either rejecting duplicated fields at the parse boundary or documenting the semantics loudly,
regardless of what you decide about the escalation.

If the boundary should stay where it is — the comparison as orchestrator judgement — then the
fix is documentation: reword `docs/guide/09-rules-and-the-learning-loop.md:90` so it does not
promise a framework-enforced stop, and say plainly that the org-contradiction escalation is
an **audit aid, not an enforcement boundary**. That stops the next person building a control
on that substrate.

If it should instead be enforceable, it needs an anchor outside the repo or outside the
user's write scope, and that is a design decision I did not want to make on your behalf.

Happy to send the docs change, or to implement whichever substrate you prefer.

## Environment

macOS 15, Claude Code harness, bun 1.3.14. Verified against `upstream/v2` at `cc08c1612`
(2.6.65) and re-verified unchanged at `75d8af80` (2.6.68).

_Found while building DocumentKB summaries; the eight rounds were against my own unmerged
attempt, not shipped code. Investigation assisted by an AI agent, claims verified against the
source._

apackeer:
The user guide promised a framework-enforced org-contradiction stop, but `aidlc-learnings.ts persist` never reads `org.md` — the comparison is deliberately the orchestrator-LLM's job and the reference docs already say so. This PR rewords the guide (and the same stale wording in the knowledge guide, harness-engineering guide, and two reference chapters) to present the admission check as an orchestrator-run audit aid rather than an enforcement boundary, and loudly documents `auditBlockField`'s first-occurrence semantics so no authenticate-then-re-read pattern gets built on it. Docs + comment only; no runtime behavior changes.

Whether to BUILD framework enforcement of the org-contradiction stop remains an open product question — this PR corrects the documentation to match today's deliberate behavior and does not foreclose that decision.

Closes #897


leandrodamascena (changes requested):
Thanks for correcting the documentation to match the existing org-conflict admission behavior and for documenting auditBlockField's first-occurrence semantics.

I reviewed the current head 0d0e92eca19bef41f4aa191ca806ccf039f0ba51 against 1c0188a2f6ecf48152bb506bd06af747fc1c5203. The change is aligned with #897: the diff updates documentation and code comments without changing runtime behavior.

[P2] Please remove the version bump. The repository policy explicitly says that pure documentation sweeps and internal changes do not bump the framework version. This PR should therefore remove its CHANGELOG entry, README badge change, core/tools/aidlc-version.ts change, and generated version-copy changes. Please do not replace them with a newer version bump after rebasing. Regenerate dist so only the aidlc-lib.ts comment projections remain.

I made no changes or commits. I confirmed that auditBlockField's executable body is unchanged, package parity passes for every harness, and the version synchronization test passes on the current head. GitHub currently reports the PR as conflicting and reports no CI checks. The remaining risk is limited to documentation accuracy; I found no runtime behavior change in this diff.

apackeer:
Addressed the requested changes in b0b8a80a4.

- Rebased the documentation/comment-only change onto current `v2` (`572dda2`).
- Removed the `CHANGELOG.md` entry, README badge change, authored version bump, and all generated version-copy changes. The repository remains at `2.6.80`.
- Regenerated distributions; the generated diff is now limited to the seven `aidlc-lib.ts` comment projections.
- Confirmed there is no runtime behavior change.

Validation: `bun scripts/package.ts --check` passes for all harnesses, and the focused `t68`, `t174`, and `t86` suites pass (20 tests).

@leandrodamascena, could you please re-review the updated head?

leandrodamascena (approved):

PR #918, opened 2026-08-24
fix: preserve summary write receipts across workspace moves

### LINKED ISSUE #863: [Bug]: checkSummaryConfirmationEvidence matches ARTIFACT_* rows by absolute path — a workspace cloned to a different directory deadlocks, and review-freeze blocks the prescribed remedy
### Description

ARTIFACT_CREATED / ARTIFACT_UPDATED audit rows record **File**: as an absolute path, and the audit shards are committed to git (the shipped guidance is to commit the whole aidlc/ tree). checkSummaryConfirmationEvidence matches those rows against the artifact's current absolute path with ===.

Clone or copy that workspace to any other directory and every pre-move ARTIFACT_* row stops matching. Every stage that had already passed its consolidated-summary confirmation becomes permanently unfinishable: the engine refuses to complete it, and the guard's own prescribed remedy — "Regenerate or re-save it after confirmation" — is refused by aidlc-review-freeze.ts on any reviewer-bearing stage.

The inconsistency is internal to one file. producesArtifactUnit (core/tools/aidlc-lib.ts:4696-4733) matches the same audit rows by suffix (norm.endsWith(suffix)), which is path-prefix agnostic. That is why review-freeze correctly identified the frozen artifacts across the move while the completion guard could not see their write records at all.

## Mechanism

core/hooks/aidlc-write-audit-log.ts writes the absolute path into **File**. The scan at core/tools/aidlc-lib.ts:4365-4392:
```
const file = auditBlockField(entry.block, "File");
if (!file) continue;
const resolved = resolvePath(projectDir, file);
if (resolved === artifactAbs) lastWrite = i;

```
resolvePath is node:path.resolve (imported at :5). Given an absolute file, resolve(projectDir, file) returns file unchanged — projectDir never re-anchors a foreign prefix. artifactAbs comes from summaryArtifactPaths (:4094-4105) and is therefore rooted at the current directory. The prefixes differ, lastWrite stays -1, lastWrite <= receiptIndex holds, and the refusal at :4382-4391 fires.

Contrast the portable sibling, :4721-4731:
```
const norm = file.replace(/\\/g, "/");
for (const name of reviewedArtifacts) {
  const suffix = `/${stage.slug}/${artifactFilename(name)}`;
  if (!norm.endsWith(suffix)) continue;

```
## The deadlock

1. The guard requires a new artifact write after the confirmation receipt.
2. aidlc-review-freeze.ts forbids writing any produces[] artifact that holds a fresh terminal review receipt.
3. That freeze only lifts on a recorded gate rejection.
4. The gate cannot be opened — the same guard also refuses report --result awaiting-approval (independently verified in #856).

The only escape found in the source is the undocumeNFIRMATION_GUARD=1 (:4001-4003), which also disables the receipt-freshness and questions-file-hash checks.

### Steps to Reproduce

1. In /path/A, run a workflow to any summary_confirmation stage that declares a reviewer (e.g. functional-design). Answer the consolidated-summary checkpoint with Looks correct, let the artifacts be generated and the reviewer record a terminal READY receipt. Do not complete the stage.
2. Commit the aidlc/ tree, including <record>/audit/, per the shipped git guidance in CLAUDE.md.
3. git clone (or cp -r) the repository to /path/B — any different absolute path.
4. In /path/B, run bun .claude/tools/aidlc-orchestrate.ts next (or /aidlc --resume).
5. Try the remedy the error message prescribes: re-save the named artifact with the native Edit/Write tool.

### Expected Behavior

The workflow resumes at step 4. Artifact write records in a committed audit log should be recognised regardless of where the workspace is checked out — the log is version-controlled by design, so it has to survive a clone to a different path.

### Actual Behavior

Step 4:

`{"kind":"error","message":"Refusing to complete \"functional-design\": artifact /path/B/aidlc/spaces/default/intents/<slug>/construction/u1-product-catalog/functional-design/entities.md has no recorded native-tool write after the human's consolidated summary confirmation. Regenerate or re-save it after confirmation, then report completion again."}`

The record it cannot find is present in the committed shard — under the old root:

**Event**: ARTIFACT_UPDATED
**File**: /path/A/aidlc/spaces/default/intents/<slug>/construction/u1-product-catalog/functional-design/entities.md

Step 5 (the prescribed remedy) is denied:

PreToolUse:Edit hook error: review-freeze: ".../functional-design/entities.md" is a
declared produces[] artifact of stage "functional-design" unit "u1-product-catalog",
which holds a fresh terminal review receipt. Writing it now would invalidate that
receipt and the engine would refuse the gate ... If the artifact genuinely needs
changes, reject at the gate (or have the human request changes); the recorded
rejection lifts this freeze ...

The gate cannot be reached, so the freeze cannot be lifted. The workflow is stuck with no supported path forward.

### AI-DLC Version

v2 (alpha)

### Release / Commit

2.6.18

### AI-DLC Phase

### Platform / IDE

Claude Code

### AI Model

Claude Opus 5

### Environment

- OS: macOS (Darwin 25.4.0)
- bun: 1.3.14
- AI-DLC: 2.6.18
- Scope: mvp, Construction Iteration: unit-major

### Additional Context

Blast radius. Every summary_confirmation stage confirmed before the move is affected, not just the current one. In our workspace that is 5 stages for one unit (functional-design, nfr-requirements, nfr-design, infrastructure-design, code-generation), each of which will hit the same wall as the engine walks the (stage × unit) grid at core/tools/aidlc-orchestrate.ts:4313.

The triggering operations are all ordinary and supported: cloning your own project onto a second machine, a colleague cloning it, CI checking it out, or renaming a parent folder.

Related, same class. **Project Root** in aidlc-state.md is also persisted as an absolute path and is silently stale after a move.

Relationship to #856. Same function, same line range, opposite failure mode — #856 is the guard being too permissive (a reviewer's ## Review append satisfies it), this is the guard being too strict (a legitimate write is invisible). #856 already documents that the prescribed re-save remedy is blocked by review-freeze, which is the second half of this deadlock. Worth noting for that issue's fix design: neither discriminator proposed there — actor identity on ARTIFACT_* rows, or an artifact SHA-256 recorded at the receipt — addresses path portability, so a content-hash design would inherit this bug unless the path comparison is normalized at the same time.

apackeer:
checkSummaryConfirmationEvidence matched ARTIFACT_* audit rows by absolute path, so a workspace cloned or moved to a different directory could never complete an already-confirmed stage — and review-freeze blocked the prescribed re-save, a full deadlock. The completion guard now also matches the full record-relative artifact suffix (composing with the <project-dir> token from #813, which only covers rows written by 2.6.69+), leaving causal ordering untouched; proven by a t185 moved-workspace test that fails without the fallback.

Note: the committed version slot is 2.6.71; v2's CHANGELOG head has since moved to 2.6.74, so this slot re-bumps at landing per the changelog conflict policy.

Closes #863

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.83] - 2026-08-25
+
+Summary-confirmation artifact-write receipts now remain valid when the same workflow workspace is moved or cloned to a different filesystem location, without weakening their causal ordering. **Upgrade:** refresh your `dist/<harness>/` shell so completion guards use the portable receipt matcher.
+
+* Completion no longer reports `no recorded native-tool write after the human's consolidated summary confirmation` solely because an audit row contains the artifact's absolute path from the workspace's previous location.
+* Moved or cloned workspaces no longer deadlock when the completion guard rejects the old absolute prefix and review-freeze blocks the prescribed artifact re-save; the full space/intent/phase/stage-relative artifact tail now identifies the same recorded write.
+
 ## [2.6.82] - 2026-08-25
 
 `/aidlc --doctor` now catches hooks that Claude Code has globally disabled. Previously an install with `"disableAllHooks": true` could pass every check yet block at runtime because doctor verified that hook files were present and wired, but not that Claude Code would run them. **Upgrade:** refresh your `dist/<harness>/` shell to pick up the new check.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.82-blue)
+![version](https://img.shields.io/badge/version-2.6.83-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -5479,6 +5479,7 @@ export function checkSummaryConfirmationEvidence(
 
     for (const artifact of summaryArtifactPaths(stage, question)) {
       const artifactAbs = resolvePath(artifact);
+      const artifactSuffix = "/" + toPosix(relative(projectDir, artifactAbs));
       const writes = events.filter((entry) => {
         if (
           entry.event !== "ARTIFACT_CREATED" &&
@@ -5487,8 +5488,14 @@ export function checkSummaryConfirmationEvidence(
           return false;
         }
         const file = auditBlockField(entry.block, "File");
-        return file !== null &&
-          resolveAuditProjectPath(projectDir, file) === artifactAbs;
+        if (file === null) return false;
+        const norm = file.replace(/\\/g, "/");
+        return (
+          resolveAuditProjectPath(projectDir, file) === artifactAbs ||
+          // #863: preserve writes across workspace moves. The leading slash
+          // and full space/intent/phase/stage tail prevent partial matches.
+          norm.endsWith(artifactSuffix)
+        );
       });
       const strictlyAfter = (
         later: AuditShardEvent,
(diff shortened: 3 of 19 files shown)


leandrodamascena (approved):
Thanks, Arden. I reviewed the current head `22771b47e` against `v2` at `572dda286`. This is aligned with the project direction as a focused bug fix for the community-reported workspace portability deadlock in #863.

I found no remaining implementation, protocol, or compatibility blocker. The change preserves the existing exact-path resolution and adds a fallback based on the complete record-relative artifact path, including space, intent, phase, Unit when applicable, stage, and filename. This allows the same committed audit receipt to survive a workspace move or clone without matching artifacts from another workflow location.

The causal contract remains unchanged: the matching write must still be strictly ordered after the applicable summary confirmation, including the identical-confirmation authorization introduced by #903. There is no persisted schema change, migration requirement, or breaking behavior.

I also verified that the regression test is effective: restoring the previous matcher makes only the moved-workspace case fail (`109 passed, 1 failed`), while the PR implementation passes all 110 `t185` cases.

I validated the expected rebase composition locally with:

- `t185-stage-artifact-guard`: 110 tests, 389 expectations
- `t320-review-confirmation-deadlock`: 7 tests, 47 expectations
- `t188-human-presence-gate`: 25 tests, 124 expectations
- `t68-version-changelog-sync`: 7 tests
- `bun run check`

Approved. The branch still needs to be rebased onto current `v2`, re-bumped to `2.6.81`, and regenerated through `bun scripts/package.ts`. The current remote head has no GitHub workflow runs, so CI must validate the rebased commit.

PR #919, opened 2026-08-24
fix: normalize positive pathspecs in codekbScopeFingerprint so a trailing slash survives an exclude (2.6.75)

tejasavora:
## The bug

`codekb-scope-diff` reports **STALE for every scope, permanently**, in the lone-repo layout.

`codekbScopeFingerprint` (`core/tools/aidlc-lib.ts`) normalizes its **exclude** pathspecs — stripping a leading `./` and any trailing slash — but spreads the **positive** paths in raw. Git matches nothing when a positive pathspec carrying a trailing slash is combined with any exclude pathspec:

```bash
git init -q repo && cd repo && mkdir -p src/payments aidlc
printf 'a\n' > src/payments/gw.ts && echo x > aidlc/f.txt

GIT_INDEX_FILE=/tmp/i1 git add -A -- "src/payments/"                           # 1 file
GIT_INDEX_FILE=/tmp/i2 git add -A -- "src/payments"  ":(exclude,literal)aidlc" # 1 file
GIT_INDEX_FILE=/tmp/i3 git add -A -- "src/payments/" ":(exclude,literal)aidlc" # 0 files  <-- empty tree
```

The empty match makes `git write-tree` return git's empty-tree hash, `4b825dc642cb6eb9a060e54bf8d69288fbee4904`, which never equals the stored fingerprint — so the staleness verdict is stuck at STALE and the check stops carrying information.

## Why it only bites one layout

The exclude list is non-empty in exactly one case. `aidlc-utility.ts` passes `["aidlc"]` when `repoDir === projectDir`, to keep the framework's own workspace tree out of a full-root fingerprint:

```ts
const fingerprintExcludes = repoDir === projectDir ? ["aidlc"] : [];
```

A sibling-repo layout passes no excludes and is unaffected. Analyzed paths are conventionally written with a trailing slash — `t248`'s own fixture uses `src/payments/` — so within that layout the affected path is the ordinary one, not an edge case.

## The fix

Run the positive paths through the same normalization already applied to the exclusions, so `src/payments/` and `src/payments` fingerprint identically.

## Testing

`tests/unit/t248-codekb-scope-diff.test.ts` goes from **32 pass / 2 fail → 34 pass / 0 fail**. No new test is added, because those two cases already asserted the correct behaviour and were failing against the shipped code:

- `matching fingerprint → CURRENT; edit inside scope → STALE` — expected `CURRENT`, received `STALE`
- `--mint prints the same fingerprint the lib computes` — expected `15aa89e9…`, received `4b825dc6…` (the empty tree)

Also green: `bun run check` exits 0 (package `--check` parity across all 7 trees, typecheck, Biome), `t68-version-changelog-sync` 7 pass, and `bun scripts/package.ts --check` clean.

Verified on **git 2.50.1 (Apple Git-155)**. The pathspec behaviour is git-version-dependent, which is the likely reason CI has not caught this — worth confirming against the git version on your runners.

## Note for the release entry

The CHANGELOG entry carries an upgrade instruction that matters beyond the code fix: any scope timestamp whose recorded fingerprint is `4b825dc642cb6eb9a060e54bf8d69288fbee4904` was written by this bug and should be re-minted. Fixing the computation alone leaves that value sitting in existing records, where it will keep comparing unequal forever.

## Base branch

Opened against `v2`, since that is where the affected code lives and the only branch `ci.yml` triggers on for pull requests. `CONTRIBUTING.md` says to work against `main` — flagging in case that guidance needs updating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.91] - 2026-08-25
+
+Lone-repository CodeKB scope fingerprints now avoid Git's unreliable combination of positive and exclude pathspecs and reject fingerprints that contain no files. **Upgrade:** refresh your `dist/<harness>/` shell before minting or checking CodeKB scope fingerprints, and re-mint (or re-run reverse engineering for) any stored scope timestamp whose fingerprint is `4b825dc642cb6eb9a060e54bf8d69288fbee4904`, the empty-tree value written by the pre-fix bug.
+
+* `codekbScopeFingerprint` stages analyzed paths first and removes framework-owned exclusions from its temporary index separately, preserving root and nested-project scope boundaries without silently rewriting malformed absolute paths.
+* Scopes that stage zero paths now mint `unknown` and report `UNVERIFIED`, never a false `CURRENT` or permanent `STALE` verdict.
+* Deterministic regression coverage simulates the affected Git pathspec behavior on CI and verifies narrow, full-root, nested-project, malformed-path, and empty-scope cases.
+
 ## [2.6.90] - 2026-08-25
 
 `/aidlc --doctor` now distinguishes hook directories that have not had a chance to populate from dead hook seams after a workflow stage ran. **Upgrade:** refresh your `dist/<harness>/` shell so doctor can detect this condition.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ All notable changes to this project will be documented in this file.
 
 Lone-repository CodeKB scope fingerprints now avoid Git's unreliable combination of positive and exclude pathspecs and reject fingerprints that contain no files. **Upgrade:** refresh your `dist/<harness>/` shell before minting or checking CodeKB scope fingerprints, and re-mint (or re-run reverse engineering for) any stored scope timestamp whose fingerprint is `4b825dc642cb6eb9a060e54bf8d69288fbee4904`, the empty-tree value written by the pre-fix bug.
 
-* `codekbScopeFingerprint` stages analyzed paths first and removes framework-owned exclusions from its temporary index separately, preserving root and nested-project scope boundaries without silently rewriting malformed absolute paths.
+* `codekbScopeFingerprint` omits exclusions outside analyzed roots, drops analyzed paths covered by exclusions, and preserves root and nested-project scope boundaries without silently rewriting malformed absolute paths.
 * Scopes that stage zero paths now mint `unknown` and report `UNVERIFIED`, never a false `CURRENT` or permanent `STALE` verdict.
-* Deterministic regression coverage simulates the affected Git pathspec behavior on CI and verifies narrow, full-root, nested-project, malformed-path, and empty-scope cases.
+* Deterministic regression coverage verifies distinct narrow fingerprints, full-root exclusions, nested-project boundaries, malformed paths, and empty scopes.
 
 ## [2.6.90] - 2026-08-25
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.90-blue)
+![version](https://img.shields.io/badge/version-2.6.91-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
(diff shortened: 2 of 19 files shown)


apackeer (dismissed):
Thank you for this contribution, Tejas - the diagnosis is excellent. You isolated a genuinely subtle git behavior (a trailing-slash positive pathspec combined with any exclude pathspec staging nothing), traced it to the exact layout that triggers it, and wrote up the empty-tree consequence and the re-mint implication clearly. The macOS git 2.50.1 evidence you provided turned out to be the key to understanding why CI never caught this.

I reviewed the current head `aacf088c719f429f949f5e6a7f7969ed4cd07661` against `v2` (currently `1c0188a2f6ecf48152bb506bd06af747fc1c5203`, version 2.6.79). I reproduced the root cause independently: on Linux git 2.43.0 the empty-match behavior is reproducible in raw probes, and on an upstream git 2.52.0 built from source it is fully deterministic, which confirms your version-dependence hypothesis. The problem is real and worth fixing. I am requesting changes for four findings, and I also have a consolidation recommendation at the end.

**Findings**

1. **P2 - full-root scopes leak outside a nested project.** `core/tools/aidlc-lib.ts:1652` normalizes the canonical full-root path `./` to an empty string and filters it out. The resulting `git add -A` runs with no positive pathspec, and since git 2.0 that means the entire enclosing worktree, not the current directory. For a project directory nested below the git toplevel (for example a package inside a monorepo), I verified that the fingerprint of `codekbScopeFingerprint(pkg, ["./"], ["aidlc"])` changes when a file outside the project changes; on the merge base it stays correctly scoped to the package prefix. Full scopes (`kind: full` requires `./` in `analyzed.paths`) would report STALE because of unrelated edits elsewhere in the worktree.

2. **P2 - the normalization silently accepts malformed absolute paths.** The regex `^\.?\//` in the new positives block makes the dot optional, so it also strips a bare leading `/`. At this head, `codekb-scope-diff --mint --paths /src/payments/` mints the fingerprint of the repo-relative `src/payments`, and a stored scope recording `/src/payments/` can report CURRENT; `/` alone normalizes to nothing and fingerprints the whole repository minus `aidlc`. On the merge base the same inputs correctly return `unknown` (git rejects the absolute pathspec, the helper returns null, callers report UNVERIFIED). Malformed scope input should stay uncomputable rather than silently selecting different files.

3. **P2 - the existing tests do not guard this fix on the CI runners.** The PR relies on the two existing cases at `tests/unit/t248-codekb-scope-diff.test.ts:294` and `:507` failing without the fix. That failure is git-version-dependent: I extracted the merge-base implementation and t248 passes 34/0 on git 2.43.0 (the Ubuntu CI substrate), while it fails under git 2.52.0. So the production change could be reverted without required CI going red. A regression guard for this defect needs to fail deterministically on the CI git version.

4. **P2 - release metadata and mergeability.** This head claims 2.6.75 (`core/tools/aidlc-version.ts:4`, `CHANGELOG.md:4`, `README.md:14`), but that slot shipped on `v2` via #880 and `v2` is now at 2.6.79, so GitHub reports the PR as conflicting. The v2 conflict needs to be resolved before this can merge: rebase onto current `v2`, keep all base changelog entries, move this entry to the next free patch slot, update the badge and title, regenerate every `dist/` tree, and rerun t68 and the package parity check.

**Overlapping PR - full disclosure**

It looks like we crossed wires on this one: #914, which is my own PR for the same issue (#896), was opened nine minutes before this one, and neither of us saw the other coming. It takes a different mechanism at the same function: it drops exclusions that no analyzed root covers, so the pathological pathspec combination never forms, keeps `./` intact (which avoids findings 1 and 2 structurally), and adds an `ls-files` zero-staged guard so the empty-tree hash can never be minted under any git behavior, plus regression tests that fail without the fix independent of git version. Since I authored #914 I should not be the one to pick between the two PRs, so I will leave the consolidation call to the other maintainers. Whichever way that lands, one part of this PR should survive regardless: your changelog upgrade instruction telling users to re-mint any stored fingerprint equal to `4b825dc642cb6eb9a060e54bf8d69288fbee4904` is exactly right, #914 currently lacks it, and I intend to adopt it there with credit to you. I would also genuinely value your review of #914, given how precisely you diagnosed this. If the maintainers continue with this PR instead, findings 1-4 above are the required changes.

**Validation performed:** t248 at this head passes 34/0 and `t68` 7/0 on git 2.43.0; t248 at this head also passes 34/0 under upstream git 2.52.0; `bun run check` (package parity across all seven dist trees, typecheck, biome) exits 0, confirming the dist projections were regenerated rather than hand-edited. Your CONTRIBUTING.md observation is accurate as well: it still says to work against `main` while `v2` is the actual working base; we will follow up on the docs separately. Note that no hosted CI has executed for this head yet (the workflow runs are awaiting first-time-contributor approval).

leandrodamascena (approved):
Thanks, Tejas, for identifying and working on this CodeKB fingerprint issue.

I reviewed the current head `d721657aa44bb30c097bfe284151e306cbfa981a` against `v2` at `ba0abc7444fe454a73753e047c6cbdf8d93d4717`. Project alignment was established by Arden’s earlier maintainer review of this bug fix.

I addressed the remaining findings and pushed the corrections:

- Preserved original positive pathspecs, including `./`, so full scopes in nested projects remain bounded to the project rather than the enclosing worktree.
- Kept malformed absolute paths invalid instead of silently converting them to repository-relative paths.
- Normalized exclusions only for coverage checks, omitted exclusions outside the analyzed roots, and removed positive paths fully covered by exclusions.
- Prevented zero-file temporary indexes from producing Git’s empty-tree fingerprint; these scopes now produce `unknown` and remain `UNVERIFIED`.
- Added deterministic regression coverage for narrow scopes, full-root exclusions, nested projects, malformed paths, exclusion-covered paths, and empty scopes.
- Rebased onto the current `v2`, updated the release metadata to `2.6.91`, and regenerated every harness distribution.

Local validation passed: 40/40 CodeKB scope tests, 7/7 version synchronization tests, package drift checks for all seven harnesses, typecheck, and lint.

GitHub CI is fully green for this exact head, including smoke/unit, integration/e2e, contract checks, changelog, markdownlint, Gitleaks, Semgrep, and Bun audit.

I found no remaining code or protocol blocker. Approved.

PR #902, opened 2026-08-24
docs: correct the balanced tier's projection and two agent-matrix rows, pinned with derived tests

tejasavora:
## What

Two documentation claims contradict the authored implementation. Both are corrected and pinned with tests that **derive** the expectation from the source of truth rather than restating it, in the existing `t239-documentation-parity.test.ts` idiom.

### 1. `balanced` pins `effort: medium`; the docs said it inherited session effort

`TIER_PROJECTIONS.balanced` projects `{model: sonnet, effort: medium}` on Claude, `model_reasoning_effort = "medium"` on Codex, and `variant: medium` on opencode. The shipped agents confirm it — both `dist/claude/.claude/agents/aidlc-product-lead-agent.md` and `…aidlc-architecture-reviewer-agent.md` carry `effort: medium`.

`core/tools/aidlc-tiers.ts`'s own header comment still described the older behaviour ("`balanced` writes `model: sonnet` and NO effort line; only `templated` pins `effort: medium`"), contradicting the `TIER_PROJECTIONS` literal ~100 lines below it. Three docs repeated it:

- `docs/reference/05-agent-system.md` — the tier Meaning row, three cells of the per-harness projection table (Claude / Codex / opencode), and "absence is the contract for judgment and balanced"
- `docs/reference/14-claude-features.md` — the Claude projection cell, and "absence is deliberate for the first two tiers"
- `docs/harness-engineering/03-adding-an-agent.md` — "templated is the one tier that steps effort down"

A consequence worth stating in the docs so the next reader doesn't re-derive it: **`balanced` and `templated` currently project identically in all six harnesses.** The docs described a three-rung ladder where the shipped projection has two. `judgment` is the only tier that inherits session effort. The two tier names are still worth keeping apart — they describe different work and can be retuned independently — so the note says that rather than proposing a merge.

### 2. Two rows of the Agent Comparison Matrix are stale

Derived from all 33 stage files' `lead_agent` / `support_agents` frontmatter:

| Agent | Documented | Shipped |
|---|---|---|
| `aidlc-architect-agent` | 6 lead / 3 support / 9 | **7 / 3 / 10** |
| `aidlc-aws-platform-agent` | 2 lead / 4 support / 6 | **2 / 5 / 7** |

The "broadest stage involvement (9 stages across 3 phases)" observation moved with it. The repo already contradicted itself here: `docs/guide/06-agents.md` says 10 stages.

## The guards

Both added to `tests/unit/t239-documentation-parity.test.ts`:

[...]

PR #944, opened 2026-08-25
test: pin the Scope Routing summary table, the one unguarded transcription in 05-scopes-and-depth.md

tejasavora:
## What

`t244` already pins the **Stage-by-Scope Matrix** in `docs/guide/05-scopes-and-depth.md` cell by cell, and its BEGIN marker tells readers not to hand-edit those cells. The **Scope Routing Table** higher up the *same file* transcribes the same compiled grid a second time — one row per scope with `EXECUTE / Total`, Depth, and Test Strategy — and nothing reads that markdown.

The neighbouring guards each stop just short of it:

| guard | what it pins | why it misses this |
|---|---|---|
| `t244` | the matrix, cell by cell | its `covers:` names `05-scopes-and-depth.md(scope-stage-matrix)` — the matrix region only |
| `t213` | `renderScopeTable`'s EXECUTE / Total vs the cost helper | that's the CLI's rendered output, not this doc |
| `t125` | each scope file's depth/testStrategy vs `loadScopeMetadata` | pins the **data**, not this rendering of it |

**No drift exists today** — all 11 rows are correct. This is preventative, and I'd rather say that plainly than dress it up.

## Why it's worth a test anyway

The precedent is concrete and recent. The agent matrix in `docs/reference/05-agent-system.md` sat at `6 | 3 | 9` for the architect while `docs/guide/06-agents.md` said 10 stages, until #902 corrected it and added the derived guard. An unpinned second transcription of derivable data is that same shape.

And this chapter has already had the triggering kind of edit: 2.6.70 moved `bugfix` 7 → 9 and `refactor` 8 → 10. That's precisely the change that updates one table and forgets the other.

## The five tests

Written in this file's existing both-directions style:

- **9 / 9b — rows.** One per compiled scope; no extra, unknown, or duplicate. `9b` pins what the scope filter is *allowed* to drop — the `(auto-detect)` sentinel, all `Varies` — so a future stray row fails loudly instead of vanishing behind the filter.
- **10 — counts.** `EXECUTE / Total` equals the compiled grid.
- **11 — depth + test strategy.** From `loadScopeMetadata`, consuming the framework's own resolution rather than a fresh frontmatter parse. Test Strategy inherits Depth when a scope declares no override, which is why `classic` reads Standard/Standard while `workshop` reads Standard/Minimal.
- **12 — the chapter against itself.** The summary count and the matrix footer total are two renderings of one number, so a partial edit that fixed only one is caught even when both happen to match the grid.

## Verification that they bite

Mutated the doc and re-ran, rather than trusting a green suite:

| mutation | tests failed |
|---|---|
| `bugfix` 9 → 7 (the pre-2.6.70 value) | **2** — the count and the cross-check |
| `workshop` Minimal test override lost | **1** |
| `express` row deleted | **2** |
| stray non-scope row added | **5** |

Doc restored byte-clean afterwards (`git diff docs/` empty).

## Housekeeping

- Mechanism: none — static files only, matching this file's existing contract (no spawn, no LLM, no env seam).
- `tests/.coverage-registry.json` regenerated for the two new `covers:` units.
- `bun run check` exits 0.
- Doc/test-only, so no version or CHANGELOG bump. **`docs/` is unchanged** — the tables are already correct.


leandrodamascena (approved):
Thanks for identifying and guarding this remaining documentation transcription. I reviewed the current head `a1e6b991` against current `v2` (`1c6e03a`). The change aligns with the existing `t244` and #902 direction: derived documentation should be checked against compiled project data, without changing runtime behavior.

During the adversarial review, I found that the parser selected the first matching table header anywhere in the document. A correct duplicate placed before the section allowed the actual `Scope Routing Table` to drift to `bugfix 7 / 33` while all 13 tests still passed.

Since this is an eligible community contribution, I fixed that bounded gap in commit `a1e6b991`. The parser is now anchored to the unique `## Scope Routing Table` section, bounded by the next H2, and rejects duplicate or misplaced table headers.

Validation:

- The reproduced false-negative mutation now fails deterministically.
- Focused scope coverage: 38 tests, 1,077 assertions, zero failures.
- The same focused suite passed on a synthetic merge with current `v2`.
- `bun run check` passed package parity, all TypeScript configurations, and Biome.
- All GitHub CI, documentation, security, smoke/unit, and integration/e2e checks are green for `a1e6b991`.

The remaining authored `Use Case` prose is intentionally outside the derived-data contract. I found no remaining correctness, compatibility, or maintenance blocker.

Approved.

PR #879, opened 2026-08-23
feat: team-owned units - parallel Construction for multiple teams

apackeer:
## Summary

Implements the team-owned units design from #795 (full proposal and discussion: the [RFC comment](https://github.com/awslabs/aidlc-workflows/issues/795#issuecomment-5351974551)): teams claim units of work and complete them independently - each team drives its own unit through the four design stages and code generation in its own checkout, with its own gates, while main pins, gates, and lands each finished unit through a resumable merge transaction. Everything is opt-in behind `Unit Ownership: team` (set at delivery-planning, requires unit-major): with the knob absent, every existing flow is byte-identical, pinned by deep-equal dormancy tests.

One user-visible release: **2.6.107**. The branch carries one feature commit plus two focused Copilot contention test follow-ups; the work landed in four increments, an independent-review hardening round, and a first maintainer-review fix round:

- **Per-unit state**: the `set-unit-ownership` / `set-unit-gate-rhythm` verbs, the derived (engine-owned, hand-edit-inert) `## Unit Progress` grid, per-unit gates in two rhythms (`per-stage` default, `unit-end`), and (stage, unit, attempt)-keyed gate/review receipt floors.
- **Atomic claims + scoped walk**: claims are unique CAS branch commits (`claim/<intent-id8>/<unit-slug>`, create-only `--force-with-lease` + nonce re-read; atomic `update-ref` for no-remote sibling-worktree topologies) with tombstoned release and attempt generations. A claimed checkout walks only its unit (guided by a typed `unit-claim` ask - nobody memorizes a command), refuses foreign units at routing and state layers, and works offline: registry reads happen only at claim-sensitive boundaries. Fork primitives bind to live claims; unscoped team-mode main emits a turn-terminal fan-out notice.
- **Pinned merge-back transaction**: a scoped checkout publishes; main pins the exact candidate OID, reads all evidence via git plumbing from that OID, gates the merge (human-turn authority from main's own audit shard only), and lands through a resumable git -> state -> audit choreography with candidate-exact pre-commit tree validation and engine-owned metadata excluded. Receipt transport means nothing re-runs on main; the per-unit block settles when the grid completes, then 3.6/3.7 run once as today. `docs/guide/workshop-mode.md` grows into the general multi-team chapter (and its previously-promised per-Bolt 3.1-3.5 loop in a claimed worktree now actually works).
[...]

PR #947, opened 2026-08-26
fix: first-class no-DAG per-Unit review continuity - merged Bolt units satisfy the gate with per-unit receipts

apackeer:
## Summary

#903 deferred one item: a per-Unit stage with no authored `unit-of-work-dependency.md` required a stage-level review receipt once its Bolts closed. A swarm-style flow only ever records per-Unit receipts, so after the last `BOLT_COMPLETED` the gate demanded a receipt the flow never produces, and the operator's only exits were hand-authoring the dependency document or a human gate rejection. The per-Unit receipts recorded while the Bolts were open were valid, fresh, and ignored.

An earlier attempt at this (persisting a conductor-supplied Unit set through `runtime-graph.json` recompiles) was built and removed during #903's review after producing a defect in three consecutive rounds: the cache could be truncated irrecoverably by a narrower `prepare`, deleted outright by a recompile, and it let conductor input reach guards that document authored-DAG authority. This PR takes the approach #903's review identified instead: derive the Unit set from the audit trail on demand and persist nothing. The current attempt's `BOLT_STARTED`/`BOLT_COMPLETED`/`BOLT_FAILED` rows already record exactly which Units ran, merged, or were discarded, and they are computed inside the same scan window that governs receipt freshness, so the demand side and the supply side cannot drift. Nothing is written to `runtime-graph.json` or the docs tree; `resolveBoltDag`, compile, and every authored-DAG authority guard (swarm settlement, autonomous batch eligibility, produces[] exemptions, `prepare`'s resolved-DAG requirement) are untouched.

**Upgrade:** refresh the whole `dist/<harness>/` shell so review admission, artifact fingerprints, receipt filtering, and gate coverage move together. An in-flight no-DAG stage-level review request stamped before this release, or while every Bolt was still open, may stop matching after a Bolt merges; re-dispatch its exact pending ordinal with `--retry-pending` before recording the verdict.

## What changes for users

* A no-DAG per-Unit stage with merged Bolts now satisfies its reviewer gate with either one fresh stage-level receipt or fresh per-Unit verdicts covering every merged Unit. Failed or discarded Bolts owe no review.
* `aidlc-log.ts review --unit <unit>` stays available after that Unit's Bolt completes (fresh requests, `--retry-pending`, and stale-receipt recovery), so post-merge re-review no longer requires authoring the dependency document. A Unit with no Bolt history in the current attempt still refuses.
* The missing-receipt refusal at the gate names each Unit owing a review and the exact per-Unit request command executable in that state, including the recovery-spent case.
* Stage-level review fingerprints on a no-DAG stage enumerate merged Bolt Units only (open-Bolt work is not yet in the main tree); with no merged Unit the single stage-level path from 2.6.76 is unchanged. Per-Unit receipt filtering recognizes both open and merged Units, and a forged receipt for a Unit outside that observed set cannot satisfy coverage.
* `aidlc-runtime.ts compile` cannot erase the continuity: the observed set is derived from audit rows, never cached.

## Verification

The limitation was reproduced against the #903 merge base before fixing it: the new `t328` continuity sequence (Bolts started, per-Unit reviews recorded while open, Bolts merged, gate) fails 5 of 6 cases on the unmodified base with exactly the stage-level demand and no-active-Bolt refusals, and passes on this branch.

An adversarial internal review round then ran against the rebased head and found three P2 defects in the first version of the tie reduction, all reproduced through the shipped CLI; each is fixed and pinned: the gate's named per-Unit command is now executable after a settled start/completion tie (the request path consumes the same reduced window instead of failing closed on a tie the reducer already resolved); same-shard append order is preserved as a causal constraint, so an unrelated same-second row in another shard can no longer flip a real merge failure to "merged"; and lifecycle truth is per attempt, so a Unit may be both merged and open when an older attempt lands while a newer one is active (merged membership drives gate demand and fingerprints, the union drives receipt filtering).

Full `smoke+unit+integration` tier on the final tree: 370 files, 9,101 assertions, 0 failures (independent builder and verifier runs). `package.ts --check`, `typecheck`, `biome`, coverage registry, and `t68` are clean. New test `t328` (11 cases): end-to-end continuity incl. mid-flow compile immunity, missing-verdict refusal executing its own advice, forged historyless receipts, discarded Bolts, post-merge admission and retries, stage-level compatibility, open-only fallback staleness after merge, cross-shard tie order-independence for the completed-vs-failed, started-vs-completed, and same-shard-causal shapes (both shard orderings asserted, extending the boundary-tie hardening from #903's final round), and joint merged-plus-open attempt membership. The behavioral cases fail on an unmodified v2 with the old stage-level refusals, not with structural errors.

## Notes for reviewers

* Worth a close look: `reviewAttemptWindow` in `aidlc-lib.ts`. It hoists the exact event filter, ordering, and attempt-floor computation that `freshReviewReceipts` already used (moved, not reimplemented) and adds the Bolt lifecycle reduction. Within a same-timestamp group, each shard's rows keep their append order (causal), and the strict started/failed/completed priority applies only between the current heads of different shard queues, so shard filename order can never decide merged-versus-failed while known causal order is never permuted. The equal-priority tie-break by shard index cannot change outcomes: equal-priority heads are same-type events whose application order cannot alter final attempt states.
* Bolt terminal rows pair to their own attempt: exact open slug first, then exact historical slug, then a name-based fallback only where one side lacks a slug. `BOLT_FAILED` rows carry `Failed Bolt` rather than `Bolt names`; both consumers read the right field. The request path (`reviewAttemptSummary`) consumes the same reduced window, and its cross-shard Bolt-tie ambiguity clears only when every tied row is that Unit's own lifecycle row and the reducer settled it merged; authority-row ties keep failing closed.
* Two behavior notes for the record: open Bolts add no separate gate block (the resolved-DAG branch has none either; this PR mirrors rather than invents), and a stage-level receipt recorded while every Bolt was open is deliberately invalidated by the first merge, because the merge changes the reviewed tree - `--retry-pending` rebinds it.
* The `--retry-pending` admission bypass from #903's final round composes with the new merged-Unit admission: retries skip membership re-admission entirely, while fresh post-merge requests go through it; `t328` pins that the first post-merge request is a normal request, not a retry.


leandrodamascena (changes requested):
Thanks, Arden. I reviewed the current head `b87b9955` against merge base `cb30d642`, and replayed it on current `v2` at `e4d17257`. The direction aligns with the no-DAG follow-up explicitly deferred by #903.

**P1 - Require merge evidence before treating a Bolt as merged**

`aidlc-bolt complete --merge` emits `BOLT_COMPLETED` before running the state and audit merges, while `reviewAttemptWindow()` immediately promotes that event to `mergedBoltUnits`.

I reproduced a crash immediately after `BOLT_COMPLETED`, with no `STATE_MERGED` or `AUDIT_MERGED`. The Unit was still considered merged, a post-merge review could be recorded against stale artifacts already present in the main tree, and a READY verdict allowed stage finalization. A slug-backed attempt can also reach the same state by running `complete` without `--merge`.

For a worktree-backed attempt, `BOLT_COMPLETED` should represent completion pending merge. It should become merged only after a matching later `AUDIT_MERGED` proves that the state and audit merge sequence landed. A name-only, non-worktree Bolt can continue treating `BOLT_COMPLETED` as terminal. A no-slug completion must not close a slug-backed attempt, and a later fragment-cleanup `BOLT_FAILED` after `AUDIT_MERGED` must not erase an already confirmed merge.

Please add regressions covering:

- A crash between `BOLT_COMPLETED` and `AUDIT_MERGED`.
- A slug-backed attempt completed without `--merge`.
- A fragment-merge failure after `AUDIT_MERGED`, which should remain merged.

The focused exact-head tests otherwise pass, including all 11 `t328` cases. The published checks are green, but those cases synthesize `BOLT_COMPLETED` as sufficient merge evidence and therefore encode rather than detect this failure window.

I also performed a diagnostic rebase onto current `v2`. The implementation merges cleanly apart from version/changelog bookkeeping, but the PR currently remains conflicting and will need the corresponding version bump and regenerated coverage metadata.

Thank you for taking on the deferred continuity work. The audit-derived approach remains the right direction, but the reducer must distinguish completion from a successfully landed merge before this can safely gate review continuity.

apackeer:
Thanks Leandro — the P1 was real and exactly where you pointed. `complete --merge` emits `BOLT_COMPLETED` before either merge primitive runs, and the reducer took completion as merge evidence.

Addressed in the updated head:

- `reviewAttemptWindow()` now distinguishes completion from a landed merge: a slug-backed (worktree) attempt's `BOLT_COMPLETED` is completion-pending-merge and becomes merged only when the matching later `AUDIT_MERGED` receipt confirms the state+audit merge sequence landed on main (audit-merge appends delta + receipt in one atomic descriptor-pinned write, so the receipt can't be lost separately). A name-only, non-worktree Bolt keeps `BOLT_COMPLETED` as terminal.
- In the crash window you reproduced, the Unit now stays unmerged: the gate doesn't demand or accept its per-unit review, and re-running `complete --merge` (idempotent) lands the evidence and unlocks it.
- A slugless completion can no longer close a slug-backed attempt, and a later fragment-cleanup `BOLT_FAILED` cannot erase a confirmed merge.
- All three requested regressions are in t328, and each fails against the previous reducer. The t328/t320 fixtures now synthesize `AUDIT_MERGED` alongside `BOLT_COMPLETED` so they detect rather than encode the window.

Also rebased onto current v2 (reconciling with #879's team-ownership floor semantics inside `reviewAttemptWindow`) and re-slotted to 2.6.109. Local validation: smoke+unit and the full deterministic integration tier green, packaging parity, typecheck, lint, and coverage registry clean.

leandrodamascena (approved):
Thanks, Arden. I reviewed the current head `4001634b` against merge base `b9fe852f` and replayed it onto current `v2` at `33e5d019`. The direction remains aligned with the no-DAG continuity follow-up deferred by #903.

The previous blocking issue is resolved: slug-backed completion now requires a later matching `AUDIT_MERGED`, and the requested crash, completion-without-merge, and fragment-failure regressions pass.

One non-blocking follow-up: `freshReviewReceipts()` currently lets a same-second, cross-shard `AUDIT_MERGED` make a team review request appear ambiguous. Since this event carries no reviewer authority, it should be excluded from receipt-freshness tie detection. A focused regression would protect that behavior.

The branch still needs rebasing onto `v2`. The conflict resolution should preserve both `captureBytes` and `mergedBoltUnits` through `reviewArtifactSnapshot()` and the `reviewArtifactFingerprint()` wrapper, followed by the version update and regenerated distributions and coverage metadata.

On the exact head, the focused tests, version synchronization, typecheck, and package parity pass. The diagnostic integration replay also passes the added follow-up regression, coverage registry, lint, and drift checks. No changes were made to the author branch. GitHub currently reports no checks for this conflicting head.

Thank you for addressing the earlier merge-evidence gap. Approved, with the freshness-tie case recorded as an additional follow-up.

PR #952, opened 2026-08-26
feat: require structured questions to be self-explanatory

### LINKED ISSUE #950: Structured questions are too abstract — bare identifiers like FR3/url1 are unanswerable without a rephrase
**Which rule, stage, agent, or harness is affected**
`core/aidlc-common/protocols/stage-protocol.md` — §3 Question Format / Depth-aware question generation (harness-neutral; observed on Kiro CLI).

**What happens**
In the same AI-DLC V2 delivery, teams found Inception questions too abstract to answer directly. Questions referenced bare shorthand — `FR3`, `url1` — with no context, forcing teams to ask the agent to rephrase before they could respond. When they did, the rephrased questions were straightforward and answerable, which shows the capability exists but is not the default.

**Expected**
Every question should stand on its own: identifiers expanded on first use, a one-line reason when it is not obvious, and concrete domain phrasing — so a reader can answer without asking for a rephrase.

**Actual**
The harness `question-rendering.md` annexes govern *format* rigorously (numbered prose, the "Other" escape, answer capture) but say nothing about question *content*. The question spec schema (`prompt`, `header`, `multiSelect`, `options[].label`, `options[].description`) has no context/rationale field and no rule requiring identifiers be expanded. Nothing in the protocol prevents a bare `Is FR3 still correct?`.

**Proposed fix**
Add a "questions must be self-explanatory" rule to Depth-aware question generation: expand every identifier on first use (write what it names, tag in parentheses), give one line of context when the reason is not obvious, and prefer concrete domain phrasing over the framework's internal vocabulary. Protocol layer, not per-stage or per-harness, so it applies everywhere via the existing "follow stage-protocol.md" mandate. Prose-only; regenerate `dist/`.

**Platform/harness/model tested:** Kiro CLI, v2 (2.6.108).

SaRedfiche:
Addresses field feedback (Indeed AI-DLC V2 delivery): questions were too abstract and bare shorthand like FR3/url1 was unanswerable without asking the agent to rephrase. Adds a rule to depth-aware question generation requiring every question to stand on its own: expand identifiers on first use, one line of context when the reason is not obvious, concrete domain phrasing over internal vocabulary. Prose-only change to core/aidlc-common/protocols/stage-protocol.md; dist regenerated, version 2.6.109, CHANGELOG + README badge bumped. Gates: typecheck, lint, drift-check, and t68 all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.112] - 2026-08-27
+
+Structured questions must now be self-explanatory, closing a top field-feedback gap where teams found questions too abstract and bare shorthand like `FR3`/`url1` unanswerable without asking the agent to rephrase. **Upgrade:** refresh your `dist/<harness>/` shell.
+
+* Depth-aware question generation gains a "questions must be self-explanatory" rule: expand every identifier in each question that uses it (write what it names, then tag it in parentheses), give each question one line of context when the reason is not obvious, and prefer concrete domain phrasing over the framework's internal vocabulary.
+
 ## [2.6.111] - 2026-08-26
 
 Plugin composition now keeps co-located tests and fixtures out of the installed engine tools tree and reports every rejected payload. **Upgrade:** refresh your `dist/<harness>/` shell and each plugin projection, then re-compose plugins; remove any already-installed file named by the new advisory and re-run compose.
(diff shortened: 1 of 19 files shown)


leandrodamascena (approved):
Thanks, Robert. This addresses a concrete field usability issue and puts the guidance in the shared stage protocol, so all supported harnesses receive the same behavior.

I reviewed the current head `dd9626e6c` against `v2` at `64399159a`. Project alignment was established from the reported delivery feedback and the maintainer decision to proceed with clearer, self-contained workflow questions.

I found no remaining blockers.

I pushed a bounded maintainer update that:

- rebases the contribution onto the current `v2`;
- clarifies that identifiers must be expanded in every question that uses them, preserving the requirement that each question stand on its own;
- adds structural regression coverage for the new protocol contract;
- re-slots the release metadata to `2.6.112`;
- regenerates all seven harness distributions.

Validation:

- 80 focused protocol and version tests passed locally.
- Packaging parity and `git diff --check` passed.
- Exact-head CI is fully green: smoke/unit, integration/e2e, parity/typecheck/lint, changelog, Markdown, dependency audit, Semgrep, and Gitleaks.

The remaining risk is limited to model adherence to prose guidance; the structural test now prevents the contract itself from disappearing or regressing silently.

PR #956, opened 2026-08-26
docs: explain Claude Bedrock default

leandrodamascena:
## Summary

- explain why the Claude distribution uses Bedrock as its shipped runtime baseline
- document the reproducibility, context-window, credential-chain, IAM, and test-baseline considerations behind that choice
- clarify that Bedrock is a distribution default rather than an AI-DLC methodology requirement
- explain how to select another Claude Code-supported provider

## Validation

- `bun test tests/unit/t68-version-changelog-sync.test.ts tests/unit/t174-docs-legacy-refs-gate.test.ts tests/unit/t246-docs-rewrite-links-fences.test.ts` (17 passed)
- `git diff --check`

No version or changelog bump is included because this is a documentation-only change.

### DIFF
--- a/docs/guide/01-getting-started.md
+++ b/docs/guide/01-getting-started.md
@@ -34,6 +34,23 @@ command -v bun    >/dev/null && echo "✓ bun installed"          || echo "✗ I
 
 This implementation ships configured for **AWS Bedrock**. The shipped `.claude/settings.json` sets:
 
+### Why Bedrock is the shipped default
+
+The Claude distribution needs a predictable runtime baseline across the
+orchestrator and its tier-pinned subagents. Bedrock lets the distribution name
+exact global inference profiles and their context variants, so a workflow does
+not silently pick different model aliases or context windows on different
+machines. It also uses the standard AWS SDK credential chain and IAM controls,
+which lets teams manage access without committing provider keys to a project.
+The repository's live Claude test environment uses the same provider baseline.
+
+This is a distribution default, not an AI-DLC methodology requirement. AI-DLC
+does not call the Bedrock API directly. To use the direct Anthropic API or
+another Claude Code-supported setup, remove the Bedrock-specific environment
+and model pins from the installed `.claude/settings.json`, then configure Claude
+Code normally. The workflow contract is unchanged; the selected models still
+need enough context for the orchestrator and delegated agents.
+
 | Variable | Value | Purpose |
 |----------|-------|---------|
 | `CLAUDE_CODE_USE_BEDROCK` | `1` | Routes Claude Code through Bedrock |
--- a/docs/guide/01-getting-started.md
+++ b/docs/guide/01-getting-started.md
@@ -32,24 +32,46 @@ command -v bun    >/dev/null && echo "✓ bun installed"          || echo "✗ I
 
 ## AWS Bedrock Setup
 
-This implementation ships configured for **AWS Bedrock**. The shipped `.claude/settings.json` sets:
+The Claude Code distribution ships configured for **AWS Bedrock**. The shipped `.claude/settings.json` sets:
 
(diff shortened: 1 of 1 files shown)


apackeer (changes requested):
Thanks for adding this rationale. The Claude-specific baseline is supported by
the shipped settings and live test setup, and the version, test, and packaging
checks all look clean. I have three documentation changes I would like to see
before merge:

1. **Please clarify how this rationale relates to the other harnesses.** The new
   section discusses only the Claude distribution
   (`docs/guide/01-getting-started.md:39-45`), but Codex also ships with Amazon
   Bedrock as its default provider
   (`docs/guide/harnesses/codex-cli.md:19-23`), and opencode's tiered personas
   include a Bedrock model pin (`docs/guide/harnesses/opencode.md:34-37`).
   Since provider setup is harness-specific (`README.md:101-104`), could we
   either put the shared rationale in a harness-neutral section with the
   per-harness differences, or explicitly title this as Claude-specific and
   cross-link corresponding explanations for the other Bedrock-backed
   surfaces?

2. **Please separate the Bedrock model profile from Claude Code's context
   selector.** The wording at `docs/guide/01-getting-started.md:40-42` currently
   attributes both exact inference profiles and context variants to Bedrock.
   The explanation at line 63, and the contract in
   `tests/smoke/t03-settings-json.test.ts:42-50`, show that `global.*` is the
   Bedrock inference-profile ID while `[1m]` is interpreted by Claude Code and
   stripped before the request reaches Bedrock. Distinguishing those two layers
   would make the reproducibility explanation more precise.

3. **Could the alternate-provider instructions be more concrete?** The PR body
   says it will explain how to select another provider, but
   `docs/guide/01-getting-started.md:47-52` currently says only to remove the
   Bedrock settings and configure Claude Code normally. The effective
   configuration includes the routing flag, four model mappings, and the
   top-level model selection (`harness/claude/settings.json:22-31`), while
   `settings.local.json` may also take precedence
   (`harness/claude/settings.local.json.example:1-12`). Please identify the
   exact keys to remove or replace, mention both settings files, and link or
   show the provider-selection step. It would also help to qualify "the
   workflow contract is unchanged": the stage protocol is provider-independent,
   but the repository ships and tests a specific model/context baseline.

The remaining review checks passed: this is correctly treated as a
documentation-only change with no version or CHANGELOG bump, no generated
`dist/` files changed, `bun scripts/package.ts --check` passed for all
harnesses, and the three listed unit files passed all 17 assertions.

leandrodamascena:
Thanks for the detailed review. I reviewed the current head `e25dda7` against `v2` and addressed all three requested documentation changes:

- Made the rationale explicitly Claude Code-specific and linked the corresponding Codex and opencode provider behavior.
- Separated Bedrock’s `global.*` inference-profile IDs from Claude Code’s `[1m]` context selector.
- Added concrete alternate-provider steps covering every effective setting, local overrides, and the provider authentication flow.

The AI-DLC stage protocol is now described as provider-independent while the shipped and tested Bedrock model/context baseline remains explicit.

Validation passed: all 17 targeted assertions, `bun scripts/package.ts --check` for all seven harnesses, and `git diff --check`. The remaining smoke/unit and integration/e2e CI jobs are currently running on this head; all other reported checks are green. Please take another look when they complete.

apackeer (approved):
Re-reviewed current head `e25dda7`. The revision addresses all three requested documentation changes: it makes the rationale Claude Code-specific with the relevant cross-harness links, separates Bedrock inference-profile IDs from Claude Code's `[1m]` selector, and gives concrete alternate-provider steps covering the shared and local settings plus authentication.

The updated wording matches the shipped configuration and test contract, `git diff --check` is clean, and `bun scripts/package.ts --check` passes for all harnesses. Approved.

PR #957, opened 2026-08-26
docs: add sbx-as-evaluator-substrate dogfooding proof

markmaxwell-Docker:
Local/fork-only proof artifact showing sbx can replace AI-DLC's bespoke docker-run sandbox mechanism, verified live end-to-end for the real install+test call path. Not a PR against upstream aidlc-workflows.

# Summary

> replace with your summary...

## Changes

> replace with a description of the changes

## User experience

## Checklist

* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented

## Test Plan

## Acknowledgment

### DIFF
--- a/scripts/aidlc-evaluator/demo_sbx_dogfood/DOGFOODING.md
+++ b/scripts/aidlc-evaluator/demo_sbx_dogfood/DOGFOODING.md
@@ -1,0 +1,89 @@
+# Dogfooding proof: Docker Sandboxes as AI-DLC's evaluator execution substrate
+
+Local/fork-only proof artifact — not a PR against `awslabs/aidlc-workflows`.
+Demonstrates that `sbx` (Docker Sandboxes) can replace the evaluator's own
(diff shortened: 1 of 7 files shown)


leandrodamascena:
Thanks for taking the time to explore this integration.

We could not determine the intended upstream contribution. The PR describes itself as a “local/fork-only proof artifact” and explicitly says it is not an upstream contribution, while the remaining PR template sections are incomplete.

Because of that ambiguity, we are closing this PR. It may simply have been opened against the upstream repository by mistake.

If this was intentional and you would like to propose an upstream integration, please open an issue first explaining what `sbx` is, which AI-DLC problem it would solve, why the existing sandbox mechanism is insufficient, and the expected integration, security, platform-support, dependency, and maintenance implications. That will give the maintainers enough context to evaluate alignment before reviewing an implementation.

Thanks for understanding.

PR #962, opened 2026-08-27
fix: stop committing machine-local absolute paths in aidlc-state.md

### LINKED ISSUE #937: [Bug]: aidlc-state.md commits machine/session-local details into shared state (absolute Project Root is the acute case)
_Suggested labels: bug, state, portability, concurrency_

### The problem
`aidlc-state.md` is committed to git as the shared, authoritative workflow-state file, but it mixes genuinely shared truth with details that are local to one machine, one session, or one agent:

- an **absolute `Project Root` path** (machine-local),
- **current status / current stage**, **session-resume info**, and **timestamps** (session-local),
- arguably the **execution/progress state** itself.

Because these all live in one committed file, they don't survive travel and they don't parallelize: the moment a second developer clones the repo, a workspace is moved, or two agents advance the same intent at once, the committed values are stale or conflicting for everyone but the last writer.

### The acute case (please fix this one)
`Project Root` is written **once** at intent-create as an absolute path, committed as shared truth, **trusted** by sensors, and **never re-derived**. On any other clone or after any move it points at a directory that doesn't exist on this machine, and the framework resolves and trusts that stale path. This is the must-fix; the rest below is a nice-to-have cleanup.

### Requested fix
- **Must:** don't treat a committed absolute `Project Root` as authoritative — make it machine-local (gitignored) or re-derive it per run.
- **Nice-to-have:** separate durable shared truth (scope, plan, per-stage/-unit completion receipts) from per-clone / per-session / per-agent runtime detail (current status, session info, timestamps, cursors), so that multiple developers or parallel agents working one intent don't collide on a single committed file.

SaRedfiche:
Addresses #937 (re-scoped with the maintainer to the low-severity readability/portability cleanup).

The committed `aidlc-state.md` embedded machine-local absolute paths in two fields. Neither is trusted as authoritative — the project root is re-derived at runtime (`projectRootFor` walks up to the real `aidlc/` dir; the committed value is an unreached fallback that `resolve()`s against cwd), and `Worktree Path` is a decorative breadcrumb doctor reconciles by existence — but on a cloned/relocated workspace they show another machine's layout to a human and churn committed diffs.

Changes:
- `Project Root` template + writer now emit a project-relative `.` instead of the absolute workspace path.
- `Worktree Path` writer now emits a path relative to the project root, matching the relative form already used in the audit log.
- Left the authoritative fields untouched (`Bolt Refs` fork/merge registry, `Session Resume Point` slugs, `Active Agent`, `Current Status` routing) — traced each from source; none carries a host/path/absolute value.

Verified: dist regenerated + drift-clean, typecheck, lint, and the affected tests (t68 version-sync, t70 greenfield-creation updated to assert the relative marker, t12 fixture-validation, t247 claim-sources reader) all pass. Version bumped to 2.6.122. Does not touch aidlc-orchestrate.ts.

Deferred (noted for a possible follow-up): `Last Updated` and `Pending Artifacts` are also unread but dropping them touches ~40 call sites for a pure-cosmetic timestamp, so left out of this cut.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.6.124] - 2026-08-28
+
+Stop committing machine-local absolute paths in `aidlc-state.md`. The `Project Root` and `Worktree Path` state fields are now written in project-relative form (`.` for the root; a `relative(projectDir, worktreePath)` breadcrumb for a worktree), so a state file shared across machines or checkouts no longer carries a host-specific absolute path. Closes #937.
+
+**Upgrade / scope:** this changes only what is written **from now on** — newly created or next-updated state records. Re-copying `dist/<harness>/` refreshes the tools but does **not** rewrite any existing `aidlc/**/aidlc-state.md` in a project; an existing state file keeps its previous absolute `Project Root` / `Worktree Path` until the engine next writes those fields in the normal course of a run. No migration is performed and none is required: both fields are re-derived at runtime (`projectRootFor` walks up to the real `aidlc/` dir; the committed `Project Root` is an unreached fallback and `Worktree Path` is a decorative breadcrumb), so a stale absolute value is inert until it is overwritten.
+
+* `Project Root` template + writer emit project-relative `.` instead of an absolute `${projectDir}`.
+* `Worktree Path` writer emits `relative(projectDir, worktreePath)` (forward-slashed), matching the audit log's existing relative form.
+* Regression coverage: `t78` pins a forked worktree's `Worktree Path` to the project-relative `.aidlc/worktrees/bolt-<slug>` (never an absolute path); `t70` and the `t152` Windows-portability guard pin `Project Root` = `.`.
+
 ## [2.6.123] - 2026-08-28
 
 Cursor now preserves delegated-agent attribution and reviewer-state integrity across POSIX and Windows path dialects, shell wrappers, Git configuration surfaces, filesystem traversal, and executable lookup changes. **Upgrade:** re-copy `dist/cursor/` into the project so the updated Cursor adapter, shared review-freeze parser, and version metadata replace the vulnerable hooks.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.6.123-blue)
+![version](https://img.shields.io/badge/version-2.6.124-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/knowledge/aidlc-shared/state-template.md
+++ b/core/knowledge/aidlc-shared/state-template.md
@@ -30,7 +30,7 @@ Authoritative generated views:
 - **Test Strategy**: [Minimal/Standard/Comprehensive]
 
 ## Workspace State
-- **Project Root**: [absolute workspace path]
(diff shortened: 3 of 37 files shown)


apackeer (dismissed):
Thanks for tracing these state fields and keeping the implementation narrowly scoped.

I reviewed the current head `13fb19db7ded07a85fa7100c3d23c6aaa96b10e5` against `v2` at `2fbee12fb29d2a6614b70b6f61f3cceeaf235245`. The direction is aligned with issue #937's narrowed readability and portability scope.

Three changes are needed:

1. **P2 - update the Windows portability guard.** `tests/unit/t152-windows-portability.test.ts:84` still requires the old absolute `Project Root` assertion. The smoke+unit tier therefore fails on this head.
2. **P2 - define the upgrade behavior for existing state.** Either normalize existing state files or clarify that this affects only newly created records. Refreshing `dist/<harness>/` does not rewrite existing `aidlc-state.md` files, contrary to the current changelog wording.
3. **P2 - add regression coverage for `Worktree Path`.** The current fork test strips that field before comparison, so it does not verify the new project-relative representation written by `core/tools/aidlc-state.ts:6616`.

No fixes or commits were pushed during review.

Validation performed:

- `bun run typecheck`: passed.
- `bun run lint`: passed.
- Focused `t12`, live Claude SDK `t70`, and `t68`/`t76`/`t247`: passed.
- Smoke+unit: 270 files passed and three files failed. The unrelated `t314` and `t328` failures passed when rerun alone; `t152-windows-portability` is the stable PR-caused failure.

SaRedfiche:
Thanks @apackeer — all three addressed on `b8b4f52d`, and the branch is rebased onto current v2 (2.6.123) and re-bumped to **2.6.124** with the README badge + CHANGELOG in sync, which clears the release-metadata conflict you flagged.

**P2 — Windows portability guard (t152).** Updated the meta-assertion at `tests/unit/t152-windows-portability.test.ts:84` to require the new project-relative form — `assertStateFieldPath(r, "Project Root", ".")` — matching the change already made in t70. The smoke+unit tier now passes on this head (t152, t68, t70, t76, t12 green locally; `package.ts --check`, typecheck, and lint clean).

**P2 — upgrade behavior for existing state.** Took the *newly-written-records-only* reading (no migration) as the proportionate fit for the narrowed readability/portability scope, and corrected the CHANGELOG wording to say so explicitly: re-copying `dist/<harness>/` refreshes the tools but does **not** rewrite any existing `aidlc/**/aidlc-state.md`. An existing file keeps its absolute `Project Root` / `Worktree Path` until the engine next writes those fields, and a stale absolute value is inert because both are re-derived at runtime (`projectRootFor` walks to the real `aidlc/` dir; the committed `Project Root` is an unreached fallback and `Worktree Path` is a decorative breadcrumb). If you'd prefer an explicit one-time normalizer for existing state files instead, I'm happy to add it — it just seemed wider than this cleanup's scope, so I left it out; your call.

**P2 — Worktree Path regression coverage.** Added a value assertion to the fork lifecycle test (`tests/integration/t78-bolt-worktree-lifecycle.test.ts`, L1): after `start --worktree`, the forked worktree's `Worktree Path` must equal the project-relative `.aidlc/worktrees/bolt-<slug>` and must not start with `/` or contain the absolute project dir — exercising the `relative(pd, wtPath)` writer at `core/tools/aidlc-state.ts:6616`. Previously only field presence was checked (t21/t64), so the relative representation was unverified.

No behavior change beyond those two fields; dist regenerated and drift-checked.

leandrodamascena (approved):
Thanks for addressing the previous review thoroughly.

I reviewed the current head `b8b4f52dec6c0e9428aebb5d758b82c5f12905f9` against `v2` at `518578adb3284898b6f9c073cf91c37322ba0fad`. The change remains aligned with the narrowed readability and portability scope agreed in #937.

The three previous blockers are resolved:

- The Windows portability guard now expects `Project Root: .`.
- The changelog clearly scopes the behavior to newly written values and states that no migration occurs.
- `t78` now verifies the project-relative `Worktree Path` value.

I found no remaining correctness blocker. As a non-blocking follow-up, the header comments in `t70` still describe `Project Root = projectDir`; those should be updated, and the upgrade prose could state even more directly that ordinary state updates do not normalize these existing fields.

No fixes or commits were pushed during this review.

Validation performed:

- `bun scripts/package.ts --check`: passed.
- `bun run typecheck`: passed.
- `bun run lint`: passed.
- Focused deterministic `t68`, `t76`, `t78`, `t152`, and `t247`: passed.
- The local live-SDK `t70` run did not reach its Bash tool call in this environment, so its new assertion was not exercised locally.
- Nine of ten GitHub checks are green; smoke + unit remains in progress as of this draft.

Proposed review event: **APPROVE**, after the remaining required check passes.

PR #964, opened 2026-08-28
fix: deliver stop-hook error directives once with the engine's exact message

### LINKED ISSUE #942: [Bug]: an `error` directive has no branch in the stop hook — the engine's reason is dropped and nothing is logged
### Description

When `aidlc-orchestrate next` returns a directive whose `kind` is not one the stop hook handles explicitly, `aidlc-continue-workflow.ts` falls through to a generic fallback in `directiveMessage`. Two consequences, and both are visible by reading the code:

1. **The engine's `reason` is discarded.** The fallback composes its own text from `kind` alone; whatever the engine said about *why* never reaches the operator.
2. **The instruction it gives may already be satisfied.** It says *"run `next`, do what the step prints, report the outcome, repeat until it answers `done`"* — but `next` had just answered, and in our case it answered `done` on every subsequent probe. An instruction that is already satisfied is not graceful degradation; it is a loop with no exit signal.

There is also no `ERROR_LOGGED` for this path. A directive with `kind: "error"` and `exitCode: 0` (the hook returns early on non-zero, so the exit code was zero) leaves **no trace in the audit ledger**. We know it happened only because the hook said so, and the hook does not keep what it said.

`grep -n '"error"' core/hooks/aidlc-continue-workflow.ts` → no branch.

### Steps to Reproduce

The structural half reproduces by reading: any directive `kind` without an explicit branch in `directiveMessage` produces the generic text, and the `reason` is not included.

The observed occurrence we cannot reproduce: with the workflow already `Completed`, the stop hook blocked a turn reporting *"pending step (a **error** directive for &lt;stage&gt;)"*. Immediately after:

```
next                                            → done
next --project-dir $PWD                         → done
AIDLC_STOP_HOOK_PROBE=1 next --project-dir $PWD → done   (the hook's exact probe)
```

Three consultations, including a literal reproduction of the hook's own probe with the env marker — `done` on all three, and `ERROR_LOGGED` unchanged in the shard. **We are reporting the shape, not the trigger**, and we did not find the trigger.

### Expected Behavior

An `error` directive gets its own branch: surface the `reason` the engine returned, and record the occurrence in the ledger. A transient condition that leaves no trace is unauditable by construction — the only place this error existed was one operator's context window.

### Actual Behavior

The reason is dropped, the operator receives an instruction the engine has already satisfied, and nothing is recorded. The way out was to measure instead of obey.

### AI-DLC Version

v2 (alpha)

### Release / Commit

`v2.3.0-154-gb44b9c9c` (`b44b9c9c1ad1`), consumed as a pinned vendored engine. The fallback and the absent `error` branch were read at this pin.

### AI-DLC Phase

Not phase-specific

### Platform / IDE

Claude Code

### AI Model

Claude Opus

### Environment

- OS: macOS 15.x

### Additional Context

Related in mechanism to #930, which is also about the stop hook giving an instruction the state has already met. There the marker regressed at the turn boundary; here an unhandled `kind` produces text unrelated to what the engine actually said. In both cases the operator's exit is to measure the engine directly rather than follow the hook — which is a reasonable thing for an engineer to do and a poor thing for a protocol to require.

apackeer:
The stop hook had no branch for an `error` directive: the engine's diagnostic `message` was dropped, the operator got the generic forwarding-loop instruction the engine may already have satisfied, and nothing reached the audit ledger. Following the design proposed in the issue thread, the hook now delivers each distinct error diagnostic exactly once (fingerprinted over intent, session, state digest, stage, and bounded message), appends an additive `ERROR_LOGGED` row best-effort, fails open on unknown directive kinds and all persistence/audit failures, and carries the bounded message through the Copilot active-directive evidence. Covered by new t121/t249/t250 cases plus fixture seeding so the adapter suites exercise genuine pending work. Closes #942


core/hooks/aidlc-continue-workflow.ts
```diff
@@ -247,6 +265,123 @@ function guardFilePath(projectDir: string): string {
...
+      ) {
+        return "failed";
+      }
+      if ((parsed as ErrorDirectiveRecord).fingerprint === fingerprint) {
+        return "duplicate";
+      }
+    }
+    const dir = stopHookDir(projectDir, intent ?? undefined, space);
+    mkdirSync(dir, { recursive: true });
+    writeFileSync(
```
leandrodamascena, line 310: [P2] Preserve every delivered fingerprint for the active occurrence.

This file stores only the last fingerprint. With the same intent, session, state, and stage, the sequence A -> B -> A delivers and audits A twice because B replaces A in error-directive.json. Interleaved sessions have the same problem because their session-specific fingerprints share this marker.

That violates the documented "each distinct error diagnostic exactly once" contract. Please retain a bounded set of delivered fingerprints, or use equivalent per-fingerprint markers, and add A -> B -> A plus interleaved-session regression coverage.

harness/copilot/hooks/aidlc-copilot-adapter.ts
```diff
@@ -538,6 +539,12 @@ export async function run(
       const directive: CopilotDirectiveMetadata = {
         kind: value.kind as CopilotDirectiveMetadata["kind"],
         ...(typeof value.stage === "string" && /^[a-z][a-z0-9-]*$/.test(value.stage) ? { stage: value.stage } : {}),
+        ...(value.kind === "error" &&
+          typeof value.message === "string" &&
+          Buffer.byteLength(value.message, "utf-8") <=
```
leandrodamascena, line 544: [P2] Apply the same message bound across direct and Copilot paths.

The direct hook truncates at 2,000 JavaScript code units, while Copilot rejects messages over 2,000 UTF-8 bytes. For example, 501 emoji are accepted by the direct path but rejected here, causing Copilot to discard the error directive and emit generic coordination recovery instead of the diagnostic.

Please use one shared UTF-8 byte-bounded normalization function and cover the multibyte boundary.

leandrodamascena (changes requested):
Thanks for addressing #942 and for carrying the diagnostic through the shared hook and Copilot evidence.

I reviewed the current head 2c06fc699c37 against base 2fbee12fb29d. The direction is aligned with the maintainer acceptance criteria documented in issue #942.

I found two blocking correctness and compatibility issues, detailed inline.

No patches or commits were pushed because this is an internal PR.

Validation performed:
- package parity: passed for all seven harnesses
- focused error/Copilot tests: 6 passed
- version/changelog synchronization: 7 passed
- typecheck and diff check: passed
- broader focused suite: 265 passed, 2 skipped, 1 unrelated pre-existing session-marker failure

GitHub currently reports no CI checks for this head. The PR also conflicts with the current v2 base and needs a rebase with the version, changelog, README, and generated distributions resolved and regenerated.

apackeer:
Rebased onto `main` (`a0ee4415e`; 47 commits) and addressed both P2s. The PR is now one commit with no `dist/`, version, CHANGELOG, or README changes - `dist/` is generated since #756 and release metadata belongs to the release PR.

**P2 - every delivered fingerprint is kept.** `error-directive.json` is now `{"fingerprints":[…]}`, a 32-entry FIFO under the existing audit lock + atomic write. Delivery and the `ERROR_LOGGED` row happen iff the fingerprint is absent; duplicates neither deliver nor refresh retention. Regressions in t121: A → B → A delivers A once; two interleaved sessions with the same diagnostic each get exactly one delivery; the 33rd distinct error evicts the oldest, and the evicted one delivering again is asserted as the documented bound. Negative proof: persisting only `fingerprints.slice(-1)` fails `expect(outputs[2], "A -> B -> A must not deliver A twice").toBe("")` - received a `decision:block` quoting A.

**P2 - one UTF-8 byte bound.** `boundDirectiveMessage` (`core/tools/aidlc-lib.ts`) cuts at ≤ 2,000 UTF-8 bytes on a code-point boundary and is the only normalization on both the direct hook and the Copilot `capturedDirective` path. 501 four-byte emoji bound to the same 500 on both; a 2,000-char ASCII message is unchanged. Negative proof: reverting the hook to `.slice(0, 2000)` fails the Copilot multibyte regression (directive discarded, generic recovery emitted).

Rebase notes worth a glance: main's typed-ask fields (`ask_type`, `remedies`, `guard_recovery_response`) and Plan Approval authority logic are retained beside the new `message` validation in `aidlc-lib.ts`; the stop hook's newer forwarding/wait logic auto-merged; the frozen `tests/fixtures/copilot-adapter-2.8.0` snapshot was restored to main's bytes after git's rename-aware merge tried to apply the live-adapter hunk to it.

Verified: t121 127 pass, t249 59 pass / 2 skip, t250 25 pass - 0 failures; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0.

Issue #945, opened 2026-08-25
Offer: Devin / Windsurf as an 8th harness (harness/devin/, 14-file review surface, live-verified on CLI + Desktop)


tejasavora:
## The offer

I ported AI-DLC to **Devin** (which is also what Windsurf Desktop runs) as a harness following the existing `harness/<name>/` contract, and verified it live on both Devin CLI and Devin Desktop. It has never been offered here. Asking before rebasing, because that part is real work and only worth doing if you want it.

If the answer is no, that's a completely fine answer — I'll keep it as a customer-facing package and stop tracking upstream for it.

## Review surface

314 files changed sounds alarming; almost all of it is the generated `dist/` trees the packager produces and this repo commits by design. The **authored** surface is:

```
 core/tools/aidlc-includes.ts                     |  ~20 +
 core/tools/aidlc-lib.ts                          |    3 +-
 core/tools/aidlc-runtime-paths.ts                |    2 +-
 core/tools/aidlc-state.ts                        |    1 +
 core/tools/aidlc-utility.ts                      |   17 +-
 harness/devin/manifest.ts                        |  101 +
 harness/devin/emit.ts                            |  194 +
 harness/devin/onboarding.fills.ts                |  117 +
 harness/devin/hooks/aidlc-devin-adapter.ts       |  168 +
 harness/devin/skills/aidlc/SKILL.md              |  246 +
 harness/devin/skills/aidlc/question-rendering.md |  155 +
 harness/devin/dot-gitignore                      |   64 +
 tests/harness/harness-matrix.ts                  |   26 +-
 tests/unit/t280-devin-packaging.test.ts          |  245 +
 14 files changed, 1357 insertions(+), 4 deletions(-)
```

The `core/` edits are ~24 lines total and all additive `if (harness === ".devin")` branches in the same shape the other harnesses already use — the includes branch is explicitly "same surface and same rewriter as Copilot; only the harness differs." No existing harness path changes behaviour.

`t280` (15 tests) covers the packaging contract, and `bun scripts/package.ts --check` was clean across all 8 trees at the time.

## Field findings you may want regardless of the answer

These came out of testing against the real product, and they're the kind of thing that's expensive to rediscover:

[...]

PR #968, opened 2026-08-28
feat: Devin harness support - one dist for Devin CLI + Devin Desktop

tejasavora:
Adds **Devin** as an eighth harness, so `bun scripts/package.ts` emits `dist/devin/` alongside the other seven. Follow-up to the offer in #945 — opened as a branch so you can judge the diff rather than a description. If you'd rather not carry an eighth harness, close it; nothing else in the repo depends on it.

**One tree serves both surfaces**: Devin CLI and Devin Desktop's "Devin Local" agent.

## Why one harness and not two (the kiro/kiro-ide question)

Worth stating up front, because it's the first thing I'd ask. The criterion this repo actually applies for splitting is **mutual format unreadability at a shared path**, not CLI-vs-IDE:

- `harness/kiro/manifest.ts` and `harness/kiro-ide/manifest.ts` **both** declare `harnessDir: ".kiro"` and diverge on **64 files** (`.kiro.hook` vs `.json`, agent JSON vs steering). They cannot coexist in one directory.
- Copilot (CLI + VS Code agent mode) and Cursor (IDE + CLI) each ship **one** tree.
- Devin Local reads the same `.devin/` paths with the same hook contract and the same subagent model, so `dist/devin-desktop/` would be a byte-identical copy of 275 files.

`docs/roadmap.md:226-231` also shows you're working to *un-split* kiro (#775), so a second Devin tree would push against your own direction. Everything CLI-vs-Desktop lives in prose in the guide and `AGENTS.md` instead.

## Cost to carry: the cheapest harness in the repo

| harness | runtime adapter | lines | authored files |
|---|---|---|---|
| claude | none needed | — | 9 |
| **devin** | **yes** | **168** | **8** |
| opencode | yes | 704 | 9 |
| codex | yes | 778 | 7 |
| kiro | yes | 1,047 | 32 |
| copilot | yes | 1,391 | 7 |
| kiro-ide | yes | 2,029 | 31 |
| cursor | yes | 3,079 | 17 |

Claude needs no adapter because core's hooks are *written in* Claude's shape — Claude is the baseline, not a template the others match. `emit.ts` is likewise established: codex, copilot, opencode and devin all have one.

Devin's adapter maps tool names, normalises `DEVIN_PROJECT_DIR` → `AIDLC_PROJECT_DIR`, denies the question picker (below), dispatches to the shared core hook, and passes everything else through untouched — exit codes and stderr included. It parses no paths and no shell commands, so unlike cursor's it owns no security-relevant logic of its own.

Core registration is additive only. Everything under `dist/` is packager output.

## Why an adapter is needed

Devin uses Claude Code's hook JSON *shape* and exit-code convention (`0` continue, `2` block with reason on stderr), so core hook **bodies** need no change. What differs is tool names — lowercase snake_case (`exec`, `edit`, `run_subagent`, `todo_write`).

Renaming only the matchers is insufficient, because three core hooks compare `tool_name` **internally**:

| hook | comparison |
|---|---|
| `aidlc-review-freeze.ts` | `toolName === "Bash"` |
| `aidlc-reviewer-scope.ts` | `"Grep"` / `"Glob"` + a 10-name allowlist |
| `aidlc-state-transition-guard.ts` | `parsed.tool_name !== "Bash"` |

Those three would load, match, and **silently no-op**. `emit.ts` also fails the build if a matcher names a tool absent from Devin's documented list.

The three read/write/shell guards now share **one** matcher set, mirroring your single matcher string on Claude — narrowing any of them to just the shell tool is a silent hole, since `review-freeze-command.ts:1023` and `plan-approval-guard.ts:95` both gate on `WRITE_TOOLS`.

## Question rendering is a correctness constraint here

Devin renders structured questions as **numbered prose**, and the PreToolUse guard denies `ask_user_question` while a workflow is `Running` (failing **open** on any throw, missing state, or non-Running status). Three measured facts force this, and I'd rather show the reasoning than just the diff:

1. **`"^ask_user_question$"` is build-rejected** by `emit.ts`'s own matcher guard — the tool is not a documented matchable name. Claude's `PostToolUse matcher="AskUserQuestion"` presence wiring cannot be reproduced.
2. A picker answer arrives as a **tool result**, so it never fires `UserPromptSubmit` and never mints `HUMAN_TURN`. Answer and approval logging would refuse the selection and the question would be re-asked.
3. **Devin's `PostToolUse` fires on a *skipped* `ask_user_question`** (CLI changelog v3000.3.22). A matcher-free `PostToolUse` arm *would* pass the guard and `aidlc-record-human-turn.ts` already reads `tool_response` — so "just wire the picker and the mint, like Claude does" looks free. It isn't: it would **forge `HUMAN_TURN` for a question no human answered**, into an append-only ledger. Your dual wiring is safe only because `AskUserQuestion` has no silent-skip path.

The annex is authored for Devin rather than inherited. It previously shipped **byte-identical to Claude's**, titled "Claude Code harness annex", binding every question to a tool Devin doesn't have — a class of defect the parity tests structurally cannot catch, since they enforce sameness. `t239` now carries negative pins for it.

## Model selection moved into your shared table

devin is a real `TIER_PROJECTIONS` column (`tierFlavor: "devin"`) rather than a harness-local post-patch, so "Tune here; every harness moves in lock-step" is now true for it too.

Its column types `model` as **non-nullable** — the only one that does. Every other harness reads an omitted `model:` as "inherit the session model"; Devin does the opposite, falling back to the **default subagent model**, explicitly *"not the parent's model"* ([docs](https://docs.devin.ai/cli/subagents)), routed to SWE-1.6. Omitting the key would silently downshift every judgment persona, so the type makes that unrepresentable. Judgment pins `opus`, balanced/templated pin `sonnet`, no tier emits an effort key. `emit.ts` now **validates** the projection instead of performing it.

I know this diverges from how cursor/kiro/copilot ship null — it has to, and the trade is explicit: a user on a Gemini session gets Opus judgment personas. Happy to revisit if you'd rather accept the downshift.

Personas also carry an `allowed-tools` allowlist omitting delegation, and core's `disallowedTools: Task` is **dropped** rather than shipped inert — Devin ignores it silently, and an inert key that reads like enforcement is worse than none. `maxTurns` gets the same treatment plus the codex-style citation rewrite.

## Known limits, measured

- **No `SubagentStop`.** Subagent completion rides `PostToolUse` on `run_subagent`/`read_subagent`: foreground delegates are audited, backgrounded ones are not. Reviewer read-scope is unaffected — it identifies a reviewer from the on-disk dispatch record, not a subagent event.
- **No `PreCompact`.** `PostCompaction` covers it, and costs little: that hook vetoes nothing on *any* harness and reads on-disk state, not the conversation. Only the audit timestamp lands later.
- **No `Notification`** — but no harness wires one and the framework never uses it, so I'm not counting it.
- **No statusline.**
- **Restricted Mode (Devin Desktop)** disables every agent and hook silently — indistinguishable from a bad install, so it's the first thing the guide says to check. Not a CLI concept.
- **`AGENTS.md` size.** The CLI changelog (v2026.4.17-0) says an oversized always-on file truncates with a path hint at 32 KiB. I'd treat that as a working budget rather than a verified constant — current docs state no limit, and I could not reproduce truncation through `devin rules show`, which returned a deliberately oversized 60 KB file in full. The shipped file is ~19.9 KB and t331 asserts the ceiling either way.
- **Devin imports Claude Code config by default, including hooks.** The hooks overview's "Where Hooks Live" lists `.claude/settings.json`, `.claude/settings.local.json`, `~/.claude.json`, `~/.claude/settings.json` and `~/.claude/settings.local.json` as hook sources "loaded when `read_config_from.claude` is enabled (the default)" — the read-config-from reference table omits hooks and is incomplete. **Consequence:** a project carrying both this install and the AI-DLC Claude install loads both hook sets and writes every audit event twice. The guide says one harness per project. This branch does not ship `{"read_config_from": {"claude": false}}` — that felt like the user's call.
- **Devin Cloud is untested and out of scope.** Cloud runs one active skill at a time, has no repo hooks and no subagents — three assumptions this distribution relies on. Rather than caveat it, every Cloud claim is absent.

## Verified live on Devin CLI 3000.6.7

This was the outstanding caveat and it's now closed. A trusted scratch workspace with the shipped `dist/devin/` tree, probed against the real binary:

| probe | result |
|---|---|
| `devin doctor --json` | `ok: true`, **14 personas load** |
| `bun .devin/tools/aidlc-utility.ts doctor` | **42 passed, 0 failed** |
| `devin skills list` | **42 skills**; 37 user-only + 5 model-invocable (orchestrator + 4 capability skills) — `triggers: [user]` is doing its job |
| `devin rules list` | `AGENTS [Standard] always-on` |
| picker deny, `Status: Running` | **blocks** — emits `{"decision":"block","reason":…}` at exit 0 |
| picker deny, `Status: Complete` | **fails open** — silent, exit 0 |
| picker deny, no `aidlc/` tree | **fails open** |
| non-picker `PreToolUse` | passes through untouched |
| `PostToolUse` `exec` → `rebuild-stage-graph` | dispatches (the transposition fix, on the wire) |
| `PostToolUse` `todo_write` → `sync-workflow-state` | dispatches |
| `PostToolUse` `run_subagent` → `log-subagent` | **writes a real `SUBAGENT_COMPLETED` audit event**, correctly attributed to `aidlc-developer-agent` |

One expected-noise item worth stating so it isn't read as a defect: `devin doctor` reports `ok: true` **with 14 `CFG005` warnings** — every persona surfaces `unsupported frontmatter key(s) ignored: display_name, examples`. I tried dropping those two for Devin and it regresses *your* schema check (`aidlc-utility.ts doctor` → 1 failed, "missing required frontmatter: display_name"). Devin's notice is informational and the host continues, so the keys ship and the guide documents the expected output. Happy to take a different call if you'd rather core stopped requiring them.

Also measured while probing: on a **Free** Devin plan only SWE-1.6 is available, so the `model: opus` pins fall back. The guide already says use an Opus-class model for the orchestrator; that now has a concrete reason behind it.

## Permissions

`.devin/config.json` ships a scoped `Exec` allowlist for the framework's own `bun .devin/tools` / `bun .devin/hooks` commands. It replaces advice to run with `--permission-mode accept-edits`, which auto-approves every workspace edit and was strictly broader than any peer's posture. It carries **only** `permissions`, so merging into an existing config is a single key. `Exec(prefix)` matches on whole words; the trailing-slash form is undocumented, so the slash is omitted.

## Test surface brought to peer parity

devin had the thinnest test surface of any harness. Closed in this branch:

- **`t332-devin-adapter`** (10 tests) — the adapter had none, where 3 peers do. It spawns the *real emitted* adapter (this one is a top-level script with no `run()` export, so there is nothing to stub) and pins: the full `TOOL_MAP`, the picker deny in Devin's dialect, fail-open on non-Running state / absent state / malformed stdin / unknown subcommand, non-picker passthrough, and that `SUBAGENT_COMPLETED` actually reaches the audit trail.
- **`tests/fixtures/devin-hook-payloads/payloads.json`** — 16 payloads shaped against 3000.6.7, matching the 4 peers that ship a corpus. Test 1 asserts the corpus covers every event the emitted `hooks.v1.json` registers, so the fixture cannot drift behind the wiring.
- **`t221` devin arm** — reviewer-scope registration was unasserted. The arm also pins that `review-freeze` and `state-transition-guard` share the *same* matcher set, which is the specific hole that existed when two of the three were narrowed to `^exec$`.
- **`MIN_DEVIN` version floor (3000.3.22)** in the doctor. Devin **auto-updates**, and the exit-2 block channel is version-gated — below that floor every guard loads, matches, and cannot refuse anything, with no diagnostic. codex and copilot both carry a floor. Devin's is **advisory when the binary is absent**, because a Desktop-only install legitimately has no `devin` on PATH; `Bun.spawnSync` *throws* rather than returning a failure there, so it is wrapped (codex's equivalent is unguarded because codex is a hard prerequisite).

## Tests

`bun tests/run-tests.ts --smoke --unit` (274 files) and `--integration` (115 files), against the same tiers on unmodified `v2` at the same commit:

| test | clean `v2` | this branch | attributable here |
|---|---|---|---|
| `t255-workspace-sync` | FAIL | FAIL | no |
| `t276-cursor-adapter` | FAIL | FAIL | no |
| `t314-source-freshness-receipts` | FAIL | FAIL | no |
| `t325-team-unit-claims` (integration) | FAIL | FAIL | no |
| everything else (385 files) | PASS | PASS | — |

**Zero failures attributable to this change.** Those four are now filed as **#989** with the clean-`v2` reproduction, so this table isn't carrying someone else's bugs unexplained. Two further files (`t272`, `t275`) failed once under sustained machine load with 5-second budgets blown at 8s, pass in isolation, and did not recur — flagging that rather than hiding it.

One methodology note that cost me time and may save yours: **run the integration tier serially when establishing a baseline.** Under `--parallel 8` I saw `t188-plugin-compose` and `t327-team-dispatcher` fail as well; both pass serially, so parallel integration results can't be trusted for attribution.

Rebased onto `82d2e304`. Note the version: you bumped to `2.6.124` while this was open, so per the changelog conflict-trap in `AGENTS.md` this re-bumps to **2.6.125** and renames its heading.

**All four tiers run**: smoke+unit (274), integration (115), and **e2e (79 files, PASS)**. Also green: `package.ts --check` (8 trees) and `bun run check` (703 files).

**Two hardcoded rosters had silently stopped covering new harnesses, and that is *how* the Claude annex shipped** — `tests/smoke/t250` pinned a six-name list omitting devin **and copilot**, and `tests/integration/t305` hardcoded `toHaveLength(7)`. Both now derive from the discovered matrix, so copilot gained coverage too.

## One bug this found in the wiring, worth a look

`devin`'s two `PostToolUse` matchers were **transposed**: `^exec$` → `sync-workflow-state` and `^todo_write$` → `rebuild-stage-graph`, where Claude and codex both pair plan-update → sync and shell → rebuild. Both hooks therefore loaded, matched, and returned 0 — `runtime-graph.json` never recompiled after a transition and Current Stage never auto-synced. Fixed, with a comment recording which payload field each core hook gates on, because nothing compared the wiring against `docs/reference/06-hooks-and-tools.md`.

## One thing I noticed and did not touch

`core/tools/aidlc-runner-gen.ts:612`'s `if (harnessName === "cursor")` is unreachable — line 603 already returns for cursor. It predates this branch (devin's arm is at 609); flagging it so it's clear I saw it rather than introduced it.

## Docs

New chapter `docs/guide/harnesses/devin.md`, plus the roster updates `t239` requires and a sweep of the 23 doc files that enumerated Copilot and never mentioned Devin — including `09-porting-to-a-new-harness.md`, the one page a new porter reads. Devin now appears in 18 doc files (was 6). Tier prose needed semantic edits rather than name insertion, since devin pins models where Kiro/Cursor/Copilot inherit.

---


wilmarques:
Hi, @tejasavora . I just saw this PR. I'm also working on it: https://github.com/awslabs/aidlc-workflows/pull/996 . I'll do a quick comparison to see if we can merge, what do you think?

tejasavora:
@wilmarques and I found we'd both built a Devin harness independently (#996) — same conventions, even the same test and fixture filenames, with no shared history. Full comparison and merge proposal in https://github.com/awslabs/aidlc-workflows/pull/996#issuecomment-5498256113.

**Short version for maintainers: #996 should be the base**, and I'm folding this branch's differentiated pieces into it rather than asking anyone to choose between two competing harnesses.

Their branch has three live end-to-end runs (77 evidence files) plus a real bug find — Devin's `PostToolUse` delivers `tool_response` as an **object**, not a string, which broke `ask_user_question` answer recording and the Plan Approval receipt triple. **That bug also exists in the codex adapter**, so it's worth attention independent of Devin.

I'm also withdrawing a decision from this PR. I bound structured questions to numbered prose and denied Devin's native picker, on the reasoning that `PostToolUse` fires on a *skipped* question and would forge `HUMAN_TURN`. #996's `hasExplicitHumanSelection` handles that properly (it requires a non-empty `answers` array matching offered option labels, so a skip records nothing), and my stated blocker — that `^ask_user_question$` is build-rejected — was true of a guard **I wrote** in my own `emit.ts`, not of Devin. Their binding is better and I'd rather say so here than leave a wrong rationale in a merged PR body.

What I'm carrying over: **Devin Desktop support** (#996 is CLI-scoped, and its doctor currently hard-fails on a Desktop-only install where no `devin` binary exists), the compiled-binary release gates and `runtime-paths` fallback, persona/skill frontmatter hygiene (`disallowedTools` and `maxTurns` ship inert on Devin; runners are model-invocable), two hardcoded test rosters that had silently stopped covering new harnesses — one of which was omitting **copilot** — and the `read_config_from.claude` double-audit disclosure.

Leaving this open until those land on their branch, then I'll close it in favour of #996.

tejasavora:
## Maintainers: please do not merge this ahead of #996

A note against an accident of timing, not a change of position.

@wilmarques and I each built a Devin harness independently ([comparison](https://github.com/awslabs/aidlc-workflows/pull/996#issuecomment-5498256113)). We agreed that **#996 is the trunk and my differentiated pieces layer onto it** — their live provenance (3 e2e runs, 77 evidence files, including an interactive run that surfaced a bug print mode could not) is irreplaceable, and their question binding and `tool_response`-object find are better than mine. The first layer is already merged into their branch (wilmarques/aidlc-workflows#1: release-engineering gates, `.devin` resolution fallback, derived test rosters).

Right now this PR is MERGEABLE and #996 is CONFLICTING. That difference is **not** a signal of readiness — #996's conflict is purely release metadata (CHANGELOG, README badge, `aidlc-version.ts` and its 7 generated copies), with nothing substantive in it. Merging this one because it happens to be green today would reverse a decision we made deliberately, and would strand the evidence base that makes #996 the better foundation.

I'd rather this sat open until #996 lands, then carry over what's still missing as reviewable layers. If you'd prefer the opposite order, say so and I'll adapt — but I don't think it should happen by default.

For the record, this branch is current: rebased onto `main` after #997, at 2.7.2, all four tiers clean apart from the pre-existing failures tracked in #989.

One correction to that issue while I'm here: of the four failures I reported, `t255` and `t276` reproduce on clean `main`, but `t314-source-freshness-receipts` fails because of **my** environment — it refuses to merge when git-lfs checkout filters are configured globally (`filter.lfs.process` / `filter.lfs.smudge`). That one is a test environment assumption, not the defect I implied. #996 carries `docs/rfcs/pre-existing-test-failures-fix-plan.md`, which actually fixes the real ones; #989 should close as superseded when it lands.

PR #888, opened 2026-08-24
fix: bind reviewer retries to immutable review evidence

apackeer:
## Summary

- declare one explicit `review_artifact` for every reviewer-bearing stage and bind the exact dispatched artifact bytes plus request-time workspace source
- validate one terminal reviewer-owned `## Review` appendix from a coherent file snapshot, rejecting arbitrary suffixes, ownership ambiguity, replacement races, and injected renderer markers
- limit pending retries to one immutable redispatch, modernize valid legacy pending requests, and invalidate malformed requests without consuming their ordinal
- update receipt consumers, stage metadata, protocol documentation, generated harness distributions, and compatibility tests

## Verification

- local `t271-review-iteration-ceiling|t279-reviewer-turn-budget`: 47 passed
- Windows `t271-review-iteration-ceiling|t279-reviewer-turn-budget`: 47 passed
- `bun run typecheck`
- focused Biome check
- coverage registry check
- `bun scripts/package.ts --check`
- `git diff --check`

## Notes

- no full-suite run was performed; verification was limited to the affected review, retry, schema, receipt, and packaging surfaces
- final adversarial closure review reported no findings and `MERGEABLE`

Issue #977, opened 2026-08-30
[Bug]: CC - express scope - workflow's own end-of-turn hook silently reverts the active step back to a rules-loading state


johnwhitegm:
### Description

Ran into another issue but with CC this time... maybe related to "Express Scope?" I'll open both issues (the one just before with Cursor, and current CC) in github for it.

CC's diagnosis:

The defect, in four lines
1. The question record and fingerprint must be created under directive epoch E.
2. Recording your approval requires a human turn that comes after that record.
3. Any human turn invalidates the continuation token, forcing a fresh next → epoch E′.
4. Under E′, the record and fingerprint from E are stale. Back to 1.

Every ordering I tried — and I tried all of them — lands back at step 1. There is no sequence of legal calls that satisfies all four constraints, because the human turn that authorizes the approval is the same event that invalidates it. That's worth reporting to whoever maintains the v2 branch: express scope, zero-Unit, --stage-level plan approval cannot be completed. The per-unit path likely doesn't hit it, which is probably why it shipped.

How should I get your transcript cleaner built?
❯ 1. Park it, then build (Recommended). Park the workflow so the guard stands down, then write clean_transcript.py, the 24-test suite and README.md here. Reversible, no restart, nothing disabled.
  2. Guard off, then restart. I add AIDLC_DISABLE_PLAN_APPROVAL_GUARD=1 to .claude/settings.local.json; you restart Claude Code and run /aidlc --resume so the stage completes inside the workflow.
  3. Stop and report upstream. Change nothing. I write the full reproduction into the stage diary so you can file it against the v2 branch.
❯ 4. Type something.

### Steps to Reproduce

Use Claude Code and got through express scope with intent such as: build me a super simple and minimalistic python application that accepts a single large transcript file and produces a losslessly cleaned transcript.

### Expected Behavior

should not error

### Actual Behavior

Confirmed the mechanism. The workflow's own end-of-turn hook silently reverts the active step back to a rules-loading state — it has swallowed 5 failures, the last at 21:42:14, exactly when the developer's writes started getting refused. Because I handed the work to a background agent, my turn ended, that hook fired, and the agent was blocked from writing anything. Your approval was never the problem.

The guard's own remedy is to re-derive the step and use it immediately, so let me do that with no handoff in between.

  Ran 2 shell commands

run-stage is live again. Probing whether the guard now permits workspace writes before committing to anything.

  Ran 1 shell command

[...]

PR #978, opened 2026-08-30
fix: state-aware guard refusals, shared attempt reducer, guard-recovery loop breaking, and directive-time preflight

apackeer:
# State-aware guard refusals, one shared attempt reducer, guard-recovery loop breaking, and directive-time preflight

## Problem

The engine's integrity guards (review freeze, summary confirmation, reviewer
preconditions, recovery budgets, lifecycle validators) each refuse
independently and each wrote its own recovery prose. Compositions of two
correct guards could produce a dead end: one refusal prescribes a remedy the
other guard refuses, and nothing names an exit that actually executes. A
field incident wedged a workshop session for over an hour in exactly this
shape; #903 and #908 fixed the known instances and the narration. This PR
fixes the class, and then makes the engine stop steering into refusals at
all.

## What changed

**1. Executable refusals (one pure evaluator).** `evaluateGuardRefusal`
computes recovery remedies from live state: checkbox lifecycle (or the team
unit-gate ledger when a team gate is in play, resolved through one shared
`teamUnitGateStatus` with the unit-end gate anchor), review budget, the
recovery slot, summary/review/source coverage, pending/repair/next review
progress, human authority, and autonomous-Bolt context. Every remedy is
marked executable from its own guards' preconditions, so a message can never
prescribe an action the engine would refuse: Request Changes remains
action-only until a separate human turn supplies exact feedback, with the
Unit and correct gate stage preserved for team gates; command-bearing
remedies are concrete packaged `bun <harness>/tools/...` invocations with no
placeholders; verdict recording and retries are offered only while a matching
request is pending; a below-cap NOT-READY routes to
apply-repairs-then-next-iteration; review requests require current summary
confirmation; and an unresolvable unit-end gate names its cause
(`no-active-gate-stage`) and the Scope repair. Gate opening, revision,
completion, review recording, summary-evidence checks, and the review-freeze
hook all route refusals through the evaluator.

**2. One shared attempt reducer.** `reviewAttemptWindow` (from the per-Unit
continuity work) becomes the single audit-scan kernel; review
budget/request accounting, the swarm worktree review check, unit candidate
coverage, and the review-brief invalidation frontier are named projections
of that one snapshot instead of five independent scans with five event sets
and five tie-break rules.

**3. Loop breaker.** Guard refusals persist a per-stage/unit streak record
keyed on a normalized guard-state signature (lifecycle, attempt floor,
budget, recovery slot, pending/repair/next review progress, coverage,
resource fingerprints, and the SET of refusal codes, so an alternating
two-guard cycle registers as one streak). A third identical refusal stops
repeating prose and returns a typed `guard-recovery` ask whose options are
exactly the executable remedies. Presentation only: it never records a
decision, restores a budget, or mints a receipt; persistence failure
under-counts and never relaxes a guard. All seven harness orchestrators
carry one byte-identical rendering clause for the new ask.

**4. Directive-time preflight (prevention).** The engine consults the same
guard predicates BEFORE instructing or spawning an action: report routes
(solo and team sequences), `next` gate re-entry, and per-unit review routing
run a pure, read-only `guardPreflight` that reuses the exact enforcement
check functions (one predicate, two consumers; a module flag turns any stray
`error()` into typed uncertainty). A blocked instruction is replaced by the
guard-recovery ask or the executable next step; the happy path is
byte-unchanged (pinned). Preflight is fail-open on any uncertainty, never
writes state or streaks, and enforcement remains the final authority at
execution time, so a race after preflight still lands on the original guard.

**5. Composition hardening from the class analysis.** Team unit rejection
requires the exact "Request Changes" choice; unit revise validates the unit
gate ledger instead of the global checkbox; `pendingSummaryDecision` gains
workflow/jump/stage-start/rejection floors; pending-decision checks are
shard-aware and an ambiguous same-second answer at an open gate now consumes
the human turn (fail closed: a turn spent on an ordinary answer can never
satisfy a later approval's presence check); source-freshness refusals
distinguish an unbindable boundary from real drift; unattended mode names
its real exit; a crashed Bolt worktree's error points at the merge path.

## Liveness is tested, not argued

`tests/unit/t330-guard-deadlock-liveness.test.ts`: a 324-state bounded
enumeration asserting every refusal advertises at least one executable
remedy, each cross-checked against its state predicate; plus preflight
equivalence (preflight refusal code equals the real attempt's, with
byte-identical audit and streak state before the attempt), team/unit-end
anchor routing, three-state review-progress mapping, tie fail-closed
behavior, alternating streak accumulation, and directive validation. Real
fixtures in t324/t278/t186/t188 drive the production paths end to end.

## Review provenance

Implemented and reviewed by separate agents with independent context, five
independent defensive review rounds against an eight-property contract (authority
preservation, preflight purity/fail-open, single-predicate consistency,
loop-breaker safety, suspension boundedness, refusal-surface consistency,
cross-harness integrity, enforcement-behavior stability). Twelve earlier
findings were raised and closed: among them an authority defect where an
ambiguous cross-shard tie could leave a human turn reusable for approval,
team-gate remedies that omitted the Unit or the unit-end gate anchor,
remedies advertised in states their own admission would refuse
(retry/verdict/review-request under invalid summary evidence,
record-verdict for already-recorded below-cap verdicts), and a streak
signature that conflated ordinary pending requests with the recovery slot.

The requested-changes review at `d9a16851c` raised three further blockers.
All are addressed: rejection feedback is bound to a separate exact human
response before `GATE_REJECTED`; every emitted command is directly executable
from a packaged harness and placeholder-free; and team pipeline revise uses
one shared predicate in preflight and both enforcement snapshots. A follow-up
defensive review also closed the open-set harness-directory validation case.
Local re-review is clean; GitHub re-review remains pending.

## Authority invariants (unchanged, by construction and by test)

No auto-rejection, no receipt minting, no budget or recovery-slot
restoration on any path, including preflight and the loop breaker. Duplicate
evidence remains idempotent only for identity-matching evidence. Cross-shard
same-second ties remain fail-closed, now uniformly in the consume direction
at the open-gate consumer. Guard-recovery feedback extends the existing
state-bound active-directive marker with response digests; it stores no
plaintext feedback and cannot be satisfied by the remedy-selection turn.
The team-unit reject path is stricter than before (exact-choice and exact
feedback checks added).

## Test evidence

Final smoke + unit + integration run covered 389 files and 9,612 assertions.
It completed with five local test-maintenance reds: one coverage-registry
refresh, three files pinned to the retired one-turn recovery wording, and one
5-second load-sensitive timeout. The coverage registry and wording pins were
updated and all affected files passed together; the timed-out pipeline test
passed when rerun alone. Every production and integration journey passed in
the full run, including `t72-stage-reverse-engineering` (18.9 minutes) and
`t276-cursor-adapter`.

Additional final checks:

- `bun run check`: package parity, all TypeScript configs, and Biome pass.
- Combined guard/Stop/Copilot authority slice: 8 files, 307 assertions, pass.
- Affected maintenance rerun: 5 files, 217 assertions, pass.
- Isolated `t315-pipeline-link-receipts`: 24 assertions, pass.
- Pre-push sensitive-data scan: 120 outgoing files, clean.

## Notes for reviewers

- `resetGuardRefusalStreak` is exported for tests; production reset is
  implicit (state change flips the signature; sanctioned boundaries change
  the reset token).
- Self-remedy suppression in preflight matches remedies by wording prefix; a
  future wording change degrades it to one redundant (harmless) ask, never a
  misroute; t115/t330 pin current behavior.
- Out of scope, deliberately: authorization IDs and moving review records
  out of the primary artifact (the embedded-appendix protocol was just
  hardened by #888; that migration belongs to its own RFC).

## Contributor statement


apackeer:
## Update: Plan Approval / Stop-hook cycle fixed

Commit `d9a16851c` adds the fix for the Code Generation approval cycle reported in #977 and hardens the broader turn-end wait handling.

### Root cause

The turn-end Stop hook consulted `aidlc-orchestrate next` before recognizing that the current turn was legitimately waiting for a human. That probe could replace the active `run-stage` directive with `load-steering`, rotate the Code Generation authority epoch, and clear the protected Plan Approval challenge before the human response was recorded.

This is distinct from #976: that issue is Cursor project hooks not loading at all; #977 occurs with hooks running, but the Stop probe invalidating an existing approval challenge.

### Changes

- Evaluate legitimate waits before the Stop hook's internal `next` probe: approval/revision gates, Plan Approval, summary confirmation, engine-issued questions, the post-skeleton autonomy choice, logged questions, compose choices, background subagents, and conversational turns.
- Require a pending Plan Approval wait to match the exact current session, target, directive epoch, source floor, fingerprint, and questions file, with no response already consumed.
- A Cursor IDE-owned challenge therefore cannot park a later Cursor CLI session indefinitely. The CLI is nudged to reissue the challenge for its own session, after which its human response records normally.
- Persist engine-issued `ask` directives as state-bound markers and consume them exactly once on the next real human turn.
- Preserve V2 interleaved unit-major directives without rewriting `aidlc-state.md`, because even a metadata-only write changes the marker-bound state digest and invalidates Plan Approval authority.
- Prevent Kiro IDE audit-tail synchronization from replacing an interleaved unit-major directive.
- Fix `git cat-file --batch` source-listing reads when a header crosses the 64 KiB buffer boundary; the prior borrowed buffer view could be overwritten on refill and make Bolt/Swarm worktree creation fail depending on source-file alignment.

### Regression coverage

- Cursor IDE challenge -> switch to Cursor CLI -> CLI challenge reissue -> `beforeSubmitPrompt` -> protected approval receipt.
- Express stage-level Plan Approval survives Stop until the human responds.
- Blank Plan Approval markdown without a protected challenge cannot suppress recovery.
- Engine-issued questions remain stable across Stop and are consumed by a real human turn.
- Unit-major audit/status synchronization preserves the active directive and approval runtime.
- Compiled-binary Bolt and Swarm worktree re-entry passes with the boundary-crossing source layout.

### Verification

- `bun run check`: pass (package parity, all TypeScript configs, Biome).
- Full smoke + unit + integration run: 385/389 files passed. The four reds were rerun individually and all passed: `t238-build-binaries`, `t276-cursor-adapter`, `t141-enterprise-scope-routing`, and `t72-stage-reverse-engineering`.
- `t314-source-freshness-receipts`: 117 assertions pass.
- `t68-version-changelog-sync`: 7 assertions pass.
- Independent defensive review: no blocking findings.

CI has been triggered on the updated head.

core/tools/aidlc-lib.ts
```diff
@@ -16733,59 +17471,662 @@ export function parseCheckboxes(content: string): CheckboxLine[] {
...
+  if (state === "in-progress" || state === "awaiting-approval") {
+    const reportStage =
+      input.teamGate?.resolved === true ? input.teamGate.gateStage : input.stage;
+    const unitArg =
+      input.teamGate?.resolved === true && input.unit
+        ? ` --unit "${input.unit}"`
+        : "";
+    const command =
+      `aidlc-orchestrate.ts report --stage "${reportStage}"${unitArg} --result rejected ` +
+      '--user-input "Request Changes" --reason "<requested changes>"';
```
leandrodamascena: **[P1] Do not publish a rejection command with placeholder feedback**

Every harness now instructs the conductor to execute a selected guard-recovery command exactly. This command hard-codes `--reason "<requested changes>"`. After the human selects this recovery option, the answered-widget or prompt-submit seam records a fresh HUMAN_TURN, and the rejection validator accepts the placeholder because it is nonblank, is not a recognized non-answer, and has no self-attribution marker. The result is a real GATE_REJECTED and STAGE_REVISING transition whose feedback was never supplied by the human, which also resets the bounded review attempt. Keep this remedy action-only until exact feedback has been collected, or add a second human-input route that binds the exact response before constructing the command. Add an end-to-end assertion that the audit Reason equals the submitted human text and never equals a template placeholder.

leandrodamascena: **[P1] Emit commands that the harness contract is able to execute exactly**

The seven SKILL.md files require exact execution of `remedy.command`, but `aidlc-orchestrate.ts` is neither installed on PATH nor directly executable. The distributed files are mode 100644 and have no shebang, so this value fails with command-not-found or permission-denied instead of performing recovery. The autonomous Bolt remedy has the same defect, and `/aidlc --scope <scope>` also publishes an unresolved placeholder as an exact command. Emit a harness-qualified invocation such as `bun ${harnessDir()}/tools/aidlc-orchestrate.ts ...`, and publish a command only after every required value is concrete. Exercise each command-bearing remedy from a packaged harness in tests instead of checking substrings only.

core/tools/aidlc-state.ts
```diff
@@ -4535,40 +4868,175 @@ function verifyReviewerPreconditionForUnit(
...
+          options.action === "complete"
+            ? "complete"
+            : "present-approval-gate",
+        );
+        if (
+          options.action === "present-approval-gate" ||
+          options.action === "revise"
+        ) {
+          for (const gateStage of team.stages) {
+            verifyPipelineLinkPrecondition(pd, gateStage);
```
leandrodamascena: **[P2] Keep team revise preflight equivalent to team revise enforcement**

For a team-owned per-unit stage, `guardPreflight(..., { action: "revise" })` requires pipeline-link evidence for every gate stage here. `handleRevise` does not run `verifyPipelineLinkPrecondition` in either of its team branches, before the sensor pass or inside the audit lock. The stage schema permits `for_each: unit-of-work` with `mode: pipeline`, including plugin stages, so this produces a reachable state where preflight refuses a revise transition that enforcement accepts. That breaks the stated fail-open property and the one-predicate/two-consumers contract. Add the same pipeline check to team revise enforcement or remove it from revise preflight, then pin the plugin-shaped case with an equivalence test.

leandrodamascena (changes requested):
Requesting changes. The guard-state reducer, Stop-hook ordering, generated-tree parity, and focused test profiles are internally consistent at d9a16851cf8b41b74b0f593fd9a67664c07067bd. The typed recovery execution contract still has two blocking defects, and the team revise preflight is stricter than its enforcement path for a supported plugin stage shape. Existing tests validate remedy shape and predicate flags, but they do not execute emitted recovery commands or verify human feedback provenance.

apackeer:
## Review findings addressed

All three requested changes are fixed on head `e4304ee5e`.

1. **Human feedback provenance**
   - Request Changes remedies are action-only until a separate human turn provides feedback.
   - The active-directive marker stores response digests, not plaintext; the remedy-selection turn cannot satisfy the feedback check.
   - Rejection refuses missing or altered feedback and accepts only the exact second response.
   - End-to-end coverage verifies that only the exact human text reaches `GATE_REJECTED`.

2. **Executable remedy commands**
   - Every command-bearing remedy is emitted as a concrete `bun <harness>/tools/aidlc-*.ts ...` invocation.
   - Unresolved placeholders and bare tool names are rejected by directive validation.
   - Tests execute restart remedies from all seven packaged harnesses and execute the autonomous Bolt abort/discard remedy against a real temporary Git worktree.

3. **Team pipeline revise equivalence**
   - One shared pipeline predicate is used by preflight, gate opening, and both team revise enforcement snapshots.
   - A plugin-shaped per-Unit pipeline fixture proves preflight and enforcement refuse with the same code.

Additional branch-transition work:

- Merged current `main`.
- Re-slotted the user-visible release metadata to `2.7.1`.
- Preserved the `main` installation/docs transition and regenerated all distributions.

Verification:

- `bun run check`: pass.
- Combined focused slice: 8/8 files, 307 assertions, pass.
- Final smoke + unit + integration: 389 files, 9,612 assertions. Five
  test-maintenance reds were fixed or isolated and all passed on rerun; every
  production and integration journey passed in the full run.
- Affected maintenance rerun: 5/5 files, 217 assertions, pass.
- Isolated load-sensitive `t315`: 24 assertions, pass.
- Pre-push scan: 120 outgoing files, clean.

leandrodamascena (dismissed):
Arden, thanks for addressing the previous review findings.

I reviewed the current head `e4304ee5e` against `main` at `96b11d390`. The three blockers from my previous review are resolved:

- Request Changes recovery now waits for a separate human feedback turn and accepts only the recorded response.
- Recovery commands use concrete packaged `bun <harness>/tools/...` invocations. Unresolved placeholders and bare tool names are rejected.
- Team pipeline revise now uses the same check in preflight and enforcement, including the plugin per-Unit case.

Validation performed:

- `t330-guard-deadlock-liveness`: 20 passed.
- `t324-team-unit-progress-gates`: 27 passed.
- `git diff --check` passed.
- CI is green for the current head.

No remaining findings from my side.

apackeer:
Pushed the systemic fix in `46f470baa`.

This treats #977 and #995 as one observer-mutation class:

- Stop-hook probes and `AIDLC_ROUTE_CHECK` are centralized as read-only engine observations. They cannot publish directives, advance continuation cursors, create token keys, bootstrap diaries, refresh state/cache files, rotate Code Generation authority, clear Plan Approval runtime, or touch engine-turn markers.
- Stop retains an exact current `load-steering`, `run-stage`, or `invoke-swarm` directive before falling back to a probe, so an end-of-turn check cannot restart rule delivery or replace execution authority.
- Code Generation markers now separate the live state digest from the authority issuance digest. Exact `unit start|pause|resume|complete` cache updates preserve the approved epoch and receipt, but only for the routed Unit and only when all non-cache state bytes are unchanged.
- Pause and completion retain reusable approval evidence but are not executable: developer dispatch/workspace mutation is allowed only while the exact lifecycle Unit is `in-progress`.
- Copilot now retains and validates `invoke-swarm` Unit batches through settlement and Stop progress.

Regression coverage includes:

- whole-project byte-purity snapshots for Stop and route-check observers;
- retained steering/run-stage/swarm behavior across Stop;
- stage-level, per-Unit, swarm, challenge/response/receipt, and generation-active Plan Approval;
- the exact issue-author path: `Approve Plan -> aidlc-state unit start -> developer dispatch`, plus pause, resume, complete, and unrelated-state-delta refusal;
- Copilot swarm retention and adapter security;
- deterministic repair of the live Reverse Engineering fixture's missing `STAGE_STARTED` boundary.

Verification:

- `bun run check`: package parity, all TypeScript configs, and lint pass.
- Coverage registry freshness and `git diff --check` pass.
- Affected authority/Copilot slice: 12 files, 279 assertions, all pass.
- Exact authority suite: 14 tests, 284 expectations, all pass.
- Copilot adapter security: 25/25 pass.
- Live Reverse Engineering integration rerun: pass.
- Independent defensive re-review: no remaining P1/P2 findings.

The earlier 389-file default run reached 387 passing files; its two reds were the Copilot test-stub import and the incomplete live RE fixture described above. Both were fixed and rerun green in isolation; the full 389-file run was not repeated after those test-only repairs.

leandrodamascena (changes requested):
Arden, thanks for the follow-up work on the observer-mutation failures from #977 and #995.

I reviewed the current head `46f470baa6f3f433515d0a652fe21f6396d79426` against `main` at `96b11d39028955d4f92375e783525db5275cdfd8`. Project alignment is established by the prior maintainer review of this internal change through `e4304ee5e`. This review covers the new authority-preservation delta and its interaction with the previously accepted guard-recovery work.

## Finding

**[P2] Do not retain the pre-checkpoint run-stage after Unit pause or completion**

`refreshActiveDirectiveUnitState` updates the marker's live state digest for every `unit start|pause|resume|complete` action, but it leaves `delivery: "issued"` and `needs_rehydrate: false` unchanged. The Stop hook therefore accepts the marker as the exact current directive in `retainedSharedDirective` and skips the read-only `next` observation.

I reproduced this with the real per-Unit Code Generation flow. After Plan Approval, `unit start`, and `unit pause --reason "session ending" --next-action "resume generation"`, the Stop hook blocked with:

> The exact delivered AIDLC run-stage for "code-generation" is still active. Complete that exact stage...

At the same state, a byte-pure `aidlc-orchestrate next` returned the correct paused-Unit ask with the exact `unit resume` command. A valid `unit complete` has the same problem. Stop retains the pre-completion run-stage while `next` routes the stage gate.

This breaks the documented pause contract for session-ending checkpoints and makes forwarding-loop recovery instruct an action that the Unit lifecycle no longer admits. Preserve the Code Generation authority issuance digest, but make pause and complete non-retainable, or make retention validate the live Unit lifecycle before reusing the marker. Add regression coverage that compares Stop behavior with the read-only engine route after both pause and complete. Keep the existing authority preservation for start and resume.

Relevant code: `core/tools/aidlc-lib.ts:5334`, `core/tools/aidlc-state.ts:2061`, and `core/hooks/aidlc-continue-workflow.ts:1073`.

## Requested release cleanup

Please remove the `2.7.1` version bump and its release metadata from this PR, including `core/tools/aidlc-version.ts`, the README badge, the `CHANGELOG.md` heading, and regenerated distribution copies.

## Validation

- Focused Stop, Plan Approval, and Copilot suites: 199 passed, 2 skipped.
- `t248-steering-content-delivery`: 43 passed.
- `bun run check`: package parity, TypeScript checks, and Biome completed successfully.
- `git diff --check`: passed.
- GitHub CI is green for the exact head.
- The pause and completion mismatch was reproduced with real packaged commands in an isolated worktree.

Remaining risk is limited to other per-Unit stages that call the same lifecycle-marker refresh helper. The same stale-retention path applies outside Code Generation even though the authority digest is Code Generation-specific.

apackeer:
Pushed follow-up `7fb479a5d`, converting the additional issue-author log failures into deterministic regressions.

### User Stories review recovery

The individual refusals were correct and already covered: review before post-confirmation save, malformed reviewer appendix, changed request-bound bytes, retry refusal, review budget, and wrong verdict iteration.

The composed path exposed one remaining #978 defect: after request-bound artifact/source bytes changed, the guard-recovery ask could still advertise both verdict recording and `--retry-pending`, even though the real review commands deterministically refuse both.

The fix adds one shared pending-request executability projection used by:

- `aidlc-log`
- `aidlc-state`
- review-freeze

It validates current artifact bytes, workspace source, Unit source manifest/fingerprint, appendix authority, required artifacts, retry consumption, and the active team claim generation before advertising either remedy. A stale request now offers the executable human Request Changes path instead of commands that will fail.

### Code Generation Plan Guard review handoff

Added an executed regression for:

`Approve Plan -> unit start -> developer dispatch -> Stop -> review request -> reviewer-dispatch record -> review appendix -> review verdict -> unit complete`

The test runs the real review and lifecycle tools, not only the PreToolUse classification. The active `run-stage`, directive epoch, and generation receipt survive, while the Plan Approval guard allows the read-only/framework/review operations observed in the supplied log.

### CI Pipeline summary sequence

The three CI summary errors in the log are intentional ordering guards: confirmation was missing, then the prompt was retried with a populated answer, then outputs had not been re-saved after confirmation. Existing tests cover those invariants, so no behavior was relaxed.

### Verification

- Final review/source/Plan Approval slice: 18 files, 475 assertions, pass.
- Final team-claim/guard slice: 7 files, 162 assertions, pass.
- Executed Code Generation review journey: pass.
- `bun run check`: package parity, all TypeScript configs, and lint pass.
- Coverage registry freshness and `git diff --check`: pass.
- Independent defensive re-review: no remaining P1/P2 findings.

apackeer:
I am closing this in favour of a replacement that fixes the same class at its source: https://github.com/awslabs/aidlc-workflows/pull/1000.

The reason is the root cause, which I did not have when I opened this. The observer-mutation fix here is correct, and it is the part users needed, but it treats the symptom: it stops four callers from destroying a Plan Approval, while leaving in place the reason a caller could destroy one at all. A recorded approval was keyed to the active directive's issuance identity, that directive was readable only while a whole-file hash of `aidlc-state.md` matched (cache fields included), and the approval store was deleted from seven call sites as a side effect of ordinary publication. So the conductor's own `next` still wiped it on this head, and so did a status sync and a compaction. I measured that on both trees before deciding.

The replacement re-binds the approval to content and attempt, deletes the whole-directory wipe primitive, and makes the state digest a projection that omits the cache layer. It carries everything this PR set out to do:

- this PR's observer slice, as a shared predicate plus a typed barrier at the write primitives, so the guarantee is a property of the code shape rather than a snapshot test;
- this PR's attempt reducer, unchanged except for the shard-ordering convention;
- this PR's guard-recovery layer, rebuilt: the refusal evaluator returns a typed refusal whose remedies each carry an `op` from a closed list, the directive-time preflight and the enforcing tools share one admission function, the refusal streak ends in a terminal ask that carries a state signature (never an error directive and never a silent counter), the guard-recovery ask is rendered identically by every conductor, and the re-presentation loop the recovery spike found here (the human's remedy selection discarded on every re-run `next`) is closed at its source rather than by the fail-open block cap.

Nothing from this PR is deferred any more.

On the pause-and-complete finding from the last review: it does not arise in the replacement, and there is a regression test for it. The cause here is that hook-side retention reuses the marker without re-routing, so a paused Unit still looks active. The replacement makes the engine idempotent instead, which recomputes routing every time and only skips the transport, so a paused or completed Unit always produces its own directive. That also means `refreshActiveDirectiveUnitState` and the second authority digest are unnecessary: under the projected state digest a Unit lifecycle write does not move the digest at all, so there is nothing to carve out.

Thanks for the four rounds of review on this branch. Every finding informed the replacement, and the two on this head are covered there: the release-metadata cleanup, and the pause/complete case above.

PR #982, opened 2026-08-31
fix: consume retired --init/--force flags instead of leaking them into intent descriptions

apackeer:
Since #847 (correctly) made unknown flag-looking tokens lossless task text, the retired `--init`/`--force` flags (Branch 3, retired in P4 — initialization is automatic) leak into the created intent's literal description: `/aidlc --init --scope bugfix` produces `intent-create --arguments='--init …'`, an unactionable task. Observed live on Kiro over ACP: the agent stops and asks the human what to build instead of proceeding. Anyone following pre-P4 docs or muscle memory hits this.

**Fix:** `parseNextFlags` consumes `--init`/`--force` as harmless no-ops. Genuinely unknown tokens keep #847's lossless behavior; the `--` delimiter still passes a literal `--init` through as task text.

**Tests:** four t114 pins — the two no-leak regressions (red without the fix, green with it), the #847 lossless guard, and the `--` delimiter escape.

Surfaced during PR #756's v2 sync validation (disclosed there); independent of that PR — reproduces on v2's committed dist.

No version or changelog bump: release metadata belongs to the release-preparation PR (repo policy).


leandrodamascena (dismissed):
Arden, the parser change looks correct and the regression tests cover the intended behavior.

I reviewed the current head `a6bbd7b99` against `main` at `96b11d390`. The fix is aligned with the retired P4 flags and preserves the lossless behavior from #847.

One change before merge: we should no longer bump the framework version in individual fix PRs. Please remove the `2.6.125` bump and its release metadata from this PR:

- `core/tools/aidlc-version.ts`
- generated `dist` version files
- README version badge
- `2.6.125` CHANGELOG section

Please keep the parser change, generated orchestrator files, and regression tests. The branch also needs a rebase onto the current `main`, since it is now conflicting.

Focused validation: all four new `t114` tests pass. The existing exact-head CI is green.

No changes were pushed because this is maintainer-owned work.

**Review event: `REQUEST_CHANGES`.**

apackeer:
Rebased onto current `main` (`a0ee4415e`) and trimmed to the two files the review asked to keep: the `parseNextFlags` no-op arm and the four t114 tests. The 2.6.125 bump, README badge, CHANGELOG section, and every `dist/` edit are gone (`dist/` is generated since #756, so nothing under it is committed any more).

One adjustment forced by the rebase: main migrated the create command from `intent-create` to `intent create`, so the four tests' command assertions now match main's convention; names and the no-leak/lossless/delimiter assertions are unchanged.

Verified on the rebased head: t114 59 pass / 0 fail. Negative proof - with only the parser arm stashed and dist regenerated, `--init with --new-intent + prose creates without leaking the flag` and `--force is likewise consumed` fail (both flags leak into `--arguments`), the other two still pass. `bun run typecheck`, `bun run lint`, `bun scripts/package.ts --check`, `git diff --check` all exit 0.

PR #986, opened 2026-08-31
fix(kiro-ide): scope plan-approval fallback guard to active workflows (#965)

### LINKED ISSUE #965: [Bug]: kiro-ide plan-approval guard denies every path-less tool unconditionally, blocking all subagent dispatch
Follow-up to #763 (arguments do arrive) and #764 (the wiring half). Since 2.6.122 ships a `plan-approval-guard` manifest for `kiro-ide`, the wiring #764 asked about appears to be in place — and registering it exposes a deny in the adapter's legacy branch that has no `state.active` condition. The result is that `invoke_sub_agent` is refused on every call, with no workflow running at all.

### What I measured, and on what

Kiro IDE (Claude Opus 5 as the chat model), AI-DLC 2.6.122 installed per `docs/guide/harnesses/kiro-ide.md`, `.kiro/` byte-identical to `dist/kiro-ide/.kiro/` at commit `2f7c18a`. macOS, bun 1.4.0.

Every attempt to delegate to a subagent is refused with:

```
```

Refused: `invoke_sub_agent` (the composer dispatch emitted by `next compose`), skill activation, and web fetch. Permitted: `execute_bash`, and any tool carrying a file path. The workflow was never started — `aidlc-orchestrate.ts next` returns `No workflow state found (no active intent)` at the same time the deny fires, so no Code Generation window exists to protect.

### Why registering the guard exposes it

The manifest shipped in 2.6.122 registers on `PreToolUse` with **no matcher**, so the guard evaluates every tool call:

```json
{ "name": "aidlc-plan-approval-guard", "trigger": "PreToolUse",
  "action": { "type": "command", "command": "bun .kiro/hooks/aidlc-kiro-adapter.ts plan-approval-guard" } }
```

In `aidlc-kiro-adapter.ts`, classification is by exclusion:

```ts
const PLAN_APPROVAL_SAFE_READ_TOOLS = new Set([
  "fs_read", "file_search", "grep_search", "thinking", "todo_list",
]);

function mutationCapableTool(name: string): boolean {
  return name.length > 0 && !PLAN_APPROVAL_SAFE_READ_TOOLS.has(name);
}
```

`invoke_sub_agent` is absent from that set, so it is mutation-capable. It then satisfies `opaqueMutation` because it has arguments but no path:

```ts
const opaqueMutation =
  toolName === "" ||
  (mutationCapableTool(toolName) &&
    (Object.keys(toolArgs).length === 0 ||
      (toolName !== "execute_bash" && paths.length === 0)));
```

Inside `if (opaqueMutation)`, every deny is gated on a live window — `state.active && !state.approved && ...`, `state.active && state.violated`, and so on — **except the last one** (line 1505 in 2.6.123, 1506 in 2.6.122):

```ts
if (Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
  return {
    hook: "__legacy_plan_approval_block__",
    input: { reason:
      "Plan Approval blocked a mutation-capable payload whose target path is missing or unsupported." },
  };
}
```

No `state.active`. With no workflow, control falls past every gated deny and lands here.

This reads as a fallback written for the argument-less payloads the comment above it describes ("Legacy planning and post-human answer recording remain usable"). On a build that sends arguments — which #763 measured on 1.0.309, and which matches the `prompt`/`explanation` keys reported there for subagent dispatch — `Object.keys(toolArgs).length > 0` is true for every path-less tool, so the intended narrow fallback becomes a blanket deny.

### Impact

Delegation is how four stages run (2.1 pipeline, 2.2 subagent, 2.4 mob, 3.5 subagent), plus every reviewer pass and the adaptive composer. All refused. Reverse Engineering is dispatched, so `/aidlc compose` on a brownfield task cannot get past the planning step.

`/aidlc --doctor` reports 40 passed, 0 failed throughout: it verifies hook bodies are present and the stage graph is coherent, not that a dispatch survives the guard.

### Suggested fix

Give that deny the same condition as its siblings:

```ts
if (
  state.active &&
  Object.keys(toolArgs).length > 0 &&
  toolName !== "execute_bash"
) {
```

`state` is already in scope from `legacyPlanApprovalGuardState(projectDir)` earlier in the branch. During a real Code Generation Plan Approval the behaviour is unchanged; outside one it goes inert. I applied this locally and subagent dispatch was permitted, then reverted to keep the tree at the shipped bytes.

A second question worth separating: `PLAN_APPROVAL_SAFE_READ_TOOLS` is an allowlist of five names used to classify by exclusion, so every tool the IDE adds — and every read tool not on the list, such as `read_file`, `list_directory`, `web_fetch` — is mutation-capable by default. Path-carrying tools escape via `paths.length`, but path-less ones do not. Whether delegation and read-only tools should be named explicitly rather than inferred seems like a design call rather than part of this fix.

### What I am not claiming

One machine, one IDE build. The trigger requires the host to deliver a tool name plus arguments and no file path, so builds that send empty `toolArgs` to PreToolUse would take the `return null` path below instead and never see this. I did not test Kiro CLI, and I did not test 0.12-era builds. I have not verified which IDE version first started sending arguments for `invoke_sub_agent`; #763's measurement on 1.0.309 is the closest data point I know of.

Verified present in both 2.6.122 and 2.6.123 (`dist/kiro-ide/.kiro/hooks/aidlc-kiro-adapter.ts`), with the allowlist unchanged between them, so upgrading does not clear it.

omaraws:
## Summary

Fixes #965. The legacy plan-approval **fallback** guard in the kiro-ide adapter blocks every path-less, mutation-capable tool call that carries arguments, regardless of whether an AI-DLC workflow is active. On Kiro IDE this blocks all MCP tool calls and subagent dispatch during normal chat when no workflow is running.

## Root cause

In `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts`, inside the `toolName === "" || mutationCapableTool(toolName)` block:

- The earlier guard (~line 1462) is correctly gated: `if (state.active && !state.approved && !LEGACY_PLANNING_WRITE_TOOLS.has(toolName))`.
- The fallback at line 1505 was **missing the `state.active` check**:

```js
if (Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
  return { hook: "__legacy_plan_approval_block__", input: { reason:
    "Plan Approval blocked a mutation-capable payload whose target path is missing or unsupported." } };
}
```

So with no active workflow, that fallback still fires for any path-less tool with arguments.

## Fix

Add `state.active &&` to the fallback so it only engages during an active workflow:

```js
if (state.active && Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
```

Code-generation protection is unchanged — the guard still applies whenever `state.active` is true.

## Reproduction

```
printf '%s' '{"toolName":"atlassian_search","toolArgs":{"query":"test"}}' \
  | bun .kiro/hooks/aidlc-kiro-adapter.ts plan-approval-guard
```

- Before: exit 2 (blocked), no active workflow present.
- After: exit 0. `execute_bash` and path-carrying write tools are unaffected either way.

Also validated live in Kiro IDE: after the change, a path-less read-only MCP tool call with arguments passes the guard and returns data.

## Notes

- Source edited in `harness/kiro-ide/`; `dist/kiro-ide/` regenerated with `bun scripts/package.ts kiro-ide`.
- `bun scripts/package.ts --check` and `biome check` both pass.

### DIFF
--- a/dist/kiro-ide/.kiro/hooks/aidlc-kiro-adapter.ts
+++ b/dist/kiro-ide/.kiro/hooks/aidlc-kiro-adapter.ts
@@ -1502,7 +1502,7 @@ function buildForward(): Forward {
               };
             }
           }
-          if (Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
+          if (state.active && Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
             return {
               hook: "__legacy_plan_approval_block__",
               input: {
--- a/harness/kiro-ide/hooks/aidlc-kiro-adapter.ts
+++ b/harness/kiro-ide/hooks/aidlc-kiro-adapter.ts
@@ -1502,7 +1502,7 @@ function buildForward(): Forward {
               };
             }
           }
-          if (Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
+          if (state.active && Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
             return {
               hook: "__legacy_plan_approval_block__",
               input: {


leandrodamascena (changes requested):
Thanks for the focused fix and the clear reproduction in #965. I reviewed the current head `157cdd73d7db56c36a207f81222aaa0914da452c` against `main` at `96b11d39028955d4f92375e783525db5275cdfd8`. The change aligns with the contract described in #965: the Plan Approval guard stays inactive outside Code Generation while the existing active-state fail-closed paths remain in place.

I found two blockers:

1. **P2 - Add regression coverage for the valid no-workflow payload.** The only behavioral change is at `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:1505`, but this PR changes no tests. I reproduced the reported path with both camelCase and snake_case payloads, plus `invoke_sub_agent`: the buggy base exits 2 and the current head exits 0. The unchanged `t218-kiro-ide-hook-adapter.test.ts` still passes on both revisions with 74 tests and 579 assertions, so the suite cannot detect this regression. Please add a `t218` case using a valid path-less tool payload with arguments and no workflow state, asserting exit 0. Keep or extend the active Code Generation contrast that asserts the same mutation-capable path remains blocked before approval.

2. **P2 - Rebase onto the current v2.7.0 baseline and add the required release metadata.** This is a user-visible Kiro IDE bug fix, but the PR does not update `core/tools/aidlc-version.ts`, the README badge, or `CHANGELOG.md`, as required by `AGENTS.md`. The head is two commits behind `main`, which now points to tagged `v2.7.0`. Please update the branch, bump to the next available patch version, add a changelog entry with the Kiro IDE upgrade instruction, and regenerate the distributions.

I made no changes to the contributor branch. Local validation on the current head included the exact MCP and subagent reproductions, `bash tests/run-tests.sh --unit --filter '^t218-'`, `bun scripts/package.ts --check`, and `git diff --check`. GitHub reports all 10 checks passing for this head, but those runs completed before `main` advanced to v2.7.0 and must rerun after the branch update.

wowzoo:
Thanks for this, and for the reproduction in #965 — the diagnosis was exactly right and the
one-line gate was the correct fix.

It has since landed by another route, so there is nothing left for this PR to merge. On `main`,
`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts` now reads
`if (state.active && Object.keys(toolArgs).length > 0 && !isKiroShellTool(toolName))`, and
`git log -S` puts that gate in #1000. I checked the behaviour rather than just the diff, using
the repro from your description: with no workflow active and a path-less payload carrying
arguments, `atlassian_search`, `invoke_sub_agent` and `disclose_context` all exit 0 on `main`,
while inside an active Code Generation window the same three still exit 2 — so the protection
you were careful to preserve is intact.

Both points from the earlier review have since been answered, and neither is anything you still
owe:

- The regression test that was asked for is present. #1000 added it in the same commit as the
  gate — it lives at the end of the `every shell name with a command and no workflow is allowed`
  case in `tests/unit/t218-kiro-ide-hook-adapter.test.ts`, sending a non-shell, path-less tool
  with arguments against a no-workflow fixture and asserting exit 0.
- The version bump, README badge and `CHANGELOG.md` entry are no longer wanted at all.
  `CONTRIBUTING.md` now says ordinary fix PRs leave those three surfaces to release prep.

So I am closing this as already fixed rather than as rejected, and to be clear about the one
thing that might look like a mark against it: the branch also carries the regenerated
`dist/` copy, which is exactly what the contributing guide required when you opened this. The
project stopped tracking `dist/` afterwards, which is why the branch now conflicts. That is our
timing, not your mistake.

Sorry it took long enough for the fix to arrive from elsewhere. The analysis in #965 and here is
what made it obvious, and that stands on its own — please do open more.

PR #990, opened 2026-09-01
chore: align v1 maintenance branch configuration

leandrodamascena:
## Summary

- Route branch-driven CI, CodeBuild, CodeQL, PR validation, release, and scanner workflows to `v1`.
- Remove CodeQL and security scanner schedules from the maintenance branch.
- Remove documentation deployment, issue forms, and the PR template from `v1` because the default `main` branch owns those repository-wide surfaces.
- Announce v2 GA on `main`, direct new installations and upgrades there, and align the v1 administrative guide.

## Validation

- `actionlint` 1.7.12 on all remaining workflows
- `markdownlint-cli2` 0.23.2 on all 59 Markdown files
- YAML parsing for every workflow
- Confirmed no `schedule` or `cron` remains under `.github/workflows/`
- Confirmed the v1 contributor-statement check exactly matches the PR template moved to `main`
- `git diff --check`


apackeer (changes requested):
Requesting changes because the default-branch transition currently leaves v1's release and PR-validation paths disconnected.

1. **Restore the dispatchable v1 release workflows on the default branch.** `.github/workflows/release-pr.yml:16` relies on `workflow_dispatch`, while `.github/workflows/tag-on-merge.yml:68` and `:99` dispatch `release.yml` and `codebuild.yml`. GitHub currently registers Release PR, Release, and CodeBuild as `deleted` because those workflow paths are absent from default `main`; `workflow_dispatch` only triggers when the workflow file exists on the default branch. Please add compatible default-branch workflow definitions/shims or redesign the dispatch path so the documented v1 release flow is executable.

2. **Restore v1 pull-request validation from the default-branch context.** `.github/workflows/pull-request-lint.yml:4-5` retargets `pull_request_target` to `v1`, but the workflow is absent from default `main` and is registered as `deleted`. This PR itself received no Pull Request Validation run, so the title, contributor statement, label, release-lock, and do-not-merge checks would remain inactive for v1 PRs. Please host the workflow on `main` with the appropriate `v1` base filter, or use a mechanism that executes from the v1 branch.

3. **Keep the CodeBuild trend artifact producer and consumer on the same branch/name.** `.github/workflows/codebuild.yml:461-465` now publishes `report-v1`, but `scripts/aidlc-evaluator/packages/trend-reports/src/trend_reports/fetcher.py:210-239` still queries branch `main` and artifact `report-main`. That drops the v1 pre-release baseline from trend reports. Please parameterize the target branch/artifact and update the associated fetcher tests.

4. **Finish the required stale-reference sweep for the removed/renamed surfaces.** At minimum, `.kiro/steering/pr-conventions.md:5` still points to the deleted local PR template, `docs/DEVELOPERS_GUIDE.md:62` still says scanners target `main` and run daily, and `docs/ADMINISTRATIVE_GUIDE.md:77` still counts the removed documentation workflow.

The existing checks are green, but they do not exercise the missing Pull Request Validation or dispatch paths above.

apackeer (approved):
Re-review complete at `d7f44960312a`.

The follow-up commit resolves the CodeBuild trend artifact mismatch and the cited stale documentation references. Default-branch PR validation is handled by the companion main-branch transition PR. The absence of the v1 release workflows from default `main` is intentional and deferred to a separate PR, so it is not a blocker for this change.

All current checks pass. Approved.

Issue #714, opened 2026-08-05
[RFC]: DocumentKB — a customer's documents become retrievable, intent-aware context


alokgp:
## Summary

Add a **DocumentKB**: a customer's documents live under
`aidlc/spaces/<space>/knowledge/documents/` as user-managed originals, are indexed
into a tool-managed `aidlc/spaces/<space>/knowledge/documentkb/`, and are then
**retrieved automatically as intent-aware context** by the agents already shipped.

```text
/aidlc knowledge onboard [path] [--intent [slug]]
/aidlc knowledge sync
```

This RFC replaces **#476** (`/aidlc-onboard`), and the PR that implemented its
first slice (**#660**) is closed as superseded. The direction comes from
@packeer's review on #660: the objective was directionally correct, but
`<space>/onboard/` was the wrong persistence boundary, and the work must be
**DocumentKB-first** rather than rule-first, grounded in the existing
`knowledge/`, `active-intent`, and `intents.json` model.

Baseline: `upstream/v2` @ `c73ee984` · **v2.5.37** · authored once in `core/`,
emitted to every harness. This is a **core skill** (the plugin mechanism has no
`skills` bucket).

## Motivation

A customer's standards, domain references, and policy documents have no home in
the engine today — they are re-explained per session. #476 framed the answer as
"onboard material, promote rules". That framing put the *rule* first and left the
documents themselves with no durable, retrievable representation, and no retrieval
layer at all.

Inverting it is the fix. The documents are the durable asset; rules are one
downstream product of reading them. Three properties keep it honest:

- **The user owns `documents/`; the tool owns `documentkb/`.** The user's folder
  structure is theirs to nest however they like and the tool never reorganises it.
  `documentkb/` is derived and rebuildable from the originals.
- **Retrieval is intent-aware.** Documents are space-wide by default; a document
  may additionally be associated with specific intents, and per-stage context
  assembly selects space-wide ∪ intent-matched.
- **Mandatory behaviour stays in `memory/`.** Approved, binding practice belongs in
  `memory/{team,project}.md`. Contextual information belongs in DocumentKB. This is
  the line that keeps the two layers from collapsing into each other.

## Proposed user experience

Users organise originals however they want beneath
`aidlc/spaces/<space>/knowledge/documents/`, then:

- **No path** → scan `documents/` and index everything not yet indexed.
- **A path inside `documents/`** → index in place.
- **An external path** → copy into `documents/` first, then index.
[...]

Issue #694, opened 2026-08-01
[Feature]: auditable supplemental-knowledge selection and delivery across stage topologies


apackeer:
## Context

#495 and PR #658 (v2.5.33) made required rule delivery deterministic. The engine resolves the active-space rule bundle, emits it as bounded `load-steering` parts, and enforces complete delivery at dispatched-worker boundaries.

Supplemental knowledge still has a different contract. `run-stage.inline_context_paths` inventories framework and active-space knowledge for conductor-owned work, but a present, readable file can still go unread without any receipt. Fully dispatched stages carry no inline roster. Kiro CLI preloads shipped methodology through native `resources`, while other harnesses and active-space team knowledge still depend on agent-directed reads.

This issue is specifically about supplemental knowledge. Dispatched personas already load through harness-native agent definitions, and reviewer checklists are absorbed into reviewer agent bodies at build time. Inline persona content delivery is a separate bounded mechanism problem and should be tracked separately. Kiro IDE native-agent work remains under #555.

## Design boundary

Knowledge is reference material an agent weighs; it is not a mandatory rule. A retrieval tool must therefore guarantee inventory, transport, bounds, integrity, and observability without deciding semantic relevance.

The ownership split should be:

- **Tool:** deterministically catalogs available knowledge, validates requested chunks, returns bounded content, and records integrity-bound delivery receipts.
- **LLM/agent:** selects the knowledge relevant to the current stage and unit of work, or explicitly selects `none` with a rationale.
- **Human/reviewer:** judges whether the selected knowledge and its application were sufficient when that affects an artifact or gate decision.

## Proposed contract

1. Build an active-space knowledge catalog for the executing agent and topology from:
   - framework shared knowledge;
   - framework agent-specific knowledge;
   - active-space shared team knowledge;
   - active-space agent-specific team knowledge.
2. Split Markdown at stable heading boundaries and identify each chunk by source path, heading identity, and content hash.
3. Allow stages to declare `required_knowledge` for true hard dependencies, such as a stage-mandated template. Required entries are tool-selected; supplemental entries are not.
4. Have the executing agent request catalog chunk IDs through a deterministic retrieval command, or explicitly record `none` with a rationale.
5. Return content through a bounded transport. Never silently truncate, omit, or claim delivery.
6. Record a receipt bound to the active space, stage, agent, dispatch attempt, catalog hash, selected chunk hashes, and completed delivery.
[...]

PR #991, opened 2026-09-01
docs: make main the v2 GA source of truth

leandrodamascena:
## Summary

- Route the v2 CI, documentation, Markdown, and security workflows to `main`.
- Mark `main` as the v2 GA branch across the README and every installation guide.
- Rewrite published documentation links to `blob/main` and update the pinned regression test.
- Refresh the roadmap through 2.6.124, move completed work out of In flight, and list current open delivery themes.
- Add issue forms and the pull request template to the default branch, with v2 listed first and v1 identified as maintenance.

## Validation

- `actionlint` 1.7.12 on every workflow
- YAML parsing and duplicate-ID checks for every issue form
- `markdownlint-cli2` 0.23.2 on the changed root documentation and PR template
- 33 targeted unit tests in `t181` and `t246`
- TypeScript typecheck and Biome lint
- Strict Zensical build after the production link-rewrite step, with all generated links targeting `blob/main`
- `git diff --check`


apackeer (changes requested):
Requesting changes for three documentation accuracy issues:

1. **P2 - The installation prose still contradicts the new commands.** The commands correctly clone `main`, but six passages immediately above them still say the checkout is on the `v2` branch:
   - `docs/guide/01-getting-started.md:195`
   - `docs/guide/harnesses/codex-cli.md:29`
   - `docs/guide/harnesses/copilot.md:55`
   - `docs/guide/harnesses/kiro-cli.md:25`
   - `docs/guide/harnesses/kiro-ide.md:35`
   - `docs/guide/harnesses/opencode.md:43`

   Please update those sentences to name `main`, matching the commands and the PR's stated branch transition.

2. **P2 - The roadmap overstates completion of the DocumentKB RFC.** `docs/roadmap.md:195` says #894 completed #714, but #894 explicitly leaves intent-aware discovery and agent delivery waiting on open issue #694. Please describe summaries and tags as completing that slice rather than the full RFC.

3. **P3 - The roadmap says all top-level plugin authoring routes remain proposed.** `docs/roadmap.md:178-180` makes that blanket claim, but `plugin validate` and `plugin build` already ship at `core/tools/aidlc-utility.ts:389-390`. Please narrow the remaining work to the routes that are actually still deferred, such as `create` and `test`.

The workflow trigger changes, issue forms, templates, `blob/main` rewrites, and named branch-comment updates otherwise look correct. Targeted `t181` and `t246` verification passed: 2 files, 33 assertions, 0 failures.

.github/workflows/pull-request-lint.yml
```diff
@@ -0,0 +1,197 @@
+# This workflow must live on the default branch for pull_request_target events.
+# It validates pull requests whose base branch is the maintained v1 line.
+
+name: Pull Request Validation
+
+on:
+  pull_request_target:
+    branches: [ "v1" ]
```
apackeer: **P1:** This restores PR validation only for `v1`, but the active default-branch ruleset on `main` requires the `Check Merge Status` and `Fail by Label` jobs. PR 991 and the other current `main` PRs report neither required check. Include `main` here as well (for example, `branches: ["main", "v1"]`) or update the ruleset atomically; otherwise the required checks on `main` remain dark.

.github/workflows/pull-request-lint.yml
```diff
@@ -0,0 +1,197 @@
...
+      - synchronize
+      - unlabeled
+  merge_group:
+    types:
+      - checks_requested
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
```
apackeer: **P2:** For `pull_request_target`, `github.ref` is shared by every PR targeting the same base branch, so unrelated PRs cancel each other's validation. This has happened in production: run `32053525109` for `docs/improve-consistency` was cancelled when run `32053658109` for a different PR started. Include `github.event.pull_request.number` in the key, with `github.ref` as the fallback for `merge_group`.

.github/workflows/pull-request-lint.yml
```diff
@@ -0,0 +1,197 @@
...
+        id: get-pr
+        run: |
+          if [ "$GITHUB_EVENT_NAME" == "merge_group" ]; then
+            PR_NUMBER=$(echo "$GITHUB_REF" | grep -oP '(?<=/pr-)\d+' || echo "")
+            PR_LABELS=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER" | jq -c '[.labels[].name] // []')
+            echo "::group::Getting Information"
+            gh api "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER"
+            echo "$PR_LABELS"
+            echo "::endgroup::"
+          elif [ "$GITHUB_EVENT_NAME" == "pull_request" -o "$GITHUB_EVENT_NAME" == "pull_request_target" ]; then
```
apackeer: **P3:** `actionlint` 1.7.12 exits nonzero here with ShellCheck `SC2166`. This workflow only receives `merge_group` or `pull_request_target`, so this can be simplified to `elif [ "$GITHUB_EVENT_NAME" = "pull_request_target" ]; then`.

.github/workflows/pull-request-lint.yml
```diff
@@ -0,0 +1,197 @@
...
+          fi
+          echo "::debug::HALT_MERGES value: $HALT_MERGES"
+          echo "::debug::This PR number: $PR_NUMBER"
+          echo "::group::Open Release Pull Requests"
+          gh pr list --state "open" --repo "${{ github.repository }}" --json "number,headRefName"
+          OPEN_RELEASES=$(gh pr list --state "open" --repo "${{ github.repository }}" --json "number,headRefName" | \
+              jq '[.[] | select(.headRefName | startswith("release/"))]')
+          echo "$OPEN_RELEASES"
+          echo "::endgroup::"
+          echo "$OPEN_RELEASES" | jq --exit-status --arg pr "$PR_NUMBER" '[.[] | select((.number | tostring) != $pr)] | length == 0' && \
```
apackeer: **P3:** `actionlint` also exits nonzero on this `A && B || C` chain with `SC2015`. Use an explicit `if`/`else` around the `jq` result so the failure path cannot depend on the success of the intervening `echo`.

.github/workflows/pull-request-lint.yml
```diff
@@ -0,0 +1,197 @@
...
+            perf
+            test
+          requireScope: false
+
+  auto-label:
+    name: Auto-label
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
```
apackeer: **P3:** All three configured labels already exist. The pinned labeler documentation requires `pull-requests: write` for existing labels and reserves `issues: write` for creating missing labels. Remove this permission to keep the `pull_request_target` token least-privileged.

apackeer (changes requested):
The earlier documentation findings are resolved. The newly added PR-validation workflow introduces branch-gating and concurrency regressions that need correction before approval; details are inline.

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          if [ -z "$INPUT_TAG" ]; then
+            echo "::error::Provide the v1 release tag in the tag input, for example v1.2.0."
+            exit 1
+          fi
+          if [[ ! "$INPUT_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::Release tag must match vMAJOR.MINOR.PATCH."
+            exit 1
+          fi
+          gh workflow run release.yml \
```
apackeer, line 49: **P2:** This adds default-branch `workflow_dispatch` runs under the same `release.yml` path as the tagged v1 child runs. The live v1 `tag-on-merge.yml` currently selects the newest `workflow_dispatch` run without checking `headBranch == TAG`, so a concurrent manual forwarder can make it watch the wrong run and start CodeBuild before the intended release finishes. Pair this with a v1 caller fix that correlates the exact tag/run.

.github/workflows/codebuild.yml
```diff
@@ -0,0 +1,38 @@
...
+      actions: write
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+      INPUT_REF: ${{ inputs.ref }}
+      REPO: ${{ github.repository }}
+    steps:
+      - name: Run the v1 workflow
+        run: |
+          if [ "$INPUT_REF" != "v1" ] && [[ ! "$INPUT_REF" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
apackeer, line 33: **P2:** This accepts every semantic-version tag, not only refs containing the maintained v1 implementation. Existing tag `v2.3.0` passes this regex but contains neither `codebuild.yml` nor `release.yml`, so the accepted dispatch fails. Restrict tags to the v1 lineage or verify that the target ref contains the expected workflow before dispatch. The same issue exists in `release.yml`.

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+      INPUT_VERSION: ${{ inputs.version }}
+      REPO: ${{ github.repository }}
+    steps:
+      - name: Run the v1 workflow
+        run: |
+          if [ -n "$INPUT_VERSION" ]; then
+            gh workflow run release-pr.yml \
```
apackeer, line 41: **P3:** `gh workflow run` only creates the child dispatch event; this forwarding run succeeds once the event is accepted even if the actual v1 release operation later fails. The same fire-and-forget behavior exists in the CodeBuild and Release shims. Either identify/watch the exact child run and propagate its result, or make the dispatch-only status explicit so a green forwarder is not read as a successful release operation.

apackeer (changes requested):
The prior documentation and PR-validation findings are resolved. The newly added v1 dispatch shims still have release-run correlation and validation gaps; details are inline.

apackeer (changes requested):
Conditional approval: I consider this PR otherwise ready and will switch my review to **APPROVED** once these two remaining findings are fixed and re-verified:

1. Correlate `v1`'s `tag-on-merge.yml` wait loop to the exact release tag/run instead of selecting the newest `release.yml` `workflow_dispatch` run.
2. Validate that a requested tag contains the full v1 workflow implementation, not merely a same-path forwarding shim, so future main-line tags cannot recurse, fall back to `v1`, or fail from missing forwarded inputs.

The dispatch-only status is now explicit, `actionlint` is clean, and the targeted tests pass. Until the two findings above are resolved, the formal state remains **CHANGES_REQUESTED**.

apackeer (approved):
Approved per maintainer direction. The remaining v1 dispatch correlation and future-tag hardening items are documented as non-blocking follow-ups.

PR #993, opened 2026-09-01
fix: correlate tagged release dispatch runs

leandrodamascena:
## Summary

- capture the existing `release.yml` dispatch run IDs for the target tag before dispatch
- select only a newly created `workflow_dispatch` run whose `headBranch` matches that exact tag
- prevent concurrent default-branch forwarders from making `tag-on-merge.yml` watch an unrelated release run

Companion fix for #991.

## Validation

Not run, per maintainer request.

### DIFF
--- a/.github/workflows/tag-on-merge.yml
+++ b/.github/workflows/tag-on-merge.yml
@@ -34,6 +34,41 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
+
+      - name: Snapshot existing release runs
+        id: release_run_baseline
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: ${{ github.event.pull_request.head.ref }}
+          REPO: ${{ github.repository }}
+        run: |
+          TAG="${BRANCH#release/}"
+          KNOWN_RUN_IDS=""
+
+          for attempt in $(seq 1 5); do
+            RUNS_JSON=$(gh run list \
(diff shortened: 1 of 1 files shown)


apackeer (changes requested):
Requesting changes for one release-correlation failure path:

**P2 - Make the pre-dispatch baseline lookup recoverable.**

The new `KNOWN_RUN_IDS` snapshot at `.github/workflows/tag-on-merge.yml:67` is a single `gh run list | jq` pipeline before the release dispatch. If the query returns no JSON, the assignment can leave `KNOWN_RUN_IDS` empty; every later `jq --argjson known "$KNOWN_RUN_IDS"` invocation then fails before inspecting the available runs, so the loop falls through and CodeBuild is dispatched without waiting for the release workflow. If the initial pipeline itself exits nonzero, the step stops before either workflow is dispatched.

At that point the preceding step has already pushed the tag, and rerunning the job stops at the existing-tag check. Please retry the baseline lookup and validate that it produced a JSON array before proceeding, with coverage for the failed/empty snapshot path.

The exact-tag selector otherwise looks correct. `actionlint` 1.7.12, `git diff --check`, and the normal selector fixtures pass.

apackeer (approved):
The prior release-run baseline finding is resolved at `26ac151`.

The baseline lookup now retries and validates its response before tag creation, so failed or malformed run data stops the job while it remains safely retryable. The exact-tag correlation and downstream baseline validation remain intact.

Verified with `actionlint` 1.7.12, `git diff --check`, live run-list shape inspection, and fixtures covering valid data, an empty run set, malformed JSON, and persistent query failure. All PR checks are green.

PR #997, opened 2026-09-01
fix: stop the Stop-hook probe from resetting solo Plan Approval (#995)

### LINKED ISSUE #995: Plan Approval deadlocks on solo workflows: Stop-hook next probe re-publishes code-generation directive, bumping authority revision and wiping the approval receipt
### Summary

On a solo (non-team) workflow, Code Generation can never start. The "Approve Plan" gate can be answered any number of times but the approval never stabilizes, because the `Stop` hook's read-only `next` probe publishes a durable active-directive marker on every turn boundary. That publication bumps `code_generation_authority_revision` (invalidating the approval fingerprint's `directiveEpoch`) and calls `resetPlanApprovalRuntime()` (deleting the receipt + challenge). Since recording the human's "Approve Plan" answer *requires* a turn boundary, the challenge minted in turn N is destroyed by turn N's own Stop probe, and a valid receipt can never be written in turn N+1.

### Why it is specific to code-generation

Every other stage approves via a plain `HUMAN_TURN` / gate-resolution entry in the append-only audit ledger, which re-publishing does not erase. Code Generation is the only stage that additionally requires a **cryptographic approval receipt** bound to the plan bytes, test-instruction bytes, testing-contract hash, **and a `directiveEpoch`** (see `resolveCodeGenerationAuthority` / `evaluateCodeGenerationApproval` in `.claude/tools/aidlc-testing-posture.ts`). Its approval artifact is therefore both *destroyable* (lives in the plan-approval runtime dir) and *fingerprinted against a value (`code_generation_authority_revision`) that the turn-boundary probe changes*. No other stage has both properties, so no other stage deadlocks.

### Root cause

The deadlock and its intended fix are already documented in-code at `.claude/tools/aidlc-orchestrate.ts` (comment block above `isStopHookProbe`):

> Publishing from a bare `next` fails `preserveCodeGenerationAuthority` (the base marker is `run-stage`, not swarm planning), so `writeActiveDirectiveMarker` bumps `code_generation_authority_revision` and calls `resetPlanApprovalRuntime`, which rm -rf's the whole plan-approval runtime dir. It also re-transports `run-stage` back down to `load-steering` part 1 … The net effect was that Plan Approval could never stabilize and code-generation could never start.

The publication-suppression guard in `emit()` was scoped too narrowly. It suppresses the durable marker write only for **team-owned** probes:

```ts
// .claude/tools/aidlc-orchestrate.ts, emit()
if (
  prepared.marker &&
  !isTeamStopHookProbe(prepared.projectDir)   // <-- team-only
) {
  ... writeActiveDirectiveMarker(...) ...
}
```

`isTeamStopHookProbe` returns `false` for a solo workflow, so a solo Stop-hook probe still publishes and re-triggers the deadlock. In `writeActiveDirectiveMarker` (`.claude/tools/aidlc-lib.ts`), `preserveCodeGenerationAuthority` is only ever true on the swarm-planning path (`baseSwarmPlanning && nextSwarmPlanning`), so a solo `run-stage` code-generation directive always takes `codeGenerationAuthorityRevision = nextRevision` and `shouldResetRuntime = true`.

### The `Stop` probe is read-only by contract

`aidlc-continue-workflow.ts` consumes the probe's directive from **stdout** (`JSON.parse(stdout)`) after spawning `aidlc-orchestrate next` with `STOP_HOOK_PROBE_ENV=1`. It never reads the durable marker the probe writes. The orchestrate comment confirms suppressing publication for the probe "costs nothing: the prepared directive still goes to stdout via `writePrepared`, which is all the hook reads, and the conductor's own `next` (not marked as a probe) publishes normally."

### Steps to reproduce

1. Start a solo workflow with a scope that includes Code Generation (e.g. `classic`, or `poc`/`bugfix`/`express`).
2. Progress to the Code Generation stage; let the conductor produce `code-generation-plan.md`, `unit-test-instructions.md`, the Testing Contract, and present the Plan Approval gate.
3. Answer **Approve Plan**.
4. Observe: the developer-agent dispatch (or any workspace mutation) is refused by `aidlc-plan-approval-guard.ts` as "not currently approved" — the approval fingerprint / receipt no longer matches because the intervening turn boundary bumped the revision and reset the runtime dir.
5. Re-approving repeats the loop indefinitely.

### Expected

An approved plan on a solo workflow yields a stable, matching approval receipt so Code Generation can proceed.

### Actual

Approval never stabilizes; Code Generation cannot start by any non-destructive route on a solo workflow.

### Proposed fix

Broaden the publication-suppression predicate in `emit()` from team-only to **all** Stop-hook probes, matching the read-only contract the probe already advertises:

```ts
if (
  prepared.marker &&
  !isStopHookProbe()          // was: !isTeamStopHookProbe(prepared.projectDir)
) {
  ... writeActiveDirectiveMarker(...) ...
}
```

The conductor's own (non-probe) `next` still publishes normally, so forward progress is unaffected. `isTeamStopHookProbe` may become dead code and can likely be removed. Suggested regression test: a solo `run-stage` code-generation workflow where a Stop-hook probe fires between the Plan Approval mint (turn N) and the receipt write (turn N+1), asserting the receipt survives and `evaluateCodeGenerationApproval(...).ok === true`.

### Workaround

Set `AIDLC_DISABLE_PLAN_APPROVAL_GUARD=1` to disable the guard entirely (documented escape hatch in `aidlc-plan-approval-guard.ts`), or drive the workflow as team-owned. Both are stopgaps, not fixes.

spraja08:
## Summary

Fixes the solo Plan Approval deadlock in #995. The `Stop` hook's read-only `next` probe published the durable active-directive marker on every turn boundary. That publication bumped `code_generation_authority_revision` and called `resetPlanApprovalRuntime()`, so the approval challenge minted in turn N was destroyed by turn N's own probe and the receipt written in turn N+1 could never match. Code Generation was unreachable on any solo workflow.

The suppression predicate in `emit()` is broadened from team-only to *all* probe-marked invocations, matching the read-only contract the probe already advertises (`aidlc-continue-workflow.ts` parses the prepared directive off stdout and never reads the durable marker). The conductor's own non-probe `next` publishes normally, so forward progress is unchanged.

```diff
   if (
     prepared.marker &&
-    !isTeamStopHookProbe(prepared.projectDir)
+    !isStopHookProbe()
   ) {
```

`isTeamStopHookProbe` is retained rather than removed — it is still used by `handleContinue` for the cursor-advance suppression.

## Evidence

Both excerpts are from real solo `run-stage` code-generation workflows; only the audit fields that carry the proof are quoted.

**Before the fix** — one solo workflow, 6 × `PLAN_APPROVAL_RECORDED`, 32 × `PLAN_APPROVAL_BLOCKED`, Code Generation never started. Every recorded approval carries a *different* `Directive Epoch`, i.e. the epoch moved across each turn boundary and no receipt could ever match its challenge:

| # | Timestamp (UTC) | Directive Epoch | Approval Fingerprint |
|---|---|---|---|
| 1 | 06:14:24 | `sha256:bcef8f59…` | `sha256:82428265…` |
| 2 | 06:17:34 | `sha256:f3070a93…` | `sha256:539601ee…` |
| 3 | 06:22:09 | `sha256:9beab8b9…` | `sha256:270cbd04…` |
| 4 | 06:31:12 | `sha256:afe80bb7…` | `sha256:a0531ed2…` |
| 5 | 06:57:45 | `sha256:c4033313…` | `sha256:33ac4383…` |
| 6 | 07:19:46 | `sha256:7b185752…` | `sha256:618bbfe5…` |

Each re-approval was answered by the human, and each was followed by another `PLAN_APPROVAL_BLOCKED` on the next workspace mutation.

**After the fix** — same shape of workflow, run on this branch. One approval, and the epoch is *identical* at mint and at receipt:

```
## Decision Recorded              ## Plan Approval Recorded
**Timestamp**: …T13:33:59Z        **Timestamp**: …T13:34:24Z
**Event**: DECISION_RECORDED      **Event**: PLAN_APPROVAL_RECORDED
**Stage**: code-generation        **Stage**: code-generation
**Checkpoint**: Code Generation Plan Approval
**Directive Epoch**:              **Directive Epoch**:
  sha256:ec2d6d61e1e77828…          sha256:ec2d6d61e1e77828…   <-- unchanged
**Approval Fingerprint**:         **Approval Fingerprint**:
  sha256:6d7fa2e06bca91fc…          sha256:6d7fa2e06bca91fc…   <-- unchanged
```

A `HUMAN_TURN` (13:34:14) sits between the two — the exact turn boundary that previously destroyed the runtime — and the receipt survives it. Code generation then proceeded normally from 13:35 to 14:00. The `PLAN_APPROVAL_BLOCKED` entries that remain in that ledger are correct guard behaviour, not the loop: 4 before the approval, and the rest after `code-summary.md` was written, from the post-completion write freeze during review.

## Tests

- New `tests/unit/t328-plan-approval-runtime-authority.test.ts` case: a solo probe fires between challenge mint and receipt write; asserts the plan-approval runtime dir is untouched, `directiveEpoch` is unchanged, and `evaluateCodeGenerationApproval(...).ok === true`.
- Updated `tests/unit/t248-steering-content-delivery.test.ts`: a solo probe must now *not* write `.aidlc-active-directive.json`.
- Local run: `bun test` on t328/t248/t68 — 58 pass, 0 fail. `bun scripts/package.ts --check` — all 7 harness trees in sync.

Closes #995.


leandrodamascena (dismissed):
Raja, thank you for isolating the Stop-probe publication defect and adding a regression that demonstrably fails on the base revision.

I reviewed the current head `2eb0bdfadc8285064c8f9d4d0934378715041661` against `main` at `96b11d39028955d4f92375e783525db5275cdfd8`. Project alignment is established by the internal #995 report and the prior maintainer analysis of the observer-mutation class in #978.

## Finding

**[P1] Preserve approved per-Unit authority across `unit start`**

The new regression exercises only the stage-level target (`unit: null`). A normal per-Unit Code Generation flow remains deadlocked.

I reproduced the complete sequence on this head:

1. Create and approve a valid plan for Unit `alpha`.
2. Confirm `evaluateCodeGenerationApproval(...).ok === true`.
3. Run `aidlc-state unit start --stage code-generation --unit alpha`; it succeeds.
4. The same approval immediately evaluates as invalid.

`handleUnit` writes `Active Unit`, `Unit State`, and `Last Updated` without refreshing the active directive marker. The marker therefore remains bound to the previous state digest, becomes unreadable as current authority, and the developer-agent dispatch is refused.

Please preserve the authority issuance digest, revision, source floor, challenge/response, and receipt while updating only the exact routed Unit's lifecycle cache fields. Add regression coverage for `Approve Plan → unit start → developer dispatch`; pause/resume/complete should retain the same authority while continuing to enforce their lifecycle-specific dispatch restrictions.

## Validation

- PR-focused tests: 58 passed.
- Broader Stop/authority suite: 122 passed; one live Claude SDK test could not authenticate because the local security token was expired.
- The PR regression fails against the base exactly as expected.
- The per-Unit reproduction fails on this head.
- The equivalent per-Unit regression passes on PR #978.
- `git diff --check` passed.
- GitHub CI is green for the exact head.

Remaining risk: this PR overlaps the observer-authority work in #978 and both currently claim version `2.7.1`.

apackeer (approved):
Thanks for the reproduction, Leandro. I dug into this today and I think the finding is real but it is a different defect from #995. My take: #997 should land as-is, and the per-Unit `unit start` case should follow in #978 (which I author, so read this with that in mind).

**What I probed.** Same fixture on clean checkouts of `main` (96b11d390), this head (2eb0bdf), and #978 (7fb479a5d): `feature` scope, Unit `alpha`, a run-stage code-generation directive, the real `aidlc-state unit start`, the real `next`/`continue` chain, a real Stop probe (`AIDLC_STOP_HOOK_PROBE=1`), `aidlc-log decision`/`answer`, then `evaluateCodeGenerationApproval` and `beginCodeGeneration`.

| Order | main | #997 | #978 |
|---|---|---|---|
| Approve Plan, then `unit start`, then dispatch (your sequence) | refused | refused | dispatch OK |
| `unit start`, fresh `next`, plan, Approve Plan with a Stop probe between mint and receipt, dispatch | deadlock (#995) | dispatch OK | dispatch OK |

The second row is the order the protocol prescribes: stage-protocol-construction.md ("Unit lifecycle receipts") brackets the Unit's work with `unit start` before the body, and Plan Approval is inside the body. On `main` that row deadlocks exactly as #995 describes: the probe re-publishes load-steering, the runtime dir is emptied, and `answer` is refused. On this head the probe publishes nothing, the epoch is unchanged, the receipt records, and the developer dispatch is authorized. So this PR fixes #995 for both stage-level scopes (bugfix, express, poc, refactor, security-patch) and per-Unit scopes (classic, enterprise, feature, mvp, workshop) in the documented order.

**Your sequence is a pre-existing gap, not a regression of this PR.** Approving before `unit start` fails identically on `main`. Two things break it: `requireEngineRoutedUnit` spawns `next` with only `AIDLC_ROUTE_CHECK=1`, so that publication still fails `preserveCodeGenerationAuthority` and resets the runtime dir; and the state write leaves the marker's `state_sha256` stale while `directiveEpoch` hashes that digest, so even a refreshed marker would change the epoch. Fixing it needs the route check made non-publishing, a unit-scoped marker refresh, a pinned issuance digest on the marker, and the epoch change. That is roughly 100 core lines across four files, and it already exists in #978 with the regression you asked for (`approved per-Unit authority survives unit start and authorizes developer dispatch`). Pulling it into this PR would duplicate that work and stall the #995 fix behind it.

**Proposal.** Merge #997 now as 2.7.1. I rebase #978 onto it, re-slot to 2.7.2, and drop the overlapping `emit()` change there. Solo workflows get unblocked today; the approve-before-start order and the rest of the observer hardening follow in #978.

For the record, one thing neither PR covers: a TaskUpdate-driven status sync (`aidlc-sync-workflow-state` -> `aidlc-utility set-status`) between Approve Plan and dispatch still resets the approval on all three heads, because `set-status` calls `refreshActiveDirectiveMarker`. Uncommon path; I will track it as a separate follow-up.

<details>
<summary>Probe output (L = approve then unit start, D = documented order with Stop probe, E = status sync after approval)</summary>

`main` @ 96b11d390
```
[L] answer exit=0
[L] after approve: ok=true reason="approved" runtimeFiles=1 markerReadable=true
[L] unit start exit=0
[L] after unit start: ok=false reason="Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`" runtimeFiles=0 markerReadable=false
[L] dispatch=REFUSED: Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`
[D] unit start exit=0
[D] after unit start: markerReadable=false
[D] authority after unit start: Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`
[D] conductor next(+continue x1) status=0 kind=run-stage stage=code-generation unit=alpha markerReadable=true
[D] stop probe exit=0 kind=load-steering runtimeFiles before=1 after=0 epochSame=false (THROWS: Code Generation approval authority requires a run-stage or invoke-swarm directive, got "load-steering")
[D] answer exit=1 {"error":"Code Generation approval authority requires a run-stage or invoke-swarm directive, got \"load-steering\""}
[D] final: ok=false reason="Code Generation approval authority requires a run-stage or invoke-swarm directive, got "load-steering"" markerReadable=true
[D] dispatch=REFUSED: Code Generation approval authority requires a run-stage or invoke-swarm directive, got "load-steering"
```

#997 @ 2eb0bdf
```
[L] answer exit=0
[L] after approve: ok=true reason="approved" runtimeFiles=1 markerReadable=true
[L] unit start exit=0
[L] after unit start: ok=false reason="Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`" runtimeFiles=0 markerReadable=false
[L] dispatch=REFUSED: Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`
[D] unit start exit=0
[D] after unit start: markerReadable=false
[D] authority after unit start: Code Generation approval authority is unavailable because the active directive is missing, stale, or legacy; run a fresh `next`
[D] conductor next(+continue x1) status=0 kind=run-stage stage=code-generation unit=alpha markerReadable=true
[D] stop probe exit=0 kind=load-steering runtimeFiles before=1 after=1 epochSame=true ()
[D] answer exit=0 {"emitted":"PLAN_APPROVAL_RECORDED","checkpoint":"plan-approval","stage":"code-generation"}
[D] final: ok=true reason="approved" markerReadable=true
[D] dispatch=OK
[E] approved: ok=true reason="approved" runtimeFiles=1
[E] set-status exit=0 {"updated":true,"phase":"CONSTRUCTION","stage":"code-generation","agent":"aidlc-developer-agent"}
[E] after status sync: ok=false reason="Code Generation approval authority requires a run-stage or invoke-swarm directive, got "error"" runtimeFiles=0 markerReadable=true
[E] dispatch=REFUSED: Code Generation approval authority requires a run-stage or invoke-swarm directive, got "error"
```

#978 @ 7fb479a5d
```
[L] answer exit=0
[L] after approve: ok=true reason="approved" runtimeFiles=1 markerReadable=true
[L] unit start exit=0
[L] after unit start: ok=true reason="approved" runtimeFiles=1 markerReadable=true
[L] dispatch=OK
[D] unit start exit=0
[D] after unit start: markerReadable=true
[D] authority after unit start: resolves
[D] conductor next(+continue x1) status=0 kind=run-stage stage=code-generation unit=alpha markerReadable=true
[D] stop probe exit=0 kind=load-steering runtimeFiles before=1 after=1 epochSame=true ()
[D] answer exit=0 {"emitted":"PLAN_APPROVAL_RECORDED","checkpoint":"plan-approval","stage":"code-generation"}
[D] final: ok=true reason="approved" markerReadable=true
[D] dispatch=OK
[E] approved: ok=true reason="approved" runtimeFiles=1
[E] set-status exit=0 {"updated":true,"phase":"CONSTRUCTION","stage":"code-generation","agent":"aidlc-developer-agent"}
[E] after status sync: ok=false reason="Code Generation approval authority requires a run-stage or invoke-swarm directive, got "error"" runtimeFiles=0 markerReadable=true
[E] dispatch=REFUSED: Code Generation approval authority requires a run-stage or invoke-swarm directive, got "error"
```
</details>

apackeer:
Thanks again for the reproduction, Leandro. I am dismissing the request-changes and merging this as 2.7.1 so solo workflows are unblocked today.

The per-Unit `unit start` finding is real, but it reproduces identically on `main` at 96b11d390, so it is a pre-existing gap rather than a regression of this PR. It is already fixed, with the `Approve Plan -> unit start -> developer dispatch` regression you asked for, in #978. I will rebase #978 onto this merge, re-slot it to 2.7.2, and drop the `emit()` change that overlaps with this one.

Keeping the #995 fix to this single predicate change also keeps this PR short and reviewable, and lets the observer-authority work in #978 be reviewed on its own terms. Happy to revisit if you disagree.

PR #1003, opened 2026-09-02
fix: Frontend mockups design artefacts are not flowing to the downstream stages (#999)

### LINKED ISSUE #999: Frontend mockups design artefacts are not flowing to the downstream stages
## Summary

Design intent captured in a mockup artifact has **no declared consumer in the construction phase**, so the layout the human approved at `rough-mockups` cannot reach `code-generation` through the artifact contract. In a real run it reached code generation only because agents opportunistically read files that were not declared inputs — and the part of the layout that had no textual carrier (vertical section order) silently inverted.

This is a traceability gap, not a model-quality problem. No sensor detects it and no gate reports it.

### 1. What was approved at `rough-mockups`

From the run's `ideation/rough-mockups/wireframes.md`, Screen 2 (empty state), grounded in the confirmed answer `[Q3:B]` — *"Single-column layout: the saved-segments list sits on top, the condition builder below it, and the live count sits beside the builder's action row."*

```
+---------------------------------------------------------------------------+
| Segments                                                                  |
+---------------------------------------------------------------------------+
| Saved segments                                          [ + New segment ] |
|                                                                           |
|   No segments yet. Build one below to see how many customers qualify.     |
+---------------------------------------------------------------------------+
| Build a segment                                                           |
|                                                                           |
|  Match  ( AND v )  of the following:                                      |
|  +---------------------------------------------------------------------+  |
|  |   ( no conditions yet )                                             |  |
|  |   [ + Add condition ]   [ + Add sub-group ]                         |  |
|  +---------------------------------------------------------------------+  |
|                                                                           |
|  Matching customers: —          [ Segment name: ____________ ] [ Save ]   |
+---------------------------------------------------------------------------+
```

The same decision was restated at the ideation gate in `ideation/approval-handoff/decision-log.md`:

> Layout: single column — saved list on top, builder below, count by action row. `[Q3:B]`

### 2. What `code-generation` actually produced

Same empty state, rendered:

```
+---------------------------------------------------------------------------+
| Travel Insurance Admin                                                    |
| Policies   Customers   Products   Insights   [ Segments ]                 |
+---------------------------------------------------------------------------+
| Customer segments                                                         |
| Build a boolean rule over non-PII customer attributes, then click         |
| Calculate count ...                                                       |
| +-----------------------------------------------------------------------+ |
| | New segment   [ New ]                                                 | |
| | +-------------------------------------------------------------------+ | |
| | | Match  [ ALL (AND) v ]                                            | | |
| | | [ Add condition ]  [ Add sub-group ]                              | | |
| | +-------------------------------------------------------------------+ | |
| | +-------------------------------------------------------------------+ | |
| | | [ Calculate count ]   —                                           | | |
| | +-------------------------------------------------------------------+ | |
| | Segment name* [____________]        Owner label [____________]        | |
| | [ Create ]                                                            | |
| +-----------------------------------------------------------------------+ |
|                                                                           |
|   ( "Saved segments" table renders HERE, below the builder )              |
+---------------------------------------------------------------------------+
```

### 3. The drift

| Wireframe decision | Generated code | Match |
|---|---|---|
| Segments as 5th top-level tab, appended after Insights `[Q1:A]` | `App.jsx` `TABS[4] = { key: 'segments', ... }` | yes |
| Single column, no new design system `[Q5:A]` | single column, existing CMS styles | yes |
| Type-aware condition rows (attribute / operator / value) | `ATTRIBUTES`, `OPERATORS`, `NO_VALUE`, `TWO_VALUE`, `MULTI_VALUE` | yes |
| Max 2 nesting levels; "Add sub-group" disabled inside a sub-group | `GroupEditor({ group, depth, ... })`, depth-gated | yes |
| Owner column, Load / Delete row actions | present in the saved-segments table | yes |
| **Saved-segments list ABOVE the builder** | **builder first; `<h3>Saved segments</h3>` renders last** | **no** |
| Empty-state guidance above the builder ("No segments yet. Build one below…") | absent from the top of the view | **no** |
| Count beside the name + Save action row | count in its own card above the name/Save row | partial* |

\* partly explained by an approved in-flight change (`BR4.3`: live count replaced by an explicit **Calculate count** button, which supersedes `FR4.2`). The vertical ordering has no such explanation.

Everything that survived had a downstream carrier — a requirement id, a component name, or a `data-testid`. Everything that lived only in the ASCII diagram did not.

### Finding 1 — the design-intent chain terminates in inception

`wireframes` is declared in `consumes` by exactly two stages:

- `core/aidlc-common/stages/ideation/approval-handoff.md` (`required: false`)
- `core/aidlc-common/stages/inception/refined-mockups.md` (`required: true`)

And `refined-mockups`' own outputs go nowhere useful:

| Artifact | Declared consumers |
|---|---|
| `mockups` | `refined-mockups`, `inception/delivery-planning` |
| `interaction-spec` | **none** |
| `design-system-mapping` | **none** |
| `accessibility-checklist` | **none** |
| `user-flow` | `refined-mockups` only |

So **no construction-phase stage declares any mockup artifact as an input, in any scope.** Even a full run that executes `refined-mockups` produces an `interaction-spec` that nothing is contractually obliged to read. Skipping `refined-mockups` (as scoped runs do) just removes the last hop the chain had.

### Finding 2 — the layout reached code anyway, off-contract

In the run traced here (scope `segments-builder-feature`, 11 of 33 stages, `2.5 refined-mockups — SKIP`):

- `inception/requirements-analysis` declares `consumes: intent-statement, scope-document, business-overview, architecture, code-structure, team-practices`. Its `requirements.md` nonetheless cites `[rough-mockups Q1:A]`, `[rough-mockups]` (×3) and `[rough-mockups R-01]` — an undeclared upstream.
- `construction/functional-design` declares `consumes: unit-of-work, unit-of-work-story-map, requirements, components, contract-summary`. Its `frontend-components.md` reproduces the wireframe's component tree and the full type-aware operator table, but cites **only** `team.md` and FR/BR ids. Zero citations to `wireframes.md`, so the provenance is lost at exactly the hop that re-authors the layout.
- `construction/code-generation` does not list `frontend-components` in `consumes` (nothing does). Its `code-generation-plan.md` Step 15 cites `frontend-components.md` explicitly anyway.

Net effect: design intent propagates by agent initiative, not by contract. That is unauditable, and it is why one dimension of the layout could vanish without anything failing.

### Finding 3 — nothing detects it

- No sensor checks layout fidelity. `functional-design/traceability.json` explicitly marks the layout requirement out of scope: `{ "id": "FR1.2", "status": "N/A", "target": "UI layout; frontend-components" }`.
- `upstream-coverage` **failed 4× on `requirements-analysis`** (fire ids `55c63db2`, `5360f013`, `5e0f8a81`, `a8e4310a`), each reporting `intent-statement` and `team-practices` as `unreferenced` — i.e. it correctly noticed the stage was not citing its declared upstreams while it was citing an undeclared one. The stage was approved regardless; the failures did not surface at the gate.

## Suggested direction

Not prescriptive, but the options as I see them:

1. **Close the contract.** Add a UI-design artifact to `consumes` for `functional-design` and `code-generation` (`interaction-spec` / `mockups`, `required: false`, `conditional_on: ui`), and add `frontend-components` to `code-generation`'s `consumes` since it is already a de-facto input.
2. **Make `rough-mockups` output a durable layout contract.** The ASCII diagram is not machine-checkable and not citable. If the stage emitted named, ordered regions (e.g. `L1 saved-segments-list`, `L2 condition-builder`, `L3 count-and-save`), downstream artifacts could cite `[L1]`/`[L2]` the way they cite `FR`/`BR`, and `traceability` could verify coverage instead of marking layout `N/A`.
3. **Require provenance when an artifact re-authors upstream content.** `frontend-components.md` restating the wireframe without citing it is the single point where the audit trail breaks.
4. **Surface `upstream-coverage` failures at the approval gate,** or make citing an undeclared upstream an explicit, recorded deviation rather than an invisible one.

## Environment

- Scope: Custom scope generated by Composer, 11 of 33 stages, brownfield
- Skipped: `2.5 refined-mockups` (among 22 skipped stages)
- Target project: React + Vite frontend, Spring Boot / Maven backend
- Stages traced: `rough-mockups` → `approval-handoff` → `requirements-analysis` → `functional-design` → `code-generation`

spraja08:
Closes #999.

## Problem

Design intent captured in a mockup had **no declared consumer downstream of inception**. `wireframes` was consumed only by `approval-handoff` and `refined-mockups`; `refined-mockups`' own outputs (`mockups`, `interaction-spec`, `design-system-mapping`, `accessibility-checklist`) had no construction-phase consumer at all. So the layout a human approved reached `code-generation` only because agents opportunistically read files that were not declared inputs — and the one dimension of that layout with no textual carrier (top-to-bottom section order) silently inverted, with no sensor or gate reporting it.

## Fix — close the contract

The chain is now declared end to end. Every addition is `required: false`, so non-UI initiatives and scopes that skip the mockup stages are unaffected.

| Stage | Added to `consumes` | Added instruction |
|---|---|---|
| `inception/requirements-analysis` | `wireframes`, `user-flow` | Read the rough mockups if present; when a requirement traces to something the wireframes decided (a screen, a flow, a layout choice), **cite the wireframe** rather than restating it as unattributed prose. |
| `construction/functional-design` | `wireframes`, `mockups`, `interaction-spec` | `frontend-components.md` must reproduce the approved screen composition — **including the top-to-bottom order of sections** — and cite the mockup it derives from. A departure is a recorded deviation, not a silent one. |
| `construction/code-generation` | `frontend-components`, `mockups`, `interaction-spec`, `wireframes` | New Critical Rule: UI code follows the approved layout, not the order that happens to be convenient in code. |

`upstream-coverage` targets were extended on `requirements-analysis` and `functional-design`, so a mockup artifact that exists on disk but is never referenced is now flagged (advisory).

This addresses directions **1**, **3**, and part of **4** from the issue. Direction 2 (emitting named, machine-checkable ordered regions such as `L1`/`L2` that `traceability` could verify) is deliberately left out of this PR — it changes the `rough-mockups` output shape and deserves its own change.

## Verification

Re-ran the affected chain on the same initiative (`260902-customer-segments`) with `refined-mockups` executed rather than skipped.

**Expectation** — the approved layout in `inception/refined-mockups/mockups.md`, chosen layout `Q1=B` (single flat, indented condition list):

```
+---------------------------------------------------------------+
| Segments                                                      |
+---------------------------------------------------------------+
| Name: [ High-value APAC leisure         ]  [New] [Save]       |
|                                                               |
|  Match [ALL v] of the groups                                  |
|                                                               |
|  Group 1  — match [ANY v]                          [x group]  |
|    - [Region        v] [is one of  v] [APAC][EUROPE +] [x]    |
|      [+ add condition]                                        |
|                                                               |
|  [+ add group]                                                |
|                                                               |
|  [ Get count ]      12 customers match                        |
+---------------------------------------------------------------+
```

followed by the `Saved segments` region — a table with `Name | Created by | Actions (Load / Edit / Delete)`.

**Result** — the generated screen (screenshot below) matches the approved mockup section by section:

| Approved in `mockups.md` | Generated | Match |
|---|---|---|
| Name input + `New` / `Save` at the top of the builder card | `Name` + `Created by` inputs, `New` / `Save` beside them | yes |
| `Match [ALL] of the groups` between the name row and the groups | present, in that position | yes |
| Group row carrying its own `ALL/ANY` + `× group`, indented condition rows below it | `Group 1 — match [ALL] × group`, then the indented `Field… / Operator… / ×` row | yes |
| `+ add condition` under the group's conditions, `+ add group` after the groups | both present, in that order | yes |
| `Get count` at the bottom of the builder | present, bottom of the card | yes |
| **`Saved segments` table below the builder, `Name / Created by / Actions`** | **renders below the builder with exactly those columns and `Load Edit Delete`** | **yes** |
| Empty-value states (`Field…`, `Operator…` placeholders) | present | yes |

The section order that inverted in the run reported in #999 is now the order the mockup specifies, and `frontend-components.md` cites the mockup it derived it from — so the fidelity is auditable rather than incidental.

## Release

Bumped to `2.7.2` with a matching `CHANGELOG.md` entry and README badge; `dist/` regenerated (`bun scripts/package.ts --check` → all harness trees in sync).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.7.2] - 2026-09-02
+
+Close the design-intent contract so an approved layout reaches code generation by declaration rather than by an agent happening to read an undeclared file. Rough/refined mockups previously had no downstream consumer: `functional-design` re-authored the layout without citing the wireframes, and `code-generation` had no declared UI-design input — so a section order that lived only in an ASCII diagram could silently invert with nothing flagging it. `requirements-analysis`, `functional-design`, and `code-generation` now declare the mockup artifacts (`wireframes`, `user-flow`, `mockups`, `interaction-spec`) and `frontend-components` as optional inputs, read them when present, and are instructed to preserve the approved screen composition — including top-to-bottom section order — or record a deviation. All additions are `required: false`, so non-UI initiatives and scopes that skip the mockup stages are unaffected. **Upgrade:** replace the `dist/<harness>/` tree; no workflow state migration is required. Closes #999.
+
+* `functional-design`'s `frontend-components.md` must now cite the mockup it derives a UI layout from; on UI units the `upstream-coverage` sensor (advisory) will flag a mockup artifact that exists on disk but is never referenced.
+* `code-generation` now lists `frontend-components`, `mockups`, `interaction-spec`, and `wireframes` as consumed inputs and carries a Critical Rule to follow the approved layout order.
+
 ## [2.7.1] - 2026-09-01
 
 Fix a Plan Approval deadlock that made Code Generation unreachable on solo (non-team) workflows. The Stop hook's read-only `next` probe published the durable active-directive marker on every turn boundary, which bumped the Code Generation authority revision and reset the plan-approval runtime, so the approval challenge minted while answering "Approve Plan" was destroyed before its receipt could be written. The probe no longer publishes that marker for any workflow, matching the read-only contract it already advertised. **Upgrade:** replace the `dist/<harness>/` tree; no workflow state migration is required. Closes #995.
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ A native implementation of the **AI-DLC methodology** (AI-Driven Development Lif
 
 The methodology lives once, in a harness-neutral `core/`; each harness adds a thin surface that decides how it shows up on that harness. So you edit the methodology in one place, and every harness distribution is generated from it — no harness gets special treatment. (See [Repository layout](#repository-layout) for how the pieces fit together.)
 
-![version](https://img.shields.io/badge/version-2.7.1-blue)
+![version](https://img.shields.io/badge/version-2.7.2-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 ![Kiro IDE](https://img.shields.io/badge/harness-Kiro%20IDE-orange)
 ![Kiro CLI](https://img.shields.io/badge/harness-Kiro%20CLI-orange)
--- a/core/aidlc-common/stages/construction/code-generation.md
+++ b/core/aidlc-common/stages/construction/code-generation.md
@@ -35,6 +35,14 @@ consumes:
     required: true
   - artifact: requirements
     required: true
+  - artifact: frontend-components
+    required: false
+  - artifact: mockups
+    required: false
+  - artifact: interaction-spec
+    required: false
+  - artifact: wireframes
+    required: false
 requires_stage:
   - units-generation
   - functional-design
@@ -70,6 +78,7 @@ outputs: application code + code-generation-plan.md, code-generation-questions.m
 - Application code goes to workspace root, NEVER to the record dir
 - Brownfield: modify files in-place. NEVER create duplicates like ClassName_modified.java
 - Add data-testid attributes to interactive UI elements for test automation
+- UI code must follow the approved layout. When `frontend-components.md`, the refined mockups (`mockups.md`/`interaction-spec.md`), or the rough `wireframes.md` exist, preserve the screen composition they specify — including the top-to-bottom order of sections — rather than the order that happens to be convenient in code. A departure from the approved layout is allowed only when an upstream design artifact records it as a deliberate change.
 - Before review, write `source-manifest.json` listing every application-source path this unit created, modified, or deleted, including shell-, scaffolding-, and generator-written files
 - Measurable quality targets from NFR Requirements, NFR Design, and the Testing
   Contract coverage floor are inputs, not suggestions. NEVER relax, lower, or
@@ -88,6 +97,7 @@ Read all design artifacts for the current unit:
 - Unit definition from `<record>/inception/units-generation/unit-of-work.md` (if exists)
 - Story map from `<record>/inception/units-generation/unit-of-work-story-map.md` (if exists)
 - Requirements from `<record>/inception/requirements-analysis/requirements.md` (if exists)
+- For a UI unit, the approved layout: `frontend-components.md` in this unit's functional-design dir, the refined mockups from `<record>/inception/refined-mockups/` (`mockups.md`, `interaction-spec.md`), and the rough `wireframes.md` from `<record>/ideation/rough-mockups/` (whichever exist). These are optional and absent for non-UI units or scopes that skip the mockup stages.
 
 Incremental scopes (bugfix, poc, refactor, security-patch) and the zero-Unit
 `express` scope skip Units Generation by design. When those inputs are absent,
--- a/core/aidlc-common/stages/construction/functional-design.md
+++ b/core/aidlc-common/stages/construction/functional-design.md
@@ -36,6 +36,12 @@ consumes:
     required: true
   - artifact: contract-summary
     required: false
+  - artifact: wireframes
+    required: false
+  - artifact: mockups
+    required: false
+  - artifact: interaction-spec
+    required: false
 requires_stage:
   - units-generation
   - contract-design
@@ -52,7 +58,7 @@ scopes:
   - refactor
   - classic
   - workshop
-inputs: unit-of-work.md, unit-of-work-story-map.md, requirements.md, domain-design components.md, contract-design contract-summary.md (if produced)
+inputs: unit-of-work.md, unit-of-work-story-map.md, requirements.md, domain-design components.md, contract-design contract-summary.md (if produced), mockups.md/interaction-spec.md or wireframes.md (if a UI unit and the mockups exist)
 outputs: "entities.md, rules.md, functional-spec.md, traceability.json, CONDITIONAL: frontend-components.md (under this stage's per-unit record dir, engine-resolved); per-kind applicability via produces_kinds (untagged unit: all). entities.md and rules.md each carry a fenced ```yaml source-of-truth block; functional-spec.md is the source of truth for workflows and state machines and carries derived ER-diagram and rules-summary views."
 ---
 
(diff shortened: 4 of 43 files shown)


apackeer (changes requested):
Thanks for tracing #999 to the real cause. "The mockups were never declared as inputs" is exactly right, and fixing it in `consumes:` means the directive, the sensors, and stage validity all pick it up without anyone remembering to. I want this wiring merged.

I reviewed head 01d2d1224 against main a277af218f. Packaging parity and t65/t66/t68 pass locally. Three changes before merging:

1. **A regression test that fails without the wiring.** The t65 change is needed, but it is graph bookkeeping: if you reverted every new consume and kept it, CI would still be green. `tests/unit/t116-directive-path-resolution.test.ts` already spawns `next` for functional-design and code-generation from fixtures and asserts on resolved `consumes`. Add cases there that seed `wireframes.md`, `mockups.md`, and `interaction-spec.md` in the record and assert their producer-keyed paths appear in both directives' `consumes` (and `frontend-components` in code-generation's). t116 has no requirements-analysis directive today, so cover that stage either with a new fixture pivot or with a graph-level assertion that `wireframes` and `user-flow` resolve requirements-analysis as a consumer. Absence semantics are already pinned in t116; no need to re-test them. The file's `covers:` header can then claim `stage:construction/functional-design` and `stage:construction/code-generation`, both `UNCOVERED` in the registry today.

2. **Backend units now get a false `SENSOR_FAILED`.** Reproduced on this head through the shipped dispatcher: feature scope, `mockups.md` / `interaction-spec.md` / `wireframes.md` present, a `billing-service` unit whose functional-design deliverables cite all four of their real inputs. Result: `upstream-coverage` fails with `unreferenced: [wireframes, mockups, interaction-spec]`, findings 3, audit row plus detail file. The same fixture passes on main. The cause is that `presentConsumes` (`core/tools/aidlc-sensor.ts:319-329`) and `resolveConsumes` (`core/tools/aidlc-orchestrate.ts:2529-2547`) filter by on-disk presence and producer kind only, and `conditional_on` knows only brownfield/greenfield, so a per-unit consumer has no way to say "UI units only". A deterministic check that fails on correct output is a regression, so this needs a fix rather than a note: for example a `consumes[].kinds` field mirroring `produces_kinds`, applied in both places, with the backend-unit fixture above as its test. If you would rather land the engine piece first as its own PR and rebase this one on it, that works too.

3. **Narrow the claim.** Nothing on this head can detect the order flip #999 reported: `upstream-coverage` only checks that the word "mockups" appears somewhere in the deliverables, and code-generation does not import it. Please change "Closes #999" to "Refs #999", title the PR as wiring the design-intent consumer chain, and reword "will flag" in the body and CHANGELOG to "records in the audit log (advisory)". #999 stays open for a machine-checkable layout contract and a fidelity sensor. One more line for the upgrade note: the validity fingerprint includes `consumes`, so in-flight workflows with a completed requirements-analysis, functional-design, or code-generation will see the stale-validity advisory on every `next` until those stages are re-completed.

Small: `docs/reference/04-stages/construction.md` still lists the old functional-design inputs.

vnlebaoduy:
Taking up the engine piece from review item 2 as its own PR, per the "land the engine piece first" option: an optional `consumes[].kinds` (same value set as `produces_kinds`, valid only on `for_each: unit-of-work` stages), applied in both `resolveConsumes` (`aidlc-orchestrate.ts`) and the upstream-coverage consume threading (`aidlc-sensor.ts`), with the backend-unit fixture from the review as the regression test. No stage `consumes:` changes in that PR — this one can then rebase onto it and tag `wireframes` / `mockups` / `interaction-spec` with `kinds: [ui]`.

@spraja08 shout if you already have this in flight and I will stand down.

vnlebaoduy:
Engine piece is up as #1132 (`consumes[].kinds`, no stage changes). Once it lands this PR can rebase and tag `wireframes` / `mockups` / `interaction-spec` with `kinds: [ui]` on functional-design and code-generation.

PR #1004, opened 2026-09-02
fix: emit permissionDecision allow on Codex PreToolUse input rewrites

jisu-im-dev:
# Summary

On Codex CLI, a PreToolUse `hookSpecificOutput.updatedInput` rewrite is applied only when the same envelope carries an explicit `permissionDecision: "allow"`. Without it the rewrite is silently dropped and the original tool input runs, so both Codex rewrite paths were inert: `spawn_agent` dispatches never received the active-stage rule bundle from `deliver-stage-rules`, and Bash commands never inherited the validated payload session prefix (`AIDLC_SESSION_OVERRIDE`).

## Changes

* `harness/codex/hooks/aidlc-codex-adapter.ts`: `wrapUpdatedInput` now emits `permissionDecision: "allow"` alongside `updatedInput`, and the `deliver-stage-rules` target re-wraps the core hook's output through a new `authorizeUpdatedInput` helper instead of forwarding it verbatim. Output without an `updatedInput` (or unparseable output) passes through untouched, and the re-wrapped response is what gets persisted for duplicate-delivery replay.
* The harness-neutral `core/hooks/aidlc-deliver-stage-rules.ts` is deliberately **unchanged**: it still emits no permission decision (as pinned by t248), so Claude Code and every other harness keep their native approval flow. The `allow` decision authorizes only the rewritten input on Codex; Codex's own approval flow still governs whether the call runs.
* `tests/unit/t149-codex-hook-adapter.test.ts`: pins `permissionDecision: "allow"` on both the `bind-bash-session` rewrite and the `deliver-stage-rules` dispatch rewrite.
* Version bumped to 2.8.2 (rebased over the 2.8.x native-distribution releases) with a matching CHANGELOG entry; local projections regenerated via `bun scripts/package.ts` (dist trees are no longer committed).

## User experience

Before: on Codex, subagent dispatches ran with the unmodified prompt (no active-stage rule bundle) and Bash commands ran without the validated session binding — both silently, since the dropped rewrite produces no error.

After: `spawn_agent` dispatches carry the exact active-stage rule bundle and Bash commands inherit the validated payload session again. Other harnesses are unaffected.

## Checklist

* [x] Changes are documented

## Test plan

* `bun scripts/package.ts && bun scripts/package.ts --check` — all harness trees in sync, no drift.
* `bun run typecheck` and `bun run lint` — clean.
* `bun test tests/unit/t149-codex-hook-adapter.test.ts tests/unit/t248-steering-content-delivery.test.ts tests/unit/t150-codex-packaging.test.ts tests/unit/t228-hook-run-exports.test.ts` — 109 pass, 0 fail (includes the two new `permissionDecision` pins and the existing t248 pin that the core hook emits **no** decision).
* Adapter-adjacent suites (t01, t02, gen-coverage-registry, t144, t146, t230, t265, t308, t318, t145, t55, t239, t68, t121, t188, t300) — all pass.
* `bun tests/run-tests.ts --smoke --unit --parallel 8` — the only failing files fail identically on an unmodified checkout of `main` in my environment (pre-existing local-environment failures; verified by stash-and-rerun diff of the failure sets).

## Acknowledgment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.8.2] - 2026-09-09
+
+Make Codex apply PreToolUse input rewrites again. Codex only honors a `hookSpecificOutput.updatedInput` rewrite when the same envelope carries an explicit `permissionDecision: "allow"`; without it the rewrite is silently dropped and the original tool input runs. The Codex adapter now emits that decision on both of its rewrite paths — the Bash session-binding prefix and the `deliver-stage-rules` core output it forwards — so subagent dispatches receive the active-stage rule bundle and Bash commands inherit the validated session again. The harness-neutral core hook is unchanged: it still emits no permission decision, and every other harness keeps its native approval flow. **Upgrade:** `aidlc update` (or `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`), then `aidlc config` to refresh the project's Codex hook tree; no workflow state migration is required.
+
+* `spawn_agent` dispatches on Codex carry the exact active-stage rule bundle instead of the unmodified prompt.
+* Bash commands on Codex inherit the validated payload session via the `AIDLC_SESSION_OVERRIDE` prefix again.
+
 ## [2.8.1] - 2026-09-08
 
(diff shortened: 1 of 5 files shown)


leandrodamascena:
Hi @jisu-im-dev, thank you for investigating the Codex PreToolUse rewrite issue and adding regression coverage for both rewrite paths.

I reviewed head `305575d5370e39adc40ab066034ca081a2bf9509` against base `c03f9e280be1d545e78e26641a4560677c82443e`. PR #1064 merged the required behavior and regression coverage as commit `227745d03bdd1879f51409eeba051e4c6c585b5f`. This PR is superseded. Merging it applies the fix twice and carries version and changelog changes based on the older base. I am closing it without merging.

Thank you again for the diagnosis and implementation.

PR #1005, opened 2026-09-03
fix: separate per-unit artifact folders from stage folders under construction/ (#862)

### LINKED ISSUE #862: [Feature]: Separate per-unit artifact folders from stage-diary folders under construction/
### Description

In v2, under a workflow's per-intent `construction/` record directory, two structurally *different* kinds of subfolder live at the **same level**:

- **Per-unit artifact folders**, named by unit slug — e.g. `construction/mined-value-schema/`, `construction/contract-synthesizer/` — each containing per-stage artifact subdirs (`functional-design/`, `nfr-design/`, `code-generation/`, ...) with the actual deliverables (`entities.md`, `rules.md`, `traceability.json`, etc.).
- **Stage-diary folders**, named by stage — e.g. `construction/functional-design/`, `construction/nfr-design/`, `construction/code-generation/` — that hold only the shared, cross-unit stage `memory.md` observation diary.

The two collide by name and by level. `construction/functional-design/memory.md` is the **shared stage diary**, while `construction/<unit>/functional-design/` is **one unit's artifacts** for that same stage. Browsing `construction/`, you cannot tell a unit folder from a stage-diary folder without already knowing the unit slugs — and the identical inner stage names (`functional-design/` appearing both as a top-level diary folder and as a child of every unit) make it easy to open the wrong one.

Proposal: nest the two axes under **distinct roots** so they never share a level, e.g.

```
construction/
  units/
    mined-value-schema/
      functional-design/ ...
    contract-synthesizer/
      functional-design/ ...
  _stages/                # or fold the diary into each unit's stage dir
    functional-design/memory.md
    nfr-design/memory.md
```

The exact scheme is open (a `units/` + `_stages/` split, a prefix convention, or moving the diary into the per-unit stage dir) — the ask is simply that per-unit artifact folders and harness/stage-diary folders not be siblings at the same level.

### Use Case

Unit-major Construction with many units. On a live v2 run with 11 units, humans and reviewers browsing the record repeatedly mistake a stage-diary folder for a unit (and vice versa), because both sit directly under `construction/` and reuse the same stage names. A clear structural separation makes the record self-describing and navigable at a glance, and removes the "which `functional-design/` is this?" ambiguity.

This is orthogonal to:
- **#795** (units-major *parallelism* / teams owning units) — that is about execution ordering, not on-disk layout.
- **#126** (multi-intent namespacing, closed) — already solved by the `intents/<slug>/` directory; this is purely the *intra-`construction/`* organization.

### Version

v2 (alpha)

### Area

Construction phase rules

### Additional Context

Observed on a real v2 unit-major run (Kiro IDE harness). The stage-diary folders are created lazily (only for stages that have run), so early in a run `construction/` shows a partial, mixed set of unit-slug and stage-name folders, which compounds the confusion. Backward-compatibility note: any change here touches the engine's artifact-path resolution (`produces[]` path templates + the stage `memory_path`), so it would want a migration/compat shim for in-flight workflows.

apackeer:
Per-unit Construction artifacts and diaries move to a dedicated `construction/units/<unit>/<stage>/` axis, so unit folders and stage-level folders (shared stage diaries, non-per-unit stage artifacts like build-and-test) never share a level and the record is self-describing at a glance.

The collision was machine-visible, not just human confusion: the plan-approval guard counted every `construction/` child as a unit (stage-diary dirs included), and the reviewer read-scope hook could not distinguish a shared stage-diary read from a sibling-unit read. All engine resolvers, enforcement hooks (reviewer read-scope now fails closed on legacy shapes), sensors, traceability, review fingerprints/receipts, and the team-owned-units surfaces now route through the units axis; zero-Unit and no-DAG fallbacks deliberately stay stage-level. `/aidlc --doctor` gains a legacy-layout advisory that detects legacy unit directories from DAG membership and directory shape (so units named `units` or after a stage slug are found, and a plain stage-diary directory is not) and prints an idempotent, record-qualified migration per unit - `if [ -d <record>/construction/<unit> ] && [ ! -e <record>/construction/units/<unit> ]; then mkdir -p <record>/construction/units && mv … ; fi`, with a two-step form through `.units-legacy` for a unit named `units`. The upgrade note lives in `docs/guide/15-troubleshooting.md` (pre-upgrade per-unit review receipts invalidate because logical fingerprint paths changed). CI Pipeline and the five Operation stages now read per-unit NFR/infrastructure artifacts from the unit axis with explicit every-Unit aggregation.

Verified with the full default tier green (smoke+unit+integration, 388 files / 0 failed) plus targeted construction-surface slices on the final rebase; `bun scripts/package.ts --check`, typecheck, and lint all pass. No version or changelog bump (release-preparation PR's job). Closes #862


rafaelgsr (changes requested):
Thanks for separating the unit and stage axes; this is aligned with linked issue #862, and I found no superseding implementation.

I reviewed the current head `e95c6373bfebd90c6b538137433e8d033cabd23e` against current `main` `1c64d9a60754dc7e0acdae6bcad6c7b01e6fb155` (merge base `a277af218f0df7f325d3b8be7b6d90fce2c5bd40`).

- **[P1] Native release construction is broken.** `t238-build-binaries` fails because `readSyncBufferedLine` retains buffer views across a refill; this tree makes a Git batch header cross the 64 KiB boundary, corrupting it. Bolt and swarm re-entry then fail because the base source listing cannot be computed. Fix the buffered reader and add a boundary-crossing regression. See `core/tools/aidlc-lib.ts:10927` and `tests/unit/t238-build-binaries.test.ts:120`.

- **[P2] The documented migration command fails as printed.** The first `mv construction/<unit> construction/units/<unit>` fails because `construction/units/` does not exist; from the usual project-root working directory, it also addresses the wrong directory. Emit an idempotent, active-record-qualified migration including parent creation. See `CHANGELOG.md:6` and `core/tools/aidlc-workspace-doctor.ts:60`.

- **[P2] Doctor silently misses valid legacy unit-name collisions.** Unit names such as `units` and `code-generation` remain valid, but the detector allowlists both the new axis and every stage slug. Such legacy directories produce no advisory; migrating unit `units` would also attempt to move a directory into itself. Use DAG/shape evidence and provide collision-safe migration coverage. See `core/tools/aidlc-workspace-doctor.ts:48` and `core/tools/aidlc-lib.ts:16083`.

- **[P2] Runtime stage instructions still consume retired paths.** CI Pipeline and five Operation stages still read NFR/infrastructure artifacts from `construction/<stage>/`, although those artifacts now live per unit. This can omit approved design context after migration. Update every live consumer to the unit axis, including how non-per-unit stages aggregate units, and correct the remaining guide examples. One example is `core/aidlc-common/stages/construction/ci-pipeline.md:47`.

No fixes or commits were pushed because this is maintainer-owned work.

Validation performed:

- 169 focused unit tests passed.
- `bun scripts/package.ts --check` passed for all harnesses.
- `t238-build-binaries` reproduced the failure locally: 2 passed, 1 failed.
- The migration and collision cases were reproduced directly.
- Current CI has failing smoke/unit and contributor-statement checks.
- The PR currently conflicts with `main`; it must be rebased, regenerated, and versioned after the current `2.8.0` baseline.

Given these blockers, I am requesting changes.

apackeer:
Rebased onto `main` (`a0ee4415e`, 41 commits) as one commit - no `dist/`, version, CHANGELOG, or README changes - and addressed the review.

**P1 - buffered reader.** Already fixed on main (`readSyncBufferedLine` copies the partial chunk before the refill); the rebase takes main's version and `t238-build-binaries` passes 4/4 here.

**P2 - migration command.** The doctor advisory and the troubleshooting note now print an idempotent, record-qualified command per legacy unit, runnable from the project root:
`if [ -d <record>/construction/<unit> ] && [ ! -e <record>/construction/units/<unit> ]; then mkdir -p <record>/construction/units && mv <record>/construction/<unit> <record>/construction/units/<unit>; fi`
and for a unit literally named `units`, a two-step form through `<record>/construction/.units-legacy`. Tests run the printed commands twice in a fixture: first run migrates, second is a no-op. Negative proof: printing the old bare `mv construction/<unit> construction/units/<unit>` fails with `mv: cannot stat 'construction/checkout-api': No such file or directory`, layout unchanged.

**P2 - collision-safe detection.** The allowlist is replaced by evidence: `construction/<name>/` is a legacy unit dir iff `<name>` is a DAG Unit and `construction/units/<name>/` is absent, or (no DAG) its children are stage slugs. Units named `units` and `code-generation` are detected; a stage-diary `construction/code-generation/` (memory.md only) and the migrated layout are not. Negative proof: reverting to the allowlist → `Expected true / Received undefined` (no advisory for the `units` unit).

**P2 - retired paths.** CI Pipeline and the five Operation stages now read NFR/infrastructure artifacts from `construction/units/<unit>/<stage>/…` with explicit every-Unit aggregation wording; `consumes` IDs are symbolic and unchanged. Build-and-test and the zero-Unit/no-DAG fallbacks stay stage-level by design.

Rebase notes: main's guard-recovery, review-record, provenance (#1052), archive (#1033), summary-authorization, and review-brief paths were carried onto the unit axis, including their fixtures; t221's main-added piped-search fixtures still carried the old paths and were corrected.

Verified: focused slice 12 files / 388 pass / 0 fail (t238 4, t256 15, t185 110, t165 58, t326 27, t327 8, t135 20, t127 24, t260 29, t115 56, t147 36, t304 1); all 65 test files this PR touches: 1,886 pass / 0 fail; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0. Contributor statement added.

apackeer:
Follow-up commit `8bc01701d` on top of the rebase, closing two gaps in `d890976bc`:

- **Zero-Unit scopes.** `artifact-resolution.ts` resolves per-unit artifacts to the stage-level path (`<record>/construction/<stage>/…`) when the scope skips Units Generation (`infra`, `express`); the six consumers I moved to the unit axis said only `construction/units/<unit>/…`. Each `inputs:` line and Step 1 bullet in `ci-pipeline` and the five Operation stages now carries the engine-resolved alternative, in `build-and-test`'s register: "…`<record>/construction/units/<unit>/infrastructure-design/` (if it exists); for a zero-Unit scope (e.g. `infra`) read `<record>/construction/infrastructure-design/` directly". t44 pins both forms for all six; t310 asserts an `infra` intent resolves `infrastructure-design` on the stage axis and a `feature` intent with Units on the Unit axis.
- **`units`-collision detector.** Absence of `construction/units/units/` was being read as proof that `construction/units/` is a legacy unit dir, so a DAG `[units, api]` with only `units/api/…` started would have nested `api` under `units/units/api`. A Unit named `units` now also needs legacy stage-shaped evidence (children are stage slugs); an empty axis or one holding unit directories is never flagged. t256: axis not migrated when the `units` Unit has not started, empty axis not flagged, migrated `units/units/` not flagged, started stage-slug-named Unit not flagged; true legacy `units` still gets the two-step command. Negative proof: the absence-only rule emitted `Construction unit layout: 1 legacy unit dir(s) … [units]` with the `.units-legacy` command against the partially-created new layout.

t256 19, t185 110, t44 47, t310 30 - 0 failures; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0.

PR #1000, opened 2026-09-02
fix: bind Plan Approval to content and attempt, add human-only exits, guard Kiro IDE's Windows shell; feat: Change Control

apackeer:
# Summary

A human approves a plan, confirms a summary, or receives a review verdict, and the framework later says that decision does not exist or cannot be acted on. Six issues reported the Plan Approval form of it, and the summary and review forms have their own reports and merged instance fixes; they share one cause: authority was stored as a derived property of mutable bytes (a directive's issuance identity, the order of two audit rows, the tail of the artifact under review), so ordinary engine activity destroyed it and ordinary human activity could not repair it.

This PR binds each authority to what a person would call a change: the plan's content and the current attempt; the confirmed answers and the outputs generated from them; the reviewed bytes and the review record beside them. It also makes the engine's read-only paths actually read-only, makes asking "what now?" twice cost nothing, and turns every guard refusal into a question the human can answer by choosing an offered remedy.

The fourth round (section 7) closes the dead ends a customer hit on 2.7.x with Codex CLI and Claude Code, where an approved plan could never be accepted: `decision` now refuses an unanswerable challenge before minting it, the source boundary says why it failed, and a human-only break-glass exit exists so no refusal is ever final. The same round fixes the Kiro IDE Plan Approval guard on Windows, where every shell call was denied because the guard only knew `execute_bash`.

Closes #995. Closes #977. Closes #975. Closes #981. Closes #987. Closes #930. Closes #1044. Supersedes #978 (see below for what it does and does not do for #931, #934, #863, and #864).

Also ships **Change Control** (section 6), a feature rather than a restoration: one per-intent setting deciding what an input change does after a human approval. Its product contract and migration contract are stated in that section and in #1032 for explicit maintainer agreement; nothing else in this PR changes a contract beyond restoring it.

## The problem

[...]

PR #1007, opened 2026-09-03
feat: centralize guard provenance proofs in a shared default-deny kernel

apackeer:
# Summary

The framework's trust chain rests on two kinds of records: receipts binding a review verdict to the exact bytes reviewed, and attribution state binding a tool call to the agent that made it. The guards protecting those records were each built as independent recognizers of specific input forms, so every guard carried its own parsing surface and each needed repeated review rounds to converge. The guard that converged fastest was the one that derived its evidence exclusively from raw Git object reads.

This PR generalizes that result into one shared kernel with a single rule, stated once in `docs/reference/06-hooks-and-tools.md`:

> **Names may deny; only proofs may allow.** A guard may use any heuristic to refuse an operation, but an allow decision may rest only on (a) canonical real-path identity, (b) raw Git object reads, or (c) fields already sealed in the append-only audit ledger. Unresolvable, ambiguous, or indirect inputs are treated as denied within the protected scope.

## What changed

**New kernel — `core/tools/aidlc-guard-kernel.ts`** (packaged into every harness `tools/` dir):

- **One Git invocation helper, `immutableGit`.** Every raw object read the kernel and its call sites make (`cat-file`, `rev-parse`, `read-tree`, `ls-files`, `check-attr`, `ls-tree`, `merge-tree --write-tree`, and every git spawn in the source-bound merge path) runs with `--no-replace-objects` as the first argument and `GIT_NO_REPLACE_OBJECTS=1` in the child environment, so `git replace` cannot substitute a tree or blob for the reviewed Source Commit.

- **Protected-store write barrier.** Given a tool call and a set of protected stores, returns a proof status: `proven-disjoint`, `proven-framework-tool`, `store-overlap`, or `unprovable`. Every candidate path is resolved to a canonical real filesystem path (separators normalized, links resolved per component, not-yet-existing suffixes resolved against their deepest real ancestor) before any comparison. Framework-tool identity is established by the resolved real file of the interpreter and the script, never by an executable's name or lookup order; invocations reached through delegation or dynamic construction never establish provenance. Attached path operands are inspected too - every `key=value` suffix and single-dash attached values (`-oPATH`) of known writers - so `dd of=<protected>/x` is `store-overlap` and an operand with an expansion is `unprovable`. The kernel supplies mechanism (proof status); each call site maps status to policy for its principal.
- **Immutable-evidence reader.** Committed-source listings and content identity derive from raw Git object reads (`cat-file`-verified commits, temp-index `read-tree` + `ls-files`), never from materialized working-tree bytes. Where a commit's own metadata declares content-transformation attributes (`filter`, `ident`, `text`, `eol`, `working-tree-encoding`) on an included path, evidence is underivable and the operation refuses rather than answering through the transformation. SHA-1 and SHA-256 repositories are supported.

**Migrated call sites (minimal-diff swaps, unchanged deny messages):**

- `harness/cursor/hooks/aidlc-cursor-adapter.ts` — the delegated-agent protection of reviewer-attribution state now takes its verdict from the kernel barrier, composing with the existing POSIX/Windows path, witness, executable, and shell protections. Deltas are narrowing-only: nothing formerly denied becomes allowed.
- `core/tools/aidlc-lib.ts` (`gitCommitSourceListing`) — same signature and null contract; the listing is now the committed tree's own identity, and transformation-bearing commits return null (fail closed) instead of folding through a pipeline.
- `core/tools/aidlc-worktree.ts` — source-bound merges prove immutable listings for the target, reviewed, and combined (`merge-tree`) commits before any target mutation. A combined-tree proof that cannot be produced (unsupported or failed `merge-tree --write-tree`) refuses the merge - `refusing to merge: cannot prove the combined merge tree; upgrade Git to support merge-tree --write-tree, resolve merge conflicts, or use AIDLC_SKIP_SOURCE_FRESHNESS=1 only with human approval` - instead of dropping that proof from the loop.

(The reviewer-retry evidence binding this design originally covered already shipped in 2.6.121 and is not duplicated here.)

## User-facing changes

- For repositories without content-transformation attributes on application source (the common case): none. Listings are byte-identical to the previous derivation, existing receipts stay fresh, and no commands or flags change.
- Repositories declaring `filter`, `ident`, `text`, `eol`, or `working-tree-encoding` on committed source paths get an explicit up-front refusal at Bolt worktree creation and source-bound merge, instead of receipts silently derived through the transformation.
- On the Cursor harness, delegated-agent access decisions are made on canonical real paths, so every spelling or indirection of a protected path receives the same verdict.

## Regression suite

Nineteen property-named tests, each pinning one invariant:

- `tests/unit/t331-guard-kernel-write-barrier.test.ts` — canonical-path resolution; link-indirection resolution; executable provenance by resolved real path; delegated-invocation denial; closed-designator allow for shell; path-separator and control-character normalization; nested-repository boundary handling; unprovable-is-never-allow.
- `tests/unit/t332-guard-kernel-immutable-evidence.test.ts` — object-read equivalence with the committed tree; materialization independence; content-transformation exclusion; gitlink identity at nested-repository boundaries; commit-provenance requirement; path-byte fidelity.
- `tests/unit/t333-guard-kernel-call-sites.test.ts` — every pre-existing deny in the Cursor corpus remains a deny; non-canonical and link-indirect spellings deny; policy mapping per principal; checkout-independent committed evidence at the worktree call sites.

`tests/unit/t314-source-freshness-receipts.test.ts` gains transformation-property coverage; `tests/unit/t276-cursor-adapter.test.ts` changes by one timeout value only (55s → 90s for the evaluator corpus, which now includes the kernel proof pass).

## Verification

- `bun scripts/package.ts --check`: all harness trees in sync.
- `bun tests/gen-coverage-registry.ts --check`: registry fresh, ratchet held.
- `bun run typecheck` (all three tsconfigs) and `bun run check`: clean (three pre-existing Biome infos).
- Post-rebase targeted slice (`--unit --filter '^(t68|t276-cursor-adapter|t314|t331|t332|t333)'`): 8 files, 206 assertions — all new suites, t314, and t68 green. `t276` shows 2 failing assertions (tests 29 and 35, plain read-only Git inspections denied by the delegated-agent evaluator); both reproduce identically on **untouched `origin/main`** in the same environment, so they are pre-existing environmental failures, not introduced by this branch. The full pre-merge gate on the original branch content (391 files) was run on 2026-08-28; content-identical code was rebased forward with conflicts confined to CHANGELOG/README/version.

## Relationship to #1000

Independent concerns that compose: #1000 binds human authority to real content changes (change control); this PR makes the records and measurements that regime relies on provable (record integrity). The branches share no code conflicts — only the release-metadata trio and the coverage instrument. No version or changelog bump per the release-metadata policy. Follow-up (not blocking either PR): route kernel denials through #1000's typed refusal/remedy surface, and adopt the kernel's provenance channel in the plan-approval guard.

## Acknowledgment


rafaelgsr (changes requested):
Thank you for consolidating the provenance and immutable-evidence logic into a shared kernel. I reviewed the current head `959d7f9c6e4aaf7b5a59822c399345beae534c82` against current `main` `f9f489458e6c8ca5decf0c209be8bae572e5faa2` (merge base `a277af218f0df7f325d3b8be7b6d90fce2c5bd40`).

This is internal maintainer-owned feature work. The default-deny/provenance direction is aligned with the framework’s existing trust-chain goals, and I found no merged or open work that supersedes it.

I found three authority-boundary blockers:

- **P1:** Git replacement refs can substitute a different tree for the reviewed Source Commit. I reproduced `gitCommitSourceListing(A)` deriving commit B’s content after `git replace A B`, followed by `git merge A` landing B’s unreviewed bytes. Immutable reads and every source-bound merge operation must disable replacement objects, with regression coverage for commit and blob replacement.
- **P1:** The shared protected-store barrier treats complete shell words as paths. `/bin/dd if=<outside> of=<protected>/entry` returns `proven-disjoint`, although executing it writes into the protected store. The current Cursor adapter remains protected only because its old command-specific parser recognizes `of=`. Attached path operands must be extracted or classified as unprovable.
- **P1:** When the combined-tree proof cannot be produced, `combinedCommit === null` simply removes it from the validation loop and merging continues. An unsupported or failed `merge-tree --write-tree` must fail closed. The target-side-attributes test currently permits an earlier baseline refusal and therefore does not pin this path.

The branch also needs rebasing onto 2.8.0. `CHANGELOG.md` contains a committed conflict marker, the version is still 2.7.3, and the PR modifies committed `dist` files that current main now treats as ignored generated outputs.

I made no changes and pushed no commits.

Validation on the exact head:

- `bun scripts/package.ts --check`: passed for all seven harnesses.
- Coverage registry check, all three TypeScript configurations, and lint: passed.
- `t331`–`t333`: 19/19 tests passed.
- The broader focused slice had two `t276` environmental failures and one `t314` timeout; all three reproduced unchanged on the merge base.
- `git diff --check`: failed on the committed changelog conflict marker.
- Exact-head implementation CI is green; the Contributor Statement check fails because the required statement is absent. GitHub reports merge conflicts.

Because the current head permits protected-store and reviewed-source authority bypasses, I am requesting changes.

apackeer:
Rebased onto `main` (`a0ee4415e`, 41 commits) as one commit - the committed changelog conflict marker, the 2.7.3 slot, README badge, and every `dist/` copy are gone (`dist/` is generated; the kernel lands in all seven `tools/` dirs via `bun scripts/package.ts`) - and addressed the three P1s.

**P1 - replacement refs.** New `immutableGit(repoDir, args, options)` in the kernel puts `--no-replace-objects` first and forces `GIT_NO_REPLACE_OBJECTS=1` in the child env. Every git spawn in the kernel (`cat-file`, `rev-parse`, `read-tree`, `check-attr`, `ls-files`, blob reads), in `gitCommitSourceListing`'s tree/materialization path, and in `aidlc-worktree.ts` (all rev-parse/cat-file/ls-files checks, `merge-tree --write-tree`, `commit-tree`, the merge/reset/checkout/worktree operations) now routes through it; no direct git spawn remains on those paths. t332/t333: after `git replace A B`, `gitCommitSourceListing(A)` is A's own listing and identity, the source-bound merge of A lands A's bytes, and a blob replacement does not alter derived content identity. Negative proof: dropping flag+env → A's listing derives B's `replacement-only.txt` and B's hash; the merge assertion receives `export const reviewed = 'UNREVIEWED B';`.

**P1 - attached path operands.** The barrier still inspects whole words and additionally every `=`-suffix regardless of key, plus single-dash attached values (`-oPATH`) when a known writer is present; operands are canonicalized, and an empty or expansion-bearing attached value is `unprovable`. Names only add potential denials. t331/t333: `dd of=<protected>/x` → `store-overlap`; `dd of=<outside>/x` → `proven-disjoint`; `dd of=$P/x` → `unprovable`; `tee --output=…`; the Cursor `of=` deny holds with the old parser bypassed. Negative proof: without operand extraction, `Expected "store-overlap", Received "proven-disjoint"` and the kernel-only Cursor case flips to `allow`.

**P1 - combined-tree proof fails closed.** A null `combinedCommit` now refuses: `refusing to merge: cannot prove the combined merge tree; upgrade Git to support merge-tree --write-tree, resolve merge conflicts, or use AIDLC_SKIP_SOURCE_FRESHNESS=1 only with human approval`. The target-side-attributes fixture was rebuilt so it pins this refusal specifically. Negative proof: restoring the skip, a forced `merge-tree` exit 129 still merged (`expect(merged.code).not.toBe(0)` received 0).

Rebase notes: main's `aidlc-lib.ts` raw tree/materialization implementation is kept with the proof boundary applied to it; main's 64 KiB-boundary and commit-provenance (#1052) tests in t314 are preserved; coverage registry regenerated (391 → 398 functions). Contributor statement added.

Verified: t331 14, t332 8, t333 8, t314 120, t276 42 (the two historical environmental denials did not occur here) - 192 pass / 0 fail; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `bun tests/gen-coverage-registry.ts --check`, `git diff --check` exit 0.

PR #1008, opened 2026-09-03
feat: add a preview release channel

apackeer:
Stacked on #756 (`feature/issue-722-rfc-reshape`). This PR targets that branch; it must be retargeted to `main` before #756's branch is deleted, because the repository deletes branches on merge and GitHub closes a stacked PR when its base branch disappears.

Version note: this branch bumps to `2.7.2` with its `## [2.7.2]` CHANGELOG heading placed above the base branch's `## [2.7.1]`. Both numbers will be re-slotted at push time once the merge order is known; `t68` is green on the branch as committed.

### What this adds

`main` is a stream of named versions (every user-visible PR bumps the patch and adds a CHANGELOG entry). The stable channel blesses one point in that stream; the new **preview** channel publishes the stream itself once a day as a GitHub prerelease that is never marked "latest". No release branch, no cherry-picks.

- **Grammar.** Stable ids stay exactly `x.y.z`. Preview ids are `x.y.z-preview.YYYYMMDD.N` (source version, UTC build date, that day's build counter). Defined once in `core/tools/aidlc-channel.ts`; every place a version is a directory name or marker value accepts both and nothing else (path escapes, whitespace, unicode digits, empty segments, and general semver prereleases are rejected, with tests).
- **Consumer commands.** `aidlc config --channel stable|preview` persists a machine channel marker beside the update cache and pins (`aidlc config --channel` prints it). `aidlc update` follows it, `aidlc update --channel <c>` overrides once, `aidlc update --check` reports "behind" against the channel's newest release with the existing exit codes (5/0/3/1). `aidlc use <id>`, `aidlc config --pin <id>`, `install.sh --version <id>`, and `install.ps1 -Version <id>` accept preview ids; project pins keep overriding the channel. Switching back (`config --channel stable` then `update`) converges on the newest stable even though its id sorts lower and is reported as a channel switch. Retention keeps the two newest complete previews on top of the active/rollback/in-use/pinned protection every release has.
- **Discovery.** Stable is unchanged (`latest/download`). Preview lists the releases of the repository behind `--release-base-url` through the GitHub API (derived for github.com; `--release-api-url` / `AIDLC_RELEASE_API_URL` for other hosts and the test fixture), picks the newest published prerelease whose tag is a preview id, and installs it through the exact-version path. API failure or rate limiting is exit 3, never a crash, never a fallback to stable.
- **Environments.** Preview runs (the schedule, or a dispatch with `channel: preview`) enter a `preview` environment: the same publication variables and App secret as `release`, the same `main`-only deployment policy, and no required reviewers, so the daily run is unattended. The authorize and promote preflights check the channel's environment and require exactly zero reviewer rules on `preview` (an approval-gated preview environment fails fast). Stable runs keep the `release` environment and its aidlc-admins approval. A workflow-level concurrency group serializes runs per channel without cancelling. The human review guarding a preview is the approval every merge to `main` already needs plus the CI gate; the accepted cost is that any commit on `main` becomes a signed preview within a day.
- **Release workflow.** `workflow_dispatch` gains a `channel` input (`stable` default); a daily `17 3 * * *` UTC schedule is always a preview run. `authorize` plans the preview from the publication repository (skip when `main` has not moved since the newest preview's source commit; allocate the day's counter from every existing preview tag; render notes from the CHANGELOG sections, or commit subjects, added since that commit). `ci.yml` becomes callable (`workflow_call`) and gates the preview build chain before anything is built. Build, smoke, and staging jobs stamp the preview id through `AIDLC_BUILD_VERSION` into every projected `aidlc-version.ts`, the projection stamps, the binary version gate, and `version.json`; the source tree keeps `x.y.z`. `promote` reuses the verified staging-draft flow, then creates the annotated `v<id>` tag (message records the source commit and build date) and publishes with `prerelease: true`, `make_latest: false`. Stable runs are unchanged.

### Deviations recorded in the report

- Preview publication keeps the staging draft rather than publishing directly: the publication repository enforces immutable releases, and assets cannot be attached after publication.
- Retention keeps rollback protection for previews and adds the two-newest window on top.

### Verification

`bun scripts/package.ts --check` clean, `bun run check` (determinism + typecheck + biome) clean on the final tree, `t68` green, new unit tests t330/t331/t332, t244 re-pinned to the environment expression and concurrency group (targeted rerun of t244/t332/t239/t68: 5 files, 0 failures), full smoke + unit + integration tier run on the final tree (405 files, 2 reds: `t243-install-mechanism` was a load-timeout flake and is green alone; `t72-stage-reverse-engineering` is a live-model journey that fails identically on the base commit unless the model improvises a stage jump, and is unrelated to this change). No GitHub run was executed; the workflow is validated by YAML parsing, job-graph checks, and the t332 contract test.

### Repository settings this needs (not code)

- The publication repository named by `AIDLC_PUBLICATION_REPOSITORY`, immutable releases, both `v*` tag rulesets, and the release App, exactly as #756 documents in `docs/reference/19-supply-chain-security.md` section 11.
- A `preview` environment on this repository with the same two variables and one secret as `release`, exactly the `main` deployment branch policy, and no required reviewers.

### Open questions

- Confirm the staging-draft publication path for previews and the retention reading.


leandrodamascena (changes requested):
I like the goal of having a preview channel. My concern is mainly the publication cadence and the user experience, not the channel itself.

GitHub shows prereleases together with stable releases. A new preview every day could quickly make the Releases page noisy and make the official stable version harder to identify, even when nobody needs that daily build.

Before choosing a daily schedule, I think we should clarify who will use previews. Are they for maintainers testing the release pipeline, contributors validating `main`, early adopters, or automated end-to-end testing? These use cases may need different solutions.

Some options we could explore:

- A local build script for contributors who only need to test the current `main`.
- A manually dispatched preview when a specific change needs broader testing.
- A periodic preview after a meaningful group of changes, instead of every day.
- A separate publication location if we need a continuous public preview channel without mixing it with stable releases.

Each option has trade-offs. A local build does not test the real installation and update flow. On-demand previews may detect regressions later. Scheduled previews provide regular coverage, but they also create more releases and more noise for users.

Could we first define the expected users, publication trigger, discoverability, and lifecycle of preview builds? I am open to a daily model if the demand supports it, but I would like us to make that decision based on the workflow we actually need.

leandrodamascena (changes requested):
Thanks for clarifying the intended cadence. I am comfortable with a daily preview channel provided that it publishes **at most one successful public preview per UTC day**.

The current implementation does not enforce that invariant. The planner only skips when the newest published preview points to the same source SHA. If `main` changes after the scheduled preview, or someone uses `workflow_dispatch` later that day, the planner allocates `.2`, `.3`, and so on and publishes another prerelease for the same UTC date. The unit test currently confirms this behavior by expecting `.10` when same-day tags already exist.

Please change the planner so that:

- It publishes when no preview has been published for the current UTC date and `main` has relevant changes.
- It skips when a non-draft preview already exists for the current UTC date, even if `main` has advanced.
- A manual dispatch cannot create a second public preview on the same day.
- Concurrent scheduled and manual runs remain serialized, with the later run observing and skipping the release created by the earlier run.
- A failed or incomplete attempt may be retried when no preview was successfully published. If the daily counter remains, it should support recovery from occupied tags or failed attempts rather than multiple successful daily publications.

Please add tests for these cases:

1. A preview exists today with a different source SHA: skip.
2. The scheduled preview succeeds, then `main` advances and a manual dispatch runs: skip.
3. A tag or draft exists from a failed attempt, but no preview was published: retry is allowed.
4. The UTC date changes and `main` has advanced: publish the next daily preview.
5. `main` is unchanged since the previous preview: skip as it does today.

The documentation should also say **“at most once per UTC day”** and explain that manual dispatches respect the same daily limit.

The PR currently conflicts with `main`, and the latest head has failing smoke/unit and deterministic integration/e2e checks. Please rebase, resolve the conflicts, and rerun the complete CI suite after implementing the daily publication guard.

Once the one-release-per-day invariant is enforced and CI is green on the rebased head, the cadence concern from my previous review will be resolved.

apackeer:
Implemented in 7fe8b1df47bd6b9dc6df6d1826f00394e1b53c2a, after rebasing onto `main` at aadc832354f2d8841489ed01b5e0d852b32b7962.

Previews now publish **at most once per UTC day** across scheduled and manual runs. The planner checks every published preview before comparing source commits, so advancing `main` cannot produce a second preview that day. Older queued runs can reach this check and skip; a new publication candidate must still match the tip of `main`.

Drafts and orphan tags reserve retry counters without consuming the daily allowance, and unchanged sources still skip. The guard counts both the build date and the actual UTC publication timestamp for overnight builds. The publisher rechecks the publication day before creating the final tag.

The [pipeline regressions](https://github.com/awslabs/aidlc-workflows/blob/7fe8b1df47bd6b9dc6df6d1826f00394e1b53c2a/tests/unit/t332-preview-release-pipeline.test.ts#L598) cover all five requested cases, plus pagination, version ordering, overnight publication, and execution of the actual workflow scripts for an older queued checkout. The documentation states the shared daily limit and retry behavior.

Local verification:

- `bun run check` passed: generated projections, independent-build determinism, all three TypeScript configurations, and lint.
- Full smoke/unit run: 299 files, initially 293 passed and 6 failed. The failures exposed host-installation dependencies in four test fixtures and insufficient time budgets in two subprocess-heavy cases. After those fixes, all six complete files passed isolated reruns (129 cases); their behavioral assertions are unchanged.
- Deterministic integration/e2e: 142 passed files, zero failures; 54 live-model files explicitly skipped by `--no-llm`.
- Final focused workflow/lifecycle check: 55 cases passed, including all 18 preview-pipeline cases.
- Changelog preservation guard passed; all 278 existing headings were retained.

These are local verification results; the checks on the pushed head will provide the fresh GitHub CI result.

leandrodamascena (approved):
Thanks for the detailed follow-up. I re-reviewed the daily publication guard and the original concern is resolved.

Scheduled and manual runs now share a strict limit of at most one published preview per UTC day. The implementation also handles queued runs, source changes, retries after failed attempts, pagination, and builds crossing UTC midnight.

I locally validated the preview channel suite: 30 tests and 545 assertions passed.

Approved from a review perspective.

Before merge, one final rebase is needed because `main` advanced and now conflicts in `scripts/install.ps1`. Please preserve both behaviors while resolving it:

- Empty `$Version` remains valid for the Windows one-liner.
- Explicit preview version identifiers remain valid.

After the rebase, please rerun CI and ensure the exact mergeable head is green.

apackeer:
Rebased onto `main` at 0865300b9de009aea3605f19af5eb6f48c1d343a and resolved the installer conflict. The resulting head is c7e16a91d8eedb840b7a34520968b9d795fef825.

Both requested behaviors are preserved in the [Windows installer](https://github.com/awslabs/aidlc-workflows/blob/c7e16a91d8eedb840b7a34520968b9d795fef825/scripts/install.ps1#L17): the omitted/empty `-Version` default works for the `irm ... | iex` one-liner, and explicit stable and preview identifiers remain valid. Manifests still require a nonempty version.

Checking the actual parameter block on Windows PowerShell 5.1 also exposed two differences from the JavaScript regex checks: default case-insensitive matching and acceptance of a trailing newline. The validator now uses explicit case sensitivity and a strict end assertion, with matching regression coverage. All 44 parameter cases and 44 manifest cases matched expectations on PowerShell 5.1; the separate omitted/empty binding probe also succeeded. These probes exercised parameter binding and validation, without running the installer download body.

Local validation on the rebased tree:

- `bun run check` passed: projection generation, independent-build determinism, all TypeScript configurations, and lint. Changelog preservation also passed.
- Smoke: 14 passed files.
- All four unit shards covered 286 files exactly once. Two files timed out in the initial run; both passed unchanged when rerun alone.
- Deterministic integration/e2e exercised 142 files, with 54 live-model files explicitly skipped by `--no-llm`. Four files failed initially; all four passed unchanged in isolation, and a focused parallel check also passed. The original deep-run failure cause remains unconfirmed.

The committed patch is byte-identical to the tested snapshot. These local results are not a claim that the original runs were entirely green; the fresh GitHub checks on c7e16a91d remain the merge gate.

apackeer:
Pre-merge live checks exposed two blockers, now fixed in `aac8ef2cb` and `1bedf08e5`:

- Kiro IDE’s audit-tail hooks matched only `execute_bash`. They now also receive Windows `execute_pwsh` and `shell` events, so runtime-graph rebuilding and workflow-state synchronization run on those tools.
- Workspace navigation through `next` could trigger workflow continuation. The Stop-hook classifier now treats static space/intent navigation as terminal while retaining enforcement for intent creation, chained mutations, and ambiguous shell forms. All seven orchestrator skills also check terminal stop instructions before another `next` or `report`; the previous loop instruction conflicted with terminal `print` behavior.

The full live SDK workspace journey passed with all three navigation turns ending naturally, each with exactly two tool results and no workflow questions or extra `next`. The Windows IDE slice passed all 11 cases, including four live cases. Windows test-driver fixes preserve the conversation across Other replies and handle the observed Claude startup modal within the isolated test profile. `fb628a1a7` also adds positive Claude plugin invocation coverage; its live run passed all four cases.

Package determinism, type checking, linting, smoke checks, and the focused classifier/hook regressions passed locally. The remaining Claude TUI slices and the SDK run with the strengthened final-tool assertion are still running. Native Windows ACP coverage remains pending Kiro CLI authentication; its two skips are not counted as coverage. Merge remains on hold until the remaining live gates are verified.

apackeer:
Final pre-merge update for `b0ee1357609c86946680068c7d6c64bd8b7bee0f`:

- GitHub CI is fully green. The follow-up test-only commit aligns the existing Kiro hook-registration expectations and CLI coverage registry with the fixes.
- The SDK workspace journey passed with the stronger assertion that each navigation command ends with the requested utility output. The remaining Claude TUI revision/depth checks also completed successfully, including the baseline/revised comparison and invalid-depth recovery.
- All 18 requested Linux Claude TUI files completed successfully across the retained runs. Interrupted aggregate runs remain marked unreported; completed per-file results were retained, and unfinished cases were rerun successfully.
- Per the maintainer’s explicit instruction, the two native Windows ACP routing cases blocked by Kiro CLI authentication are waived for this merge. They are not counted as passed, and their tests are unchanged. The Windows IDE and Claude TUI checks completed separately.

Proceeding with the approved head through GitHub’s normal merge process.

PR #1019, opened 2026-09-04
fix: isolate Cursor background reviews from workflow steering

apackeer:
## Summary

- keep Cursor background stop events out of the foreground forwarding loop; background identity is taken from the lifecycle events that carry `cursor.is_background_agent` (`sessionStart`, `beforeSubmitPrompt`, `sessionEnd`) and stored per conversation, then consulted by `preToolUse` and `stop`, which never carry the field; a conversation with no lifecycle event is foreground, as today
- deny background lifecycle and routing commands with an allowlist-literal rule: only one direct, literal Bun/`aidlc` invocation of the read-only utility set is permitted; variables, substitutions, backticks, shell control syntax, `sh -c`/`node -e`/`python -c`/`pwsh`/`cmd`, `find -exec`, `xargs`, helper scripts, `eval`, and Bun `-e`/`-r`/`--preload` are denied
- preserve read-only utilities and existing stale, replay, wrong-project, and wrong-state rejection
- add deterministic concurrency and repeated-background-review coverage
- document the Cursor background-review boundary and regenerate distributions

## Verification

- `bash tests/run-tests.sh --debug -P 8 --unit --filter '^(t68-version-changelog-sync|t242-state-transition-guard|t248-steering-content-delivery)$'`
- `bash tests/run-tests.sh --debug -P 8 --unit --filter '^t276-cursor-adapter$'`
- `bun run check`
- native Windows core/version slice: PASS
- native Windows changed background-review regression: PASS

The complete Windows `t276` file has one repeatable failure in the unchanged `beforeSubmitPrompt` rebind assertion; the adapter code for that path is byte-identical to `main`. All 41 other cases, including the new background-review case, passed.

## Upgrade

Re-copy `dist/cursor/`; active workflows require no state migration.

No version or changelog bump: release metadata belongs to the release-preparation PR.

## Acknowledgment


rafaelgsr (changes requested):
Thanks for tightening the Cursor background-review boundary and adding concurrency coverage.

I reviewed the current head `28d15b43d2983d480dc4dcc0580624e81aaf709e` against `a277af218f0df7f325d3b8be7b6d90fce2c5bd40`. The merge base is the same base commit. This is an internal bug fix aligned with the existing foreground-only workflow authority contract.

**P1 — Background identity is read from events that do not carry it**

`backgroundWorkflowCommand()` and the Stop branch inspect `cursor.is_background_agent` directly on `preToolUse` and `stop` events (`harness/cursor/hooks/aidlc-cursor-adapter.ts:747-759`, `3093-3098`).

Cursor’s published schemas and the repository’s field-verbatim fixtures carry this field on session lifecycle events, but not on `preToolUse` or `stop` (`tests/fixtures/cursor-hook-payloads/payloads.json:43-56`, `157-168`). The regression test instead adds the field manually at `tests/unit/t276-cursor-adapter.test.ts:1048-1059`.

Using the documented event sequence, the direct background `aidlc-orchestrate.ts next` call received `permission: "allow"` and reset the active directive from part 2 to part 1 with a new continuation token. A field-verbatim Stop payload also still emitted a workflow `followup_message`.

Please persist background identity from a supported lifecycle source—such as session-scoped environment or protected session state—and consult that identity for later events. Cover the real event shapes, including environments where `sessionStart` is unavailable, rather than injecting the flag into tests.

**P1 — Computed lifecycle commands bypass strict classification**

In `core/hooks/aidlc-state-transition-guard.ts:862-890` assignments are resolved only for the executable, while `:952-978` requires the Bun script and lifecycle verb to remain literal.

Consequently:

```sh
verb=next; bun .cursor/tools/aidlc-orchestrate.ts "$verb"
```

is classified as safe. The Cursor adapter returns `permission: "allow"`, and executing it reproduced the same part-2-to-part-1 cursor reset. Similar fail-open paths exist through command substitution, Node/Python evaluation, PowerShell/CMD, `find -exec`, helper scripts, and Bun preloads.

Strict background mode must resolve or fail closed when the script, lifecycle verb, or execution host is dynamic or uninspectable. Please add regression coverage for variable-derived scripts/verbs and non-Bun execution hosts.

**Validation**

- Packaging parity and `bun run check`: passed.
- `t68`, `t242`, and `t248`: passed.
- The added `t276` background regression passed with its synthetic payload.
- Full local `t276`: 40/42 passed. One failure reproduces unchanged on the base commit; the other is a sandbox denial when creating a home-directory symlink.
- No fixes or commits were pushed.
- GitHub’s code, test, security, and documentation checks pass. `Require Contributor Statement` is failing because the required license confirmation is absent from the PR description.

Remaining risk: documented Cursor background events can still enter the forwarding loop and execute workflow-routing commands, invalidating the foreground continuation.

**Proposed review event: `REQUEST_CHANGES`**

apackeer:
Rebased onto `main` (`a0ee4415e`) and addressed both P1s in `cd34fb7bd` (the intermediate commit `7fe6262d5` had persisted identity but treated unknown as blocked and expanded variables before classifying; both are corrected here).

**P1 - identity from lifecycle events.** `sessionStart`, `beforeSubmitPrompt`, and `sessionEnd` - the events that carry `cursor.is_background_agent` in the field-verbatim fixtures - now store `{background: true|false}` per conversation at `aidlc/.aidlc-cursor-subagents/session-<sha16(conversation_id)>.marker`. `preToolUse` and `stop` read that stored identity; the direct event-field reads are gone. A conversation with no lifecycle event is foreground, exactly today's behaviour (documented in the adapter and `docs/guide/harnesses/cursor.md`). t276 regressions use the fixture shapes verbatim: background `sessionStart` → field-verbatim `preToolUse` `orchestrate next` is denied and the active directive is untouched; the same with `beforeSubmitPrompt` as the only lifecycle event; background `sessionStart` → field-verbatim `stop` emits no workflow `followup_message`; foreground and no-lifecycle-event conversations behave as before. Negative proof: restoring the event-field read on `preToolUse` fails the first regression - `expected permission="deny", part=2, original continue_token; received permission="allow", part=1, changed continue_token`.

**P1 - fail closed on dynamic commands.** Background classification in the state-transition guard is now allowlist-literal: one direct, literal invocation of Bun or the compiled `aidlc` dispatcher whose script and verb resolve to the read-only utility set. Variables, `$(...)`, backticks, assignments and control syntax, `sh -c`/`bash -c`, `node -e`, `python -c`, `pwsh`, `cmd /c`, `find -exec`, `xargs`, helper scripts, `eval`, and Bun `-e`/`-r`/`--preload` are denied; foreground classification is unchanged. t242 carries the matrix. `verb=status; bun .cursor/tools/aidlc-utility.ts "$verb"` is denied even though the resolved verb is read-only, while the literal `... status` is allowed; the literal allowlist independently rejects an unexpanded `$verb`, so the two checks are layered.

Contributor statement added to the body; the release-slot sentence removed. Verified: t242 23, t248 44, t276 51 (home-symlink case passed here) - 0 failures; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0.

PR #1023, opened 2026-09-04
fix: accept the documented (Recommended) label decorator at gate replies

### LINKED ISSUE #1126: [Bug]: Gate approval rejects the documented "(Recommended)" label decorator
### Description

At a protected human gate, the engine matches the human's reply against the offered choices with an exact string comparison. On Codex CLI the documented question-rendering convention tells the conductor to append `(Recommended)` to the recommended option's label, and `request_user_input` returns that decorated label verbatim. When the recommended option is `Approve`, the reply comes back as `Approve (Recommended)`, which does not equal `Approve`, so the engine refuses the approval and asks the conductor to re-present the same gate. Re-presenting it reproduces the same decorated label, so the gate cannot be approved through the documented path.

The convention that produces the decorator is documented in `harness/codex/skills/aidlc/question-rendering.md`:

> When a question has a recommended option, list it FIRST and append "(Recommended)" to its label — the tool renders recommended-first natively.

The two gate-reply comparisons on `main` are exact:

- `core/tools/aidlc-orchestrate.ts:8661-8663` — `const approvalChoice = flags.userInput?.trim();` then `approvalChoice === "Approve"`
- `core/tools/aidlc-state.ts:5331,5355` — `const approvalInput = userInput?.trim();` then `approvalInput === "Approve"`

This class of bug is already recognized in the codebase: `stripRecommendedDecorator` was added to `core/tools/aidlc-lib.ts:7148` in #1000 with the comment "Plan Approval matches offered labels exactly, so the one trailing decorator is removed before that match". That helper is applied in `core/tools/aidlc-testing-posture.ts:1837` for Plan Approval, but not on the stage gate-reply path above, so the same defect remains there.

### Steps to Reproduce

1. Install AI-DLC on Codex CLI and start any workflow that reaches a gated stage.
2. At the approval gate, have the conductor render the question per `question-rendering.md`, marking `Approve` as the recommended option, so the picker label is `Approve (Recommended)`.
3. Choose that option.
4. The reply reaches `report --result approved --user-input "Approve (Recommended)"`.

### Expected Behavior

The reply is recognized as the offered `Approve` choice and the gate is approved. The `(Recommended)` suffix is presentation the framework itself asked for, not part of the choice content.

### Actual Behavior

The engine rejects the reply:

```
report --result approved for "<slug>" received reply "Approve (Recommended)" which did not match an offered choice at the held gate. Re-present the original held gate with every offered choice and wait for the human to choose one.
```

Re-presenting the gate yields the same decorated label, so the workflow cannot advance past the gate while following the documented convention. Work around it today by answering with the undecorated label or by not marking a recommended option.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.8.2 / 0865300b

### AI-DLC Phase

Not phase-specific

### Harness

Codex CLI

### AI Model

Claude Opus 4.8

### Environment

- OS: macOS 14.2.1
- Harness: Codex CLI 0.152.1

### Additional Context

The narrow fix is to reuse the existing `stripRecommendedDecorator` helper on the gate-reply path instead of comparing verbatim, keeping the error message showing the reply exactly as received. Note the existing helper takes a non-nullable `string`, so the call sites need `?? ""`. The same decorator also reaches the `Request Changes` comparisons on that path.

I have an open PR that predates this issue (#1023). It currently carries a duplicate copy of the helper because it was written before #1000 landed; I will rebase it onto `main` to reuse the existing `core/tools/aidlc-lib.ts` helper and link it here. Happy to close the PR instead if you would rather implement it yourselves.

jisu-im-dev:
# Summary

Closes #1126.

At a protected human gate the engine matched the human's reply against the offered choices verbatim. The question-rendering guide tells the conductor to append `(Recommended)` to the recommended option's label, and the picker returns that decorated label, so answering a gate through the documented path came back as "did not match an offered choice". Re-presenting the gate reproduced the same decorated label, leaving no way to approve it without dropping the decorator by hand.

The codebase already recognizes this class of defect: `stripRecommendedDecorator` was added in #1000 for Plan Approval's exact-label match. This change applies that same existing helper on the stage gate-reply path, which was still comparing verbatim.

## Changes

* `core/tools/aidlc-orchestrate.ts` and `core/tools/aidlc-state.ts`: strip the one trailing decorator before matching the offered approval choices, reusing the existing `core/tools/aidlc-lib.ts` helper. The helper takes a non-nullable `string`, hence the `?? ""` at the call sites.
* `core/tools/aidlc-lib.ts`: fold the same strip into `isRequestChangesChoice`. That matcher already tolerates case, an option prefix, quotes, and trailing punctuation, but not the decorator, so `Request Changes (Recommended)` failed the same way. Folding it in covers every caller of that choice at once.
* The refusal message still shows the reply exactly as received (`formatReceivedReply(flags.userInput)`), so a genuinely unoffered reply is still reported verbatim.
* Plan Approval's exact-label anti-forgery binding is untouched.
* In `isRequestChangesChoice` the decorator strip runs after the quote and trailing-punctuation tolerance. The strip is end-anchored, so running it first let a surrounding quote or a trailing period hide it (`"Request Changes (Recommended)"`, `Request Changes (Recommended).`) — the ordering defect flagged in the #1133 review — fixed here so the fallback is correct as written.
* No release metadata changes: `core/tools/aidlc-version.ts`, the README badge, and `CHANGELOG.md` are left to the release-preparation PR per CONTRIBUTING.

## User experience

Before: marking `Approve` as the recommended option — which the framework's own guide asks for — made the gate unapprovable. The reply `Approve (Recommended)` was refused, and re-presenting the gate produced the same label. The workaround was to answer with the undecorated label or to not mark a recommended option at all.

After: `Approve (Recommended)` approves the stage, and `Request Changes (Recommended)` is recognized as Request Changes. Nothing else about the reply text is changed, and no other harness behavior moves.

## Checklist

* [x] Changes are documented

## Test plan

Regression coverage lands with the fix and was confirmed to fail before it and pass after (verified by stashing only the three `core/tools/` edits and re-running):

* `tests/unit/t188-human-presence-gate.test.ts` — new `B2` pins that `approve --user-input "Approve (Recommended)"` commits the gate and emits exactly one `GATE_APPROVED`.
* `tests/unit/t331-guard-deadlock-liveness.test.ts` — adds `Request Changes (Recommended)`, `request changes (recommended)`, `"Request Changes (Recommended)"`, `Request Changes (Recommended).`, and `B. Request Changes (Recommended)` to the accepted forms in the existing `isRequestChangesChoice` table. The quoted and punctuated forms fail with the decorator strip placed first and pass with it placed after the quote/punctuation tolerance.

Commands run on the current revision (rebased onto `main` at `891669aa`):

* `bun tests/run-tests.ts --unit --no-llm --filter '^(t188-human-presence-gate|t331-guard-deadlock-liveness|t68-version-changelog-sync|t242-state-transition-guard|t328-plan-approval-runtime-authority)\.test\.ts$'` — PASS, 0 failed files.
* `bun scripts/package.ts && bun scripts/package.ts --check` — deterministic, no drift.
* `bun run typecheck` and `bun run lint` — clean.

## Acknowledgment


wowzoo:
https://github.com/awslabs/aidlc-workflows/pull/1023

Thank you for this — and for #1126 before it. The issue named the two comparison sites and the helper that #1000 had already added for Plan Approval, which is why the fix is three lines in the right places rather than a new matcher.

I owe you a piece of bookkeeping. #1133 makes the same change, and I reviewed that one first without noticing yours had been open for a week. The source edits are the same in both, down to reporting the raw reply in the orchestrate error rather than the stripped one. #1133 additionally carries the stage-protocol and state-machine documentation and a wider test table, so we are going to consolidate there rather than carry two PRs for one issue.

To be straight about the grounds: that is a project decision about not reviewing the same change twice, not a judgement that anything here is wrong. The closure guidance in `CONTRIBUTING.md` covers PRs that are out of scope or off-guideline, and yours is neither.

Two things worth having on the record before this closes.

**The approach has a defect that both PRs share.** `stripRecommendedDecorator` is anchored at the end of the string and runs before the quote and punctuation removal, so the decorator does not compose with the tolerances next to it:

```
"Request Changes (Recommended)"        true
"B. Request Changes (Recommended)"     true
'"Request Changes (Recommended)"'      false
"Request Changes (Recommended)."       false
```

The option prefix survives because it is at the start; quotes and punctuation sit where the anchor looks. I have asked for that on #1133, so it will be fixed once — but it came out of reading your diff as much as theirs.

**Your diff reads worse than it is.** Against `main` it is only the gate change. The `isWorkspaceNavigationNext` deletion and the `isEngineEngagementSegment` signature change that show up here are relative to your base at `d66b9ad6`; `main` removed that function itself in #1043, and merging your head into `main` today produces no conflicts and keeps main's current signature. Anyone reviewing this later should not read those hunks as scope creep.

One note for a future PR: the version constant, README badge and `CHANGELOG.md` entry belong to the release-preparation PR — `AGENTS.md` said so at your base too, so that one is worth remembering rather than a policy that moved under you.

I will leave this open until #1133 actually merges, so there is a fallback if it stalls. Your precedence is recorded there.

jisu-im-dev:
Rebased onto `main` (`891669aa`), single commit `c9a9a12e`. Two changes from the previous revision: the version/README/CHANGELOG edits are dropped per CONTRIBUTING (release metadata belongs to the release-preparation PR), and the decorator strip in `isRequestChangesChoice` now runs after the quote/punctuation tolerance — the ordering defect flagged in the #1133 review — with the quoted, punctuated, and prefixed forms pinned in t331. Staying open only as the fallback noted above; happy to close as soon as #1133 lands.

wowzoo:
I have approved the three gated workflows on `c9a9a12e9` — CI, Security Scanners and Markdownlint were all sitting at `action_required`, so the test suite had never run against your rebase. That is on us, not on you. While it runs, here is what I measured locally, because two of these results matter to you specifically.

**Your ordering change did not inherit the regression I later blocked #1133 on.** This is the part worth being explicit about, since it was my wording that sent you here. My #1133 review said the decorator should be stripped "in the same normalization pass as the quotes and punctuation, or after them", and you took the second half — reasonably, because that is what it says. On #1133 that second half turned out to be a trade: moving the strip to the end fixed the forms where the decorator sits inside the quotes and punctuation and broke the two where it sits outside them, so I had to block that head and the author has since replaced the ordered passes with a single shape match.

Here it is not a trade. Your base never accepted the decorator in any position, so an ordered chain can only add. Running my own 65-form corpus against `be94bde7c` and against your head — `be94bde7c` was the default branch when I measured; `main` has since advanced to `2931ef02a`, and `isRequestChangesChoice` is byte-identical between the two, so the comparison still holds — your head accepts **12 forms that `be94bde7c` refuses, and loses none**:

`Request Changes (Recommended)`, `request changes (recommended)`, `Request Changes (RECOMMENDED)`, `Request Changes(Recommended)`, `Request Changes  (Recommended)`, `Request Changes\n(Recommended)`, `"Request Changes (Recommended)"`, `'Request Changes (Recommended)'`, `Request Changes (Recommended).`, `Request Changes (Recommended)!`, `B. "Request Changes (Recommended)"`, `2) request changes (recommended).`

So the diagnosis and the fix are sound as far as they go. What they do not reach is the other composition direction — the decorator sitting *outside* the wrappers. These six are the gap, and they are the only rows that belong to this PR's subject:

| still refused on `c9a9a12e9` | where the decorator sits |
|---|---|
| `Request Changes. (Recommended)` | outside the trailing punctuation |
| `Request Changes! (Recommended)` | outside the trailing punctuation |
| `"Request Changes" (Recommended)` | outside the quotes |
| `'Request Changes' (Recommended)` | outside the quotes |
| `"Request Changes (Recommended)".` | inside the quotes, punctuation outside |
| `"Request Changes." (Recommended)` | outside the quotes, punctuation inside |

No fixed order of single-pass replacements covers both directions, which is the thing neither of us saw at the time. #1133 ended up at a shape regex that allows wrapper noise on both sides of at most one decorator.

For completeness rather than as anything to fix: #1133's head also accepts four forms that carry no decorator at all and so are outside what this PR set out to do — `"Request Changes".` and `Request Changes".`, which are a quote/punctuation asymmetry that pre-dates both PRs, and `Request Changes ... !!! "" '''` and a reply whose second line is only `"..."`, which are incidental to the shape regex being broader in its tail than the old chain.

**Everything else on this head checks out.** `bun run check` is rc=0 over 774 files, `t331` is 25/25 and `t188` is 28/28 through `tests/run-tests.sh --unit --filter`, your `package.json` and `bun.lock` are byte-identical to `2931ef02a`, and `git merge-tree --write-tree` against it produces a clean tree with no conflicts. The two source hunks outside `aidlc-lib.ts` are byte-identical to #1133's, including reporting the unstripped reply in the `report` refusal, which is the precedence I recorded earlier.

**Where this stands.** I approved #1133 on `253ab53b9` a short while ago; it is still blocked on two other change requests, so nothing has merged. This stays open as the fallback exactly as agreed, and if it is ever the one that lands, the six rows above and the missing `docs/reference/12-state-machine.md` and `04-stage-protocol.md` updates are what a reviewer would ask for. Nothing here needs a push from you today.

Thank you for #1126, for this fix, and for coming back to rebase it. Both of those are the reason the defect is understood at all.

PR #1025, opened 2026-09-04
chore: upgrade bun to 1.4.0

apackeer:
## What

Upgrades the toolchain from Bun 1.3.14 to Bun 1.4.0.

- Bumps every `bun-version` pin to 1.4.0 - 27 across `ci.yml`, `security-scanners.yml`, `release.yml`, `preview-release.yml`, and `ai-pr-review.yml` (the last three grew their pins after this PR was opened) - and `bun-types` to `^1.4.0` (lockfile stays v1; the only lock movement is `bun-types`).
- `tests/run-tests.sh`: stops unconditionally prepending `~/.bun/bin`, so a caller-pinned bun on `PATH` is respected; the home install remains the fallback when no bun is on `PATH`. The t05 guard builds both competitors - a `HOME`-installed fake bun and a `PATH` fake bun with distinct sentinels, `HOME` pinned to the fixture - and asserts the `PATH` one is selected; it fails 0/1 against the previous runner and passes with this one, on 1.4.0 and 1.3.14 alike.
- t298: the original fix here (detach the holder's stdout, parse the numeric PID line) was superseded on `main`, which now launches the holder with `Bun.spawn` and takes the OS PID; that version passes 59/59 on both Bun versions, so this PR carries no t298 change.

No `core/`, `harness/`, or `dist/` changes; `bun scripts/package.ts --check` passes. No user-facing behavior changes (users install their own bun), so no version/CHANGELOG bump.

## Why now

Bun 1.4.0 is the current stable line (Zig→Rust rewrite). The full 1.3→1.4 upgrade guide and breaking-changes tracker were swept item-by-item against this repo's ~545 Bun-API files; only the two test-infra issues above were genuinely affected. The JUnit XML is not byte-stable across versions (1.4.0 adds a `message` attribute and body to `<failure>`), but the root `<testsuites tests= failures=>` counts that `tests/lib/bun-junit-to-meta.ts` reads are unchanged, and both versions convert to the same STATUS/TESTS/FAILED/RC.

## Validation

All tiers and live slices run under Bun 1.4.0 (children verified on `bun test v1.4.0`):

| Gate | Result |
| --- | --- |
| Linux smoke+unit+integration | 354 pass / 10 fail → 9 pass rerun solo (load flakes); 1 real (t298) fixed |
| Claude SDK integration slice | 9/9 |
| Deterministic e2e (`^t[0-9]`) | 32/33 — t138 times out identically on a 1.3.14 control of the same tree (bun-independent; live-latency budget) |
| Claude TUI live | 19/20 — red passed rerun solo |
| Kiro ACP live | 9/10 — red passed rerun solo |
| Kiro TUI live | 3/3 |
| Codex exec live | 5/5 |
| Native Windows smoke | 13/14 — both t05 reds reproduce identically on a 1.3.14 baseline (pre-existing) |
| Windows Kiro IDE live | 2/3 — remaining red reproduces identically on 1.3.14 (pre-existing) |
| `bun install --frozen-lockfile` (Linux + Windows) | clean, node-pty prebuilds fine |
| `bun run check` (packager drift ×7, 3 tsconfigs, biome) | pass |

Every red was rerun solo and/or controlled against bun 1.3.14 on the same tree and box; none is attributable to Bun 1.4.

## Acknowledgment


wowzoo (changes requested):
Thanks — the upgrade itself looks right, and the two runtime issues you isolated are real. I checked
both rather than taking them on faith, and ran the current tree on 1.4.0 and 1.3.14 side by side.
Two things need a pass, and both come from `main` having moved since you opened this.

## What holds up

**The t298 fix is correct and I reproduced why it's needed.** Head separates the holder's
stdout/stderr to `/dev/null` and locates the numeric line (`:600-605`). Restoring the base
implementation makes the targeted test fail 0 pass / 1 fail: stdout arrives as `PID\nREADY PID`, so the
old `.pop()` picks `READY PID` and yields `NaN`. Head runs 59 pass / 0 fail / 268 expects on **both**
1.4.0 and 1.3.14, so the fix isn't version-locked.

**`tests/run-tests.sh:7-16` does what the description says** — a bun already on `PATH` is preserved, and
`$HOME/.bun/bin` is prepended only when none is found, with a re-check after.

**The lockfile claim holds.** `bun.lock:2` stays v1, and `bun install --frozen-lockfile
--ignore-scripts` is accepted by both versions (1.3.14: "Checked 111 installs across 126 packages (no
changes)").

**And the toolchain is compatible where I could test it.** `t244-install-management` — which postdates
your sweep and leans on `Bun.spawn`, `Bun.sleep`, `Bun.spawnSync`, `Bun.which`, and `Bun.YAML.parse` —
runs 35 pass / 1 Windows-only skip / 0 fail on 1.4.0 and identically on 1.3.14. That's a useful data
point for the upgrade rather than against it.

## Blocker 1 — `release.yml` has five more pins now

`.github/workflows/release.yml` had **zero** `bun-version` pins at this PR's base and head. Current
`main` has **five**, all `1.3.14`, at `:54`, `:80`, `:132`, `:183`, `:485` — added by `12b8d6e0` (#756).
This PR bumps only `ci.yml:45,64,86,116` and `security-scanners.yml:163`.

So this isn't something you missed; the file grew those pins under you. But a plain rebase leaves the
release verify, native build, stage, and publish paths setting up 1.3.14 while CI tests on 1.4.0 —
"the runtime we tested" and "the runtime we shipped with" diverge. For a repo whose release lifecycle
was just hardened, that seems worth closing in the same change.

## Blocker 2 — the new t05 case passes against the unfixed runner

`tests/smoke/t05-run-tests-parallel.test.ts:160-187` builds the fake-bun sentinel on `PATH` but never
creates the `$HOME/.bun/bin/bun` competitor the fix exists to lose to, and doesn't pin `HOME` to a
fixture. On my machine `$HOME/.bun/bin` exists as a directory with no bun executable in it — so with
the **base** runner restored, the new test still reported 1 pass / 0 fail. It isn't detecting the
change; it's detecting my home directory.

I confirmed that's the cause by strengthening it temporarily: create both competitors (a `HOME` bun and
a `PATH` bun, each printing a distinct sentinel) and point `HOME` at the fixture. Then the base runner
gives 0 pass / 1 fail — it selects the home bun — and head gives 1 pass / 0 fail (4 expects). That
version fails before and passes after regardless of the machine.

I found no defect in the `run-tests.sh` change itself; this is only about the test that guards it.

## Non-blocking

- **"the JUnit reporter shape … verified unchanged" is broader than what I see.** The XML is not
  byte-stable across versions: on the same 1-pass/1-fail probe, 1.3.14 emits an empty
  `<failure type="AssertionError" />` while 1.4.0 adds a `message` attribute and a body. What *is*
  stable is the root `<testsuites tests="2" failures="1" …>`, which is all the converter reads
  (`tests/lib/bun-junit-to-meta.ts:172-182`) — both versions convert to `STATUS=FAIL, TESTS=2,
  FAILED=1, RC=1`. So the *consumed* shape is unchanged and there's no compatibility problem; the
  sentence just claims more than that.
- **The sweep's delta needs a second pass after the rebase**, not because the new files are dangerous
  but because there are a lot of them: `tests/` has 164 changed files and 21 new ones since your base.
- **`Require Contributor Statement` is failing** because the PR body doesn't carry the required
  acknowledgment sentence. Worth adding while you're in there.
- Two red signals that aren't yours: the current smoke+unit failure is a single `t241` EPIPE (I ran
  that file 20× on 1.4.0 — 20/20 pass, and it passes on 1.3.14 too, so I see no reproducible 1.4
  regression), and the `Bun audit` failure is an upstream `registry.npmjs.org` 503 rather than a schema
  change. Both should be re-run after the rebase.

## What I ran

At `b4a67f0b` in a detached worktree, plus a second worktree at current `main` for the version
comparison. `bun scripts/package.ts` first in each (since `dist/` is generated now), then:

| check | result |
|---|---|
| t298, head, Bun 1.4.0 | 59 pass / 0 fail / 268 expects |
| t298, head, Bun 1.3.14 | 59 pass / 0 fail / 268 expects |
| t298, base implementation restored, targeted | 0 pass / 1 fail (`READY PID` → `NaN`) |
| t05, head, Bun 1.4.0 | 22 pass / 1 fail / 65 expects — the one failure is a sandbox tmux socket, 1 pass / 1 skip / 0 fail unsandboxed |
| new t05 case, **base runner restored** | **1 pass / 0 fail** ← blocker 2 |
| same case, strengthened with both competitors + pinned `HOME` | base 0/1 fail · head 1/0 pass |
| t244, current main, Bun 1.4.0 | 35 pass / 1 skip / 0 fail |
| t244, current main, Bun 1.3.14 | 35 pass / 1 skip / 0 fail / 503 expects |
| t241 ×20, Bun 1.4.0 | 20/20 pass |
| `bun scripts/package.ts --check` | 7/7 OK |
| `bun run typecheck` / `bun run lint` / `git diff --check` | exit 0 / exit 0 / exit 0 |

Requesting changes on the two items above. The upgrade and both runtime fixes are sound — it's the
release-pipeline pins and the guard test that need another pass, and the rebase forces a visit anyway.

apackeer:
Rebased onto `main` (`a0ee4415e`) and addressed both blockers.

**Blocker 1 - pins.** All 27 `bun-version` pins on current main now read 1.4.0 in each file's own quoting: `ci.yml` 5, `security-scanners.yml` 1, `release.yml` 8, `preview-release.yml` 10, `ai-pr-review.yml` 3. `bun.lock` stays `lockfileVersion: 1`; the only lock movement is `bun-types` (`^1.3.13` → `^1.4.0`, resolved 1.3.14 → 1.4.0). `bun install --frozen-lockfile` exits 0 on both versions.

**Blocker 2 - the t05 guard.** It now creates both competitors (`<fixture>/.bun/bin/bun` → `HOME-BUN-SENTINEL`, `<fixture>/path-bin/bun` → `PATH-BUN-SENTINEL`), pins `HOME` to the fixture and `PATH` to `<fixture>/path-bin:/usr/bin:/bin`, and asserts the PATH sentinel appears and the HOME one does not. With main's runner copied over `tests/run-tests.sh` it fails 0 pass / 1 fail (`Expected to contain: "PATH-BUN-SENTINEL" · Received: "HOME-BUN-SENTINEL\n"`); with this PR's runner 1 pass / 0 fail - on Bun 1.4.0 and 1.3.14.

**t298.** Main already replaced the shell PID capture with `Bun.spawn` + OS PID (readiness/release files, guaranteed reaping), so this PR's t298 hunk was dropped in the rebase rather than regressing that redesign. Resolved t298: 59 pass / 0 fail on both versions.

Non-blocking items: contributor statement added to the body; the JUnit sentence now says what is actually stable (the root counts the converter reads), not "unchanged".

Gates on both Bun 1.4.0 (installed side-by-side, verified in child test output) and 1.3.14: t05 28/28, t298 59/59, `bun scripts/package.ts --check` 7/7 deterministic, `bun run typecheck`, `bun run lint`, `git diff --check` all exit 0. Note for later: Bun 1.4.2 is the current latest; this PR deliberately stays on the 1.4.0 that was swept and validated.

PR #1036, opened 2026-09-07
feat(intent): add abandon/restore status for in-flight intents

### LINKED ISSUE #980: [Feature]: No in-flight way to abandon (cancel + discard) an in-flight intent
## Summary
AI-DLC provides no first-class way to abandon an in-flight intent. There is no
command and no terminal status for "I want to stop this intent partway through."
The intent status graph is a single path — `in-flight` → `complete` — and every
call site that transitions status passes `"complete"`, so there is no wiring to
any other terminal status. As a result, a partially-done intent that the user
gives up on stays `in-flight` forever in the `/aidlc intent` list. The only way
out is to manually archive the record directory and hand-edit `intents.json`.
This is **not a bug** (nothing is broken or corrupted). The archive-don't-delete
policy for records is a deliberate, correct design choice for auditability. What
is missing is the *abandon* use case: a supported way to move an intent to a
terminal, non-`complete` state while preserving its record and audit trail.
- Type: Feature request / UX gap
- Severity: Medium (missing capability; no data loss or misbehavior)
- Environment: Kiro CLI harness, space `default`
## Expected behavior
A first-class command — e.g. `/aidlc intent abandon <name>` (or `archive` /
`discard`) — that safely retires an in-flight intent. The abandoned intent gets
a terminal status such as `abandoned` (or `archived`), is hidden from the default
`/aidlc intent` listing (or clearly marked as retired), and its record directory
and audit shard are preserved intact. Optionally a `restore <name>` to bring it
back.
## Actual behavior
1. No abandon/discard command exists. `aidlc-utility.ts` exposes only
   `intent-create` and `intent` (list / create) — nothing equivalent to
   delete / remove / archive / discard / abandon.
2. The only documented guidance is "archive the record directory and start a new
   `/aidlc`" (docs/guide/01-getting-started.md → Re-initialization;
   docs/reference/04-stages/initialization.md). That is the *start-over* story,
   not "remove one specific in-flight intent from the list."
3. Status transitions are a single path in practice: create → `in-flight`, and
   completion → `complete`. All three `updateIntentStatus()` call sites in
   `aidlc-state.ts` pass `"complete"`, and all are on the workflow-completion
   path. There is no route to any terminal status other than `complete`.
4. Consequently an abandoned intent remains `in-flight` in the `/aidlc intent`
   list indefinitely.
## Reproduction
1. Have an in-flight intent (e.g. `learnings-gate-split`) you decide to stop.
2. Run `/aidlc intent` — no abandon/discard/delete action is offered.
3. Manually archive the record directory — the intent still shows as `in-flight`
   because its row remains in `intents.json`.
4. Hand-edit `intents.json` to remove the row — only now does it disappear from
   the list. (This hand-edit can only be done in a workflow-inactive session.)
## Evidence (code / docs)
- `aidlc-utility.ts`: subcommand switch has no abandon-family verb (only
  `intent-create` / `intent`).
- `aidlc-lib.ts:3938` / `:4107`: intent creation always sets `status: "in-flight"`.
- `aidlc-lib.ts:3951-3970`: `updateIntentStatus(status: string)` is generic and
  accepts any status. Its comment says it flips to a "terminal/other status
  (e.g. \"complete\")" — `complete` is only an example, so the primitive already
  supports terminal states other than `complete`.
- `aidlc-state.ts:1438` / `:4396` / `:5714`: all three `updateIntentStatus` calls
  pass `"complete"`, all on the workflow-completion path ⇒ no wiring to any
  non-`complete` terminal status.
- `aidlc-utility.ts` doctor Check 7 (registry ⇄ record-dir reconciliation, near
  line 4590): treats both "row present, dir gone" and "dir present, no row
  (orphan)" as advisory (pass=true), noting "a rename is a user action, not a
  framework fault." ⇒ registry hand-editing is treated as expected operator
  behavior, yet users are forced to rely on it to abandon an intent.
- docs/guide/01-getting-started.md (Re-initialization) /
  docs/reference/04-stages/initialization.md: framing is record-dir *archive*,
  not intent *delete*.
## Is this spec or defect?
- Keeping records (archive, don't delete) is a deliberate, correct design for
  auditability. No change proposed there.
- Leaving an abandoned intent as `in-flight` is **not** an intentional design
  decision — it is the consequence of there being no wiring to a non-`complete`
  terminal status. The primitive (`updateIntentStatus`'s genericity) can already
  accept an abandoned state; only the command/route that would call it is missing.
- Therefore this is best characterized as a **missing feature for the abandon
  use case**, not a broken behavior.
## Proposed solution
- Add `/aidlc intent abandon <name>` (and optionally `restore <name>`) that
  internally calls `updateIntentStatus(..., "abandoned")` to move the intent to a
  terminal status. Preserve the record directory and audit shard; hide abandoned
  intents from the default list (show with `--all`).
- Minimal alternative: give the intent list a "retired (abandoned/archived)"
  distinction plus a no-hand-edit archive-move command, so users no longer have
  to edit `intents.json` by hand.
## Workaround (current version)
In a workflow-inactive session (Kiro CLI exited), from a terminal:

### Use Case

What problem does this solve or what workflow does it improve?

Developers routinely start an intent, get partway through, and then decide to stop it — the direction was wrong, priorities changed, it was an experiment, or a better intent superseded it. Today there is no supported way to say "I'm done with this one, retire it." The intent is stuck at in-flight and clutters the /aidlc intent list forever, and the only escape is to exit the workflow, manually move the record directory, and hand-edit intents.json — an error-prone, undocumented-as-a-delete operation that also risks leaving the registry and record directory out of sync.

This feature gives the abandon/discard workflow a first-class, safe path: retire an in-flight intent to a terminal state with a single command, keep the /aidlc intent list showing only live work, and preserve the record and audit trail for later reference — no hand-editing, no exiting the session, no risk of a half-cleaned registry. It closes the gap between "complete an intent" (fully supported) and "give up on an intent" (unsupported), so the intent list stays an accurate picture of what is actually in progress.

### Version

v2 (current)

### Area

Platform support

### Additional Context

_No response_

fsatsuki:
Closes #980

## Summary

AI-DLC has no first-class way to abandon an in-flight intent. The intent status graph is a single path — `in-flight` → `complete` — and every call site that transitions status passes `"complete"`, so there is no route to any other terminal status. An intent the operator gives up on stays `in-flight` in `/aidlc intent` forever, and the only escape is to exit the workflow, manually move the record directory, and hand-edit `intents.json`.

This is not a bug. Archive-don't-delete is a deliberate, correct design for auditability, and nothing here changes it. What was missing is the **abandon** use case: a supported way to move an intent to a terminal, non-`complete` state while preserving its record and audit trail.

This PR adds `/aidlc intent abandon <name>` and `/aidlc intent restore <name>`, plus `--all` on the listing. Nothing is ever deleted or moved: only the registry row's `status` changes, and the record directory and audit shard are preserved intact.

## What changed

**New commands**

| Command | Behavior |
|---|---|
| `/aidlc intent abandon <name>` | Move an in-flight intent to the terminal `abandoned` status. Drops out of the default listing; record dir and audit shard preserved. Emits `INTENT_ABANDONED`. |
| `/aidlc intent restore <name>` | Return an abandoned intent to `in-flight`. Emits `INTENT_RESTORED`. Does not switch to it — restoring and selecting stay separate moves. |
| `/aidlc intent list --all` | Include abandoned intents. The default listing reports how many are hidden, so the visible count is never silently partial. |

**Status vocabulary.** `INTENT_STATUS_IN_FLIGHT` / `COMPLETE` / `ABANDONED` name every lifecycle value in one place. Creation, the completion path in `aidlc-state.ts`, the abandon/restore verbs, the listing filter, and the doctor reconciliation all spell the status through them, never as a bare literal.

**An abandoned intent is never treated as live.** Four separate layers could have kept one behaving like live work:

- `activeIntent()` refuses to *infer* an abandoned record on both inference routes — the cursor and the lone-record fallback. Because abandon preserves the record directory, abandoning a space's only intent would otherwise hand that record straight back as active: exactly the "terminal yet active" state the verb exists to remove. An explicit override (session binding, `--intent`) stays exempt, and such an intent stays visible in the listing so the operator is never stranded.
- Switching to an abandoned intent is refused with a message pointing at `intent restore`, rather than silently making a terminal record the active cursor.
- Abandoning the active intent releases the session binding, not just the active-intent cursor.
- The delegated-agent guard still covers both verbs. They rewrite the shared registry row and release the cursor and binding, so they stay main-session-only: a reviewer subagent must not abandon the intent it was dispatched into.

**Audit.** `INTENT_ABANDONED` / `INTENT_RESTORED` are written to the target intent's **own** shard, named explicitly rather than resolved from the ambient cursor. An omitted target resolves through `activeIntent()`, so abandoning a non-active intent would file the event under whichever intent happened to be active, and abandoning the active one would land it in the bare space root once the cursor was released. Preserving the intent's own auditable history is the point of abandon-not-delete. Both events are registered in the canonical registry (`audit-format.md`) and the emitter registry (`docs/reference/12-state-machine.md`), taking the taxonomy from 91 to 93.

**Doctor.** The registry ⇄ record-directory reconciliation counts abandoned rows, so a preserved record directory is not mis-reported as an orphan.

**Routing.** Both verbs are wired into the deterministic-dispatch seam, and `--all` is carried on the parsed listing. A verb or flag the parser does not model is dropped from the reconstructed argv, so without this the `<name>` never reached the tool and `--all` silently downgraded to the default listing. A nameless `intent abandon` is a seam-level error instead of a nameless dispatch, and a leading `--all`/`--json` is the listing rather than a record literally named `--all`. `abandon` and `restore` join the reserved record names, so `intent abandon <name>` cannot be ambiguous with a record called `abandon`.

## Verification

Verified end-to-end through the real `/aidlc` dispatcher on a scratch workspace: the name reaches the tool, the default listing hides the intent and reports the hidden count, `--all` reveals it as `abandoned`, switching is refused, the audit event lands in the intent's own shard, and restore returns it to the listing.

New regression tests cover the behavior that was wrong, not just the happy path:

- the audit event lands in the target intent's own shard — for a non-active intent, for the active intent, and for restore — with nothing leaking into the space root
- a lone abandoned record is not resolved as active, including when a stale cursor still names it
- switching to an abandoned intent is refused and points at `restore`
- `abandon`/`restore` forward both the verb and the `<name>` through the seam, and a nameless invocation is an error
- the listing forwards `--all` alone, with `--json`, and in bare form; `--all` is intent-only
- the delegated-agent boundary still covers both verbs, via `aidlc.ts` and `aidlc-utility.ts`

Existing pins updated with the source: the event count (91 → 93) in t28, t81, t111, and t239, and the parser shape in t229.

Full local runs: `bun scripts/package.ts --check` clean (all seven harness trees in sync), `biome check` clean on every touched file, and 293 tests green across the intent-lifecycle, audit, parser, guard, and documentation-parity files.

## Notes for reviewers

- The event count previously drifted: the two new events were in `VALID_EVENT_TYPES` with no registry row, which failed the `t48` emitter-drift gate. That is fixed here.
- The repository sets `"formatter": { "enabled": false }` in `biome.json`. This diff carries no reformatting — every changed line is a semantic change, so the diff is reviewable line by line.
- No version bump or CHANGELOG entry is included in this PR.


leandrodamascena:
Thanks for working through this, @fsatsuki.

We merged #1033, which solves the same #980 lifecycle with `intent archive` and `intent unarchive`. It also includes the runtime and concurrency guards added during review.

The remaining difference here is the command naming, `abandon` and `restore`, so I am closing this PR as superseded by #1033.

I appreciate the work you put into this. It helped confirm the behavior and edge cases we needed to cover.

PR #1038, opened 2026-09-07
fix: type every engine ask so conductors never answer through report

apackeer:
## Summary

A customer on 2.7.0 with Codex CLI typed a short description containing "workshop" on a fresh workspace. The engine emitted the cold-start scope-confirm question, and the codex orchestrator skill told the conductor to feed the answer back through `report --user-input` without naming a `--result`. Codex invented one:

```
bun .codex/tools/aidlc-orchestrate.ts report --result answered --user-input Workshop
{"kind":"error","message":"Unknown --result \"answered\". accepted outcomes: approved, completed, complete, done, awaiting-approval, rejected, revised, resume, resumed, skipped."}
```

The same failure reproduces with any freeform first prompt ("fix the login bug" hits the same ask for `bugfix`; longer prose hits the compose offer). Even a valid verdict fails there, because no workflow state exists yet.

## Root cause

The `ask` row of the directive table in all seven harness skills said "For every other ask, feed the human's answer back on the next `report` via `--user-input`". The engine has exactly one `report` round-trip for an ask, the resume menu (`--result resumed`). The scope confirm, compose offer, intent picker, and paused-Unit asks all expect their answers to come back through `next` or a command named in the question. The sentence dates from the first v2 commit; every ask that broke since then received its own typed exception (new-work routing, unit claim, legacy plan-approval recovery, the resume menu) while the fallthrough sentence stayed. No test exercised the cold-start confirm through a conductor, and the codex journey test bypasses it with an explicit `--scope`.

## Changes

Engine (`core/tools`):

- Every `ask` that `next` emits now carries `ask_type` and `response_route` plus the exact follow-up command(s): `scope-confirm` (`confirm_command`, `compose_command`, `scope_command_template`), `compose-offer`, `intent-pick` (`select_commands`, one complete POSIX-quoted command per selector), `unit-paused` (`resume_command`, route `command`). Main's four existing typed asks (`new-work-routing`, `unit-claim`, `legacy-plan-approval-recovery`, `guard-recovery`) keep their fields and routes.
- Request text is transported once. At ask time the engine stores the pending request (`{id, description, proposedScope, createdAt}`, 8-hex id, one per-person slot at `aidlc/.aidlc-sessions/pending-request.json`) and every command carries `--pending-request <id>` instead of embedding the text; `next`/`intent create` resolve the id back to the exact original and delete the slot on creation. A missing, replaced, or consumed id is a typed `error` (`Pending request <id> is no longer available; restate the request.`). The question echoes at most 240 characters of the request; `intent_text` carries the full text once. A 6,000-character request that previously exceeded the 28 KiB directive limit now emits the ask, and a 20,000-character one round-trips to an intent with the exact text.
- Branch 9a no longer limits the typed `new-work-routing` ask to Kiro: on every harness, cold-start prose in a clone with existing intents and no cursor gets `scope-confirm`, and the second hop after `confirm_command` is `new-work-routing` carrying the same description and proposed scope, not a bare picker that drops the request.
- The untyped `ReportAskDirective` variant is removed from the directive union and `validateDirective` requires the typed fields per subtype.
- `report` errors for an unknown `--result` or a missing state file now add that an ask's answer is never reported except for the resume menu.

Prompts and docs:

- The `ask` row in every harness `SKILL.md` describes the typed route contract (`next` runs the named command and never calls `report`; `command` runs the named command then re-runs `next`; `report` is the resume menu only; `claim` unchanged).
- The codex, kiro, and kiro-ide `question-rendering.md` annexes, the generated runner instructions (`aidlc-runner-gen.ts`), `docs/reference/03-orchestrator.md`, and `docs/reference/17-skill-system.md` follow.

Tests:

- t171: typed fields on the scope-confirm, compose-offer, and intent-pick asks, quoting of an apostrophe in the intent text, and the round trip of executing the emitted `confirm_command` to the intent-create print.
- t115: `report --result answered` with and without state names the ask route.
- t181: no harness ask row may route an ordinary ask through `report`; every row and the docs row must mention `response_route`.
- t113: validator accepts each new subtype and rejects an untyped ask. t260: paused-Unit ask fields.

Release: version 2.7.2, README badge, CHANGELOG entry. `dist/` regenerated; `bun scripts/package.ts --check` passes.

## Verification

All runs: `bash tests/run-tests.sh --debug -P 8 <tier>` on this branch (macOS, another session's test run was loading the machine at the same time, so reds were rerun alone at `-P 6` and compared against the base commit e7689885f in a separate worktree).

| Run | Files | Result | Reds and classification |
|---|---|---|---|
| smoke + unit | 273 | 270 pass, 3 red | t255, t298: red on base too (git calls killed at the 5s test timeout; a `$!` output-ordering race in the test). t314: green alone. |
| integration | 115 | 105 pass, 10 red | t163, t52, t49, t99, t326, plugin.test.ts: green alone (5s hook timeouts). t327: green alone and green on base. t188: red on base too (macOS `/var` vs `/private/var` realpath in a temp path). t325: red on base too. t72: live SDK, see below. |
| e2e deterministic (`^t[0-9]`) | 33 | 32 pass, 1 red | t138: live SDK full-workflow drive, see below. |
| integration `--filter t171` | 1 | pass | The new round-trip test: emitted `confirm_command` executes to the intent-create print. |
| `bun run typecheck` | | pass | |
| `bun scripts/package.ts --check` | | pass | all seven harness trees in sync |
| `bun run lint` | | pass | |

Live SDK reruns (t138, t72): t138 timed out at its 45-minute budget twice (full security-patch workflow, `/aidlc security-patch ...`, so it never reaches the cold-start ask); its trace shows zero ask directives, zero engine errors, and steady progress through two gated stages into Code Generation plan approval. t72 timed out at its 25-minute budget twice on this branch and passed once on the base commit in the same hour. All three t72 traces hit the same pre-existing friction: the developer `aidlc-log.ts link` call is refused with "Cannot record reverse-engineering developer link", both conductors then read engine source for about five minutes, and `aidlc-learnings.ts surface` later fails on the runtime graph. The base conductor recovered by asking the human a question at minute 19 (the test's first assertion); the branch conductor chose a redo jump and ran out of time. No changed path (typed asks, report errors) appears in any of the traces.

Reproduction against the regenerated codex/claude tool: a bare prose start now returns `ask_type: "scope-confirm"` with `confirm_command`, `compose_command`, and `scope_command_template`; long prose returns `compose-offer`; `report --result answered` and a no-state `report --result approved` both error with the ask-route sentence appended.

No live test drives a conductor through the cold-start ask end to end (that gap is how this shipped). Adding one is a follow-up; the SDK driver would need to answer the rendered question.

## Upgrade

Native installs pick this up with the binary; copy-channel installs re-copy `dist/<harness>/`. No workflow state migration.

## Acknowledgment


rafaelgsr (changes requested):
Thanks for tackling the ask-routing failure and for carrying the contract across the harnesses.

I reviewed the current head `9fae19c5fe7fa88e677877c9657d0ad2a5debb43` against `e7689885fb98d380421ec574a7fa9f3301a22205`. This bug fix is aligned with the project direction: it restores the existing contract that engine asks follow their typed response route, and the work has not been superseded on current `main`.

[P1] Quote selectors before exposing an executable template — `core/tools/aidlc-orchestrate.ts:968`

`available_intents` contains raw on-disk record-directory names, including surfaced hand-created or migrated orphan records, but `select_command_template` inserts `<selector>` unquoted. The Kiro instructions replace that placeholder verbatim and execute it. A selector such as `customer work` is split into separate argv values, while shell metacharacters can execute additional commands. Please emit shell-safe commands per selector or structured argv, and cover spaces and shell metacharacters in a regression test.

[P1] Preserve pending work when confirmation reaches an unselected workspace — `core/tools/aidlc-orchestrate.ts:1944`

On Claude, Codex, Cursor, Copilot, and OpenCode, freeform work in a clone with existing intents but no active cursor first emits `scope-confirm`. Running the new `confirm_command` reaches Branch 9a with `pendingWork`, but this fallback drops the description and proposed scope and emits only an `intent-pick` for existing work. The confirmed request therefore disappears. Please preserve the pending request through new-work routing on every harness, or add an explicit new-work route to this picker, and test the second hop after `confirm_command`.

[P2] Keep typed asks below the directive transport limit — `core/tools/aidlc-orchestrate.ts:934`

The full intent is repeated in the question, `intent_text`, and each command field. A 6,000-character request now exceeds the 28 KiB directive limit and exits without emitting JSON, while the base revision emitted the ask successfully. Please deduplicate or structurally encode the command data so valid detailed requests stay within the transport limit, with a boundary regression test.

Validation performed on this head: targeted `t113`, `t115`, `t181`, `t260`, and integration `t171` passed; `bun run typecheck`, `bun scripts/package.ts --check`, and `git diff --check` passed.

Remaining state: current `main` is 21 commits ahead and GitHub marks this PR conflicting, so it needs a rebase and release-metadata resolution. GitHub code, test, and security checks are green; `Require Contributor Statement` is still failing because the required statement is absent from the PR description.

Requested review event: REQUEST_CHANGES.

apackeer:
Rebased onto `main` (`a0ee4415e`) - one commit, no `dist/`, version, CHANGELOG, or README changes - keeping main's four typed asks and the guard-recovery contract intact, and addressed all three items.

**P1 - quoted selectors.** `select_command_template` is gone; `intent-pick` emits `select_commands: [{ selector, command }]`, each a complete invocation with the selector POSIX-single-quoted (`shellArg`). Regression in t171 runs the emitted commands through `sh -c`: a record named with a space is selected; `x; touch <fixture>/pwned` creates no `pwned`. Negative proof: an unquoted selector in one command → `pwned` created, test fails.

**P1 - pending work in an unselected workspace.** Branch 9a's `isKiroRoutingHarness()` condition is removed: any harness with `pendingWork` gets the typed `new-work-routing` ask (description, proposed scope, selectors, route `next`); the plain picker is now the typed `intent-pick` and only appears when nothing is pending. Regression: non-Kiro cold-start prose → `scope-confirm`; executing `confirm_command` → second hop is `new-work-routing` with the same description and scope; route 2 creates the intent with that description. Negative proof: restoring the Kiro-only condition → second hop is `intent-pick`, test fails.

**P2 - transport size.** Commands no longer embed the request. The engine stores the pending request once (`aidlc/.aidlc-sessions/pending-request.json`, per-person, 8-hex id, one slot per project) and commands carry `--pending-request <id>`; `next`/`intent create` resolve it and delete the slot on creation; stale/consumed/mismatched ids return a typed `error` naming the remedy. The question echoes ≤ 240 chars; `intent_text` carries the full text once. Regressions: 6,000-char and 20,000-char requests emit the ask and round-trip to intents with the exact text (apostrophe and `$(...)` included); a stale id creates nothing. Negative proof: restoring the original full repetition (question + three commands + `intent_text`) for the 6,000-char case → `refusing to emit a directive larger than 28672 bytes`.

Cutover notes: t118 (unit+integration) and t198 pinned the old embedded `--arguments '<text>'` shell spelling; those pins are replaced by real round trips through the token. Contributor statement added to the body; release slot removed from the title.

Verified: t113 77, t115 57, t171 21, t181 28, t260 29, t198 19, t118 27+18 - 0 failures; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0.

Issue #1031, opened 2026-09-07
[Bug]: Code-Generation guard blocking on poc, bugfix, refactor, security-patch, express cause there isn't a plan


michael-proctor:
### Description

Everything up to generation is done and approved. But the dispatch itself is unreachable, and it's a defect in your framework rather than anything missing here.

The bug. knownUnits() in .claude/hooks/aidlc-plan-approval-guard.ts enumerates <record>/construction/*/ directories as unit names. On a zero-Unit workflow — which poc is by design, since it skips units-generation — the engine's own produces paths put the stage artifacts at construction/code-generation/. So the guard reads the stage directory as a unit named code-generation, then looks for its approval evidence at construction/code-generation/code-generation/, which cannot exist:

unit "."               -> ok      | approved
unit "code-generation" -> BLOCKED | code-generation-plan.md is missing or empty

. has valid evidence but isn't in knownUnits(). code-generation is in knownUnits() but has no evidence. evaluatePlanApprovalDispatch requires the marker to be both, so no value of AIDLC-UNIT can pass.

This hits five of the eleven shipped scopes — poc, bugfix, refactor, security-patch, express all skip units-generation.

What I won't do: create a nested construction/code-generation/code-generation/ to feed the hook. That would duplicate the artifacts, contradict the engine's declared produces paths, and is precisely the "make the check pass" move the construction guardrails forbid. Your plan approval is genuine and verifiable — the guard's unit matching is what's broken.

### Steps to Reproduce

Run a scope of poc, bugfix, refactor, security-patch, express when it gets to Construction the knownUnits code blocks the construction due to lack of plan

### Expected Behavior

Should check if the plan is needed and not block if no plan is part of scope.

### Actual Behavior

Currently Blocks 

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.6.99

### AI-DLC Phase

Construction

### Harness

Claude Code

### AI Model

Opus 5

### Environment

- Windows 11
- Running in Powershell

spraja08:
Thanks for reporting @michael-proctor. This is fixed in 2.7.1 as part of https://github.com/awslabs/aidlc-workflows/pull/997. Could you please check the latest version and comment further?

apackeer:
[...]

PR #1040, opened 2026-09-07
fix(kiro-ide): treat Kiro read tools as read-only in the plan-approval guard

### LINKED ISSUE #965: [Bug]: kiro-ide plan-approval guard denies every path-less tool unconditionally, blocking all subagent dispatch

### LINKED ISSUE #1039: [Bug]: kiro-ide plan-approval guard classifies Kiro IDE read tools (read_file, read_files, list_directory) as mutation-capable — recovery from a lost approval receipt is read-blocked
### Description

The Kiro IDE plan-approval guard denies the IDE's read tools during a Code Generation Plan Approval window.

The adapter classifies tools by exclusion. `PLAN_APPROVAL_SAFE_READ_TOOLS` (`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:167-173`) lists only `fs_read`, `file_search`, `grep_search`, `thinking`, and `todo_list`. Kiro's tool catalog (https://kiro.dev/docs/tools/) names the read tools `read_file`, `read_files`, `list_directory`, `read_code`, `web_fetch`, and `web_search`. `fs_read` is only an alias. The set does not contain these names. `mutationCapableTool()` (`:1081-1083`) therefore returns true for every one of them.

Two deny paths follow:

- **Path-bearing read** (`read_file { path }`, `read_files { paths }`). The adapter does not recognise the name. It forwards the call verbatim to the core guard (`:1594-1602`). The core guard's `SAFE_READ_TOOLS` (`core/hooks/aidlc-plan-approval-guard.ts:96-109`) also lacks the name. The guard treats the call as an opaque mutation (`:701-705`) and denies it.
- **Path-less read** (`list_directory`, `web_fetch`). `opaqueMutation` (`:1283-1294`) is true. With an active unapproved window, the legacy branch (`:1459-1475`) denies the call.

The Kiro CLI adapter already maps `read`, `fs_read`, `read_file`, and `read_files` to `Read` (`harness/kiro/hooks/aidlc-kiro-adapter.ts:844`). The IDE adapter does not. The two Kiro harnesses disagree on what a read is.

**Impact.** To re-run Plan Approval, the conductor must read `code-generation-questions.md`. When the receipt is missing (see #975), the guard denies that read until approval is recorded. Approval cannot be recorded without the read. The workflow cannot recover. A customer hit this on Kiro IDE on macOS and on Windows. The only exit was to set `aidlc-plan-approval-guard.json` to `"enabled": false`.

**Related.** #965 and #986 remove the unconditional deny for path-less tools. They do not change the allowlist. #1000 fixes the receipt loss. It does not change either allowlist (`aidlc-kiro-adapter.ts:171`, guard `:98`). This defect survives all three.

### Steps to Reproduce

1. Install `dist/kiro-ide/.kiro/` into a project. Keep `aidlc-plan-approval-guard.json` enabled.
2. Run `/aidlc express <intent>` to Construction → code-generation.
3. Let the conductor write `code-generation-plan.md` and `unit-test-instructions.md`.
4. Let the conductor present Plan Approval. Do not approve.
5. Ask the conductor to read the questions file. It calls `read_file` or `list_directory`.
6. Observe the PreToolUse deny.

Deterministic variant, no IDE needed:

1. Create a fixture with an active `run-stage` / `code-generation` directive and no receipt.
2. Pipe this payload to `bun .kiro/hooks/aidlc-kiro-adapter.ts plan-approval-guard`:
   `{ "hook_event_name":"PreToolUse", "tool_name":"read_file", "tool_input":{"path":"<questions file>"} }`
3. Observe exit code 2.

### Expected Behavior

The guard never classifies a read-only tool as mutation-capable. `read_file`, `read_files`, `list_directory`, `read_code`, `web_fetch`, and `web_search` return `null` from the adapter, as `fs_read` does. The core guard accepts these names if the adapter forwards them. The conductor can re-run Plan Approval with the guard enabled.

### Actual Behavior

The core guard denies path-bearing reads with: `Code generation cannot run mutation-capable shell command: unknown mutation-capable tool: read_file for the zero-Unit stage-level implementation ...`

The adapter denies path-less reads with: `Legacy Plan Approval permits only single-file planning writes before approval; this mutation-capable tool is not safely attributable.`

The conductor cannot read the artifacts it must edit. The workflow stops until a human disables the guard.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.7.1 / `e7689885`. The allowlists have not changed since #965 closed on 2.6.123.

### AI-DLC Phase

Construction

### Harness

Kiro IDE

### AI Model

Claude Opus 4.8

### Environment

- OS: macOS and Windows (both reproduced)
- Harness: Kiro IDE v1.0.242
- bun: v1.4.2

### Additional Context

**Suggested fix.** The change is small and does not conflict with #1000.

1. In `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:167`, add the Kiro read tools and their aliases to `PLAN_APPROVAL_SAFE_READ_TOOLS`: `read`, `read_file`, `read_files`, `read_code`, `list_directory`, `glob`, `grep`, `web_fetch`, `web_search`.
2. In `core/hooks/aidlc-plan-approval-guard.ts:96`, add the same names to `SAFE_READ_TOOLS`. The adapter forwards unknown names verbatim, so the core guard needs the list too.
3. Add a unit test. Seed an active unapproved code-generation window. Assert that each read payload exits 0. Assert that the `fs_write` deny is unchanged.

**Out of scope.** `execute_bash`, `control_bash_process`, MCP tools, and sub-agent dispatch stay outside the safe set. This issue covers only Kiro's documented first-party read tools.

**Cross-references.** #975 (receipt loss that this defect compounds). #965 / #986 (earlier fix next to this allowlist). #1000 (does not address this). #1031 (also affects express at the same gate).

AdrianBegg:
# Summary

Fixes #1039. On Kiro IDE, the plan-approval guard denied the IDE's own read tools during a Code Generation Plan Approval window. The kiro-ide adapter classifies tools by exclusion, and its safe-read allowlist named only the `fs_read` alias. Kiro's documented read tools (`read_file`, `read_files`, `list_directory`, `read_code`, `web_fetch`, `web_search`, per https://kiro.dev/docs/tools/) were therefore classified as opaque mutations.

This made the #975 receipt loss an unrecoverable stop. To re-run Plan Approval, the conductor must read `code-generation-questions.md`, and the guard denied that read until approval was recorded. A customer reproduced this on Kiro IDE v1.0.242 on macOS and Windows. The only exit was to disable `aidlc-plan-approval-guard.json`.

This PR closes #1039 only. It does not close #965, #975, #1031, or #1044. See "Related issues" below.

## Changes

The behavior change is Kiro IDE only. The version bump is global by design.

* `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts` — extend `PLAN_APPROVAL_SAFE_READ_TOOLS` with Kiro's documented read tools and the `read`/`glob`/`grep` aliases. For a well-formed payload, a listed tool returns `null` before any forward to the core guard. The malformed-payload branch in `buildForward` runs first and stays fail-closed.
* `tests/unit/t218-kiro-ide-hook-adapter.test.ts` — two regression tests:
  1. Seed an unapproved `run-stage` / `code-generation` window. Assert each read payload (1.x stdin shape and legacy 0.12 `toolArgs: {}` shape) exits 0 while the `fs_write` deny still exits 2.
  2. Open the legacy write-recovery latch (`activeWriteWindows`). Assert each read exits 0 in both payload shapes while `fs_write`, `str_replace`, and `execute_bash` still exit 2 on the latch.
* `core/tools/aidlc-version.ts`, `README.md`, `CHANGELOG.md` — version `2.8.1` with a changelog entry and an **Upgrade** instruction. The branch is rebased on `main` after #756 (`2.7.2`) and #1046 (`2.8.0`), so the patch number is `2.8.1`.

`dist/` is no longer tracked since #756, so this PR does not commit generated files. `bun scripts/package.ts --check` passes.

The core guard, other harnesses, and the write, shell, MCP, and sub-agent dispatch enforcement are not changed.

## Related issues

| Ref | Effect of this PR |
|---|---|
| #1039 | Closed. The read-tool deny on Kiro IDE is fixed. |
| #965 | Not closed. Reads now pass, but `invoke_subagent`, `invoke_sub_agent`, and `delegate` still exit 2. |
| #1044 | Not closed. `execute_pwsh` and `shell` still exit 2. This is a different classification path in the same file. |
| #1031 | Not related. The `code-generation` directory is already excluded in `knownUnits()` since #891. |
| #975 | Not closed. Recovery after a lost receipt is now possible. The loss itself is not fixed. |
| #986, #1000, #1005, #1008, #1010 | No conflict. None of these change this allowlist. |

## User experience

**Before:** during Code Generation on Kiro IDE, `read_file` on the questions file fails with `Code generation cannot run mutation-capable shell command: unknown mutation-capable tool: read_file ...`. `list_directory` fails with `... this mutation-capable tool is not safely attributable.` The conductor cannot read the artifacts it must edit to re-approve. The workflow stops until the guard is disabled.

**After:** these reads pass the guard in the same way `fs_read` already did. The conductor can reset `[Answer]:`, regenerate the fingerprint, and present Plan Approval again with the guard enabled.

## Checklist

* [x] Changes are documented

## Test plan

1. `bun scripts/package.ts` then `bun scripts/package.ts --check` — all 7 harness trees in sync.
2. `bun test tests/unit/t68-version-changelog-sync.test.ts` — passes.
3. `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts` — passes. Against the pre-fix adapter, the unapproved-window test fails with `unknown mutation-capable tool: read_file`, and the latch test fails with `a legacy write did not complete PostToolUse mediation`.
4. `bun run typecheck` and `bun run lint` — clean.
5. `bun test tests/unit/t145-packaging-parity.test.ts tests/unit/t239-documentation-parity.test.ts` — passes.
6. Manual: install `dist/kiro-ide/.kiro/`, run an express workflow to Code Generation, present Plan Approval, ask the conductor to read `code-generation-questions.md`. The read is allowed. An `fs_write` outside the record directory is still denied.

## Acknowledgment


wowzoo (changes requested):
Thank you for this — the diagnosis is the hard part and you got it right. Tracing the deny to
`PLAN_APPROVAL_SAFE_READ_TOOLS` naming only the `fs_read` alias, separating the path-bearing
forward-to-core path from the path-less legacy path, and connecting it to the #975 receipt loss as an
*unrecoverable* stop rather than a nuisance is a genuinely good piece of work. Scoping the second
commit down to the kiro-ide adapter was the right call too, and I traced that reasoning rather than
taking it on faith.

I reviewed the current head `7504f2ca` against `e7689885`. The fix itself is correct and I found
nothing wrong with the adapter change or the test. One thing is missing before it can merge.

On direction: this is aligned as a narrow Kiro IDE bug fix that restores the guard's existing
read-only contract without changing another harness or weakening write, shell, MCP, or dispatch
enforcement. That is a reviewer assessment, not a claim that a maintainer has already settled it.

## Requested change — restore the release metadata that commit 2 dropped

`AGENTS.md:54-58` and `CONTRIBUTING.md:51` both require it for user-visible changes, and this is
user-visible: a deny users hit disappears, and adopting the fix means replacing `dist/kiro-ide/`.

1. `core/tools/aidlc-version.ts:4` — `AIDLC_VERSION` `"2.7.1"` → next patch (`"2.7.2"` unless
   something else lands first)
2. `README.md:18` — `badge/version-2.7.1-blue` → the same number
3. `CHANGELOG.md` — a new `## [2.7.2] - <merge date>` heading directly above `## [2.7.1]`, with the
   one-paragraph summary carrying the `**Upgrade:**` instruction, then user-facing bullets. For a
   starting point:

   > Fix a Kiro IDE Plan Approval stop that denied the IDE's own read tools during a Code Generation
   > approval window. **Upgrade:** replace the `dist/kiro-ide/` tree; no workflow state migration is
   > required. Closes #1039.
   >
   > * Kiro IDE read tools no longer fail during Code Generation with `unknown mutation-capable
   >   tool` or `this mutation-capable tool is not safely attributable`.
   > * Disabling `aidlc-plan-approval-guard.json` is no longer needed to recover a Code Generation
   >   plan approval on Kiro IDE. Writes, shell, MCP, and sub-agent dispatch stay guarded.

4. Regenerate and re-run:

```bash
bun scripts/package.ts
bun scripts/package.ts --check
bun test tests/unit/t68-version-changelog-sync.test.ts
bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts
```

**One thing to expect, so it doesn't look wrong:** step 4 legitimately re-touches every harness.
`aidlc-version.ts` is copied into all seven `dist/` trees, so a version bump necessarily produces
`dist/{claude,codex,copilot,cursor,kiro,kiro-ide,opencode}/**/aidlc-version.ts` changes and
`--check` fails without them. I suspect that is why commit 2 dropped the bump along with "the
regenerated non-kiro-ide dist copies" — but that churn is the version file being global by design,
not scope creep, and it puts no behavior change into another harness. It adds 10 files and leaves
the adapter and the test untouched:

```
 CHANGELOG.md                                |  7 +++++++
 README.md                                   |  2 +-
 core/tools/aidlc-version.ts                 |  2 +-
 dist/claude/.claude/tools/aidlc-version.ts  |  2 +-
 dist/codex/.codex/tools/aidlc-version.ts    |  2 +-
 dist/copilot/.aidlc/tools/aidlc-version.ts  |  2 +-
 dist/cursor/.cursor/tools/aidlc-version.ts  |  2 +-
 dist/kiro-ide/.kiro/tools/aidlc-version.ts  |  2 +-
 dist/kiro/.kiro/tools/aidlc-version.ts      |  2 +-
 dist/opencode/.aidlc/tools/aidlc-version.ts |  2 +-
```

I applied exactly that at this head and ran it, so it is known-good rather than suggested:
`package.ts --check` all 7 trees in sync, `t68` 7 pass / 0 fail, `t218` 75 pass / 0 fail.

Worth noting that `t68` passes on the current head only because `aidlc-version.ts`, the README badge,
and the latest `CHANGELOG.md` heading all agree at `2.7.1`. It checks that the three are consistent,
not that a user-visible diff came with a bump — so no CI job flags this omission.

## What I verified holds up

- **The regression test is a real regression test.** Reverting only the adapter (both the authored
  file and the dist copy) and re-running your test reproduces the exact deny from the report:
  `read_file: Code generation cannot run mutation-capable shell command: unknown mutation-capable
  tool: read_file ...`. Restored, it passes. That is exactly the shape a fix like this needs.
- **Dropping the core-guard change in commit 2 is correct.** A listed name makes
  `mutationCapableTool()` return false, so the branches at `:1272`, `:1301`, `:1444`, and `:1477`
  are all skipped and `:1539` returns `null` — the call never reaches the core guard. It holds
  across harnesses too: `harness/kiro/` only forwards Task / Bash / Write / Edit and returns 0 for
  everything else (`:602-634`), and `harness/copilot/` maps `read_file` → `Read` (`:179`). kiro-ide
  is the only exclusion-based classifier, so the core list is not load-bearing for these names.
- The guarantee in your description holds: writes, `execute_bash`, `control_bash_process`, MCP, and
  dispatch all still exit 2 in the seeded window. I checked `invoke_subagent`, `invoke_sub_agent`,
  `delegate`, and `control_bash_process` specifically.
- No existing test was weakened — the PR has zero deletions.
- `dist/kiro-ide/` is byte-identical to the authored file, and `bun scripts/package.ts --check`
  reports all seven harness trees in sync.
- Your no-conflict claim about #986 and #1000 checks out — see the table below.

## What this closes, and what it does not

Your description cross-references #975, #986, #1000, and #1031, which could read as this PR
sweeping them up. `Fixes #1039` is the only closure here.

| ref | | effect of this PR |
|---|---|---|
| **#1039** | issue | **Closes it.** The reported and accepted read-tool defect is fixed on this harness. |
| #965 | issue, `CLOSED`/completed | **Partially mitigated, do not close.** Path-less *reads* now pass, but its headline symptom does not: `invoke_subagent`, `invoke_sub_agent`, and `delegate` all still exit 2 at this head. Worth flagging that it was closed as completed on 2026-08-28 with no fix commit, and @omaraws reported on 2026-08-31 that it was still reproducible. |
| #1044 | issue, opened 2026-09-08 | **Not closed.** Same file and same target, different classification path: `execute_pwsh` and `shell` exit 2 at this head while `execute_bash` exits 0. That lives in the nine `execute_bash` comparisons at `:1283`–`:1554`, none of which this PR touches. (`:941` already handles `execute_pwsh`/`shell` for a different target, so the names are known to the adapter.) |
| #1031 | issue | **Unrelated.** The directory-classification path it reports is already excluded on the current base — `core/hooks/aidlc-plan-approval-guard.ts:363` skips the `code-generation` directory in `knownUnits()`, landed in `83efa07d` (#891). |
| #975 | issue | **Not closed, but de-escalated.** This PR makes recovery *possible* after the receipt is lost; it does not stop the loss. The severity drops from unrecoverable to recoverable, which is the real value of this change. |
| #986 | open PR, same file | **No conflict, no supersession** either way — one hunk at `:1502`, and it does not touch this allowlist. |
| #1000, #1005, #1008, #1010, #756 | open PRs | All on `core/hooks/aidlc-plan-approval-guard.ts`. Unaffected: a listed name returns `null` in the adapter before any forward, so this PR and those never meet. |

I'd suggest keeping `Closes #1039` as the only closing keyword, so #965 and #1044 don't get swept
shut by association.

## Non-blocking, take them or leave them

1. **The new comment slightly overstates its invariant.** It says a listed tool "returns null before
   any forward to the core guard, so this set is the single classification point for reads on this
   harness". The malformed-payload branch at `:1150-1196` runs first and can still deny a listed
   read: `read_file` with a non-object `tool_input` during code-generation exits 2 with `Plan
   Approval denied a malformed PreToolUse payload (toolArgs).`, while the well-formed control exits
   0. The behavior looks like deliberate fail-closed handling — it is the comment that needs the
   qualifier.
2. **No coverage for the legacy write-recovery latch.** Your test covers the unapproved-window path
   in both payload shapes, which is the important one. Reads also flipped from deny to allow on the
   `activeWriteWindows` latch at `:1272-1300`, and nothing pins that behavior.

## Validation I ran at `7504f2ca`

| check | result |
|---|---|
| `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts` | 75 pass / 0 fail (594 expect) |
| your regression test vs. reverted adapter | fails with the exact reported deny |
| `bun scripts/package.ts --check` | all 7 harness trees in sync |
| `bun run typecheck` (3 tsconfigs) | clean |
| `bun run lint` (biome, `--error-on-warnings`) | exit 0 |
| t265, t328, t245, gen-coverage-registry, t68 | 143 pass / 0 fail |
| t147, t321, t327, t264, t221 | 135 pass / 0 fail |
| t145-packaging-parity, t239-documentation-parity | 28 pass / 0 fail |
| smoke tier | 14 files, 1737 assertions, PASS |
| t261-audit-authority-floor | 27 pass / 39 fail — **pre-existing**, identical on `main` |
| e2e `t-ide-kiro-*` | not run |

To be clear on that t261 row: it is not yours. I ran it on `main` at `e7689885` and got the same
27/39, so it is a pre-existing failure in my environment, not something this PR touches.

One note for whoever merges this: the six checks running on this PR are all PR-validation only
(title, labels, contributor statement, merge status). No test, lint, or packaging job runs here, so
everything above is local rather than CI-confirmed.

Requesting changes on the release metadata only. The fix and its regression test are solid, and the
recipe above is validated end to end, so this should be a short round trip — ping me and I'll
re-review.

AdrianBegg:
@wowzoo Thank you for the review. I applied all the requested changes. The new head is `2629407d`.

## Requested change: release metadata

The version is `2.8.1`, not `2.7.2`. Two releases landed on `main` after you reviewed: #756 took `2.7.2` and #1046 took `2.8.0`. I rebased the branch on `main` and re-bumped the patch number, as the conflict-trap rule in `AGENTS.md` requires.

* `core/tools/aidlc-version.ts` — `AIDLC_VERSION` is `"2.8.1"`.
* `README.md` — the badge is `version-2.8.1-blue`.
* `CHANGELOG.md` — a new `## [2.8.1] - 2026-09-08` heading is above `## [2.8.0]`. The summary has the **Upgrade** instruction and the two user-facing bullets you proposed.

One difference from your recipe: #756 removed `dist/` from git. The tree is now generated and ignored. So this PR does not commit the seven `dist/**/aidlc-version.ts` copies. `bun scripts/package.ts --check` confirms all seven trees are in sync from the authored source.

## Closing keywords

`Fixes #1039` is the only closing keyword. The PR description now has a table that lists #965, #975, #1031, and #1044 and states that this PR does not close them.

## Non-blocking items

1. **Comment on `PLAN_APPROVAL_SAFE_READ_TOOLS`.** I corrected the comment. It now states that a listed tool returns `null` for a well-formed payload, and that the malformed-payload branch in `buildForward` runs first and stays fail-closed.
2. **Coverage for the legacy write-recovery latch.** I added a test. It opens the `activeWriteWindows` latch with a guarded `fs_write` and a PostToolUse that deletes the state file. Then it asserts that each Kiro read exits 0 in both payload shapes, and that `fs_write`, `str_replace`, and `execute_bash` still exit 2 on the latch. Against the pre-fix adapter, this test fails with `a legacy write did not complete PostToolUse mediation`.

## Validation at `2629407d`

| Check | Result |
|---|---|
| `bun scripts/package.ts --check` | all 7 harness trees in sync |
| `bun test tests/unit/t68-version-changelog-sync.test.ts` | 7 pass / 0 fail |
| `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts` | 76 pass / 0 fail |
| both #1039 tests against the pre-fix adapter | fail with the reported deny strings |
| `bun run typecheck` (3 tsconfigs) | clean |
| `bun run lint` (biome, `--error-on-warnings`) | exit 0 |
| t145-packaging-parity, t239-documentation-parity | 13 pass / 0 fail |

The diff against `main` is five files: the adapter, the test, and the three version files. Ready for re-review.

wowzoo (approved):
Thanks for turning this around so quickly, and for two calls that were better than what I asked for.

The version is right: `2.8.1` after #756 took `2.7.2` and #1046 took `2.8.0`. And you were right not to
commit the seven generated `dist/**/aidlc-version.ts` files — my recipe was written against the
pre-#756 tree, where `dist/` was still tracked. Following it would have put generated files back into
git. Good catch on both.

Approving. I re-verified everything rather than reading your table, and both non-blocking items landed
well.

## I resolved the conflict — it was ours, not yours

While CI was running, `mergeable` flipped to `CONFLICTING`. That was not your branch: `main` rewrote
`README.md` wholesale in #1053 (491 → 180 lines) after your merge-base, so your one-line version-badge
bump had nowhere to land. The `Resolve conflicts` web editor would have shown the whole file as a
conflict and made you rebuild 180 lines by hand, which isn't a reasonable thing to ask for a change we
caused.

So I pushed the resolution to your branch as `17e21fcf`. Specifically:

- **Merged `main` in rather than rebasing**, so `cd7b23d7` and `2629407d` keep their SHAs and remain
  ancestors of the head. Your history is untouched; there is one merge commit on top, authored by me.
- **Resolved `README.md` by taking `main`'s new version and bumping only the badge** to `2.8.1`. The
  file now differs from `main` by exactly that one line.
- **Nothing else touched.** The diff against `main` is still the same five files, `+189/-2`. `main`
  had not touched `CHANGELOG.md`, so your `## [2.8.1]` heading still sits correctly above `## [2.8.0]`,
  and `2.8.1` is still free.

Sorry for the churn on your branch — shout if you'd rather I had left it to you, and I'll ask first
next time.

## Verification

| check | result |
|---|---|
| CI (hosted, at `17e21fcf`) | **16/16 pass** — `Tests (smoke + unit)` 25m17s, `Tests (integration + e2e, deterministic)` 3m44s, `Contract checks (determinism + typecheck + lint)`, `Changelog completeness`, Gitleaks, Semgrep, Bun audit, Markdownlint |
| `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts` | 76 pass / 0 fail |
| both #1039 tests vs. the pre-fix adapter | fail with `unknown mutation-capable tool: read_file` and `a legacy write did not complete PostToolUse mediation` |
| `bun test tests/unit/t68-version-changelog-sync.test.ts` | 7 pass / 0 fail |
| typecheck / lint | exit 0 / exit 0 |
| version consistency | `aidlc-version.ts`, README badge, latest `CHANGELOG.md` heading all `2.8.1` |

The three `pull_request`-triggered workflows were parked at `action_required` because this is a
first-time contribution; I approved the runs on both heads, which is why the hosted signal exists.
Worth knowing for next time: the six `Pull Request Validation` checks run as `pull_request_target` and
never need that approval, which is why the PR looked half-checked.

## On the two non-blocking items

The comment qualifier is accurate — scoping the invariant to a well-formed payload and naming the
malformed-payload branch as fail-closed is exactly the distinction I had measured.

And the latch regression test pins the right behaviour rather than freezing an accident: the
`activeWriteWindows` branch gates on `mutationCapableTool()` and its reason says "before any legacy or
modern write", so safe reads skipping the latch while `fs_write`, `str_replace`, and `execute_bash`
stay at exit 2 is the intended contract.

## Why it still won't merge immediately

The `default-branch` ruleset also requires an approving review from the `aidlc-admins` team, and I'm
not a member — so my approval clears my earlier `CHANGES_REQUESTED` and satisfies the general
one-approval requirement, but not that one. A maintainer on that team still needs to approve, and
merges go through the merge queue with `ALLGREEN` grouping. Nothing further is needed from you.

## Non-blocking, ignore if you like

The `#975` row in your Related-issues table says "Recovery after a lost receipt is now possible."
True for the read-denial path this PR fixes, but it can read as though #975 as a whole is addressed —
something like "removes the read-denial blocker on recovery" would be tighter. Not worth a push on its
own.

Thanks again. The reproducible fixture in #1039 and the two regression tests are what made this quick
to verify on both sides.

wowzoo:
Reverted the release metadata at the maintainers' request — the version constant, README badge, and CHANGELOG entry are set at release time rather than per pull request. Pushed as `ba279d82`.

That one is on me. I asked for the bump in review because `CONTRIBUTING.md:56` and `AGENTS.md:54-60` still say a user-visible PR bumps all three in the same commit, and `core/tools/aidlc-version.ts:2-3` says the same in its own header comment. Those docs are out of date relative to how releases actually work now. Sorry for the round trip — you did exactly what was asked.

Your adapter fix and both regression tests are untouched. The diff against `main` is now just the two files that matter:

```
 harness/kiro-ide/hooks/aidlc-kiro-adapter.ts  |  20 ++++
 tests/unit/t218-kiro-ide-hook-adapter.test.ts | 160 ++++++++++++++++++++++++++
```

Re-verified after the revert: `t68` 7 pass / 0 fail (the three values agree at `main`'s `2.8.0` again), `t218` 76 / 0, both `#1039` regression tests still fail against the pre-fix adapter, `t239` 13 / 0, `bun scripts/package.ts --check` deterministic across two builds for all seven harnesses, typecheck and lint exit 0. I also ran `scripts/ci-changelog-guard.ts` against `main` — OK, 275 entries preserved, since `## [2.8.1]` was never on the base branch.

My approval stands. The remaining gate is an approving review from the `aidlc-admins` team; nothing further is needed from you.

wowzoo:
Merged — thank you for this, and for the write-up in #1039 that came with it. The diagnosis there was precise enough that it named the mechanism, the classification point, and the exit code, and the fix that ended up on `main` matches what you proposed.

A few notes on how it landed, because it did not land as a clean fast-forward and you deserve to know why.

**Your adapter change was overtaken while it waited.** #1097 added the same nine names to `PLAN_APPROVAL_SAFE_READ_TOOLS`, so by the time review finished, `read`, `read_file`, `read_files`, `read_code`, `list_directory`, `glob`, `grep`, `web_fetch` and `web_search` were already on `main`. I checked the behaviour rather than only the diff: on the commit before #1097 those reads exit 2 in an active unapproved Code Generation window, and on `main` they all exit 0 while `fs_write` still exits 2. That left your branch conflicting in exactly the two files it touched, through nothing you did — the review simply took longer than the parallel fix.

**The one thing still missing was the one you flagged yourself.** You pointed out on #1039 that the diff did not add `disclose_context`, so the deny would survive. That was right. It is in now, with the reasoning recorded beside it: Kiro documents `disclose_context` under Context tools next to `introspect` and `knowledge` and gives it no write surface, anything an activated skill then asks for is gated by its own `PreToolUse` call, and approval authority comes from the active directive and disk receipts rather than from activated context — so it cannot mutate the workspace during a Plan Approval window, while denying it stopped a Windows workflow mid-run.

**Your latch test was kept, not replaced.** The write-recovery regression you added in your second commit was rebuilt against `main`'s current helpers and kept all three of its deny assertions: `fs_write` and `str_replace` are still denied on the open latch, and `execute_bash` is still denied too. One detail worth recording — `execute_bash` asserts only the exit code, because a shell name routes into legacy recovery and fails closed with a different message (`recovery failed closed: No workflow state found`) rather than the mediation message the two write tools produce. Without the allowlist entry the case fails with `a legacy write did not complete PostToolUse mediation`, so it is a real regression test rather than a passing decoration.

**The release-metadata commit came back out.** Version constant, README badge and `CHANGELOG.md` are set at release time rather than per pull request — `CONTRIBUTING.md` and the Release Metadata Policy in `AGENTS.md` both say so. Nothing wrong with the reasoning in your commit message; the policy just moved that step elsewhere.

All four of your commits are on the merged branch and the squash is authored to you, with `Co-authored-by` for the assisted work.

One more thing your report earned. It split into two independent halves and both were real. The no-workflow half — path-less tools denied before any Intent exists — was fixed by #1000, which is why #986 could be closed. This half, the deny inside an active window, is the one that just merged. That is a genuinely good bug report, and the second one is the harder half to see.

PR #1043, opened 2026-09-08
fix: make parallel test execution deterministic

apackeer:
## Summary

Make the deterministic test suite reliable under `--parallel 4`, and fix two core issues exposed by the stress runs. This PR is rebased onto current `main` and updates AI-DLC from 2.8.1 to 2.8.2.

No assertion was weakened and no retry was added. Timing-sensitive tests either receive a realistic Bun test/hook budget or use the runner's `.serial.` mechanism when wall-clock timing is the behavior under test.

### Core

* `source-manifest.json` validation now resolves HEAD once, reads the Git tree once, and batches ignore-rule checks per repository instead of spawning three Git processes per claimed path. Cache misses retain the previous per-path fallback and accept/refuse semantics.
* SessionStart writes the current process ancestry before stale-entry cleanup, resolves each ancestor identity once, skips entries written by the current walk, reaps dead processes without `ps`, and keeps live entries whose identity cannot be verified within the 50 ms budget.
* Added deterministic SessionStart regressions and updated the 2.8.2 version, badge, and changelog.

### Test reliability

* Detached the t298 background holder's stdio so `spawnSync` returns while the holder is alive and `$!` is the only stdout line.
* Added file-level Bun timeout budgets for spawn-heavy real-Git cases and cleanup hooks in t255, t314, t325, t326, t327, and t52.
* Increased t326's two-cycle full-round-trip budget to reflect its measured standalone runtime.
* Marked t311 and t188 serial because their assertions intentionally depend on bounded wall-clock/process-lock timing.
* Updated coverage-registry and documentation references for renamed serial tests.

## Test plan

* `bun scripts/package.ts`
* `bun scripts/package.ts --check`
* `bun run typecheck`
* Session/version/coverage unit checks: 67 passed
* t188/t311/t52 serial integration checks: 109 passed
* t314 Windows command-line batching case: passed in 23.18 s
* Existing CI will rerun the repository's complete required matrix on the rebased branch.

## Checklist

* [x] Changes are documented

## Acknowledgment


spraja08 (dismissed):
APPROVED TO MERGE with this advisory:

- Confirm in testing (t325/t327 logs) that readUnitSourceManifest() receives the same manifest across all three validation calls (review/verdict/finalize). If it receives different manifests, the caching strategy needs a note in the CHANGELOG about behavior change (more spawns for subset manifests).

- Monitor release notes to users about the dramatic speedup in unit source-manifest validation (minutes → ~23 seconds). This is good, but early adopters with large manifests should be aware.

- Consider adding a code comment in ignoredSourceClaimReason() explaining why the cache is created at the readUnitSourceManifest() call level (per-validation-session), not globally (would violate freshness across independent manifest checks).

No bug fixes or rework required. All three bugs were real, deterministically reproducible, and correctly fixed at the source.

Issue #1044, opened 2026-09-08
[Bug]: kiro-ide plan-approval-guard not working on Windows


Yukinobu-Mine:
### Description

In [`plan-approval-guard` of `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts`](https://github.com/awslabs/aidlc-workflows/blob/main/harness/kiro-ide/hooks/aidlc-kiro-adapter.ts#L1238-L1603), shell execution is identified by tool name 'execute_bash'.
But the actual tool name is 'execute_pwsh' on Windows environment (PowerShell).
So, `plan-approval-guard` blocks execution of PowerShell.

Self analysis by Opus 4.8 is as follows.

---

Issue Occurring

The AI-DLC workflow is essentially a loop of "executing `bun .kiro/tools/aidlc-orchestrate.ts next ...` in the shell and proceeding according to the returned directive." However, all shell executions needed to drive this workflow are being blocked.

Upon investigation, the `plan-approval-guard` (a safety hook that enforces approval before code generation) in `aidlc-kiro-adapter.ts` has a Windows-specific incompatibility.

This guard treats only `execute_bash` (assumed for macOS/Linux) as a special case for shell execution tools.
However, the tool name sent by Kiro on Windows is `execute_pwsh`.
As a result, `execute_pwsh` is judged as a "modification operation whose target path cannot be identified" and is **always blocked** regardless of whether code generation is in progress (relevant location: around line 1505 of the adapter).
Consequently, even the `next` command that serves as the starting point of the workflow cannot be executed, and the AI-DLC loop cannot be started. File editing tools (`fs_write`/`str_replace`) work, but since the CLI engine responsible for decision logging, state management, and auditing can only be run via the shell, it is by design impossible to bypass the engine and proceed manually (reproducing routing in prose is prohibited).

Proposed Fix

Modify the `plan-approval-guard` in `aidlc-kiro-adapter.ts` so that `execute_pwsh` and `shell` are treated the same as `execute_bash`. This is a legitimate bug fix that **preserves** the guard's safety purpose (stopping the shell / placing it on a recovery path before plan approval for code generation) while ensuring the same behavior in PowerShell environments. It does not weaken the security boundary.

However, since this is a change to a security-related file of the framework, I would like your confirmation rather than proceeding on my own.

[...]

Issue #987, opened 2026-08-31
code-generation: the Plan Approval fingerprint is unsatisfiable after any compliant execute+review cycle


Apethor:
## Summary

In `code-generation`, `fingerprintValid` can never be `true` once the stage has been executed and reviewed as the stage file instructs. The stage fingerprints `code-generation-plan.md`, and then mandates two separate mutations of that same file. Any conductor that follows the instructions ends the stage with an invalidated approval, and an adversarial reviewer correctly reports it as a finding — every unit, every time.

This is not a conductor-discipline problem. It is three instructions in the shipped stage/protocol that cannot all hold.

## The three instructions

`.claude/aidlc-common/stages/construction/code-generation.md`:

- **Step 2** — the plan is authored "with checkboxes for each implementation step".
- **Step 3** — `aidlc-testing-posture.ts fingerprint --unit <u>` hashes the plan, and: *"Any post-approval change to the plan, unit test instructions, `## Testing Posture` memory, scope, test strategy, or project type invalidates the fingerprint and reopens Plan Approval."*
- **Step 4** — the delegation prompt must include *"Instructions to execute each plan step sequentially and mark checkboxes as completed"*.

`.claude/aidlc-common/protocols/stage-protocol-reviewer.md` §12a step 2:

- the reviewer *"Appends exactly ONE `## Review` section to the primary artifact file"* — and for `code-generation` the primary artifact is `code-generation-plan.md`, the same file Step 3 fingerprints. Step 1 further requires deleting the prior `## Review` section before every dispatch, so the file changes on each review iteration too.

`approvalFingerprint(plan, instructions, contractHash)` in `aidlc-testing-posture.ts` hashes the plan's full text, so both mutations invalidate it.

## Measured, on a real unit

Unit `u2-plugin-skeleton`, scope `research-harness`, unit-major Construction iteration.

| Plan variant | `approvalFingerprint` |
|---|---|
| Recorded at Plan Approval | `sha256:a49e90fe45ab9179661f3a5a089419bc134a44886e78d7dc56348d5a71625439` |
| Current (executed + reviewed) | `sha256:c12344e30f1fd20f44859e1869fc903e5f8ac664f0dc0f22026eaad05950e9dc` |
| Current, `## Review` section removed | `sha256:f8c41c0e7c5ee6f45ff932c407dd02688189b09c39cd95348c3cbf7fafca568a` |
[...]

Issue #981, opened 2026-08-31
[Bug]: Code Generation pre-planning source floor can become permanently unsatisfiable


fsatsuki:
### Description

Code Generation binds a **pre-planning source floor** (`code_generation_source_sha256` in
`.aidlc-active-directive.json`) and refuses to record Plan Approval unless the live workspace
source still matches it. The guard itself is reasonable — it proves no source was written between
planning and approval — but there is **no path that re-baselines the floor without an existing
Plan Approval receipt**. A workspace that legitimately changes during the planning window
therefore locks Code Generation permanently.

Two properties combine into a deadlock:

1. `codeGenerationSourceFloorForPublication` (`core/tools/aidlc-lib.ts:4893`) keeps the previous
   floor for as long as the previous directive's stage is already `code-generation`:

   ```ts
   if (!rotate && base.stage === "code-generation" &&
       base.code_generation_source_sha256 !== undefined) {
     return base.code_generation_source_sha256;   // sticky
   }
   return workspaceSourceFingerprint(...);         // fresh capture
   ```

2. `rotate` comes from `rotateCodeGenerationFloor` (same file, around line 5127), which is true
   only when `planApprovalRuntimeHasReceiptForMarker` finds a receipt with
   `status === "generation"`. That receipt is exactly what cannot be created while the floor check
   fails.

Rotating the floor therefore requires an approval receipt, and obtaining an approval receipt
requires a matching floor. `--resume`, a same-stage redo jump, and regenerating the plan all keep
`base.stage === "code-generation"` and so preserve the stale floor.

The condition is easy to hit because the floor is a **filesystem** identity, not a Git one
(`workspaceSourceState` → `filesystemSourceIdentity`, made Git-independent in #904). Any addition,
removal, or modification anywhere in the project tree outside the record directory drifts it — an
unrelated `git pull`, a plugin projection writing to the project root, a build output, or even
creating a temporary directory while trying to recover.

Recovery from inside the session is also blocked: while Plan Approval is pending, the
plan-approval guard refuses mutation-capable shell commands and workspace writes, so the workspace
cannot be restored toward the floor from the same session.

Observed on a zero-Unit (`--stage-level`) run under the `express` scope. The floor helper is
stage-scoped rather than unit-scoped, so per-unit runs appear to share the same code path.

### Steps to Reproduce

1. Install AI-DLC Workflows 2.6.124 into a project and start a workflow whose scope reaches Code
[...]

PR #1051, opened 2026-09-08
test(kiro-ide): pin the legacy execute_pwsh populated-command approval flip

fsatsuki:
# Summary

Narrowed to a **test-only** change, per review.

`#1000` landed on `main` and already carries everything this PR originally implemented: the Kiro IDE shell-alias predicate (`isKiroShellTool()` covering `execute_bash` / `execute_pwsh` / `shell`), the Kiro CLI `Bash` canonicalization, and the CLI `state-transition-guard` change. `#1044` is closed. Both adapter edits and the `t147` `execute_pwsh` case here are therefore redundant and have been dropped.

What is **not** yet pinned on `main` is the legacy `{ toolName, toolArgs }` channel with a *populated* command. Every legacy assertion in `t218` uses `execute_bash` with empty arguments, and an empty-argument payload is decided by the opaque-mutation branch both before and after approval — it never reaches the code that resolves a shell alias. A Windows host names the same tool `execute_pwsh` and does supply `toolArgs.command`. This PR keeps that one regression test.

## Changes

* `tests/unit/t218-kiro-ide-hook-adapter.test.ts`: add `legacy execute_pwsh with a populated command flips from blocked to permitted across approval` (+104 lines, no other edits). It walks the same legacy mediation flow as the neighbouring `execute_bash` test and asserts the approval-boundary transition:
  * before approval — exit 2, and *not* the opaque `target path is missing or unsupported` refusal, proving it took the recovery path;
  * after approval — exit 0, so the AI-DLC loop can advance.
* Merged current `main` into the branch and resolved both adapter files and `t147` back to `main`'s content. The merge tree is identical to `main`; the PR diff is the single test file.
* No implementation change, no `dist` change, no version/CHANGELOG/README bump — release preparation owns version and changelog consolidation.

## User experience

Unchanged. This PR adds coverage only; the behaviour it pins already ships on `main` via `#1000`.

## Checklist

* [ ] Changes are documented

## Test plan

* `bash tests/run-tests.sh --unit --filter "t218"` → **87 pass / 0 fail**, including the new case.
* Fail-first evidence: narrow `isKiroShellTool()` in `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts` to `return toolName === "execute_bash";`, run `bun scripts/package.ts`, then `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts -t "legacy execute_pwsh with a populated command"`. The post-approval assertion fails with `Expected: 0 / Received: 2`. Restoring the predicate makes it pass again. The same test with empty `toolArgs` passes either way, which is exactly the gap flagged in the earlier review.
* `bunx biome check tests/unit/t218-kiro-ide-hook-adapter.test.ts` → clean.
* `bun test tests/unit/gen-coverage-registry.test.ts` → 37 pass / 0 fail (committed coverage registry stays fresh).

## Acknowledgment

### DIFF
--- a/harness/kiro-ide/hooks/aidlc-kiro-adapter.ts
+++ b/harness/kiro-ide/hooks/aidlc-kiro-adapter.ts
@@ -1082,6 +1082,12 @@ function mutationCapableTool(name: string): boolean {
   return name.length > 0 && !PLAN_APPROVAL_SAFE_READ_TOOLS.has(name);
 }
 
+const SHELL_TOOLS = new Set(["execute_bash", "execute_pwsh", "shell"]);
+
+function shellTool(name: string): boolean {
+  return SHELL_TOOLS.has(name);
+}
+
 function inputPaths(input: Record<string, unknown>): string[] {
   const paths: string[] = [];
   const add = (value: unknown) => {
@@ -1262,7 +1268,7 @@ function buildForward(): Forward {
         } catch {
           // Missing host identity remains fail closed below.
         }
-        if (toolName === "execute_bash") {
+        if (shellTool(toolName)) {
           const recovery = runLegacyRecoveryNext(
             projectDir,
             recoverySession,
@@ -1287,7 +1293,7 @@ function buildForward(): Forward {
           (
             Object.keys(toolArgs).length === 0 ||
             (
-              toolName !== "execute_bash" &&
+              !shellTool(toolName) &&
               paths.length === 0
             )
           )
@@ -1300,7 +1306,7 @@ function buildForward(): Forward {
           (!state.active || state.target === null) &&
           writeWindows.length > 0
         ) {
-          if (toolName === "execute_bash") {
+          if (shellTool(toolName)) {
             const recovery = runLegacyRecoveryNext(projectDir, approvalSession);
             return {
               hook: "__legacy_plan_approval_block__",
@@ -1334,7 +1340,7 @@ function buildForward(): Forward {
           }
         }
         if (interruptedWrite && !state.approved) {
-          if (toolName === "execute_bash") {
+          if (shellTool(toolName)) {
             const recovery = runLegacyRecoveryNext(projectDir, approvalSession);
             return {
               hook: "__legacy_plan_approval_block__",
@@ -1359,7 +1365,7 @@ function buildForward(): Forward {
               ?.trim()
               .toLowerCase()
               .replace(/\s+/g, "-") === "code-generation";
-          if (durableCodeGeneration && toolName === "execute_bash") {
+          if (durableCodeGeneration && shellTool(toolName)) {
             const recovery = runLegacyRecoveryNext(projectDir, approvalSession);
             return {
               hook: "__legacy_plan_approval_block__",
@@ -1379,7 +1385,7 @@ function buildForward(): Forward {
           }
         }
         if (state.active && state.violated) {
-          if (toolName === "execute_bash") {
+          if (shellTool(toolName)) {
             const recovery = runLegacyRecoveryNext(projectDir, approvalSession);
             return {
               hook: "__legacy_plan_approval_block__",
@@ -1444,7 +1450,7 @@ function buildForward(): Forward {
           !state.approved &&
           (
             toolName === "" ||
-            toolName === "execute_bash" ||
+            shellTool(toolName) ||
             toolName === "fs_append"
           )
         ) {
@@ -1502,7 +1508,7 @@ function buildForward(): Forward {
               };
             }
           }
-          if (Object.keys(toolArgs).length > 0 && toolName !== "execute_bash") {
+          if (Object.keys(toolArgs).length > 0 && !shellTool(toolName)) {
             return {
               hook: "__legacy_plan_approval_block__",
               input: {
@@ -1533,7 +1539,7 @@ function buildForward(): Forward {
           },
         };
       }
-      if (toolName === "execute_bash") {
+      if (shellTool(toolName)) {
         return {
           hook: "aidlc-plan-approval-guard.ts",
           input: {
--- a/harness/kiro/hooks/aidlc-kiro-adapter.ts
+++ b/harness/kiro/hooks/aidlc-kiro-adapter.ts
@@ -590,7 +590,9 @@ if (target === "guard-tool-call") {
 // --- state-transition-guard: engine ownership of lifecycle mutations -------
 if (target === "state-transition-guard") {
   const tool = kiro.tool_name ?? "";
-  if (tool !== "shell" && tool !== "execute_bash") process.exit(0);
+  if (tool !== "shell" && tool !== "execute_bash" && tool !== "execute_pwsh") {
+    process.exit(0);
+  }
   const command = String(kiro.tool_input?.command ?? "");
   const registeredAgent = extraArgs[0] ?? "";
   const executable = process.env.AIDLC_COMPILED_EXECUTABLE;
@@ -870,7 +872,9 @@ function canonicalTool(
   }
   if (name === "str_replace" || name === "fs_append") return "Edit";
   if (["read", "fs_read", "read_file", "read_files"].includes(name)) return "Read";
-  if (name === "shell" || name === "execute_bash") return "Bash";
+  if (name === "shell" || name === "execute_bash" || name === "execute_pwsh") {
+    return "Bash";
+  }
   return name;
 }
 
--- a/tests/unit/t147-kiro-hook-adapter.test.ts
+++ b/tests/unit/t147-kiro-hook-adapter.test.ts
@@ -332,10 +332,16 @@ describe("t147 Kiro hook adapter (live-captured payload fixtures)", () => {
           tool_name: "execute_bash",
           tool_input: { command: "sort input.txt -o src/blocked.txt" },
         },
+        {
+          hook_event_name: "preToolUse",
+          cwd: dir,
+          tool_name: "execute_pwsh",
+          tool_input: { command: "Sort-Content input.txt > src/blocked.txt" },
+        },
       ]) {
         const r = runAdapter(dir, "plan-approval-guard", payload);
-        expect(r.code).toBe(2);
-        expect(r.stderr).toContain("Code generation cannot");
+        expect(r.code, payload.tool_name).toBe(2);
+        expect(r.stderr, payload.tool_name).toContain("Code generation cannot");
       }
     } finally {
       rmSync(dir, { recursive: true, force: true });
--- a/tests/unit/t218-kiro-ide-hook-adapter.test.ts
+++ b/tests/unit/t218-kiro-ide-hook-adapter.test.ts
@@ -1779,11 +1779,16 @@ describe("t218 Kiro IDE plan-approval enforcement", () => {
 
       // Before approval the Windows shell tool is blocked exactly like
       // execute_bash: exit 2, and NOT with the opaque "target path" prose — it
-      // takes the recovery-shell path.
+      // takes the recovery-shell path. A { command } payload (not empty args)
+      // reaches the branches shellTool() actually changes; empty args short-
+      // circuit through opaqueMutation before that code runs.
       const preApproval = runIde(
         dir,
         "plan-approval-guard",
-        JSON.stringify({ toolName: "execute_pwsh", toolArgs: {} }),
+        JSON.stringify({
+          toolName: "execute_pwsh",
+          toolArgs: { command: "bun .kiro/tools/aidlc-orchestrate.ts next" },
+        }),
       );
       expect(preApproval.code).toBe(2);
       expect(preApproval.stderr).not.toContain(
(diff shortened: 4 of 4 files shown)


wowzoo:
Thanks — this is a clean, well-scoped fix, and I checked completeness rather than assuming it: all nine
`toolName === "execute_bash"` comparisons in the kiro-ide plan-approval path go through `shellTool()`
(`:1271`, `:1296`, `:1309`, `:1343`, `:1368`, `:1388`, `:1453`, `:1511`, `:1542`), with no literal
comparison left in that path. Both CLI sites are covered (`:593` `state-transition-guard`, `:875`
`canonicalTool`), the core guard correctly needs no change because it matches on the canonical `Bash`,
and no other harness references `execute_bash`. CI is 16/16 at this head.

I'm leaving this as a comment rather than a blocking review, because there's a triage question that
isn't yours to resolve — see the last section. Everything below is measurement you can use either way.

## What #1044 actually was — the direction differs per harness

Your summary says `execute_pwsh` calls "slipped past the code-generation plan-approval gate and the
lifecycle state-transition guard". That's accurate for the CLI and backwards for the IDE:

- **kiro CLI — both guards genuinely bypassed.** `plan-approval-guard` let `execute_pwsh` through
  (`t147:343` in `1bb`: expected 2, **received 0**), and `state-transition-guard` skipped it entirely
  via the early `process.exit(0)` at `:593`. This is the security-relevant half, and it's worth stating
  as two separate bypasses rather than one.
- **kiro IDE — the opposite.** `execute_pwsh { command }` was *blocked* (exit 2, "target path is
  missing or unsupported") even in a fully approved window, where `execute_bash { command }` exits 0.
  Nothing slipped past; the AI-DLC loop simply couldn't advance. A liveness failure, not a bypass.

And `shell` was broken on the IDE too. Your summary says the adapters "classified only `execute_bash`
and `shell` as shell tools", but on the kiro-ide plan-approval path only `execute_bash` was
special-cased — I measure `shell { command }` at **exit 2 pre-fix** in the approved window, going to 0
after. So this change also repairs `shell` there, which the description doesn't claim credit for. (The
terminal-command guard at `:923` did already list all three, which is probably where that wording came
from.) Since #1044 came from a Windows user, being precise about which half was a bypass matters to
anyone assessing exposure.

## The new t218 case passes against the unfixed adapter

Worth knowing regardless of what happens to this PR. I ran both suites against the pre-fix adapters to
check the regressions fail first. `t147` does its job — **34 pass / 1 fail** at `t147:343`. But `t218`
returns **75 pass / 0 fail both pre- and post-fix**, so the kiro-ide half has no failing-first coverage.

The cause is the payload shape: both assertions use `toolArgs: {}`, and with empty args
`opaqueMutation` is already satisfied by its first clause (`Object.keys(toolArgs).length === 0`), so
control never reaches the branches `shellTool()` changes.

- Pre-approval, `{}`: pre-fix falls to the `:1484` "not safely attributable" deny, post-fix takes the
  `:1453` shell deny. Both exit 2, and neither message contains "target path is missing or
  unsupported", so `expect(preApproval.stderr).not.toContain(…)` holds either way.
- Post-approval, `{}`: `:1511` is gated on `Object.keys(toolArgs).length > 0`, which is false, so both
  revisions fall through to `return null` and exit 0.

A realistic payload is also the case that actually regressed:

```diff
-        JSON.stringify({ toolName: "execute_pwsh", toolArgs: {} }),
+        JSON.stringify({
+          toolName: "execute_pwsh",
+          toolArgs: { command: "bun .kiro/tools/aidlc-orchestrate.ts next" },
+        }),
```

With `{ command }` in the approved window I measure **exit 2 pre-fix, exit 0 post-fix**, and the same
for `shell { command }`.

## The triage question — #1000 also claims this issue

**#1000 declares `Closes #1044`** as well, alongside #930, #975, #977, #981, #987, and #995. It isn't a
stale claim: at head `ac6a86f6` it carries an `isKiroShellTool()` helper (`:184-185`) covering all three
names, applies it at the plan-approval sites *and* at `:962` — which also collapses the duplicate
triple still sitting at `:923` here — includes `execute_pwsh` in the CLI `canonicalTool` (`:875`), and
its `:1556` reads `state.active && Object.keys(toolArgs).length > 0 && !isKiroShellTool(toolName)`.

Its coverage is also already in the shape I described above: `tool_input: { command: … }` payloads and a
`SHELL_NAMES = ["execute_bash", "execute_pwsh", "shell"]` table (`t218:1559`), plus `t147:345`
`1bc: execute_pwsh normalises to Bash in the plan-approval-guard path like execute_bash`. So it does not
have the gap this PR's test has.

That makes it a real decision rather than a formality. #1000 is 172 files with seven rounds of
`CHANGES_REQUESTED` behind it and two reviewers still active, so it is moving but not imminent; this PR
is four files and green today. If the maintainers want the small fix in first, the test payload above is
the one thing I'd change and #1000 rebases over it easily. If they'd rather #1000 own it, nothing here is
wasted — the direction correction above should go into #1000's description, since it inherits the same
wording problem.

I'm deliberately not gating this either way. Flagging it now so you don't polish something that might be
superseded.

## Non-blocking — the duplication that caused #1044 is still here

`SHELL_TOOLS` is now the set for the plan-approval path, but `:923` keeps its own
`tool !== "execute_bash" && tool !== "execute_pwsh" && tool !== "shell"` triple for the terminal-command
guard, and the CLI adapter keeps two ad-hoc literals (`:593`, `:875`) rather than a shared set. No
correctness hole today — all three already list all three names — but it's three independent spellings
of "is this a shell tool" across two files, which is the shape that let `execute_pwsh` go missing.
#1000 solves this by construction at `:962`.

One more thing for whoever sequences these: this PR and **#986 edit the same non-empty-args line**, so
they conflict, and #986's `state.active &&` alone is insufficient — I applied it to the base adapter and
approved-window `execute_pwsh { command }` still exits 2. The correct condition needs both that and the
shell test, which is what #1000's `:1556` already does. **#1051 + #1040 is clean.**

## Validation I ran at `d401ec96`

| check | result |
|---|---|
| `bun test tests/unit/t147-kiro-hook-adapter.test.ts` | 35 pass / 0 fail |
| `bun test tests/unit/t218-kiro-ide-hook-adapter.test.ts` | 75 pass / 0 fail |
| **the same two suites against the pre-fix adapters** | t147 **34 / 1** ✅ · t218 **75 / 0** ❌ |
| approved-window probe, `{ command }` payloads | pre-fix `execute_pwsh` 2 / `shell` 2 / `execute_bash` 0 → post-fix all 0 |
| `bun run typecheck` / `bun run lint` | exit 0 / exit 0 |
| hosted CI at this head | 16/16 SUCCESS |

Nice piece of work either way — the `SHELL_TOOLS` shape is the right one, and it's what #1000 arrived at
independently.

leandrodamascena (changes requested):
Thanks for addressing the Windows shell-tool handling.

I reviewed the current head `d401ec967aa81643d8a6d1f83d70788958b6c721` against base `0d399dd828b59e84d90f7cc198c69fab9ad8f1a7`. The change is aligned with #1044: Kiro CLI needs to prevent `execute_pwsh` from bypassing its guards, while Kiro IDE needs to permit recognized shell commands after plan approval.

**[P2] The new Kiro IDE regression test does not exercise the fixed path**

In `tests/unit/t218-kiro-ide-hook-adapter.test.ts:1786` and `:1845`, the `execute_pwsh` payload uses `toolArgs: {}`. With no command, both assertions pass against the adapter before this PR because unrelated empty-input handling produces the expected exit codes. Therefore, the test does not fail when the implementation fix is removed.

Please provide a realistic shell payload in both calls, for example:

```ts
toolArgs: {
  command: "bun .kiro/tools/aidlc-orchestrate.ts next"
}
```

This exposes the relevant behavioral transition: before the fix, an approved `execute_pwsh` call remains blocked with exit code 2; after the fix, it is permitted with exit code 0.

Please also update the PR description to distinguish the two behaviors:

- Kiro CLI had a guard-bypass/security issue.
- Kiro IDE had a liveness issue where `execute_pwsh` and `shell` remained blocked after approval.

The focused `t147` coverage is valid, the implementation direction is sound, and all 16 current CI checks are green. No version bump is needed because release preparation owns version and changelog consolidation.

fsatsuki:
@leandrodamascena @wowzoo 
Thanks for the detailed measurement — especially confirming the fail-first gap.

Fixed the t218 test in d44d2ba5: both the pre- and post-approval assertions now send a realistic `{ command: "bun .kiro/tools/aidlc-orchestrate.ts next" }` payload instead of `toolArgs: {}`, so control reaches the branches `shellTool()` actually changes instead of short-circuiting through the empty-args `opaqueMutation` clause.

I verified fail-first the way you described, using a `git worktree` checked out at `origin/main` with only the updated test files copied in:

- t218 `treats execute_pwsh identically to execute_bash (#1044)` — fail against the unfixed adapter, pass against this branch.
- t147 `1c` — same (fail → pass).

On the direction correction: you're right that the summary is imprecise. It's a genuine two-way bypass on the CLI (plan-approval-guard let it through, and state-transition-guard skipped it at the early exit) but a liveness failure on the IDE (an approved `execute_pwsh { command }` was blocked), and this change also repairs `shell` on the IDE path. I'll correct the PR description to say exactly that rather than the one-line "slipped past both guards" wording.

On the #1000 overlap and the remaining three-spellings duplication: understood and happy to defer to the maintainers on which PR owns #1044. If they prefer the small fix first, this is ready; if #1000 lands first, the direction note above should move into its description.
 
I also updated the description of PR.

apackeer (changes requested):
Thank you for updating the regression test after review. #1000 has now merged and includes the Kiro IDE shell-alias handling, CLI Bash canonicalization, and CLI state-transition guard change covered here. I checked those changes against current main, and #1044 is already closed.

There is still a useful test to preserve: your t218 case sends a populated `toolArgs.command` through legacy `execute_pwsh` before and after human approval. Main's corresponding approved legacy shell assertion still uses `execute_bash` with empty arguments, so it does not pin that exact regression.

Could you narrow this to a test-only PR against current main: drop the two adapter changes and the redundant t147 addition, retain/adapt the t218 populated-command regression, update the title and description to that scope, and rerun the focused t218 suite? Then we can review the remaining test contribution.

fsatsuki:
@apackeer 
Thank you for fix this issue.

OK, I will do it.
> Could you narrow this to a test-only PR against current main: drop the two adapter changes and the redundant t147 addition, retain/adapt the t218 populated-command regression, update the title and description to that scope, and rerun the focused t218 suite? Then we can review the remaining test contribution.

fsatsuki:
@apackeer
I left only the t218 test case and updated the PR description.

wowzoo:
Thank you for narrowing this to the single case, and I am sorry the thread went quiet for five days
afterwards — you did what the review asked and then had to wait.

Four things from re-measuring at `5134a9d50`.

**The red check is not a branch-test failure.** The only failing check is `Check Merge Status`; every other
check passes, including all four unit shards. That job does not inspect or execute this diff — it can go red
on repository-wide merge controls, on an unavailable PR number, or on a failure while retrieving or parsing
the open-PR list — so the result alone is not evidence of a defect here.

**One correction to the description.** It says every legacy assertion in `t218` uses `execute_bash` with
empty arguments. Two are already there: the `terminal-command-guard` case drives the legacy
`{ toolName, toolArgs }` channel with a populated command and asserts exit 2, and the test
`execute_pwsh and shell are routed to legacy recovery exactly like execute_bash` already covers
`execute_pwsh` on `plan-approval-guard` through the legacy channel. The gap is narrower than stated and
still real: before this PR, `plan-approval-guard` has no legacy **populated-command** assertion across the
approval boundary — `execute_pwsh` is covered there for empty-argument recovery, and this case adds the
populated-command transition. Saying it that way keeps the test's reason legible to the next reader.

**One request on the assertion.** The case turns on
`expect(preApproval.stderr).not.toContain("target path is missing or unsupported")`. The preceding
`fs_write` leaves an active legacy write window, and from inside that window both the recognized-shell
recovery branch and the non-shell fallback branch return the legacy plan-approval block: both exit 2, and
neither emits the target-path wording, which belongs to the opaque-mutation branch further down. So the
negative assertion does not show that `execute_pwsh` was recognized as a shell.

`legacyRecoveryBlockReason` is the branch worth pinning, and the neighbouring
`execute_pwsh and shell are routed to legacy recovery exactly like execute_bash` test already pins it
positively with `toContain("recovery requires a human response")` while cross-checking all three shell
names against each other. `isKiroShellTool` matches on the tool name alone, so a populated command reaches
that same recognition branch — following the neighbouring shape here would make the case prove what it is
for, without needing a new message.

**One cross-PR note, with no action requested here.** #1157 consolidates `kiro-ide` into `kiro`: it renames
this test and repoints its scratch tree to `dist/kiro/.kiro`. The two therefore overlap on this file and
will need reconciling in whichever order they land, but the renamed test itself points at the tree #1157
produces. I am raising that on #1157.

leandrodamascena (commented):
Approve

leandrodamascena (approved):

PR #1052, opened 2026-09-08
feat: Commit Provenance

rafaelgsr:
# Summary

Adds a durable reverse lookup from any git commit or diff range to the reviewed units of work that own its changed paths. Attribution is computed from committed content alone — the commit's tree plus the intent record the repository already carries — so any clone answers identically: no hooks need to have fired, no commit trailers are parsed, and no session or machine state is read. The same `(base, head)` pair resolves the same way on a colleague's laptop, in a fresh checkout, or years later.

That makes "does this change trace to a reviewed unit, and does it still match what was reviewed?" a question anyone can ask, however they want to ask it: a reviewer at a terminal before approving a diff, a merge check, a periodic audit over the last release, a compliance report, a bot, or a script. The tool answers in JSON with an exit code and takes no opinion on the caller.

Three constraints ruled out the familiar mechanisms. Most commits are manual `git commit` runs from arbitrary machines, often long after the authoring session, so anything session-observed can only be enrichment. A receipt's `Unit Source Fingerprint` proves *what* was reviewed but cannot name the unit starting from a changed file. Commit message formats belong to the team and are never relied on.

## Changes

**Committed review evidence.** Per-unit source review now dual-writes the reviewed-source snapshot into the committed intent record at `construction/<unit>/<stage>/reviewed-source-<hash12>.tsv`, alongside the existing machine-local copy. The receipt's `Unit Source Fingerprint` is the SHA-256 of exactly those bytes — no new fingerprint semantics — and rewrites are content-addressed: idempotent on match, refusing on collision. Units finalized in a swarm worktree carry that evidence home with the manifest, hash-checked on the way out.

**`aidlc attest resolve`** (read-only) attributes every changed path in a commit or range to its owning reviewed unit and classifies it `verified`, `drifted`, `unattested`, `unverifiable`, `indeterminate`, or `excluded`, emitting JSON with per-path and per-unit detail. `--fail-on <statuses>` turns any subset of the four failable statuses into exit code 3, so a caller can act on the answer without parsing the report.

**`aidlc attest anchor`** appends `SOURCE_COMMITTED` audit events for commits that landed reviewed claims: a forward pointer in the audit trail, enrichment only — `resolve` never reads them, so a team that never anchors loses nothing but audit-trail readability. Deduplicated per intent on (commit, repo); `--reconcile` walks first-parent history bounded by `--max-commits`. It also runs automatically on every real session start (last 25 first-parent commits), because humans commit mostly *between* sessions; best-effort, idempotent, and disabled by `AIDLC_SKIP_SESSION_ANCHOR=1`.

**Semantics.** Ownership is the newest READY `REVIEW_COMPLETED` receipt per unit, mirroring the completion-time rule that a newer reviewed claim can own an intentional shared-file integration. Attribution keys on blob content rather than commit ancestry, so squashes and rebases are stable and re-landing a reverted file at its reviewed bytes returns it to `verified`.

**Fail-closed throughout.** Verification requires bytes that hash to the receipt fingerprint, so tampering can only fail, never falsely verify. Missing or mismatched evidence is `unverifiable`; so is evidence found only in the gitignored machine-local snapshot, because a verdict no other clone can reproduce is not a verdict. Same-timestamp receipts that cannot be ordered — in different shards, or in different records claiming one path — are `indeterminate`. A shallow-clone boundary is refused with an actionable error rather than diffed against the empty tree, which would classify every file in the repository; `--reconcile` reports such commits in `boundaries` and continues. Each verb rejects the other's flags, so a caller cannot believe it enabled a check it did not.

**Honest about what it cannot see.** `recordSource` and `warnings` flag repository byte-form conversion (`core.autocrlf`, `.gitattributes`, LFS) and an intent record whose working-tree state differs from the queried commit. Both underlying gaps are documented limits with roadmap entries rather than silent behavior.

The audit taxonomy grows to 92 events (new Commit Provenance category). New reference chapter `docs/reference/19-commit-provenance.md`; CLI guide, state-and-audit guide, hooks-and-tools, audit-format registry, overview, nav, and roadmap updated. The branch also merges `origin/main` (2.7.0/2.7.1), with the version at 2.7.3 and count pins re-baselined to the merged taxonomy.

## User experience

**Before.** A reviewed unit's per-path evidence existed only in a gitignored per-machine file, so nothing outside the authoring checkout could connect a landed commit to the review that approved it. Anyone holding a commit range — a reviewer, an auditor, a script — had no way to ask whether those changes were reviewed; the available answers were manual inspection of the record or trusting the process. Commit provenance was an open design question in the roadmap.

**After.** From any clone:

```bash
# a range: what landed between two refs?
aidlc attest resolve --diff origin/main...HEAD

# one commit's first-parent delta
aidlc attest resolve <commit>

# make it a hard check, wherever you run it
aidlc attest resolve --diff origin/main...HEAD \
  --fail-on drifted,unattested,unverifiable,indeterminate
```

The report names, per path, which unit and intent own it and whether the landed bytes match the reviewed bytes; per unit, whether its whole claim landed. With `--fail-on` it exits 3 when any named status appears, so the same command works as an interactive question, a pre-merge check, a `git bisect`-style sweep over history, or a scheduled audit. Audit trails also gain human-readable `SOURCE_COMMITTED` rows tying commits to the units they landed, with no command for anyone to remember. Nothing changes for users who never run the new verb, except that per-unit review writes one additional small file into the intent record.

Two behaviors to know about when upgrading: records reviewed before this version have no committed evidence and honestly report `unverifiable` until their next per-unit review writes it, and a shallow clone must be deepened before it can be resolved — the tool refuses the boundary instead of guessing.

## Checklist

* [x] Changes are documented

## Test plan

Automated:

```bash
bun scripts/package.ts --check     # dist byte-parity with core/ + harness/
bun run lint
npx tsc --noEmit -p tsconfig.json
bun test tests/unit/t311-committed-reviewed-source-evidence.test.ts
bun test tests/unit/t312-attest-resolve-anchor.test.ts
bash tests/run-tests.sh --ci       # smoke + unit + integration
```

`t311` pins the evidence grammar, dual-write byte parity, collision refusal, and the exclusion truth table. `t312` covers usage errors and per-verb flag rejection, the verified/drifted/unattested flows, two- and three-dot ranges, squash stability, local-only evidence never verifying, fail-closed `unverifiable` and cross-record `indeterminate` (including the tie broken by one second restoring ownership), workspace-shell exclusion, a real `git clone --depth 1` boundary refusal plus recovery after `git fetch --deepen 1`, anchor dedupe/reconcile/swarm-skip, and the session-start sweep fired as the real spawned hook. `t135` asserts a swarm-merged record carries evidence hashing to the receipt. Updated and green: t10, t28, t48, t68, t81, t145, t227, t230, t239, t304, t305.

Manual, in a project with at least one completed per-unit code-generation review:

1. Confirm the record now carries `construction/<unit>/code-generation/reviewed-source-<hash12>.tsv`, and that its SHA-256 equals the `Unit Source Fingerprint` on the unit's `REVIEW_COMPLETED` block.
2. Commit the reviewed source, then run `aidlc attest resolve` — the reviewed paths report `verified` and the owning unit is named.
3. Edit one reviewed file, commit, re-run — that path reports `drifted` while the rest stay `verified`; `--fail-on drifted` exits 3.
4. Add an unreviewed file, commit, re-run — it reports `unattested`; framework shell and record paths report `excluded`.
5. Corrupt or delete the committed evidence TSV and re-run — the unit's paths report `unverifiable` with the reason, never `verified`.
6. Clone the project fresh (and separately with `--depth 1`) and re-run: the full clone reports exactly what the authoring checkout reported; the shallow one exits non-zero naming the boundary and the fix.
7. Start a fresh session and check the audit trail for `SOURCE_COMMITTED` rows covering the commits above; start another and confirm no duplicates.

## Acknowledgment


leandrodamascena (changes requested):
**Request changes**

The underlying need makes sense: mapping landed source changes back to reviewed units would improve traceability. However, the current design does not yet support the guarantees described by this PR.

1. **Historical resolution is not deterministic.** The resolver uses receipts, manifests, and evidence from the current worktree rather than the queried commit tree. Consequently, the same `(base, head)` can produce different results as the repository’s review records evolve. A drift warning does not address this correctness issue.

2. **There is no trust root.** Receipts, manifests, and evidence are ordinary files controlled by the branch author. A contributor could modify source and commit matching fabricated review records in the same PR, while record paths are excluded from attribution. The result could still be reported as `verified`. This can be useful as informational metadata, but it cannot currently provide a trustworthy CI or compliance attestation.

3. **Automatic anchoring should be separated.** Running reconciliation on every session start adds startup work and can modify audit records implicitly, although anchors are enrichment data and are not required by resolution. This should be explicit or opt-in.

Before continuing, I suggest revisiting the design to:

- Define the threat model and clarify whether this is informational reporting or an enforcement mechanism.
- Resolve historical commits using records from the corresponding Git tree.
- Establish a trusted authority for review receipts.
- Separate the resolver from automatic session-start anchoring.
- Rebase after the design is settled, since this currently conflicts with recent Change Control work on `main`.

The capability is valuable, but these are foundational correctness and trust concerns rather than implementation details.

rafaelgsr:
Thanks — the three concerns were correct, and each one changed the design rather than the wording. Summary of what the branch now does, then the specifics.

**The design question first: informational or enforcement?** Both, explicitly, and the report always says which. Resolution answers an *integrity* question — do the landed bytes equal the bytes some receipt approved? It cannot answer an *authenticity* question on its own, because receipts are ordinary files in the repository. So a report with no verifier-supplied anchor is informational metadata by construction, and it now says so in a `trust` object instead of leaving the reader to infer it. It becomes an enforcement mechanism only when the verifier supplies a trust root. The full threat model — assets, in-scope and out-of-scope adversaries, and the guarantee boundary — is `docs/reference/20-commit-provenance.md` §2.

**1. Historical resolution is not deterministic.** Fixed structurally, not with a warning. The resolver now reads receipts, manifests, and evidence out of a **git tree** rather than the worktree: a `RecordView` seam with two backings (`treeRecordView`, one `git ls-tree -r -z --full-tree` plus one batched `git cat-file --batch`; `worktreeRecordView` only for the one layout where the record provably cannot be inside the queried repository, and that report is marked `informational` with a warning). The same `(base, head)` is therefore stable as the repository's review records evolve, and the drift warning is gone because the drift is gone.

The guard test does exactly the thing that used to break it: resolve a commit, then review more units, then tamper with the committed evidence in the checkout, then re-resolve the same commit and require byte-identical output. Committing the tamper does change the verdict — and only from that commit forward, where it fails closed as `unverifiable`.

**2. There is no trust root.** Agreed, and the adversarial shape you describe — modify source and commit a fabricated matching receipt in the same change — is now a test case. Two anchors, both supplied by the verifier rather than by the change:

- `--record-ref <ref>` reads the record from a ref the change under test cannot write (a protected branch, a records-only ref, a mirror the review system maintains). With the record pinned, a self-written receipt is simply not in the record source, so its paths report `drifted`/`unattested` instead of `verified`.
- `--require-trust <level>` gates on the report's own basis (exit 3, same code as `--fail-on`), over a four-rung ladder: `informational` (worktree record) < `reproducible` (tree-read, but the range writes record paths, so the change may have authored the receipts judging it) < `independent` (the receipts relied on did not come from the change under test) < `signed` (independent, and every relied-upon evidence file arrived in a signed commit, `%G?` ∈ {G, U}).

`trust{}` reports the achieved level, what was required, where the record came from (`recordSource`/`recordRef`/`recordCommit`/`recordPinned`), which record paths the range itself touches, `selfAttested`, and the signature verdict per relied-upon evidence file. So a stored report can never read as stronger than the anchor that produced it. On your point that record paths were excluded from attribution: they still are (they are framework files, not source), but the report no longer stays silent about the fact that the range wrote them — that is precisely what caps the level at `reproducible`.

What remains out of scope, stated as such in the chapter: anyone who can write to the pinned ref, or forge signatures on record commits, can produce receipts the resolver honours. Establishing *who may write approvals* is branch protection, required signatures, and access control on the hosting platform. The tool reports the record's provenance; it does not create it. `signed` is also commit-level — an accepted key wrote the evidence — not an authorized-reviewer identity check; §10 lists that as a limit and the roadmap carries per-approval signing and an approver identity policy as future work.

**3. Automatic anchoring should be separated.** Done — it is now off by default, behind `AIDLC_SESSION_ANCHOR=1`. The reasoning matches yours: anchors are enrichment the resolver never reads, so an implicit write to the append-only audit trail on every session start is pure cost. An untouched install does no provenance work at session start. The explicit `attest anchor [--reconcile]` verb is the way to write anchors; the opt-in switch exists only for teams who want landed commits named in the trail without remembering a command.

**Rebase.** The branch is merged up to date with `main` (2.8.1 plus the installation/quick-start docs simplification), and the version is re-bumped to 2.8.2 per the changelog conflict policy.

Docs: the new chapter now leads with the threat model and adds two worked trust setups (records on a protected branch, and records alongside the change with signatures — including why the first gives separation of duties and the second does not). The guide's gate recipe shows the report form and the enforced form side by side, and says out loud that without `--record-ref` a change can verify itself.

leandrodamascena (changes requested):
Thanks for addressing the previous review. The new head fixes the three issues I raised earlier: historical records now come from a Git tree, trust is explicit and gateable, and session-start anchoring is opt-in.

I reviewed the current head `27ce31f98b23799d27a8cfc981a81fc7181e30c8` against base `52da70ada01dfbadd6f28a3d28ebfce4bc6a6079`. Project alignment comes from the accepted commit-provenance goal and the prior maintainer review. This head remains within that scope.

Two blocking issues remain:

1. **P1: `signed` trust does not cover the receipt that selects the evidence.** `buildOwnershipIndex` takes the authoritative unit, timestamp, stage, and fingerprint from a `REVIEW_COMPLETED` audit block, but `computeTrust` checks signatures only for the evidence file. I reproduced this with two signed reviews. I then appended an unsigned forged receipt with a later timestamp so it selected the older evidence and committed a matching source change. `aidlc attest resolve --record-ref <unsigned-receipt-commit> --require-trust signed --fail-on drifted,unattested,unverifiable,indeterminate` exited 0 and reported both `verified` and `signed`. The signed gate must cover every authority-bearing input that determines the verdict, including the selected receipt and its bound review record, or the `signed` level must be removed or renamed to avoid claiming receipt authenticity.

2. **P2: harness-shell exclusion still depends on the current checkout.** `classifyPath` calls `sourcePathIsExcluded` with `query.dir`, which discovers harness directories from the working tree. I committed a `.claude` shell and resolved that commit. The path reported `excluded`. I then deleted `.claude` only from the working tree and resolved the same SHA again. The same path changed to `unattested`. Derive the exclusion context from the queried Git tree and use that immutable context for classification.

Validation performed:

- Official `t311` and `t312`: 16 tests passed.
- Dispatcher, documentation, audit taxonomy, and version tests: 145 tests passed.
- Swarm evidence transfer test: passed.
- `bun run check`: package determinism, TypeScript, and Biome passed.
- Two added diagnostic regression tests fail on the issues above. No changes were committed or pushed.
- Remote CI is green except `Tests (smoke + unit)`, which is still running on this head.

rafaelgsr:
# Reply to second review (PR #1052)

Both findings reproduce, both were real defects in the round-1 fixes, and both are now fixed structurally rather than documented. Thanks for going after the *fixes* rather than the original code — the first one in particular was a hole in the trust model, not a bug in a check.

**1. `signed` did not cover the receipt that selects the evidence (P1).** Correct, and your reproduction is now a test. The authority chain here is receipt → `Unit Source Fingerprint` → evidence file → manifest digest → claimed paths. The receipt is the **root** of that chain, and git guarantees nothing about append-only-ness of an audit shard: whoever last wrote the shard blob could have written anything in it. Checking signatures only on the evidence therefore let an unsigned receipt borrow the standing of signed evidence it merely pointed at — exactly the claim nobody could honour.

Fix: ownership resolution now records, for the receipt it chose, the audit shard it came from, the commit that last wrote that shard, and git's `%G?` for it (a new `shardOrigin()` method on the `RecordView` seam, so the tree backing answers it from the queried tree). `computeTrust` builds `trust.signatures[]` from **both** inputs per relied-upon unit, each entry carrying `role: "receipt" | "evidence"`, and `signed` holds only when every entry is good (`%G?` ∈ {G, U}) and at least one exists. `receipt`/`receiptCommit`/`receiptSignature` are also reported per unit, since a reader auditing a `signed` verdict needs to see which receipt selected which evidence.

The guard test is your scenario: two signed reviews of one unit (both evidence files survive, since evidence names are content-addressed), a signed revert to the older content, then an unsigned commit that appends a forged `REVIEW_COMPLETED` block with a year-2099 timestamp whose fingerprint is the sha256 of the *superseded* evidence. The path still reads `verified` — that is honest, the bytes do match the receipt that now owns them — but the level drops to `independent`, `--require-trust signed` exits 3, and `signatures[]` shows `{role: "receipt", code: "N"}` next to `{role: "evidence", code: "G"}`. A control assertion checks the legitimate two-signed-review case still reaches `signed`, so the test would fail if the fix simply made `signed` unreachable.

`signed` keeps its name because it now means what it says: every input the verdict rests on arrived in a signed commit. What it still does *not* mean is an authorized-reviewer identity check — signatures are commit-level and any key git accepts counts — and §10 of the chapter states that limit, with per-approval signing and an approver identity policy on the roadmap.

**2. Harness-shell exclusion still depended on the current checkout (P2).** Also correct — the round-1 change moved the *record* onto a tree read and left this one path reading the filesystem, which is precisely the inconsistency you found: same SHA, different verdict depending on which harnesses happen to be installed where you run it.

Fix: the exclusion context is derived from the queried git tree. A root-level `git ls-tree` finds dot-directories, then one pathspec'd `ls-tree` plus a batched `cat-file` looks for a committed `<dir>/tools/data/harness.json` in each, validated by the same predicate the checkout-side rule uses — the shape test and the manifest test are now exported from `aidlc-lib.ts` (`isHarnessDirName`, `HARNESS_SHELL_MANIFEST_REL`, `isHarnessShellManifest`) so there is still exactly one definition of "what is a harness shell", applied to a tree instead of a directory. `sourcePathIsExcluded` takes an optional context that supplies the shell set instead of discovering it. For a range, the context is the **union of both endpoints**, so installing or removing a harness inside the range does not push framework paths into `unattested` from the side that lacks the manifest; `anchor` does the same per commit against its first parent.

The determinism test now does both halves of your reproduction: delete `.claude/` from the working tree only and require byte-identical output for the same SHA, then commit the deletion and assert the path in the *earlier* commit is still `excluded` with `unattested: 0`.

Verified on the branch: `bun scripts/package.ts --check` (two-build determinism), `bun run lint`, `npx tsc --noEmit`, and the unit tier. Changelog bullets and `docs/reference/20-commit-provenance.md` (§2 threat model in-scope list, §4 determinism, §6 statuses and report shape, §9 worked setups, §10 limits) are updated in the same commit.

leandrodamascena:
I updated the branch and merged the current `main`.

The latest head `1cf46d809` addresses the remaining review findings:

- Harness exclusions are now fail-closed. Only a harness shell already established in the range’s base tree can be classified as `excluded`.
- A commit can no longer add an arbitrary `<dot-directory>/tools/data/harness.json` and use it to hide sibling files from attestation.
- The regression case adds a fake `.github` harness manifest and an unreviewed workflow. The strict gate now exits 3 and classifies both paths as `unattested`.
- Updates and removals of previously established harness shells remain `excluded`.
- The feature PR no longer changes version metadata or adds a CHANGELOG release entry.
- The README tool inventory was updated to 70 after merging `main`.

Validation completed:

- `t312`: 13 passed
- `t311` and `t239`: 18 passed
- `t135`: 19 passed
- `t68`: 7 passed
- TypeScript, Biome, package determinism, changelog guard, and coverage-registry checks passed

The branch is ready for CI and another review.

leandrodamascena (approved):

leandrodamascena (approved):
Thanks for the work on commit provenance. I reviewed the current head `090b330df` against base `355903d6d`; the feature remains aligned with the previously reviewed direction.

The upgrade compatibility concern is resolved. In-flight swarms reviewed by an older runtime can now promote their existing receipt-bound local snapshot into committed provenance evidence during finalization. The bytes must still match the receipt fingerprint and manifest; missing or altered evidence continues to fail closed.

Validation completed:
- Swarm pre-upgrade compatibility regression: passed
- `t311` and `t312`: 18 passed
- TypeScript: passed
- Biome: passed
- Package determinism: passed
- CI smoke, unit, integration/e2e, contract, changelog, documentation, and security checks: passed

No version bump or release metadata is included; that remains part of the release process.

PR #1054, opened 2026-09-08
fix: accept Enter for wizard defaults and make same-version update umask-independent

apackeer:
Two defects found while exercising the published v2.8.0 native install end to end on Linux (arm64) and Windows Server (PowerShell 5.1, standard user), plus the remediation/doc text they exposed. Bumps to 2.8.1.

## 1. First-run wizard cancels on Enter (P1)

Every bracketed default in the bare `aidlc config` wizard - harness picker, provider, region, preset, plugins, MCP, record layer, and the final `Apply? [Y/n]` gate - printed `Nothing written.` and exited 2 when Enter was pressed, contradicting the wizard's own `Enter accepts the [default]` line. Reproduced on Linux (pexpect pty) and Windows (ssh -tt / conpty).

**Root cause:** Bun's global `prompt()` returns `null` for an empty line (`bun -e 'prompt("Q:")'` + Enter -> `null`), and `firstRunPromptValue` in `aidlc-init.ts` treated `null` as `FirstRunCancelled`. The scripted-answer test seam (`AIDLC_TEST_CONFIG_TTY=1`) never reaches `prompt()`, which is why t299 passed.

**Fix:** `readTerminalLine` in `aidlc-command.ts` reads one line from fd 0, byte-wise so nothing past the newline is consumed: `""` for Enter, `null` only on EOF (Ctrl-D / Ctrl-Z). Both interactive prompts (wizard `configPrompt` and the uninstall `Continue [y/N]` confirmation) use it so there is one prompt convention; the uninstall prompt's observable behavior is unchanged (empty or anything but y/yes cancels). Verified through a real conpty on Windows: Enter -> `""`, text -> text, Ctrl-Z -> `null`.

**Tests:** t299 gains a real-pty run (util-linux `script`, Linux-only) that presses Enter at the recommended-defaults gate - fails on the old code with `Nothing written.` / exit 2 - plus a unit test for the reader: empty line vs EOF, CRLF, and three consecutive answers on one descriptor.

## 2. `aidlc update` on a current install fails under a normal umask (P2)

`aidlc update` when already on the latest release failed with `error: existing 2.8.0 runtime does not match the verified release` (exit 4) after downloading ~110 MB, unless the shell umask was `077`. `--dry-run` said `Would update aidlc from 2.8.0 to 2.8.0.`

**Root cause:** `install.sh:288` sets `umask 077`, so the installed runtime is 0600/0700. A later `aidlc update` re-extracts the tarball (header modes 0644) under the user's umask, and `treesMatch` compared `mode & 0o777` alongside content. Deterministic: `(umask 077; aidlc update)` succeeds, `(umask 022; aidlc update)` exits 4. The installed tree was byte- and mode-identical to its own `runtime-integrity.json` - nothing was corrupt.

**Fix:** same-release identity is path set + content. The installed tree's modes are still enforced against its own recorded baseline by `completeVersion()`, which the same branch calls right after, so integrity is not loosened - t243 now also asserts that mode drift is rejected by the baseline. Extraction is unchanged, so trees installed by 2.8.0 under any umask keep working after upgrading. The same-version `--dry-run` message now says `You're on the latest version of aidlc (X); nothing to update.`

**Tests:** t244 gains the install-under-`077` / update-under-`022` regression (fails on the old code with the exact production error).

## 3. Text and hermeticity

- `aidlc doctor` no longer says `copy the workspace shell from dist/<harness>/` (retired channel); remediation is `aidlc config`. t22/t165 updated accordingly.
- README: removed the pre-2.8.0 IMPORTANT note telling users to install from a source checkout until native assets ship.
- t299/t304 children now get an isolated `AIDLC_INSTALL_ROOT`/`AIDLC_BIN_DIR`. On a machine with a real `aidlc` install, `installedSources()` adds the active machine runtime on top of `AIDLC_RUNTIME_ROOT`, so the wizard listed every harness twice and 10 tests failed. Verified: 30/30 pass with a real 2.8.0 install present and no env overrides.

## Validation

- `bun scripts/package.ts --check`: deterministic, zero drift.
- Dispatched CI on the first head (b25536f84): all jobs green - contract checks, smoke + unit, deterministic integration + e2e. Re-dispatched on the current head; link in the checks tab.
- Local: t68 version sync, t230, t243, t244, t299, t304 green; t22 and t165 (deterministic integration, `AIDLC_NO_LLM=1`) 49 pass. A local full smoke+unit run had three failing files (t238, t244 Windows-journal, t304) that all traced to the host's real machine install leaking into the run; they pass with the hermeticity fix and on CI.
- Regression tests confirmed to fail on the pre-fix code with the production error strings.

Not in this PR (observed during the same test pass, product decisions rather than defects): `install.ps1` refuses the built-in Administrator (documented, but a hard stop on Windows Server/EC2 default accounts); `aidlc doctor` outside any project emits four Claude-specific `fail` rows; the runtime PATH remediation is generic because the check's baseline is `getconf PATH`, which no shell startup file changes; Cursor scaffold drops a Bun `install.ts` at the project root; fresh scaffolds on five harnesses warn `Plugins: 1 need attention` before any session; the shipped `.gitignore` block carries generic `dist`/`node_modules` rules.


leandrodamascena (changes requested):
Thank you for investigating these two release regressions against the published 2.8.0 native installation paths. I reviewed the current head `003b0c094` against current base `78a6879e0` and merge base `0d399dd828`.

The scope is aligned as an internal bug fix. It restores two existing contracts rather than introducing new product behavior:

- Pressing Enter at a bracketed wizard prompt accepts the displayed default.
- Reinstalling or updating to the same verified release succeeds when the original installation and later update ran under different umasks.

I traced both implementation paths.

For the wizard, Bun's global `prompt()` returned `null` for an empty line. `firstRunPromptValue()` interpreted that as cancellation, causing the wizard to print `Nothing written.` and exit 2. The new `readTerminalLine()` correctly distinguishes an empty line (`""`) from EOF (`null`) and consumes exactly one line so queued answers remain available. Reusing it for uninstall confirmation preserves the existing default-no behavior.

For same-version updates, the previous `treesMatch()` compared paths, bytes, and complete permission modes. An installation extracted under the installer's `umask 077` therefore differed from an identical candidate extracted later under a normal `umask 022`. Comparing path inventory and content in `treesMatch()` is appropriate because `completeVersion()` subsequently validates the installed runtime, including modes, against its recorded `runtime-integrity.json` baseline.

I found one material gap in the test that is intended to protect that integrity argument.

In `tests/unit/t243-install-mechanism.test.ts`, the test first corrupts one runtime file's content and confirms that inspection fails. It then leaves that content corrupted, changes a different file's mode, and calls `inspectInstalledVersion()` again. The second inspection still fails because of the earlier SHA mismatch. Consequently, the test remains green even if mode validation is removed entirely.

Please make the mode case independent: restore the corrupted content, assert that the installation is complete again, change only one file's mode, and then assert that inspection fails because that specific file no longer matches the installed baseline. Splitting content and mode drift into separate pristine-install cases would also work.

The branch also conflicts with `main` after #1053. Please rebase while preserving the new simplified README. The final README diff should retain the current Quick Start structure and apply the 2.8.1 badge; it must not restore the removed pre-#1053 sections. The obsolete pre-2.8.0 installation note is already absent from current `main`.

Validation performed on `003b0c094`:

- `bun scripts/package.ts --check` passed across all seven harnesses.
- Version, changelog, and README synchronization passed 7/7.
- t299 passed all 11 applicable tests; the Linux-only real-PTY test was skipped on macOS.
- The new differing-umask regression passed.
- The focused installed-runtime integrity test passed, subject to the blind spot above.
- The exact-head contract and deterministic integration/e2e CI jobs are green.
- Smoke/unit CI is still running.
- Local failures outside the focused paths came from this sandbox refusing `Bun.serve(port: 0)`; the unrelated load-sensitive timeout passed when rerun alone.

I made no source changes and pushed no commits.

I do not currently see a functional defect in either implementation fix. However, the test does not verify the mode-integrity property that makes the relaxed `treesMatch()` comparison safe, and the current head cannot merge until it is rebased. I am therefore requesting changes.

apackeer:
Addressed the requested changes in `821f676ce`.

- Rebased onto current `main` after #1053. The README now preserves the simplified Quick Start exactly, with only the intended `2.8.1` badge change.
- Made the runtime mode-integrity case independent: the content mutation is restored, completeness is re-established, then a mode-only mutation is checked against the exact affected file and restored to a complete install.
- While exercising the Linux real-PTY regression, fixed its detection fixture to provide a complete synthetic harness map. This prevents shell startup PATH additions from inserting an unrelated host CLI prompt before the single Enter intended for `Choice [1]`.

Verification:

```text
STAMP:    tests/logs/2026-09-08T20-19-29Z-p714055
TRACES:   tests/logs/2026-09-08T20-19-29Z-p714055/*.ndjson  (0 files; unit tier)
SUMMARY:  tests/logs/2026-09-08T20-19-29Z-p714055/summary.txt + failures.txt - Result: PASS; Failed files: 0
RESULT:   unit . 3 pass/0 fail . reds: none . live vars set: none . invariant grep hits: 0 (path-excluded)
```

The focused files passed 103 assertions across `t243-install-mechanism`, `t299-first-run-wizard`, and `t68-version-changelog-sync`. `bun run check` also passed: all seven harnesses regenerated deterministically, all three TypeScript configurations passed, and Biome checked 736 files with no findings.

leandrodamascena (approved):

leandrodamascena (approved):
Thank you for addressing the previous review findings and rebasing the branch.

I reviewed the current head `821f676ce` against base and merge base `78a6879e0`.

The scope remains aligned as an internal bug fix. It restores the documented wizard default behavior and permits a same-version update across different umasks without weakening the installed runtime baseline.

I found no remaining code-review findings.

The previous blockers are resolved:

- The branch is rebased onto current `main`. The README keeps the simplified Quick Start from #1053 and changes only the version badge.
- The runtime-integrity test now restores the corrupted content, confirms that the installation is complete, changes only that file's mode, checks the exact mode-drift failure, restores the mode, and confirms completeness again.
- The real-PTY detection fixture now supplies all seven harness rows, preventing shell startup PATH entries from introducing an extra prompt.

I rechecked both implementation paths. `readTerminalLine()` returns an empty string for Enter, returns `null` for EOF, and consumes one line without dropping queued answers. Its uninstall use preserves the existing default-no contract. Same-release matching compares the path inventory and content, while `completeVersion()` still checks the installed tree's content and modes against `runtime-integrity.json`.

Validation performed on `821f676ce`:

- The five changed or directly affected unit files passed with 156 tests, 2 platform skips, and 0 failures.
- The version, changelog, and README synchronization checks passed 7/7.
- All required CI passed, including package determinism, typecheck, lint, smoke, unit, deterministic integration, and e2e.
- Documentation, security, title, contributor statement, and changelog checks passed.
- The local package check stopped with host `ENOSPC`. The equivalent CI contract check passed on this head.

I made no source changes and pushed no commits. I did not independently rerun the Windows conpty path from this macOS review host. The byte-reader tests, Linux real-PTY CI case, and reported Windows conpty validation cover that path.

Issue #723, opened 2026-08-06
RFC: The plugin marketplace experience: a first-party marketplace repo, remote discovery in the CLI, core as a choosable plugin, and a graduation path


apackeer:
*RFC: Plugins*

The plugin mechanism and its git-repo marketplace pattern
already shipped; this RFC builds the experience on top: a first-party
marketplace repo (the template private customer marketplaces copy), remote
discovery and update in the CLI across all five harnesses, the core-plugin
choice surfaced instead of silently defaulted, and a documented path for
community plugins to graduate into core.

Status: the plugin mechanism shipped on v2; this RFC is about
the experience layer on top. One new repo, four decisions (one contested:
remote trust), and a process. Depends on the Distribution RFC (#722) for the CLI front door.

## What it feels like: every path, as a transcript

Before the mechanism, the experience each option buys. Transcripts are
illustrative; each carries its honest status - the local flows are shipped
or ride the Distribution RFC's install branch, the remote flows are this RFC's proposals.

### 1a. Installing a plugin, per harness - hosts with stores today; ONE surface everywhere after D-1

```
# Claude Code - the host's own marketplace flow works today (host trust prompt gates it)
$ /plugin marketplace add awslabs/aidlc-plugins
$ /plugin install aidlc-test-pro

# Codex - same shape, host-native, works today
$ codex plugin marketplace add <repo>/dist/plugins/test-pro/codex
$ codex plugin add aidlc-test-pro

# Every harness, after D-1 - the same three verbs, Kiro and opencode included:
$ aidlc plugin marketplaces add https://github.com/awslabs/aidlc-plugins
$ aidlc plugin install test-pro
  -- fetched tag test-pro--v0.2.0, checksum verified
  -- this plugin installs 2 hook files that run in your shell:
       hooks/compose.kiro.hook, hooks/compose.ts
     Install? [y/N] y
  -- composed: 2 new stages + 4 contributions merged into the project graph
```

The experience is deliberately uniform: on Claude and Codex,
`aidlc plugin install` hands off to the host's own flow so its trust
prompt still gates execution; on Kiro and opencode - no host store - the
SAME command is the managed path, and its explicit confirm (naming every
hook file) is the trust gate those platforms otherwise lack (D-3). Nobody
types `git clone`: the tag-pinned, checksum-verified fetch is the managed
path's job. (Cloning a plugin repo by hand keeps working as the
air-gapped/source fallback - `aidlc plugin sync` composes whatever is on
[...]

PR #1056, opened 2026-09-08
feat(plugins): ddd — domain model as an enforced deliverable

johndo9:
# Summary

Adds the `ddd` plugin: Domain-Driven Design as an installable AIDLC plugin that makes the domain
model a first-class, machine-checkable deliverable — produced early, approved by the business, and
enforced against generated code as a hard gate. 28 files, ~2,275 lines, no `core/` edits: everything
is additive through the plugin mechanism, following the `test-pro` reference plugin's conventions
including its own `tests/` suite on the plugin test kit.

Generated code is only as trustworthy as the constraints around it. This plugin treats the domain
model as a formal specification — ubiquitous language, bounded contexts, a typed context map,
aggregates with declared invariants and finite state machines, domain events, structured rules — in
machine-checkable YAML with stable IDs (`{project}.{context}.{type}.{name}`). The `ddd-conformance`
gate compiles that model into executable tests and runs them after Build and Test. Code either
conforms, or a human adjudicates: fix the code, or amend the model as a business decision with
sign-off. Nothing passes silently, and the model is never weakened in place.

This builds on the #711 domain/contract design restructure. #711 gave every unit machine-checkable
design artifacts — `entities.md` and `rules.md` with YAML source-of-truth blocks, `functional-spec.md`
owning workflows and state machines, `components.md` with single-owner entity catalogues. What was
missing is a *global* model that bounds them all, plus enforcement. Roadmap context: #723 names a
design plugin (#527) as a first-party candidate; this is offered as that candidate or as input to it.

## Changes

**New stages**

* `ddd-domain-modeling` (2.15, Inception) — produces the `ddd-domain-model` artifact; brownfield path
  seeds from the CodeKB (`business-overview`, `component-inventory`); business approval gate with the
  product-lead reviewer bound to the model artifact.
* `ddd-conformance` (3.9, Construction) — hard gate after `build-and-test` via a real `requires_stage`
  edge (not the deferred `adds.requires_stage`); generates and RUNS the domain-conformance suite in
  the workspace (`workspace_requires: true`); adjudicates violations as code fixes or human-approved
  model amendments; emits `ddd-conformance-report.json`.

**Contributions (additive-only)** onto `units-generation` (bounded contexts bound decomposition),
`functional-design` (projection binding: `entities.md` entries trace to model element IDs, realized
invariants in `rules.md` trace to model rule IDs, `functional-spec.md` state machines must be subsets
of the owning aggregate's FSM), `code-generation` (runtime enforcement of invariants and guarded FSM
transitions), `build-and-test` (fold domain tests into the standing suite so they re-run in CI), and
`requirements-analysis` (`adds.scopes`, see below).

**Scope `ddd-modeling`** — a standalone modeling-only engagement: init → requirements-analysis →
ddd-domain-modeling, deliverable being the approved domain model (workshops, discovery, KB seeding).
The conformance gate deliberately stays out of this scope; the full path runs under
`enterprise`/`feature`/`mvp`/`workshop`.

**Sensors** — `aidlc-ddd-conformance` is advisory; it reads the conformance report.
`aidlc-ddd-model-schema` is deterministic and **blocking** (`fire_on: gate`): it validates the
authored domain model against the normative schema before the modeling gate can pass. This changes
gate behaviour, so this PR is more than a docs change — see the follow-up comment for why it was
added.

**Knowledge and docs** — `ddd-modeling-method.md` under the architect agent (additive, no-clobber);
plugin README; and `ddd-model-and-rule-schema.md`, the normative schema, now shipped as architect-agent
knowledge so it reaches the agent at authoring time. It covers the frontmatter schema with a worked
example, the required/optional field table, stable IDs, the four rule tiers and their compilation
targets, the derived rule-0 set, projection mapping into core artifacts, and the report shape.

## User experience

**Before:** a team can capture entities, rules, and state machines per Unit, but nothing binds those
per-Unit artifacts to a single agreed domain model, and no gate verifies generated code against it.
Domain drift between Units is invisible until someone reads all the designs.

**After:** with the plugin installed and enabled, Inception gains a business-approved domain model
that bounds downstream decomposition and design, and Construction gains a hard gate that compiles
the model into executable tests (structural boundary/naming, invariant property tests, exhaustive FSM
transition matrices, example-based rules) and runs them. A violation stops the workflow for
adjudication rather than shipping. The authored model is itself gate-checked against the schema, so a
model that drifts from the specification cannot pass the modeling gate. Teams that only want the model
can run `--scope ddd-modeling` and stop after Inception. Teams that do not enable the plugin see no
change.

## Checklist

* [x] Changes are documented

## Test plan

Per the current contributing checklist (post-#756, generated trees are ignored locally):

```bash
bun install --frozen-lockfile
bun core/tools/aidlc-plugin-validate.ts plugins/ddd    # 0 errors, 1 expected warning
bun scripts/package.ts                                 # materializes dist/ + dist-release/
bun test plugins/ddd/tests                             # 27/27 across 2 files
bun run check                                          # exit 0
bun test tests/unit/t68-version-changelog-sync.test.ts # 7/7
```

Run these serially. `bun test plugins/ddd/tests` and `bun run check` both touch `dist/` and the
disposable install fixtures, so running them concurrently produces spurious failures.

The plugin validator reports one expected warning (no vendored `hooks/compose.ts`; the build injects
the bundled one). The compose smoke builds the claude projection into a disposable install and
asserts both stages land, all five contributions merge, the `ddd-modeling` scope join happens, and
the sensor/scope/knowledge files copy over.

For the schema sensor specifically, every defect class has a fixture observed failing on exactly that
class (`ev-agg`, `rule-agg`, `fsm`, `id-ctx`, `ev-past`, `rule-kind`, `top-version`), the schema
document's own example is observed passing, and any non-model artifact is a clean pass-through.

The full L1 suite was run locally on an earlier revision of this branch; the targeted verification
above was re-run on the current head. The L1 failures at that point were environment-specific —
5-second budgets on git-heavy workspace-sync/dispatcher tests, one Windows-evaluator timeout, and the
native binary compile — and they **reproduce identically on a clean `main` checkout with no plugin
present**. The live-LLM sdk families are excluded from the PR gate by `--no-llm`.

**Dogfooded end to end.** The `ddd-modeling` scope was run against a real brownfield TypeScript
service (a Nova Sonic / Bedrock real-time translation service, ~240 components) on a Kiro IDE
install: install upgrade → plugin compose → scope join → 5-stage plan → both stages to completion.
Every enforcement seam held (summary-confirmation receipts, artifact-write freshness against the
audit ledger, two-phase adversarial review, a gate rejection/revision cycle, human-presence minting).
Two defects in the *modeling work* were caught by two different layers: the adversarial reviewer
refuted an FSM state that was unsatisfiable for one of two delivery directions (it would have failed
the conformance gate's transition-matrix test against correct code), and the human business gate
caught reverse-engineering documentation that had recorded a config default as domain behavior. Two
independent safety nets, each catching what the other missed.

That run also surfaced a defect in the plugin itself, which is what the schema sensor fixes — the
follow-up comment has the detail.

## Known limitations / follow-ups

* `aidlc-ddd-conformance` (the Construction report sensor) is still advisory — write-fired sensors
  carry no blocking severity; binding it gate-bound blocking per #836 is a natural follow-up. The new
  `aidlc-ddd-model-schema` sensor is already gate-fired and blocking, so the design-time check is
  hard today.
* Inline mode with core reviewer agents keeps compose clean on every harness; a `mob` topology for
  live model-storming is a later enhancement.
* The ubiquitous-language glossary is a natural query vocabulary for semantic code search (model term
  → implementing components); out of scope here, noted in the plugin README.

## Acknowledgment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.8.1] - 2026-09-08
+
+Add the `ddd` plugin: Domain-Driven Design as an installable AIDLC plugin that makes the domain model a first-class deliverable. It adds a `ddd-domain-modeling` stage in Inception (ubiquitous language, bounded contexts, aggregates, invariants, state machines, domain events, business rules), a `ddd-conformance` hard gate after Build and Test that generates and runs a domain-conformance test suite, contributions onto Units Generation, Functional Design, Code Generation, and Build and Test, an advisory `ddd-conformance` sensor, a `ddd-modeling` scope, and architect-agent modeling methodology knowledge. **Install:** copy `dist/plugins/ddd/<harness>/` into the project and run `/aidlc plugin sync`.
+
+* New stages `ddd-domain-modeling` (2.15, Inception) and `ddd-conformance` (3.9, Construction) run when the plugin is selected under the `enterprise`, `feature`, `mvp`, or `workshop` scopes. The plugin's `ddd-modeling` scope is a standalone modeling-only path (initialization, requirements analysis, domain modeling — no build, no gate) whose deliverable is the approved domain model.
+* The conformance gate adjudicates violations as either code fixes or human-approved domain-model amendments, and folds the generated domain tests into the standing suite.
+
 ## [2.8.0] - 2026-09-08
 
 AI-DLC 2.8.0 consolidates the 2.7.x release cycle into a new minor baseline without changing runtime behavior from 2.7.2. **Upgrade:** use `install.sh --version 2.8.0`, `install.ps1 -Version 2.8.0`, or replace a manual copy with `runtime/<harness>/` from `aidlc-runtime-2.8.0.tar.gz`. Existing 2.7.2 workflow records require no migration. Upgrades from earlier releases must still apply every intervening **Upgrade**, **Breaking**, and migration note below.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ structured, verifiable software-delivery workflows. One harness-neutral core
 runs natively in Claude Code, Kiro CLI, Kiro IDE, Codex CLI, Cursor, opencode,
 and GitHub Copilot.
 
-![version](https://img.shields.io/badge/version-2.8.0-blue)
+![version](https://img.shields.io/badge/version-2.8.1-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 
 The Quick Start below installs the latest stable AI-DLC release.
--- a/core/tools/aidlc-version.ts
+++ b/core/tools/aidlc-version.ts
@@ -1,4 +1,4 @@
 // Hand-edited single source of truth for the AIDLC framework version.
 // Bumped in the same commit that adds the matching ## [N.N.N] heading
 // to CHANGELOG.md. Pinned by tests/unit/t68-version-changelog-sync.test.ts.
-export const AIDLC_VERSION = "2.8.0";
+export const AIDLC_VERSION = "2.8.1";
--- a/plugins/ddd/.aidlc-plugin/plugin.json
+++ b/plugins/ddd/.aidlc-plugin/plugin.json
@@ -1,0 +1,17 @@
+{
+  "name": "ddd",
+  "version": "0.1.0",
+  "description": "Domain-Driven Design as a first-class, inviolable control plane. Adds a domain-modeling stage that produces a structured ddd-domain-model (ubiquitous language, bounded contexts, aggregates, invariants, finite state machines, domain events) and a ddd-conformance gate stage that compiles the model into executable tests (structural rule-0 + invariants + FSM + business rules) and runs them as a hard gate. Reuses aidlc-architect-agent (lead) and aidlc-product-agent (business voice).",
+  "author": { "name": "AWS AIDLC" },
+  "dependencies": ["core"],
+  "aidlc": {
+    "contributes": {
+      "stages": "stages/",
+      "overlays": "contributions/",
+      "scopes": "scopes/",
+      "knowledge": "knowledge/",
+      "sensors": "sensors/",
+      "tools": "tools/"
+    }
+  }
+}
--- a/plugins/ddd/README.md
+++ b/plugins/ddd/README.md
@@ -1,0 +1,50 @@
+# ddd — Domain-Driven Design plugin
+
+Makes the domain model a **first-class, inviolable deliverable** and enforces it with **deterministic,
+blocking controls**: the model compiles into executable tests, and a failing test is a hard stop.
+
+## What it adds
+
+- **Stage `ddd-domain-modeling`** (inception) — produces the structured `ddd-domain-model` artifact
+  (ubiquitous language, bounded contexts, context map, entities/value objects, aggregates with finite
+  state machines, domain events, and rules). Facilitated event storming: architect leads, product
+  supplies business language, business stakeholders confirm at the gate.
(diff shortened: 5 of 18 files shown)


johndo9:
Pushed two commits.

**`76992fb93`** — ddd: ship model schema as knowledge; add blocking `ddd-model-schema` gate sensor.

Heads-up that this widens the review scope beyond docs: it adds a blocking sensor, so it changes gate
behaviour. The description is updated accordingly.

**Why.** Validating this plugin on a real 24-stage `mvp` project surfaced a defect in the plugin
itself. The `ddd-domain-modeling` stage produced a model that diverged from
`ddd-model-and-rule-schema.md` — missing `aggregate:` citations on events and invariants, every FSM
transition keyed `event` instead of `on`, and shared-kernel IDs in an undeclared context — and it
passed the modeling gate *and* a two-round adversarial review. Root cause was specification delivery,
not the agent: the schema was never shipped into the harness (it lived only in the plugin source),
the stage prose paraphrased just its top-level keys, and nothing enforced it. The stage ran
`mode: inline` and cleanly.

The omission was load-bearing. The schema says invariants compile to "property test + runtime-assertion
presence"; without `aggregate:` the generator has no declared home for that assertion. And because the
model and the generated suite were both written from the model rather than from the schema, they agreed
with each other while both disagreed with the spec — a suite written against `on` would have derived an
empty FSM.

**What the commit does**

* Moves the schema into `knowledge/aidlc-architect-agent/` so it ships to the architect agent, and
  makes it normative (required/optional field table, `{ from, on, to }`, `aggregate:` on events and
  invariants, declared context per ID prefix, any-token past-tense).
* `ddd-domain-modeling`: cites the schema as the shape to author against, names the easily-dropped
  fields, and declares the new sensor.
* New `sensors/aidlc-ddd-model-schema.md` (deterministic, `default_severity: blocking`,
  `fire_on: gate`) + self-contained `tools/aidlc-sensor-ddd-model-schema.ts`; clean pass-through on any
  non-model artifact.
* `ddd-conformance`: compiles from the schema's field names instead of adapting to drift; adds
  "aggregate citation" to derived rule-0.

**Verification.** `bun test plugins/ddd/tests` 27/27, `bun run check` exit 0, plugin validator 0 errors.
Every defect class has a fixture observed failing on exactly that class (`ev-agg`, `rule-agg`, `fsm`,
`id-ctx`, `ev-past`, `rule-kind`, `top-version`), and the schema's own example is observed passing. On
the validation project, `engine sensor fire ddd-model-schema` returned `result: failed` against the
drifted model with zero false positives; the corrected model passes both the generated suite and an
independent validator.

**`ce42bffb3`** — drops a stale `2.8.1` changelog entry this branch had carried since Sep 8. It
collided with the released 2.8.1, which documents unrelated native-install fixes, and sat above 2.9.0
in the file, failing `t68` 3/7. The branch now carries no release-metadata delta against `main`, per
the release-metadata policy. `t68` is 7/7.

**What I need to move this forward**

1. **Workflow approval.** Pull Request Validation is green, but `CI`, `Markdownlint`,
   `Security Scanners`, and `Adversarial AI PR Review` are all at `action_required` for this fork, so
   no test, lint, or security suite has run against the PR yet. Happy to act on whatever the runs
   surface.
2. **Do you want this as one commit or two?** The schema move and the blocking sensor are separable;
   I'll split them on request.
3. **Is this the #527 design-plugin candidate, or input to it?** @apackeer — given #1104, you're
   probably the right person to call this. If `ddd` should land as a marketplace plugin rather than in
   `plugins/`, I'd rather restructure now than after review.

Minor: still labelled `documentation`, which no longer fits now that it adds a blocking gate sensor.

PR #1057, opened 2026-09-08
test: bound cleanup hooks in t243 and t248

vivibui:
## What

Two test files remove every temporary tree they staged in a single `afterAll`,
whose cost scales with the file rather than with any one case — so it outgrew
bun's 5s hook default and the file failed on cleanup alone. Both cleanup hooks
now carry an explicit `120_000` budget.

Test-only, so no version bump or CHANGELOG entry per the Changelog Policy.

## Measured

| File | Alone | Under load | Cleanup hook |
|---|---|---|---|
| `t243-install-mechanism` | fails **deterministically even alone** — 84 pass / 1 fail, `hook timed out` | same | 6464–6703 ms observed |
| `t248-steering-content-delivery` | **20 sequential runs, 20 pass / 0 fail** — clean at rest | **base 0/12, head 12/12** across 4 processes × 3 rounds, every base failure being 42 passing tests then the hook timeout | 8686 ms at `--parallel 8`; mean 3181 ms unloaded |

The `t248` pair of numbers is the useful one: the file is fine in isolation and
reliably broken under the conditions the suite actually runs in, so testing it
alone and concluding "passes for me" is the trap.

A timed-out hook does two kinds of damage: it fails the file for a reason
unrelated to anything under test, and it abandons whatever it had not yet
deleted. One `t243` run left 606 files / 6.5 MB behind in `$TMPDIR`. After the fix
only zero-byte dangling `*-alias` symlinks remain — `t243:573-574` creates that
symlink as a sibling of `machine` and never pushes it onto `temporary`, so
`rmSync` takes the target and leaves the link. Pre-existing; deliberately left
alone.

## Why the budget is on the hooks, not the file

An earlier revision used a file-wide `setDefaultTimeout(120_000)`. Scoping it to
the two `afterAll` hooks is better, and `t243` itself is the argument: the file
**already carries 39 explicit `}, 60_000)` case budgets**, so per-case budgets are
its local convention, and a file-wide default would let every unannotated case run
for two minutes before failing. The one case a load probe showed crossing the 5s
case default, `--force does not replace a pre-existing user-owned JSON entry`, gets
its own explicit budget instead.

## Known pre-existing failure this did not fix

`t243 release lifecycle > project pins require the retained OpenCode runtime
selected by project metadata` times out under macOS load — 8 of 12 runs at
5002–5019 ms. It sets its own budget at `t243:1900`
(`process.platform === "win32" ? 30_000 : 5_000`), and an explicit per-case value
overrides both a file-wide default and a hook budget, so no timeout mechanism in
this PR could reach it. Verified pre-existing: the same 4-way load against
unmodified `origin/main` fails 4/4 at 5004–5006 ms. Left for whoever owns that
budget.

## Verification

- `bun test` both files together — **126 pass / 0 fail / 1547 expects**.
- 4 concurrent processes × 3 rounds on macOS — **zero hook timeouts**; 125/1 in
  rounds 1–2 and 126/0 in round 3, the single failure being the pre-existing case
  above.
- `bun run typecheck` (3 projects), `bun run lint` (736 files), `git diff --check`
  — all exit 0.
- `bun scripts/package.ts --check` — deterministic across two independent builds,
  all seven harnesses.

Verified on macOS; CI covers Linux and Windows. Note that this class of failure
does not appear in Linux CI, so this fixes a failure developers hit locally rather
than a red pipeline.

## Relationship to #1043

#1043 is the umbrella effort for this failure class. It touches neither of these
two files, so this complements rather than overlaps it — happy to have it folded
in there instead.

An earlier revision also rewrote the `t298` live-holder launch. Dropped: #1043 and
#1025 both already fix it, and their one-line `>/dev/null 2>&1` redirect is the
better fix — detaching the holder's stdout is what lets `spawnSync` return while
the holder is still alive, and it keeps the `READY <pid>` line out of the buffer
the pid is parsed from.

---

### DIFF
--- a/tests/unit/t243-install-mechanism.test.ts
+++ b/tests/unit/t243-install-mechanism.test.ts
@@ -1,7 +1,7 @@
 // covers: tool:aidlc-init, tool:aidlc-lifecycle, file:core/tools/aidlc-archive.ts
 // covers: file:core/tools/aidlc-transaction.ts, file:scripts/package.ts
 
-import { afterAll, beforeAll, describe, expect, setDefaultTimeout, test } from "bun:test";
+import { afterAll, beforeAll, describe, expect, test } from "bun:test";
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
 import {
@@ -104,10 +104,6 @@ const NEXT_VERSION = (() => {
 const temporary: string[] = [];
 const originalPath = process.env.PATH;
 
-// The default also governs afterAll removal of 128 staged roots, several of them
-// dist-sized, which exceeds bun's 5s hook default.
-setDefaultTimeout(120_000);
-
 beforeAll(() => {
   process.env.PATH = `${join(REPO_ROOT, "tests", "fixtures", "bin")}${delimiter}${
     originalPath ?? ""
@@ -118,11 +114,12 @@ function releaseBinaryName(): string {
   return `aidlc-${targetTriple()}${process.platform === "win32" ? ".exe" : ""}`;
 }
 
+// Removing the accumulated temporary trees can exceed bun's 5s hook default.
 afterAll(() => {
   if (originalPath === undefined) delete process.env.PATH;
   else process.env.PATH = originalPath;
   for (const path of temporary) rmSync(path, { recursive: true, force: true });
-});
+}, 120_000);
 
 // Production emits canonical project and machine paths, so fixtures live under
 // the canonical temp root (macOS aliases /var to /private/var).
@@ -1313,7 +1310,7 @@ describe("t243 project initialization", () => {
     };
     expect(merged.mcpServers.context7).toEqual(custom);
     expect(merged.projectSetting).toBe(true);
-  });
+  }, 60_000);
 
   test("MCP consent and managed AGENTS blocks preserve user-owned configuration", () => {
     const claudeProject = temp("aidlc-t240-mcp-matrix-");
--- a/tests/unit/t243-install-mechanism.test.ts
+++ b/tests/unit/t243-install-mechanism.test.ts
@@ -1,7 +1,7 @@
 // covers: tool:aidlc-init, tool:aidlc-lifecycle, file:core/tools/aidlc-archive.ts
 // covers: file:core/tools/aidlc-transaction.ts, file:scripts/package.ts
 
-import { afterAll, beforeAll, describe, expect, test } from "bun:test";
+import { afterAll, beforeAll, describe, expect, setDefaultTimeout, test } from "bun:test";
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
 import {
@@ -103,6 +103,10 @@ const NEXT_VERSION = (() => {
 })();
 const temporary: string[] = [];
 const originalPath = process.env.PATH;
+
+// The default also governs afterAll removal of 128 staged roots, several of them
+// dist-sized, which exceeds bun's 5s hook default.
+setDefaultTimeout(120_000);
 
 beforeAll(() => {
   process.env.PATH = `${join(REPO_ROOT, "tests", "fixtures", "bin")}${delimiter}${
--- a/tests/unit/t248-steering-content-delivery.test.ts
+++ b/tests/unit/t248-steering-content-delivery.test.ts
@@ -4,7 +4,7 @@
 // bounded load-steering directives before run-stage; optional persona/knowledge
 // remains path-loaded with actionable warnings.
 
-import { afterAll, describe, expect, test } from "bun:test";
+import { afterAll, describe, expect, setDefaultTimeout, test } from "bun:test";
 import { spawnSync } from "node:child_process";
 import { createHash, createHmac } from "node:crypto";
 import {
@@ -49,6 +49,10 @@ const REVIEWER_AGENTS = [
   "aidlc-architecture-reviewer-agent",
   "aidlc-product-lead-agent",
 ] as const;
+
+// The default also governs afterAll removal of every staged project, which
+// exceeds bun's 5s hook default.
+setDefaultTimeout(120_000);
 
 type RuleContent = { path: string; text: string };
 type WireDirective = {
--- a/tests/unit/t248-steering-content-delivery.test.ts
+++ b/tests/unit/t248-steering-content-delivery.test.ts
@@ -4,7 +4,7 @@
 // bounded load-steering directives before run-stage; optional persona/knowledge
 // remains path-loaded with actionable warnings.
 
-import { afterAll, describe, expect, setDefaultTimeout, test } from "bun:test";
+import { afterAll, describe, expect, test } from "bun:test";
 import { spawnSync } from "node:child_process";
 import { createHash, createHmac } from "node:crypto";
 import {
@@ -50,10 +50,6 @@ const REVIEWER_AGENTS = [
   "aidlc-product-lead-agent",
 ] as const;
 
-// The default also governs afterAll removal of every staged project, which
-// exceeds bun's 5s hook default.
-setDefaultTimeout(120_000);
-
 type RuleContent = { path: string; text: string };
 type WireDirective = {
   kind: string;
@@ -88,9 +84,10 @@ function project(): string {
   return proj;
 }
 
+// Removing every staged project can exceed bun's 5s hook default under load.
 afterAll(() => {
   for (const proj of projects) cleanupTestProject(proj);
-});
+}, 120_000);
 
 function invoke(
   proj: string,


wowzoo (approved):
Thanks — this is a well-scoped fix with the diagnosis already done, and the two things you chose *not*
to do are as good as the change itself.

I reviewed the current head `a6a416bf` against `c03f9e28` (which is `main`'s tip, so no rebase is
needed) and reproduced both failures rather than taking the table on faith. Then I had a second pass
attack my own conclusions, which corrected two of my numbers and turned up a better justification for
your approach than either of us had written down.

## Both defects are real, and your `--parallel 8` framing was the right one

- **`t243-install-mechanism` fails deterministically even alone** — 84 pass / 1 fail, `a
  beforeEach/afterEach hook timed out for this test`. Instrumenting the `afterAll` gives **6635 ms**,
  and re-measurement puts it at 6464–6703 ms. Close to your 6680 ms isolated figure.
- **`t248-steering-content-delivery` needs load, exactly as your table says.** I first thought I had
  caught it in isolation, but a larger sample says otherwise: **20 sequential base runs, 20 pass / 0
  fail**, cleanup mean 3181 ms. Under load it is not flaky at all — it is deterministic. Base and head
  run 4 processes each, three rounds: **base (5 s default) 0 pass / 12 fail, head (120 s) 12 pass / 0
  fail**, every base failure being 42 passing tests followed by the hook timeout. So the file is fine
  at rest and reliably broken under the conditions the suite actually runs in. Worth putting that pair
  of numbers in the description — "0/12 under load, 20/20 alone" answers the "passes for me" objection
  more directly than a single parallel-8 timing does.

At head both files are green: 84/0 and 42/0, or 126 pass / 0 fail / 1547 expects in one invocation.

## `setDefaultTimeout` at file level is not just convenient — a narrower fix fails

Your note says the fix belongs in one call per file rather than a timeout argument on each hook. That is
correct, and it is worth stating as a measurement because the narrow alternative looks tempting. bun's
types do accept `number | { timeout }` on hooks, so I tried scoping it to the hook alone under the same
load:

- t248: 28/28 pass — the hook-only fix is sufficient there.
- **t243: 83/84.** `--force does not replace a pre-existing user-owned JSON entry` times out at
  **5023.20 ms** against the untouched 5000 ms *case* default.

So under load this file needs the case budget raised as well, which only the file-level call does.

I also checked what I might have asked for instead. At file-wide 30 s, t243 passes 84/0 under load — but
individual reported durations reach **31,756 ms** (then 26,415 / 23,859 / 20,124), so 30 s is not a
comfortable margin here, and I have no evidence that 120 s is more than necessary. Nothing to change.

And the placement is safe: raising the default during module evaluation applies to tests and hooks
already registered, so calling it before or after the first `describe` makes no difference — both of
yours are before it anyway. Neither file has a wall-clock assertion that a larger budget could disturb
(zero `Date.now` / `performance.now` / elapsed comparisons); the one
`expect(...).rejects.toThrow("timed out after 10ms")` at `t243:2300` is the implementation's own
timeout, and it passes at head.

## One thing to fix, non-blocking

**The comment this PR adds says 128 staged roots; the array holds 182.**

```ts
// The default also governs afterAll removal of 128 staged roots, several of them
// dist-sized, which exceeds bun's 5s hook default.
```

Instrumenting `temporary.length` at the top of the `afterAll` gives `roots=182 unique=182` on **every**
successful run — four independent observations across two machines, normal and loaded. The only runs
that come in lower are ones that fail early and never reach the later allocations (178 when six
`Bun.serve({ port: 0 })` cases hit `EADDRINUSE`). Since you verified on macOS too, it is not a platform
difference. Same number in the PR body. Purely cosmetic — but it is the one figure a future reader will
trust when deciding whether this budget is still right, so it is worth being accurate.

## Also non-blocking — what 120 s costs on a bad day

Raising the default raises the ceiling per *case*, not per file: a probe with three 250 ms tests at a
100 ms default fails 0/3 in 0.40 s wall, so budgets accumulate. If t243 ever went systematically wrong,
84 cases × 120 s is ~168 minutes against ~42 at 30 s, and the smoke+unit job has no explicit
`timeout-minutes` (`.github/workflows/ci.yml:55-82` — the 90-minute cap is on integration/e2e). I could
not construct a reachable failure that consumes all 84 (the shared spawn helper caps children at 60 s,
`t243:137-145`), and a normal run is ~193 s, so this is failure-diagnostic latency rather than a
correctness issue. Flagging it for whoever ends up setting the band in #1043 rather than asking you to
change anything.

## Two good calls

**Dropping the `t298` rewrite was right.** #1025 carries the same `>/dev/null 2>&1` fix, and I verified
in that review why it's the correct root cause: with the holder's stdout attached, `spawnSync` blocks
until the holder dies and stdout arrives as `PID\nREADY PID`, so `.pop()` picks `READY PID` and yields
`NaN`. Restoring the base implementation there fails the targeted test 0 pass / 1 fail.

**No version bump.** `AGENTS.md:56` is explicit that test-only changes do not bump, and the `Changelog
completeness` job only guards against *deleted* entries — I ran `bun scripts/ci-changelog-guard.ts`
against your base and it reports 276 entries preserved, 0 new, exit 0.

The scope honesty helps too: the `*-alias` residue you describe is real and pre-existing —
`t243:573-574` creates the symlink as a sibling of `machine` and never pushes it onto `temporary`, so
`rmSync` takes the target and leaves a dangling link. Correctly left alone.

## What I ran

At `a6a416bf` in a detached worktree, with `bun install --frozen-lockfile --ignore-scripts` (your
`package.json` and `bun.lock` are byte-identical to `main`) and `bun scripts/package.ts` first, since
`dist/` is generated now:

| check | result |
|---|---|
| head, both files together | 126 pass / 0 fail / 1547 expects |
| t243 with the fix reverted | 84 pass / **1 fail** (hook timeout), alone |
| t248 base, sequential ×20 | 20 pass / 0 fail — no isolated failure |
| t248 under 8-way load | base **0/12**, head **12/12** |
| t243 hook-only 30 s under load | **83/84** — a case times out at 5023 ms |
| t243 file-wide 30 s under load | 84/0, max individual duration 31,756 ms |
| instrumented `afterAll` | t243 6464–6703 ms / 182 roots · t248 mean 3181 ms / 35 projects |
| `bun run typecheck` / `bun run lint` / `git diff --check` | exit 0 / exit 0 / exit 0 |
| `bun scripts/ci-changelog-guard.ts c03f9e28` | exit 0 |

All instrumentation was temporary and reverted; the worktree is clean at your head.

One extra check, because it decides whether this fix is version-locked: the release pipeline pins bun
`1.3.14` in five places in `release.yml` while CI tests on `1.4.0`, so a `bun:test` behaviour that only
held on the newer runtime would not survive a release build. It holds on both — a file whose only
`afterAll` sleeps 7 s passes with the call and fails with the hook timeout without it, on **1.3.14 and
1.4.0 alike**. The negative control matters as much as the positive one: removing the call fails on
both, so the probe measures the mechanism rather than confirming itself.

## One thing that isn't yours

The PR looked half-checked because only the six `Pull Request Validation` jobs had run — those are
`pull_request_target` and always execute. The three `pull_request`-triggered workflows, including the
test suites, were parked at `action_required` because this is a first-time fork contribution, so the
test suite had never run on this branch in CI. I approved those runs, and this head is now **16/16
green** — `Tests (smoke + unit)` 24m53s, `Tests (integration + e2e, deterministic)` 4m1s, `Contract
checks (determinism + typecheck + lint)` 1m13s, plus `Changelog completeness`, both Gitleaks jobs, both
Semgrep jobs and `Bun audit`. Worth knowing for next time: the six validation checks never need that
approval, which is why the PR looked stuck.

Approving. Nothing here needs changing for correctness — the `128` in the comment and body is the only
thing I'd fix, and the load numbers are worth adding to the description so the next reader doesn't
repeat my mistake of testing it in isolation and doubting it.

It still won't merge immediately: the `default-branch` ruleset also requires an approving review from
the `aidlc-admins` team and I'm not a member, so my approval satisfies the general requirement but not
that one. Nothing further is needed from you.

wowzoo (approved):
Duplicate submission of the review above — identical content, posted twice by mistake. Please read the earlier one; nothing extra here.

wowzoo:
Correction to my review above: I wrote that CI tests on 1.4.0 while `release.yml` pins 1.3.14 — that has it backwards.

Current `main` pins 1.3.14 in all ten places: `ci.yml` ×4, `release.yml` ×5, `security-scanners.yml` ×1. Moving CI to 1.4.0 is what #1025 proposes; it isn't the state today.

The check itself stands, and the coverage is better than I described rather than worse: this PR's green CI ran on **1.3.14**, my local runs were on **1.4.0**, and the isolated probe confirms `setDefaultTimeout` governs hook budgets on **both** — with the negative control failing on both when the call is removed. So the fix is not version-locked either way, and it will still hold if #1025 lands and moves CI to 1.4.0.

One related thing worth recording while I'm correcting myself, because it affects how this is prioritised rather than whether it's right: I do not see this class of failure in Linux CI at all. #1025's most recent full smoke+unit run (273 files, 7797 assertions, on 1.4.0) contains **zero** `hook timed out` occurrences, and t243 and t248 both pass there; its single failure is an unrelated `t241` EPIPE. That is consistent with #1043 scoping this class to "macOS under load" and with your own "Verified on macOS" note. So this PR fixes a real failure that developers hit locally, not a red CI — which is still worth merging, just for a different reason than a reader might assume from my review.

Nothing changes in the approval.

apackeer (dismissed):
Requesting changes for one test-signal regression; the cleanup-timeout fix itself is valid.

## [P2] Keep the enlarged budgets explicit

`setDefaultTimeout(120_000)` in both files changes the timeout of every
unannotated test case, although the measured defect is accumulated `afterAll`
cleanup. Bun already supports a timeout argument on `afterAll`, and `t243`
already uses explicit 60s/120s budgets for cases known to be slow.

Please put `120_000` on the two cleanup hooks instead:

```ts
afterAll(() => {
  // cleanup
}, 120_000);
```

The earlier load probe identified one additional `t243` case,
`--force does not replace a pre-existing user-owned JSON entry`, crossing the
5s case default. Give that case its own explicit budget as well. This fixes the
observed load behavior without allowing every other unannotated case to run for
two minutes before failing.

I tested that narrower hook-specific form locally at `-P 8`: both changed files
passed, 126/126 assertions, with durations effectively identical to the PR
head. This was Linux rather than the reported macOS environment, so the
case-specific timeout should retain the macOS load coverage established in the
earlier review.

Locations:

- `tests/unit/t243-install-mechanism.test.ts:109`
- `tests/unit/t248-steering-content-delivery.test.ts:55`
- Bun's installed types distinguish hook options from the file-wide default at
  `bun-types/test.d.ts:359-403`

## [P3] Remove the fixed “128 staged roots” count

The `temporary` array records every `temp()` invocation, including invocations
inside loops and helper calls, so its runtime size is not the number of static
call sites and varies by executed paths. My run had already accumulated 159
live roots before cleanup; the earlier completed instrumentation reported 182.

Please replace the number in
`tests/unit/t243-install-mechanism.test.ts:107` with qualitative wording such
as “accumulated temporary trees”, and update the matching PR-description row.

apackeer:
@leandrodamascena could you review the scoped fix on
`fix/pr-1057-review-comments` at `6fe8d462e` against PR head `a6a416bfb`?

The PR head is contributor-owned, so I pushed the fix to a maintainer branch
rather than modifying Vivian's branch. The commit:

- scopes both 120-second budgets to the `afterAll` cleanup hooks;
- gives the demonstrated load-sensitive `--force` case an explicit 60-second
  budget; and
- removes the brittle fixed temporary-root count.

Verification: the targeted `t243` + `t248` unit slice passed 126/126 assertions
at `-P 8`; `bun run typecheck`, `bun run lint`, and `git diff --check` are clean.

vivibui:
Both points applied — I cherry-picked your `6fe8d462` onto the branch rather than
re-typing it, so the commit keeps your authorship. Head is now `c3f38e96`.

You were right on [P2], and for a reason stronger than the one in your comment:
**`t243` already carries 39 explicit `}, 60_000)` case budgets.** I matched the
repo-wide `setDefaultTimeout` convention and missed the file-local one, which is
the better signal. Net diff is now smaller than mine was: +5/−3.

## The macOS load coverage you asked me to retain

4 concurrent processes × 3 rounds, both files per invocation, macOS 26.6 / bun
1.4.2, at your `c3f38e96`:

| | result |
|---|---|
| **Hook timeouts across all 12 runs** | **zero** — your 120 s `afterAll` budget is sufficient |
| Rounds 1–2 (8 runs) | 125 pass / 1 fail |
| Round 3 (4 runs, load eased) | 126 pass / 0 fail |
| Sequential, unloaded | 126 pass / 0 fail / 1547 expects |

So the hook-scoped form holds on macOS. One case fails under load, and it is
**not** one either of us can fix with a timeout mechanism:

```
(fail) t243 release lifecycle > project pins require the retained OpenCode runtime
       selected by project metadata
  ^ this test timed out after 5000ms
```

8 of 12 runs, durations 5002.61 / 5002.66 / 5003.48 / 5004.78 / 5004.98 / 5008.56
/ 5009.40 / 5018.78 ms — clustered 3–19 ms over the limit, so it fails whenever
load nudges it across.

It is **pre-existing and unrelated to this PR**. At `t243:1900` the case sets its
own budget:

```ts
}, process.platform === "win32" ? 30_000 : 5_000);
```

An explicit per-case value overrides both the file-wide default and a hook
budget, so neither my original `setDefaultTimeout` nor your hook-scoped form ever
touched it. Confirmed by running the same 4-way load against unmodified
`origin/main` (`c03f9e28`): **4/4 fail** at 5004.40–5006.24 ms.

You authored that line in #756 and the non-Windows arm of this pattern is picked
per-test by expected cost (5 s / 10 s / 45 s / 60 s / 240 s across `t243` and
`t244`), so 5 s presumably means "this should be fast" — raising it is your call,
not something I'll slip into a cleanup-hook PR. Flagging it as data, and happy to
add it here if you'd rather it not wait.

## [P3] — and where 128 came from

Applied. Worth recording why the three numbers disagreed: **128 was the count of
static `temp()` call sites, not staged roots** — I measured one thing and labelled
it another. `fixtureRelease()` / `fixtureReleaseBytes()` each call `temp()`
internally and are invoked from many cases, so the runtime array exceeds the site
count, which is how you got 159 mid-run and @wowzoo got 182 on completed runs.
Your qualitative wording is the right fix; the number was never going to be
stable.

## Verification at `c3f38e96`

`bun test` both files 126/0/1547 · `bun run typecheck` exit 0 · `bun run lint`
exit 0 (736 files) · `git diff --check` exit 0. The cherry-pick touches only test
files, so `dist/` is unaffected.

vivibui:
Thanks for the scoped fix, @apackeer — cleaner than what I had, and the 39 existing per-case budgets in `t243` were exactly the detail I'd missed. I've cherry-picked `6fe8d462` so it keeps your authorship; head is now `c3f38e96`, and both [P2] and [P3] are in.

Two small things whenever you have a moment:

- GitHub still shows the review as *Changes requested* — it doesn't clear on its own, so a fresh look next time you're in here would sort it out.
- Pushing the new head reset `CI` / `Markdownlint` / `Security Scanners` to `action_required`, so the suite hasn't run on `c3f38e96` yet. No rush.

@wowzoo also mentioned the ruleset wants an `aidlc-admins` approval — any thoughts on who's best to loop in for that?

leandrodamascena (approved):
Thanks for working on this @vivibui! APPROVED

PR #1067, opened 2026-09-09
fix: allow native Plan Approval prerequisites before approval

### LINKED ISSUE #1062: [Bug]: Native install blocks a command needed for code-generation plan approval
### What happened

I’m using the native AI-DLC 2.8.0 release with Claude Code. During code generation, Step 2 asks me to run:

```sh
aidlc engine testing-posture render
```

The plan-approval guard blocks this command as potentially able to write files. That leaves the workflow stuck: I need its output to prepare the plan for approval, but the guard tells me to approve the plan before running it.

### Steps to reproduce

1. Install the native AI-DLC 2.8.0 release for Claude Code.
2. Start a small workflow:
   ```text
   /aidlc --scope express "a small HTTP API that returns hello world"
   ```
3. Approve Requirements Analysis and continue to code generation.
4. Run `aidlc engine testing-posture render`, as instructed in Step 2.

### Expected behaviour

The guard allows this prerequisite command before plan approval, while continuing to block implementation changes until approval is recorded.

### Actual behaviour

The guard reports:

```text
Code generation cannot run mutation-capable shell command: aidlc engine
testing-posture render for the zero-Unit stage-level implementation because
the plan, unit-test instructions, and current Testing Contract are
fingerprinted and approved. Writes inside the selected code-generation record
directory remain available for Steps 2-3. Record the human's explicit
"Approve Plan" answer before beginning Step 4 generation.
```

It asks me to record an “Approve Plan” answer, and describes the plan as already approved when nothing has been approved yet. Both make the failure confusing.

My audit log contains 10 `PLAN_APPROVAL_BLOCKED` entries during code generation. Other blocked commands included `bun --version` and `bun test src/app.test.ts` — these are caught because the guard decides what may write from a list of safe command names (`READ_ONLY_SHELL_COMMANDS`, L111-140), and `bun` is not on it, so every `bun` invocation is blocked regardless of its arguments. Writes inside the code-generation record folder were allowed (L734-738).

### Workaround

Running the same tool through its original Bun entry point worked:

```sh
bun .claude/tools/aidlc-testing-posture.ts render
```

I had to use this form for subsequent framework commands too. It works in this installation because the native release still includes `.claude/tools/`.

### Suspected cause

The native packager (`rewriteNativeInvocations()` in `scripts/package.ts`, L1065) rewrites framework commands from:

```text
bun <harnessDir>/tools/aidlc-<name>.ts
```

to:

```text
aidlc engine <name>
```

However, `isFrameworkToolInvocation()` in `core/hooks/aidlc-plan-approval-guard.ts` (L465-490) still recognises only the Bun form. The packaged stage instructions and the guard therefore disagree about how framework tools are invoked.

This looks less like a missed file than a structural mismatch, which is why I think it will recur. The rewrite is a find-and-replace over text, but the guard does not hold that text — it works the form out in code (`normalizedCommandName(name) !== "bun"`, then a path check against `<harnessDir>/tools/`). There is no string for the packager to find, so it never touched the hook. My installed copy is byte-for-byte identical to `core/hooks/aidlc-plan-approval-guard.ts`, and searching it for `aidlc` as a command name returns nothing.

### Suggested fix and regression coverage

Update the guard to recognise the native invocation form, including `aidlc.exe`, while preserving the existing restrictions on trusted framework tools. Matching it against the same delegate-name list `rewriteNativeInvocations` already uses (`scripts/package.ts:1072-1095`) would let the packager and the guard share one definition, so they cannot drift apart again.

It would also help to test the commands required before plan approval against both installation methods. Include negative cases to confirm that unrelated commands and implementation writes remain blocked.

### Environment

- AI-DLC: 2.8.0, native release
- Harness: Claude Code
- OS: macOS, Darwin 25.6.0
- Bun: 1.4.0
- Scope: `express`, with no Units

### Related issues

- #1047 introduced the native distribution change.
- #1031 also affects zero-Unit code generation, but appears to have a different cause.
- #1039 reports a similar command-classification problem with Kiro IDE.

leandrodamascena:
## Summary

- recognize the native `aidlc engine` Plan Approval prerequisites in the code-generation guard
- keep the exemption limited to read-only testing-posture verbs and the `code-generation` / `plan-approval` decision and answer receipts
- preserve blocking for unrelated engine and system routes, project test execution, local `./aidlc` binaries, PATH overrides, and non-Plan-Approval checkpoints
- correct the refusal text so it says approval is not current
- add native projection coverage for Claude Code, Codex, Copilot, Cursor, Kiro CLI, Kiro IDE, and opencode

## Root cause

The native packager rewrites the stage instructions from Bun tool entry points to `aidlc engine` commands. The guard still recognized only the Bun form, so it classified the required native commands as opaque mutation-capable shell invocations.

The fix classifies the native form by route and verb. It also checks the exact stage and checkpoint for the two log writes. Package parity tests pin the commands emitted by every supported native projection.

## Validation

- `bun test --timeout 30000 tests/unit/t265-plan-approval-guard.test.ts tests/unit/t68-version-changelog-sync.test.ts`
- `bun test --timeout 60000 tests/integration/t145-packaging-parity.test.ts`
- `bun run check`

Fixes #1062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.8.2] - 2026-09-09
+
+Restore the native Code Generation Plan Approval prerequisites that 2.8.0 rewrote to `aidlc engine` commands but the plan-approval guard still recognized only in their Bun form. **Upgrade:** run `aidlc update`, or use `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`. No project migration is required.
+
+* Native projects can run the required `testing-posture render`, `testing-posture fingerprint`, `log decision`, and `log answer` engine commands before Plan Approval on the seven supported harnesses, including the Windows `aidlc.exe` spelling. The guard still blocks generation, system routes, unrelated engine mutations, and arbitrary project test execution until approval is current.
+* Plan Approval mutation refusals now state that approval is not current instead of describing an unapproved plan as already approved.
+
 ## [2.8.1] - 2026-09-08
 
 Fix defects found while exercising the 2.8.0 native install: the guided `aidlc config` setup cancelled itself when Enter was pressed to accept a default, `aidlc update` on an already-current install failed its integrity check under a normal shell umask, and every native GitHub Copilot and Cursor hook was dead because the 2.8.0 packager projected those adapters onto the one-argument core-hook route. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.1` / `install.ps1 -Version 2.8.1`. Copilot and Cursor projects configured by 2.8.0 work as soon as the binary is updated: their existing `aidlc engine hook <harness>-adapter ...` wiring is accepted, and the 2.8.0 Copilot adapter still installed in the project (whose core-hook calls are the bare `aidlc hook <name>`) is accepted too. `aidlc config` in the project then rewrites the wiring to the canonical `aidlc engine adapter <harness> ...` spelling and installs the current adapter — Cursor's merged `.cursor/hooks.json` collapses every earlier AI-DLC spelling of an entry (bun-era and 2.8.0) into the one shipped entry instead of leaving duplicates that keep executing, and Copilot's `.github/hooks/aidlc.json` is regenerated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,6 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [2.8.2] - 2026-09-09
-
-Restore the native Code Generation Plan Approval prerequisites that 2.8.0 rewrote to `aidlc engine` commands but the plan-approval guard still recognized only in their Bun form. **Upgrade:** run `aidlc update`, or use `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`. No project migration is required.
-
-* Native projects can run the required `testing-posture render`, `testing-posture fingerprint`, `log decision`, and `log answer` engine commands before Plan Approval on the seven supported harnesses, including the Windows `aidlc.exe` spelling. The guard still blocks generation, system routes, unrelated engine mutations, and arbitrary project test execution until approval is current.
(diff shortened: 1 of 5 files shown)


apackeer:
I found three blocking issues in the current head:

1. **PATH changes in an assignment-only shell segment bypass the guard.** `core/hooks/aidlc-plan-approval-guard.ts:506,657` checks `executableResolutionChanged` only on the parsed invocation. For `PATH=.; aidlc engine testing-posture render`, the shared parser detects the aggregate resolution change, but the assignment-only segment produces no invocation and the later `aidlc` invocation carries no flag. With no write targets, the guard allows the command, so a project-local fake `aidlc` can perform arbitrary writes before Plan Approval. Please reject commands when `shellCommandAltersExecutableResolution(command)` is true (or preserve that state across segments) and add this exact regression case alongside the existing same-segment PATH tests.

2. **Data-driven wrappers can change the trusted route after inspection.** The new `shellInvocationNeedsApproval` shape at `core/hooks/aidlc-plan-approval-guard.ts:549-554` ignores the parser's `dataDriven` marker. For example, `xargs aidlc engine log decision --stage code-generation --checkpoint plan-approval` is classified from its static arguments, but `xargs` can append a later `--checkpoint` or `--project-dir`. `aidlc-log.ts:137-152` uses the last flag value, so this escapes the intended Plan-Approval-only boundary. Please fail closed for data-driven native exceptions and cover the override case.

3. **Release metadata was removed from this user-visible fix.** The PR head still reports `2.8.1` in `core/tools/aidlc-version.ts:4`, `README.md:8`, and `CHANGELOG.md:4`. The repository policy requires the matching version bump and changelog entry in the same user-visible PR.

The PR is also currently conflicting with `main`, so these findings should be addressed while rebasing the guard and test changes.

apackeer (approved):
Conditional approval: please address all three findings from my review comment before merge:

- block executable-resolution changes across assignment-only shell segments, including `PATH=.; aidlc ...`;
- fail closed for data-driven native invocations such as `xargs` that can append route-changing flags;
- restore the required version, README badge, and changelog updates.

Please also rebase and resolve the current conflict with `main`. This approval is contingent on those corrections remaining covered by regression tests.

PR #1065, opened 2026-09-09
fix: route native Copilot and Cursor hooks through the adapter dispatcher

apackeer:
Closes #1061. Closes #1058 (same root cause).

## Problem

The 2.8.0 native channel (`scripts/package.ts` `rewriteNativeInvocations`) rewrites `bun <dir>/hooks/aidlc-<name>.ts` into `aidlc engine hook <name>`, special-casing only the kiro and codex adapters onto `aidlc engine adapter <harness> <target>`. The Copilot and Cursor adapters fell through to the one-argument `hook` route, which ignores the target and calls `run(stdin)`:

- **Copilot** (#1061): `run(target, input)` receives `input === undefined` and crashes on every event with `aidlc: undefined is not an object (evaluating 'input.length')`. The whole hook layer was dead; `aidlc doctor` was green because it never exercises the dispatcher.
- **Cursor** (#1058): `target` is the raw stdin, matches no `case`, exits 0 with no stdout behind a `failClosed: true` `preToolUse`, so Cursor blocked every tool.

Two more defects sat underneath and would have kept Copilot dead even with the wiring corrected:

- `aidlc engine adapter copilot` did not exist (`AdapterHarness` in `core/tools/aidlc.ts` had no copilot entry) → `unknown verb 'copilot'`.
- The Copilot adapter spawned its core hooks as `[executable, "hook", <name>]` under the compiled binary. Bare `hook` is not a dispatcher command (exit 2); stderr is ignored and hooks fail open, so audit, sensors, guards, and Stop enforcement silently no-oped.
- `runAdapter` restored `AIDLC_HARNESS_DIR` but left the startup-pinned `AIDLC_HARNESS_NAME` in place, so children could resolve a foreign packaged runtime when the hook cwd carried no harness metadata (`.aidlc` is shared by copilot and opencode).

## Change

- `scripts/package.ts`: the four authored adapters (`kiro|codex|cursor|copilot-adapter`, an explicit set) rewrite to `aidlc engine adapter <harness>` (plain and escaped-JSON forms); any other hook file keeps the generic `engine hook <name>` rewrite.
[...]

PR #1068, opened 2026-09-09
test: shard unit CI across isolated jobs

### LINKED ISSUE #1066: [Feature]: Shard smoke and unit CI tests without reducing coverage
## Description

Split the deterministic smoke and unit CI gate into isolated serial shards. Preserve the existing `Tests (smoke + unit)` required check as an aggregate gate.

## Use Case

The current main-line suite contains 285 test files and 8,097 assertions. The latest measured successful run spent 20 minutes and 30 seconds in smoke and unit, while deterministic integration and e2e completed in 3 minutes and 55 seconds.

Running unit files concurrently in one checkout is unsafe. A diagnostic `--parallel 8` run reproduced missing `dist/` files, fixed five-second timeout failures, and a hidden ordering dependency between `t238-build-binaries.test.ts` and `t249-copilot-adapter.test.ts`.

## Version

v2 (current on main)

## Area

Other: CI and testing

## Proposed approach

- Run smoke in its own job.
- Split unit tests across four isolated GitHub Actions jobs.
- Keep each shard serial inside its own checkout.
- Balance shards using measured test durations.
- Keep the compiled-binary producer and consumer in the same ordered shard, or provision the compiled binary explicitly.
- Preserve the required `Tests (smoke + unit)` name through an aggregate job.

## Acceptance criteria

- Every discovered smoke and unit file runs exactly once.
- No test is moved to a weaker tier, skipped, or given a larger timeout to make the change pass.
- The aggregate assertion count matches the unsharded baseline for the same commit.
- Shard assignment is deterministic and rejects invalid shard arguments.
- Documentation describes local serial execution and CI sharding.
- The smoke and unit gate completes within eight minutes on standard GitHub-hosted runners, excluding queue time.

## Alternatives considered

Running unit files concurrently in one checkout was rejected because repository-level packaging tests mutate `dist/` and CPU contention trips existing per-test timeouts.

leandrodamascena:
Closes #1066

## Summary

- Split smoke into its own CI job.
- Run unit tests in four isolated GitHub Actions jobs while keeping each shard serial.
- Assign every unit file exactly once using measured duration weights.
- Keep the native binary producer and compiled Copilot adapter consumer in the same ordered shard.
- Reject shard counts that create an empty shard.
- Require compiled coverage in sharded unit runs instead of silently skipping it.
- Preserve the required `Tests (smoke + unit)` check through an aggregate job.

## Why

The measured smoke and unit gate took 20 minutes and 30 seconds for 285 files and 8,097 assertions. Direct `--parallel 8` execution in one checkout reproduced `dist/` races, fixed-timeout failures, and skipped compiled coverage. Isolated weighted shards reduce wall time without weakening coverage or running packaging tests concurrently in one tree.

## Review fixes

- `t249` now reads the producer output `build/binaries/build-results-native.json`.
- Sharded unit execution sets `AIDLC_REQUIRE_COMPILED_COVERAGE=1`, so a missing native artifact fails the suite.
- `assignWeightedShards` rejects `M` above the number of assignable groups, and the runner exits 2 without reporting `RESULT: PASS`.
- The t249 weight was recalibrated after enabling the compiled cases.
- The Copilot native dispatcher correction is already present in the current base through #1065.

## Validation

- `t238-build-binaries`: 4 passed, 0 failed.
- `t249-copilot-adapter` with compiled coverage required: 58 passed, 0 failed, 0 skipped, including 11a, 21b, and 26b.
- Focused smoke, version sync, and Copilot packaging tests: 41 passed, 0 failed.
- `bun scripts/package.ts --check`.
- `bun run typecheck`.
- `bun run lint`.
- `git diff --check`.

No version or changelog bump. This PR changes CI, test infrastructure, and test documentation only.

## Checklist

- [x] I have reviewed the [contributing guidelines](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of this change
- [x] Changes have been tested
- [x] Changes are documented

## Acknowledgment

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,8 +52,8 @@ jobs:
       - name: Run contract checks
         run: bun run check
 
-  test:
-    name: Tests (smoke + unit)
+  test_smoke:
+    name: Tests (smoke)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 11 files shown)


apackeer (changes requested):
Two blocking findings:

1. **Compiled Copilot coverage is still silently skipped.**

   `tests/unit-shard-weights.json:22` places `t238-build-binaries.test.ts` before `t249-copilot-adapter.test.ts`, but ordering alone does not restore the compiled cases. `scripts/build-binaries.ts:2293` writes `build-results-native.json`, while `tests/unit/t249-copilot-adapter.test.ts:181` only reads `build-results.json`. Consequently `COMPILED_BINARY` remains null and both compiled cases at lines 1353 and 2191 skip. The current green `Tests (unit shard 1/4)` log confirms `54 pass`, `2 skip`, `0 fail`.

   Please make the producer/consumer contract agree and add a guard that proves those compiled cases execute rather than merely checking that the two files share a shard.

2. **A valid shard specification can produce a zero-test green run.**

   `tests/lib/test-sharding.ts:67` accepts any positive shard count. With 271 unit files and the two-file affinity group, there are only 270 assignable groups, so:

   ```text
   bash tests/run-tests.sh --debug -P 8 --unit --shard 271/271
   ```

   exits 0 with `Test files: 0` and `RESULT: PASS`. Please reject a shard configuration whose selected shard is empty, and cover this boundary in the runner tests.

apackeer (approved):
Approved. The previously requested fixes are addressed.

PR #1077, opened 2026-09-09
fix: a thematic break before Assumption Confirmation no longer voids the summary receipt

### LINKED ISSUE #1074: [Bug]: a --- before the sanctioned "Assumption Confirmation" section changes the confirmed-content digest and refuses stage completion
### Description

The consolidated-summary receipt binds to a **scoped** digest (`Hash Scope: confirmed-content-v1`) whose whole purpose is to let a stage append sanctioned sections after the summary without invalidating the confirmation. The parser explicitly excludes the `## Assumption Confirmation` section from that digest. But the exclusion range starts **at the heading line** — so a Markdown thematic break (`---`) placed *between* the summary and the `## Assumption Confirmation` heading is **not** excluded, lands in the confirmed-content body, and changes the digest. The `.trimEnd()` applied to the assembled body strips trailing whitespace but does not strip a `---`. On the next `report --result approved`, the summary-confirmation guard recomputes the hash, finds a mismatch, and refuses the stage with `... changed after the human confirmed its summary`.

The sanctioned shape (append `## Assumption Confirmation` after the summary) is exactly what triggers the failure when the author separates it with a horizontal rule — a near-universal Markdown convention for setting off an appended block.

### Steps to Reproduce

1. In a stage that carries an assumption gate, get the consolidated summary confirmed so a `confirmed-content-v1` receipt is recorded.
2. Append the sanctioned `## Assumption Confirmation` section, separating it from the summary with a `---` thematic break (blank lines above and below it).
3. Run `aidlc-orchestrate.ts report --stage <slug> --result approved`.
4. Observe the refusal.

### Expected Behavior

Appending the sanctioned `## Assumption Confirmation` section — with or without a leading `---` — leaves the confirmed-content digest unchanged, because a thematic break is presentation, not confirmed semantic content. The stage completes.

### Actual Behavior

Stage completion is refused:

```
Refusing to complete "<slug>": <questions>.md changed after the human confirmed its summary. ...
```

- The digest excludes only the `## Assumption Confirmation` **section body**: the exclusion range opens at the heading line (`core/tools/aidlc-lib.ts:7637`, `openExcludedAssumption = line`), so a `---` on a line before that heading is never in any excluded range.
- The confirmed body is assembled at `core/tools/aidlc-lib.ts:7686-7689` (`lines.filter(!excluded).join("\n").trimEnd()`). `.trimEnd()` at `:7689` strips trailing whitespace but not a `---` — the break is content, so the SHA-256 at `:7690-7692` changes.
- The guard at `core/tools/aidlc-lib.ts:8593-8601` compares `receipt.block "Questions SHA-256"` against the recomputed `currentHash` (`:8578-8584`) and returns `SUMMARY_CONTENT_STALE` on mismatch.

### AI-DLC Version
v2 (current on main)

### Release / Commit
2.8.1 (`52da70ad`)

### Phase
Not phase-specific (any stage carrying an assumption gate — e.g. Inception requirements/design stages)

### Harness
Claude Code (defect is harness-neutral; lives in `core/`)

### AI Model
Reproduced originally under Claude; the digest logic is model-independent.

### Additional Context

The current recovery is expensive for a formatting artifact: re-present the consolidated summary and take a fresh confirmation so a new receipt records the current hash — a full re-confirmation for inserting a `---`.

**Proposed fix (PR offered against `main`):** exclude a trailing thematic break (a line matching `^---+$` / `^\*\*\*+$` / `^___+$`) that immediately precedes an excluded `## Assumption Confirmation` heading from the confirmed-content body — or, equivalently, extend the exclusion range backward over blank lines and a single thematic break adjacent to the excluded heading. Small, localized to `summaryConfirmationContentHash`, with a regression test that appends `---\n\n## Assumption Confirmation` after a confirmed summary and asserts the digest is unchanged.

*Migration note:* the fix keeps the same hash-scope name `confirmed-content-v1`, so a receipt recorded over a body that already contained a `---` *before* the fix would mismatch *after* it — a rare case, recovered by one re-confirmation. Worth a changelog line so a maintainer isn't surprised.

tejasavora:
## Summary

Appending `## Assumption Confirmation` after a confirmed summary can include a blank-delimited Markdown thematic break (`---`, `***`, or `___`) without invalidating the summary receipt and refusing stage completion with `SUMMARY_CONTENT_STALE`.

The exclusion scan uses the original normalized source lines, so it crosses only actual blank lines before an adjacent divider. Fenced code, indented code, and preformatted HTML between the divider and the assumption heading remain confirmed content: changing a JSON requirement such as `"tls_required":true` to `false` still requires a fresh confirmation. Setext headings remain invalid after the consolidated summary, and follow-up questions remain covered by the receipt.

Closes #1074.

## Compatibility

The hash scope remains `confirmed-content-v1`. A receipt recorded before this change over a body that already contained the newly excluded divider may need one fresh confirmation after upgrading. At the maintainer’s request, the version remains **2.8.2** after rebasing onto #1079: the version source, README badge, and current changelog heading agree. Both PRs’ release notes are preserved. The unpublished 2.8.6 development entry is explicitly marked superseded and retained as history.

## Validation

Verified on `ddde5285ece6d95b85eba1d0c412c951c4befed7`, based on `main` at `0c12f99ba6ce97485f77670e809e9fd5c57d4edb`:

- The repository test runner, with `--debug -P 8 --unit --integration --no-llm` and a filename filter, reran the receipt/version suites and the pre-existing CI failure. The receipt/version suites executed 32 tests with no failures or skips. Full output and per-file results were captured.
- `t332-summary-authorization.test.ts`: 25 tests, including a real receipt surviving a divider-only append, real receipts rejecting substantive edits in all three code formats, follow-up answer coverage, supported divider forms, CRLF normalization, and setext rejection.
- `t68-version-changelog-sync.test.ts`: 7 tests, including the shipped CLI reporting `2.8.2`.
- `bun scripts/ci-changelog-guard.ts origin/main`: all 278 existing changelog entries preserved.
- `t328-authority-rebinding.test.ts`: 19 pass, 3 fail in parking/resuming and Unit lifecycle approval checks. The identical three failures are present in [the already-merged #1079 CI run](https://github.com/awslabs/aidlc-workflows/actions/runs/34400168953/job/102630982764); they are not introduced by this PR.
- `bun scripts/package.ts --check`: deterministic output across all seven harnesses.
- `bun run typecheck` and `bun run lint`: passed.

Generated distributions were rebuilt for validation and are not committed.

## Acknowledgment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,17 @@ All notable changes to this project will be documented in this file.
 
 ## [2.8.2] - 2026-09-10
 
-Improve review-findings table diagnostics so a missing cell no longer produces a misleading status error or guesses which column was omitted. Malformed rows report their cell count and the expected column order, with a repair hint when the last cell looks like a shifted status. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`. No migration is required.
+Preserve summary confirmations when an Assumption Confirmation section is appended with a decorative divider, and improve review-findings table diagnostics so malformed rows report their cell count and expected column order without guessing which column was omitted. The intended development release version is 2.8.2. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`. A summary receipt recorded before this fix over a body that already contained the newly excluded divider may need one fresh confirmation after upgrading; no other migration is required.
 
 * Short review-findings rows show the expected columns and, when applicable, suggest checking earlier cells for a missing value or `|` separator. Review completion remains refused until the row is corrected. Closes #1076.
 * Review-findings rows with surplus cells report the extra count. Well-formed rows, including escaped pipes and explicit blank cells, retain their existing behavior.
+* `aidlc version` reports `2.8.2`; the unpublished `2.8.6` development version is superseded.
+* Appending `## Assumption Confirmation` after a confirmed summary with an adjacent, blank-delimited `---`, `***`, or `___` no longer causes `SUMMARY_CONTENT_STALE` during stage completion. Code examples before the assumption heading and follow-up answers remain covered by the receipt, so substantive edits still require confirmation. Closes #1074.
+* Content appended under a `## ` heading is separated from the following `## ` heading by a blank line, keeping method files well-formed. Closes #1075.
 
 ## [2.8.6] - 2026-09-09
+
+**Superseded development entry:** No 2.8.6 release was published. The intended release version is 2.8.2, documented above; this entry is retained as development history.
 
 Fix a markdown-hygiene defect in the shared section writer: `appendUnderHeading` inserted new content flush against the following `## ` heading, so a bullet written under a section (for example a self-learning entry under `## Decided` in `project.md`) abutted the next heading with no separating blank line. The engine re-reads `project.md` every stage and the file is human-editable, so the malformed markdown could misgroup for a re-reading model or a stricter markdown tool. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.6` / `install.ps1 -Version 2.8.6`. No migration is required; existing files are corrected the next time content is appended to an affected section.
 
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -7580,6 +7580,27 @@ function visibleHeading(
 // Hash the normalized semantic questions content the human confirmed. The
 // shared protocol does not impose names on pre-checkpoint sections, while
 // follow-up Q<n> sections after an assumption decision remain hashable.
+// A thematic break (`---`, `***`, `___`) placed before the sanctioned
+// `## Assumption Confirmation` heading is presentation, not confirmed content.
+// It must fall inside the excluded range so appending that section — with or
+// without a leading rule — leaves the confirmed-content digest unchanged.
+const ASSUMPTION_THEMATIC_BREAK = /^ {0,3}(?:-{3,}|\*{3,}|_{3,})\s*$/;
+
+function assumptionExclusionStart(lines: string[], headingLine: number): number {
+  let i = headingLine - 1;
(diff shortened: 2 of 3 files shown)


apackeer:
Thanks @tejasavora for the fix and the clear reproduction in #1074. This is a good direction: a decorative divider before the sanctioned Assumption Confirmation section should not force another summary confirmation. Maintainers will take over rebasing and merging this PR from here; no further action is needed from you.

The review found one issue to fix before merge:

- **[P1] Preserve substantive content before the assumption heading.** At [`core/tools/aidlc-lib.ts:7658`](https://github.com/awslabs/aidlc-workflows/blob/cd87f52c9c8993327ab4d8699398fe13fc27b26e/core/tools/aidlc-lib.ts#L7658), `assumptionExclusionStart` receives `visibleLines`, which replaces code blocks with empty lines for heading detection. If a blank-delimited `---` is followed by a fenced JSON block and then `## Assumption Confirmation`, the backward scan crosses and excludes that JSON block too. Changing `"tls_required":true` to `false` leaves the digest unchanged. Indented code and `<pre>` blocks reproduce the same problem. A test using a real summary receipt confirmed that the completion validator still accepts the changed file; the merge-base implementation detects the change.

We'll restrict the scan to actual blank lines in the original normalized source and add regression coverage, preserving the intended formatting tolerance without allowing substantive changes to bypass confirmation.

apackeer (approved):
Reviewed the rebased head ddde5285ece6d95b85eba1d0c412c951c4befed7 on main at 0c12f99ba6ce97485f77670e809e9fd5c57d4edb (#1079). The two code commits are unchanged by the rebase; the changelog conflict preserves both sets of 2.8.2 notes. The exclusion scan uses raw normalized source lines, so code/configuration changes remain covered by the summary receipt.

The 25 summary-authorization and 7 version tests pass without skips. Type checking, lint, packaging determinism, and changelog preservation pass. The three t328 authority-rebinding failures reproduce identically in the already-merged #1079 CI run; this existing baseline issue is documented in the PR validation section.

Approved for the requested merge.

PR #1079, opened 2026-09-09
fix: diagnose malformed review findings rows without guessing columns

### LINKED ISSUE #1076: [Bug]: reviewer findings-table parser is positional with no arity check — a short row yields a misdirecting "invalid finding status \"\"" that names the wrong cause
### Description

The review-findings table is parsed **positionally** with no row-arity validation. A row missing a column (e.g. omitting `Required action`) has every later cell shift left, so `Status` resolves to `""` and the parser throws `... invalid finding status ""`. The message names the invalid *status* rather than the real cause — a **missing column** — so it points a reader away from the fix.

### Scope (what this is and is NOT)

This is a **diagnosability defect**, not a workflow wedge. On the current **record-era** review path a malformed table is caught safely: `parseReviewSection` is wrapped in try/catch and a parse failure **refuses `REVIEW_COMPLETED`** (`core/tools/aidlc-log.ts:2406-2417`), so no terminal receipt arms over an unparseable table, and the gate presentation reads the already-structured `record.findings` (`aidlc-lib.ts:11138`) rather than re-parsing raw markdown. The unguarded throw only reaches a user on the **legacy `## Review` appendix** path (`aidlc-review-brief.ts:148`, "read for migration only"). So the harm is: the refusal/throw **misdirects** (wrong cause named), and there is no arity check to produce an actionable message.

### Steps to Reproduce

1. Author a review-findings table whose header declares six columns (`ID`, `Severity`, `Location`, `Finding`, `Required action`, `Status`).
2. Write a findings row with only five content cells (omit `Required action`).
3. Have it parsed (record-era: `REVIEW_COMPLETED` is refused; legacy appendix: the parser throws).
4. Observe the message is `... invalid finding status ""` — naming the status, not the missing `Required action` column.

### Expected Behavior

The parser validates each row's arity against the header **before** the positional read and, on mismatch, names the offending column — e.g. `row R-05 has 5 cells, header declares 6: missing "Required action"`. The record-era refusal then carries that actionable reason instead of `invalid finding status ""`.

### Actual Behavior

- `splitMarkdownRow` (`core/tools/aidlc-lib.ts:10560`) splits a row into cells with **no padding** to header width.
- Header→index map at `:10642`: `const index = new Map(headers.map((name, position) => [name, position]));`
- Positional read at `:10646-10647` returns `""` for any shifted/missing cell:
  ```ts
  const value = (name: string): string =>
    cells[index.get(name) ?? -1]?.trim() ?? "";
  ```
- `:10652-10656` throws on the empty status:
  ```ts
  const status = value("Status");
  if (!validReviewFindingStatus(status)) {
    throw new Error(`${artifact}#${id}: invalid finding status ${JSON.stringify(status)}`);
  }
  ```
- No `cells.length` vs `headers.length` check anywhere in the loop.

### AI-DLC Version
v2 (current on main)

### Release / Commit
2.8.1 (`52da70ad`)

### Phase
Construction (reviewer-backed approval gate)

### Harness
Claude Code — source-confirmed; defect is harness-neutral (`core/`).

### AI Model
Claude Opus 4.8

### Additional Context

Shares the family root cause "the framework does not check its own contracts for self-consistency," and the misdirecting message also belongs to the misleading-failure class. Kept deliberately narrow: the current record-era path already fails safe (refuses completion), so this is about a **wrong error message + a missing arity guard**, not an unrecoverable gate.

**Proposed fix (PR offered against `main`):** validate row arity against the header before the positional read and name the offending column; the record-era refusal then surfaces that reason. Localized to the brief/section parser, with a regression test for a short row. (`t304` does not pin the current message, so the wording change is safe.)

tejasavora:
## Summary

A findings row with too few cells previously failed with `invalid finding status ""`, even when a value such as `New` was present and an earlier cell had been omitted. Validate the row width before reading its values, report the expected column order, and give a conditional repair hint without claiming to know which field was omitted.

For example:

```text
requirements.md#R-05: row has 5 cells, header declares 6.
Expected columns: ID | Severity | Location | Finding | Required action | Status.
The last cell "New" looks like Status; check earlier cells for a missing value or "|" separator.
```

The trailing-status hint is shown only when `Status` is the final declared column and the last cell contains a recognized status. Other short rows receive a general missing-cell or separator hint; long rows report their surplus cells. Malformed reviews remain refused before `REVIEW_COMPLETED` is recorded. Valid tables, including escaped pipes and explicit blank cells, retain their existing behavior.

Closes #1076.

## Validation

- `bun run check` — projection generation, two-build packaging determinism, all three TypeScript configurations, and Biome passed.
- Focused `t304-review-brief` and `t68-version-changelog-sync` unit slice with `--debug -P 8 --no-llm` — 39 passed, 0 failed.
- TypeScript and Biome checks passed again after the final test assertion change.
- After setting the release target to `2.8.2`, `t68` passed all 7 checks, packaging remained deterministic, and the changelog guard confirmed that all existing entries were preserved.

## Checklist

- [x] I have reviewed the contributing guidelines.
- [x] I have performed a self-review of this change.
- [x] Regression coverage includes shifted statuses, missing status values, reordered headers, extra cells, escaped pipes, explicit blank cells, and refusal to record a malformed review.
- [x] No fingerprint, epoch, or receipt identity inputs change.
- [x] The version, README badge, and latest changelog entry target `2.8.2`; existing changelog entries are preserved.

## Acknowledgment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [2.8.2] - 2026-09-10
+
+Improve review-findings table diagnostics so a missing cell no longer produces a misleading status error or guesses which column was omitted. Malformed rows report their cell count and the expected column order, with a repair hint when the last cell looks like a shifted status. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.2` / `install.ps1 -Version 2.8.2`. No migration is required.
+
+* Short review-findings rows show the expected columns and, when applicable, suggest checking earlier cells for a missing value or `|` separator. Review completion remains refused until the row is corrected. Closes #1076.
+* Review-findings rows with surplus cells report the extra count. Well-formed rows, including escaped pipes and explicit blank cells, retain their existing behavior.
+
 ## [2.8.6] - 2026-09-09
 
 Fix a markdown-hygiene defect in the shared section writer: `appendUnderHeading` inserted new content flush against the following `## ` heading, so a bullet written under a section (for example a self-learning entry under `## Decided` in `project.md`) abutted the next heading with no separating blank line. The engine re-reads `project.md` every stage and the file is human-editable, so the malformed markdown could misgroup for a re-reading model or a stricter markdown tool. **Upgrade:** `aidlc update`, or `install.sh --version 2.8.6` / `install.ps1 -Version 2.8.6`. No migration is required; existing files are corrected the next time content is appended to an affected section.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ structured, verifiable software-delivery workflows. One harness-neutral core
 runs natively in Claude Code, Kiro CLI, Kiro IDE, Codex CLI, Cursor, opencode,
 and GitHub Copilot.
 
-![version](https://img.shields.io/badge/version-2.8.6-blue)
+![version](https://img.shields.io/badge/version-2.8.2-blue)
 ![license](https://img.shields.io/badge/license-MIT--0-green)
 
 The Quick Start below installs the latest stable AI-DLC release.
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -10643,6 +10643,28 @@ export function parseReviewSection(
   const findings: ReviewFinding[] = [];
   for (const line of table.slice(2)) {
     const cells = splitMarkdownRow(line);
+    // Validate row arity against the header before reading positionally. A row
+    // with the wrong cell count shifts every later column, so a positional read
+    // would silently return "" for a missing cell and later throw a misdirecting
+    // "invalid finding status" that names the wrong cause. Name the offending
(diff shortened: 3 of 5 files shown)


apackeer (approved):
Reviewed the row-width diagnostics, conditional trailing-status hint, and refusal to record malformed findings at 7755eda75d36a92b4016ecbef623104183bf1c23. The focused regression and version checks passed; the rebased commit tree matches the validated candidate. Version is 2.8.2.

CI contract checks, smoke tests, all four unit shards, Markdownlint, and Security Scanners passed. The deterministic integration/e2e job failed only in three existing t328-authority-rebinding scenarios: parking/resuming, Unit start, and Unit pause/completion. An isolated rerun on this PR and an independent checkout of unchanged main (6af101a0e9920733c803ef21c804431f983ae094) each produced 19 passing tests and the same 3 failures. This establishes that the failure predates this PR.

CI failure log: https://github.com/awslabs/aidlc-workflows/actions/runs/34400168953/job/102630982764

PR #1080, opened 2026-09-09
fix: treat a piped no-operand grep/rg as reading stdin, not a recursive search of "."

### LINKED ISSUE #1073: [Bug]: reviewer read-scope hook refuses a piped grep (opens no file) as a recursive search of ".", while the same-position cat is allowed
### Description

The reviewer read-scope hook judges a `grep` with **no file operand** as a recursive search rooted at `.`. But no-file-operand is exactly what the **downstream side of a pipe** looks like: that `grep` reads **stdin** and opens **no file at all**. Because `.` as a search root spans every sibling unit, the hook refuses it — recording `.` as the target, a path the reviewer never typed.

So a command that accesses **zero files** is refused as a project-wide sweep, while the unpiped form against the same in-scope file succeeds. The ambiguity is real (a static reader cannot distinguish `grep pat` reading stdin from `grep pat` recursing in some tools), but it is resolved toward the blocking interpretation, and the safe reading is the far more common one in a pipeline.

The framework already gets the adjacent case right one branch away: `cat`/`less`/`more`/`head`/`tail` are judged in `target` mode, where the no-operand default never fires, so `… | cat` is allowed and `… | grep` is blocked **from the same pipeline position in the same dispatch**. That asymmetry is unintended by construction.

### Steps to Reproduce

1. Under an active reviewer dispatch, run a pipeline whose later segment is a `grep` with no file operand, e.g. `grep pat components.md | grep other`.
2. Observe the first segment (carries an in-scope operand) is allowed and the second segment (no operand) is refused, with `.` recorded as the target.
3. Compare with `… | cat`, which is allowed from the same position.

### Expected Behavior

A search command with no path operand that is **not the first segment of a pipeline** is treated as reading **stdin** (opens no file) and is allowed. A first-segment no-operand search may keep the conservative `.` reading. When the target is a default rather than a typed token, the refusal message should say so.

### Actual Behavior

- `judgeGrepLike` at `core/hooks/aidlc-reviewer-scope.ts:489` defaults a no-operand grep to a recursive search of `.`:
  ```ts
  if (!rootSeen) return judgePathAccess(".", "search-root", scope, bases);
  ```
- `judgeRipgrep` at `:528` (`rg`/`ripgrep`) is `if (!rootSeen && !constrainedToCurrent) return judgePathAccess(".", "search-root", …)` — the added `!constrainedToCurrent` guard only spares the case where a `-g`/`--glob` constrains scope; a bare no-glob `rg pat` reading stdin still defaults to `.`.
- The dispatch loop `judgeCommandText` iterates `for (const segment of shellSegments(text))` (`:594`) but passes **no pipeline-segment index** to the judges, so a judge cannot tell a downstream (stdin) segment from the first segment.
- Contrast: `judgeSimpleFileCommand` (`cat`/`head`/…) only defaults to `.` when `mode === "search-root"` (`:574`), and those commands are judged in `target` mode — so `… | cat` is allowed from the same position.

### AI-DLC Version

### Phase
Construction (reviewer dispatch)

### Harness / Model
Reviewer-observed on Claude Code; source-confirmed. Defect is harness-neutral (`core/`).

### Additional Context

Shared root cause across this family: **the framework does not check its own contracts for self-consistency** — two branches of the same dispatch reach opposite verdicts for the same pipeline position. Not reproduced end-to-end (the hook only arms under a live reviewer dispatch), so this is reviewer-observed behavior + source-confirmed mechanism + a matching audit row, without a standalone live re-run.

**Proposed fix (PR offered against `main`):**
1. Thread the pipeline-segment index through `judgeCommandText` into `judgeGrepLike`/`judgeRipgrep`; treat a no-path search command as **stdin** when it is not the first segment.
2. When the judged target is a **default** (not a typed token), say so in the refusal message — `Target: .` for a command containing no `.` is what misleads readers.

Happy to open the PR (localized to `aidlc-reviewer-scope.ts`, with a regression test for a piped no-operand grep).

tejasavora:
## Summary

Per-unit reviewers can filter an in-scope search through `grep pat components.md | grep other` or `… | rg pat` without the downstream filter being refused as a project-wide search of `.`. Refusals for an implicit search root now explain that the path was assumed.

The stdin exception applies only to ordinary filters. Recursive `grep`, `rg --files`, and `rg -f -` still check their filesystem search roots. Explicit file operands and pattern files remain scoped, including attached/bundled `-e` and `-f`, long option values, empty patterns, and operands after `--`. Logical and sequential command separators do not grant the pipe exception.

Rebased onto current `main` and restored the scope safeguards and regression tests omitted by an intervening branch rewrite. Release metadata is unchanged from `main`, including version `2.8.2` and all existing changelog entries.

Closes #1073.

## Validation

- Focused reviewer-scope, reviewer-protocol, and version tests (`--debug -P 8 --unit --no-llm`): 134 passed, 0 failed, no skips.
- `bun run check`: passed packaging, determinism across seven harnesses, all TypeScript projects, and repository lint.
- Broad/live runs were stopped and their logs preserved. They are incomplete and are not counted as passing coverage.

## Acknowledgment


apackeer:
Thanks for the contribution, @tejasavora! The maintainers will now prepare this PR for merge and take care of the remaining fixes and validation.

apackeer (approved):
Reviewed the final rebased diff and the restored reviewer-scope safeguards. Ordinary piped filters are allowed while filesystem search modes, explicit operands, and pattern files remain scoped. The focused tests passed (134/134), contract checks passed, and release metadata is unchanged from main at 2.8.2.

PR #1081, opened 2026-09-09
fix: exempt high-specificity keywords from the scope-inference word-count veto

### LINKED ISSUE #1072: [Bug]: scope keyword-inference is vetoed for any description >5 words, so detect-scope silently returns the default scope
### Description

`inferScopeFromText()` collects keyword matches across every scope, then **discards them** whenever the input is longer than 5 words. The guard is:

```ts
// core/tools/aidlc-utility.ts:8379
if (allMatches.length === 0 || wordCount > 5) {
  return {
    scope: selectionAwareDefaultScope().scope,
    source: "freeform",
    matches: allMatches,
  };
}
```

The matches are already found at `:8365-8375` (the alphabetical scan that pushes into `allMatches`), so on a >5-word input the function returns the **selection-aware default scope** with `source: "freeform"` while still carrying the discarded matches in `matches` — the scope it computed is thrown away, but the evidence that it *had* a match is handed back. The comment at `:8377-8378` justifies the veto as "keyword + >5 words → likely a project description containing the keyword incidentally."

The problem: **essentially every real task or vision description exceeds 5 words.** A five-word ceiling means the keyword path only fires for terse inputs like `refactor this code` or `sap` — never for the kind of one-sentence description a user actually types. So keyword auto-detection effectively never fires in practice; the feature is dead for its intended input.

The blast radius is contained *only* by the interactive cold-start menu (the orchestrator offers the scope list when detection is inconclusive). **Any non-interactive consumer** of `detect-scope --from-text` — a script, a CI harness, a wrapper that pipes a vision statement in — silently receives the selection-aware default scope (documented default `feature`) and proceeds with the wrong stage graph. There is no error and no warning; `source: "freeform"` is the only tell, and a caller that only reads `scope` never sees it.

### Steps to Reproduce

Against an installed AI-DLC tree (data files present), run the scope detector two ways with the *same* keyword:

```bash
# short input (<=5 words) — keyword path fires
bun .claude/tools/aidlc-utility.ts detect-scope --from-text --input "refactor this code"
# → {"emitted":"SCOPE_DETECTED","scope":"refactor","source":"keyword","matches":["refactor"]}

# realistic input (>5 words) — same keyword, vetoed
bun .claude/tools/aidlc-utility.ts detect-scope --from-text \
  --input "refactor the legacy authentication module to improve long-term maintainability"
# → {"emitted":"SCOPE_DETECTED","scope":"feature","source":"freeform","matches":["refactor"]}
```

The second call detected the `refactor` keyword (`matches:["refactor"]`) yet returned the default scope `feature` with `source:"freeform"`. The word count — not the keyword — decided the scope.

### Expected Behavior

A description that contains a clear, high-specificity scope keyword resolves to that scope regardless of length. A 45-word SAP-migration description that mentions `abap` should detect the SAP scope, not fall back to the default. At minimum, the "incidental keyword" heuristic should not be a hard word-count veto that fires for every real-world sentence.

### Actual Behavior

Any input longer than 5 words returns `selectionAwareDefaultScope().scope` with `source: "freeform"`, discarding the keyword matches it already found. Measured during a real run (register #20): a ~45-word SAP run description returned `matches: ["sap"]` but `scope: "classic"`, `source: "freeform"` — on a **non-interactive** `detect-scope` consumer that default is taken silently — the wrong stage shape, zero SAP stages, no warning. (On the interactive path the orchestrator surfaces a compose menu when `source` is `freeform`, which contained the blast radius in that run, so the *silent* wrong-scope outcome is specific to non-interactive callers.)

| Input | Result |
|---|---|
| `sap` | scope `sap-build`, source `keyword` ✓ |
| `abap` | scope `sap-build`, source `keyword` ✓ |
| `refactor this code` | scope `refactor`, source `keyword` ✓ |
| full SAP run description (~45 words, contains `sap`/`abap`) | `matches:["sap"]` but scope **`classic`**, source **`freeform`** ✗ |

### AI-DLC Version
v2 (current on main)

### Release / Commit
2.8.1 (`52da70ad`)

### AI-DLC Phase
Initialization

### Harness
Claude Code (defect is harness-neutral — the code is in `core/`, so it ships to every harness)

### AI Model
Claude Opus 4.8

### Environment
- Verified against source at `main` @ `52da70ad` (2.8.1).
- `git log -S "wordCount > 5"` on `core/tools/aidlc-utility.ts` shows the guard has been present and untouched since the founding v2 commit (`7b824b34`).

### Additional Context

This is not fixable by adding more keywords — a longer keyword list still gets vetoed by the same word-count gate. It affects **all scopes**, not just plugin-contributed ones (I confirmed the scope loader globs `*.md`, so plugin scopes with bare names are loaded and their keywords parsed; the veto, not the loader, is the fault).

Two things verified as **not** bugs: the `break` after the first keyword per scope (`:8372`) intentionally caps `matches` at one entry per scope; and scope files are loaded regardless of prefix.

**Proposed fix (PR offered against `main`):** replace the word-count veto with one of —

2. **High-specificity-keyword exemption (preferred)** — mark keywords that are never incidental (`s/4hana`, `abap`, `spro`, `odata`, and similar domain tokens) as veto-exempt, while low-specificity words (`feature`, `fix`) still defer to the word-count heuristic. Preserves the incidental-keyword heuristic for generic prose. Or
1. **Match-count rule** — resolve to the detected scope whenever there is exactly one distinct matching scope. *Caveat:* this flips any long prose containing a single low-specificity keyword (`fix`, `feature`) to `source: "keyword"`, defeating the incidental-keyword heuristic and changing the cold-start UX for generic descriptions (exercised by t67/t118) — which is why the exemption is preferred.

Either is a single-function change confined to `inferScopeFromText()`; the `InferResult` shape (`scope` / `source` / `matches`) is preserved, so `handleDetectScope` and the t67 unit tests need no signature change. Happy to open the PR with a regression test asserting that a long description carrying a high-specificity keyword resolves to `source: "keyword"`.

tejasavora:
## Summary

Long task descriptions containing affirmative core scope keywords now resolve to the matching scope. For example, `refactor the legacy authentication module to improve long-term maintainability` returns `refactor / keyword` instead of `classic / freeform`. Interactive routing proposes the matching plan for confirmation; generic long descriptions retain the compose offer.

Closes #1072 for the core keyword path. Plugin-specific keyword metadata remains a follow-up.

## Changes

- Preserve the core keyword allowlist: `refactor`, `mvp`, `minimum viable`, `poc`, `proof of concept`, and `CVE`.
- Check every keyword for the exemption, so `security` cannot hide `CVE` and `prototype` cannot hide `poc`.
- Exclude occurrences with nearby preceding negation, while allowing later affirmative mentions and requests such as `refactor without changing behavior`. This remains a lexical heuristic.
- Preserve short-input precedence and the result/audit schema. Update scope documentation.
- Distinguish the exact named keyword policy from scope-identity coupling in t225, with runtime regressions proving that `refactor`, `mvp`, and `poc` still resolve a renamed scope.
- Preserve main's release metadata unchanged: version **2.8.2**, README badge, and all existing changelog entries.

## Validation

- Rebased onto main at `ca190330d3e529d0a4162af3a2590f75cac60140`, including #1080.
- Focused captured unit run: **87 pass, 0 fail** across t67, t198-compose-surfaces, t225-scope-name-decoupling, and t68-version-changelog-sync.
- The inference implementation is byte-for-byte identical to the previously reviewed implementation; the rebase preserves the negation and masked-keyword fixes.
- Local checks cover CLI and interactive routing, keyword masking, negation, whitespace, substring boundaries, precedence, renamed scopes, and version metadata agreement.
- No broad local suites or live-agent gates were run for this rebase.

## Acknowledgment


apackeer:
Thanks @tejasavora for the contribution and for investigating #1072. The maintainers will address the remaining review findings, prepare this PR for merge, and handle the merge.

apackeer (approved):
Reviewed the final rebased diff at 1bd66f5bd96fc85184f3351ce5a1ecbddaff007b. The focused t67, t198, t225, and t68 run passed 87 tests with no failures. The keyword exemption preserves short-input behavior, handles the reported masking/negation cases, and still resolves renamed scopes; the t225 exception is restricted to the exact named keyword policy. Release metadata is unchanged from main at 2.8.2. Approving this scoped change; normal GitHub merge gates still apply.

PR #1095, opened 2026-09-09
fix: protect engine-routed waves from serial lifecycle calls

### LINKED ISSUE #1071: [Bug]: bare unit start is accepted in wave mode, mutates state, and removes the batch path for the stage's remaining units
### Description

In the wave path a builder does not call the serial lifecycle verbs; the wave directive **is** the batch checkpoint, and every entry settles through `unit complete --wave`. The engine documents this. But the only argument-level guard is the **inverse** one: `aidlc-state.ts` rejects `--wave` on any action other than `complete`. **Nothing rejects a bare `unit start` while the workflow is *in* wave mode.**

So the invalid call succeeds: it mutates state and writes an audit entry (which cannot be voided). Worse, the accepted serial `unit start` sets the stage's lifecycle mode to serial/mixed, which **disables the wave branch for every remaining unit of the stage** — the deterministic batch fan-in is replaced by a manual, hand-deduplicated copy. A single accepted-but-invalid call cascades into losing a deterministic guarantee.

The verb set is `start|pause|resume|complete`; the argument-level guard is inverse-only for all of them, so `start`, `pause`, and `resume` are all plausibly accepted bare in wave mode. The engine already holds the fact that a wave is active — the verb simply does not ask.

### Steps to Reproduce

1. In a construction stage routed as a wave (batch checkpoint present), call `aidlc-state.ts unit start --stage <slug> --unit <name>` (no `--wave`).
2. Observe the call is accepted rather than refused.
3. Observe the resulting state mutation and the audit entry, and that subsequent `unit complete --wave` for the stage is now locked out.

### Expected Behavior

When a wave is active for a stage, `start`/`pause`/`resume` are refused, naming `complete --wave` as the route. The refusal message should name the blast radius — that the call would drop the batch path for every remaining unit of the stage.

### Actual Behavior

- The inverse-only guard at `core/tools/aidlc-state.ts:1899-1900`:
  ```ts
  if (waveMode && action !== "complete") {
    error("unit --wave is supported only with the complete action.");
  }
  ```
- The `start` branch at `core/tools/aidlc-state.ts:1952-1967` checks only the serial checkpoint, idempotent re-entry, and `requireEngineRoutedUnit` — it **never consults wave/batch mode**, so a bare `start` in wave mode passes.

### AI-DLC Version

### Phase
Construction

### Harness / Model
Observed on Claude Code; defect is harness-neutral (`core/`).

### Additional Context

Shared root cause across this family: **the framework does not check its own contracts for self-consistency**. The engine's own invariant ("a wave has no single active unit; every entry settles with `complete --wave`") is enforced in only one direction.

**Proposed fix (PR offered against `main`):**
1. Have `start`/`pause`/`resume` consult the active batch/wave state and refuse when a wave is active, naming `complete --wave` as the route.
2. Make the refusal message name the stage-wide blast radius, since a per-unit mistake has stage-wide scope.

Happy to open the PR (localized to the `unit` verb dispatch, with a regression test for a wave-active `start`).

tejasavora:
## Summary

Reject serial `unit start`, `pause`, and `resume` whenever the orchestration engine currently routes that stage as a wave, including before the first wave completion. Refused calls leave state and audit unchanged and direct the caller to `unit complete --wave`, preserving deterministic batch completion for the remaining units.

An explicit switch to unit-major iteration still permits the serial lifecycle, including when a revised stage has existing wave completion receipts. The guard uses the existing read-only engine route probe and reuses its directive for serial-start validation.

Closes #1071.

## Review fixes

- Cover fresh waves and waves with an existing completion receipt; verify all three serial verbs refuse without changing state or audit, and the next directive remains a wave.
- Cover the rejected-wave → first completion → unit-major switch, including serial start, pause, resume, and completion of remaining units.
- Preserve serial checkpoint handling when another stage routes a wave, and configure serial-lifecycle tests explicitly for unit-major iteration.
- Rebase onto current `main` after #1081. Version remains 2.8.2; the changelog conflict preserves the merged release summary and all history, adding only the two wave-lifecycle bullets.

## Validation

- After the final rebase: `bash tests/run-tests.sh --debug -P 8 --unit --filter '^t(68|260|278)-'` — all 58 tests passed across version sync, serial lifecycle, and wave lifecycle suites.
- The lifecycle implementation, protocol, and regression files are unchanged by the rebase. Prior captured package determinism, TypeScript/lint, smoke, 11-file focused unit, and isolated `t49` integration evidence remains applicable.
- Full outgoing diff reviewed for sensitive data. Required GitHub checks will be verified on the exact rebased head before merge; full local merge slices are deferred to the final PR in this release sequence.

## Acknowledgment


apackeer:
**[P2] Preserve a usable lifecycle when a revised wave switches to unit-major**

At [core/tools/aidlc-state.ts:1958–1960](https://github.com/awslabs/aidlc-workflows/blob/5a3b3a8e33b6c5656f8a51f96f077f76889c8de2/core/tools/aidlc-state.ts#L1958-L1960), a current wave receipt is treated as proof that the engine still routes a wave. That stops being true after an accepted iteration-mode switch.

Reproduction on this head, with `alpha` and `beta` in a focused `functional-design` workflow (other Construction stages skipped, review override `none`):

1. Request Changes through `aidlc-orchestrate.ts report --stage functional-design --result rejected --user-input "Request Changes" --reason "Revise the design"`.
2. Write alpha’s required artifacts and record `aidlc-state.ts unit complete --wave --stage functional-design --unit alpha`.
3. Run `aidlc-state.ts set-construction-iteration unit-major`.
4. `next` now emits a serial `run-stage` for beta, with no `wave`. After writing beta’s required artifacts, `unit start --stage functional-design --unit beta` is refused by the new guard, while the suggested `unit complete --wave --stage functional-design --unit beta` is also refused because the engine exposes no wave entry.

The switch does not create a new attempt boundary, and the existing `GATE_REJECTED` floor applies in both iteration modes, so the ledger still says `wave`. Remaining work is therefore blocked in the selected mode. In the equivalent fixture, the exact merge-base state handler (`52da70ad`) successfully starts and completes beta; the PR handler refuses its start.

Please coordinate this guard with actual wave routing, or handle existing wave receipts when accepting the iteration-mode switch, so the accepted configuration always leaves a usable completion route.

leandrodamascena (dismissed):
Thank you for tracing the lifecycle-mode transition and adding focused regression coverage.

I reviewed the current head `5a3b3a8e33b6c5656f8a51f96f077f76889c8de2` against its recorded base `3cbc00e7bac13673796ea97bf228a4ec93a75de5`, and also compared it with current `main` at `54d6d09eb3c1e79d3d2e4bdc597470243fb52f67`.

The linked issue is aligned with the project direction: accepting serial lifecycle verbs during an engine-routed wave violates the harness-neutral wave contract and can silently discard deterministic batch fan-in. However, the current implementation leaves the original failure path reachable.

### [P1] Reject serial verbs when the engine routes the wave, not after its first completion

The guard in `core/tools/aidlc-state.ts` checks:

```ts
currentUnitLifecycleMode(pd, slug) === "wave"
```

That only becomes true after the first `unit complete --wave` receipt. Before that receipt, the engine can already have emitted `directive.wave`, but the ledger mode remains `none`.

A bare `unit start` during this initial wave state is therefore still accepted. It appends a serial receipt, changes the lifecycle mode, and disables the wave path for the remaining units. This is the same irreversible state mutation and stage-wide loss of batch behavior described by #1071.

The new test currently asserts that this call succeeds from a fresh wave-routed stage, which preserves the defect rather than covering it.

Please derive the guard from the live engine route/wave directive and reject `start`, `pause`, and `resume` before any mutation. The regression should cover:

1. The engine emits a wave for a fresh multi-unit stage.
2. A bare serial lifecycle verb is attempted before any wave completion.
3. The command fails without changing state or appending an audit event.
4. The next engine directive remains a wave.

If serial fallback from an emitted wave is an intended product capability, it should require an explicit transition rather than occur implicitly through a bare lifecycle verb.

### [P2] Remove release metadata from this fix

Current `main`, through #1096, makes release preparation responsible for version, README badge, and changelog updates. After rebasing, please remove:

- the `2.8.4` change in `core/tools/aidlc-version.ts`;
- the README badge change;
- the `2.8.4` changelog entry.

No files or commits were changed or pushed during this review. I inspected the complete diff and the existing wave contract; local validation was intentionally deferred because the lifecycle interpretation above must be resolved first.

apackeer:
Thank you @tejasavora for the contribution and for tracing the wave lifecycle issue. The maintainers will now prepare this PR for merge, address the review findings, rebase it onto the current base, and complete validation. Merge will remain gated on final maintainer approval.

tejasavora:
Thanks for the detailed review. I've implemented the P1 guard as requested — and enforcing it surfaces exactly the product decision you raised, so I want your call before finalizing.

**Guard (done).** The refusal now keys on the **live engine wave-route** rather than `currentUnitLifecycleMode`. A new `engineRoutesWaveForStage(stage)` inspects the current directive (`kind: run-stage`, matching stage, non-empty `wave.entries` — the same directive `requireEngineRoutedWaveUnit` already reads) through a **non-mutating** observer, so `start`/`pause`/`resume` are refused **before any state write** the moment the engine routes a wave — including the fresh, pre-first-`complete --wave` state #1071 describes. Serial/mixed ledgers and open checkpoints don't spawn the check; a transport failure resolves to "not a wave," so a transient engine hiccup never blocks a legitimate verb (and can't reintroduce #1071, since a failed engine can't route a wave either).

**t278 (done, green).** Reworked to your 4-point spec: engine routes a wave for a fresh multi-unit stage → a bare serial verb is attempted before any wave completion → it's refused with no lifecycle receipt appended (mode doesn't flip) → the next engine directive is still a wave. Added a second case for the settled-wave (`mode: "wave"`) arm.

**The t260 question — your call.** Enforcing this exposes that every **stage-major inline** per-unit stage (functional-design, nfr-requirements, nfr-design, infrastructure-design) *always* routes a wave, so the bare-verb serial lifecycle becomes unreachable on all of them (it survives only on unit-major iteration and code-generation's `mode: subagent`). About ~10 t260 tests use serial `unit start` as **setup** for receipt/artifact/run-floor mechanics, so they can't simply be flipped to "expect refusal." Which direction do you prefer?

- **(a) Serial-on-inline is deprecated** — wave is the sole path. I delete/convert those t260 mechanics tests (their coverage lives in the wave path via t278 and unit-major via t209) and convert the guard-subject tests to expect the refusal.
- **(b) Add an explicit serial-fallback transition** (your "explicit transition rather than a bare verb") — I add the transition verb and rehome t260's serial fixtures behind it.

I've held the branch (guard + t278 + the metadata strip you asked for in P2 are all green locally) rather than push red t260 CI. Happy to take whichever way you choose.

apackeer (approved):
Reviewed 8a69f589a13ab324593c2b4945614960eada6690 rebased onto main after #1081. The live-route guard rejects serial lifecycle verbs before any wave receipt and preserves state and audit bytes. Explicit unit-major selection remains usable after a revised wave completion. The prior review findings are addressed.

The rebase leaves the reviewed implementation and regression files unchanged; the changelog resolution preserves the existing 2.8.2 release entry and all historical content. Focused t68, t260, and t278 validation passed: 58 tests, zero failures. Required GitHub checks remain a separate merge gate.

PR #1097, opened 2026-09-09
fix: restore Plan Approval forwarding and harness hook execution

apackeer:
## Summary

Plan Approval could block the forwarding needed to present and answer the plan. Lifecycle bookkeeping and harness input/path handling also caused valid workflow operations to fail. Restore that transport while retaining approval and evidence requirements. Addresses #1047.

- Allow `orchestrate next` and `continue` through native `aidlc` and the real installed Bun dispatcher. Keep the narrow planning allowlist and reject wrappers, shadow/symlinked entry points, preloads, changed executable lookup, and data-driven calls.
- Preserve shell descriptor redirections, escaped operators, empty quoted operands, and actual shell token boundaries, including Unicode filename cases.
- Exclude strictly empty Markdown separators from the state digest so parking/resuming and Unit lifecycle bookkeeping preserve authority. Routing fields, stage progress, unknown content and nonempty whitespace still bind it.
- Anchor Claude source hooks/statusline to quoted `$CLAUDE_PROJECT_DIR` paths. Normalize unified Bun workflow commands for engagement and runtime-graph classification; retain native packaging and runtime-compile recursion checks.
- Preserve Kiro IDE hook stdin for Plan Approval and human-turn recording, and recognize documented read aliases while continuing to reject unknown tools, malformed payloads and mutations.
- Require actual composer dispatch once selected. After pipeline rejection, direct fresh ordered work and receipts while retaining reviewer policy; stop suggesting an evidence opt-out as recovery. Render recovery commands through the native-aware invocation helper so native installs use `aidlc engine ...` and source installs retain Bun commands.

The full-gate investigation also repairs the test infrastructure and fixtures: owned-fixture Claude trust-menu readiness, env-wrapped launch recognition and current effort labels; Kiro IDE chat focus/obstructions and completed-continuation observation; native ACP dispatch and separate reviewer evidence; SDK transcript persistence; Codex inner-command/workspace checks with a scoped `.codex` write grant; semantic-digest fixtures; child-PID and bounded setup/cleanup handling; and generated coverage/hook-registration expectations.

T51 now charges startup against its original 40-minute cap and reserves cleanup time, with a watchdog for its owned driver. T139 waits for approval's cursor advance before comparing phases, adds cursor equality, retains its existing state/revision assertions and deadline, and checks recorded native requests for model-issued opt-outs. Its separate final live confirmation passes; the original phase failure remains preserved.

[...]

PR #1101, opened 2026-09-09
fix: keep the current provider unless Bedrock is selected

### LINKED ISSUE #1098: aidlc config can record a non-Bedrock provider while keeping Bedrock active
## Problem

`aidlc config` can record or display a non-Bedrock provider choice while the installed harness still uses Amazon Bedrock.

This reproduces on v2.8.1 and `main` at `0c12f99ba6ce97485f77670e809e9fd5c57d4edb`.

## Reproduction

Run the first-use wizard for Claude Code and select:

```text
Provider: other
Preset: balanced
MCP: off
Record in: this project, committed
```

The wizard confirms `Provider other` and says the choices were recorded. The resulting `.claude/settings.json` still contains:

```json
"CLAUDE_CODE_USE_BEDROCK": "1",
"AWS_REGION": "us-east-1",
"ANTHROPIC_DEFAULT_OPUS_MODEL": "global.anthropic.claude-opus-4-8[1m]"
```

Running the explicit follow-up also leaves those values in place:

```bash
aidlc config providers --provider other --acknowledge --yes
aidlc config providers --check
```

The second command exits 0 and reports the provider configuration as clean.

## Additional affected surfaces

- Codex records `provider: other` while `.codex/config.toml` retains `model_provider = "amazon-bedrock"` and the Bedrock model.
- opencode records `provider: other` while balanced reviewer agents retain provider-qualified Bedrock model IDs.
- The first-use wizard does not persist `provider: other`. It only prints the choice and adds a manual follow-up.
- Step 6 says it selects where to record "these choices", but the selected settings layer receives only the model preset. Plugin, MCP, completion, and provider state remain project-level.
- The live `/aidlc --config trust` SDK test can fail before opening config because the project-level Bedrock settings take effect first.

## Expected behavior

- Shipped project configuration does not select a provider.
- The first provider choice is `keep current`; it preserves the harness provider and model already configured by the user and records `provider: current`.
- The second provider choice is Amazon Bedrock.
- Selecting `other` records a manually configured non-Bedrock provider.
- Selecting Amazon Bedrock applies the required Bedrock settings explicitly.
- The first-use wizard persists the provider answer.
- `config providers --check` compares the recorded answer with AI-DLC-owned provider settings and fails on a mismatch.
- The settings-layer prompt states which choices use the selected layer and which remain project installation choices.

## Acceptance criteria

- Provider-neutral defaults for Claude Code, Codex, and opencode.
- Upgrade handling removes stale AI-DLC-shipped Bedrock defaults when `current` or `other` is selected.
- Bedrock opt-in remains functional from a provider-neutral baseline.
- Provider records and effective harness settings cannot disagree while `--check` exits 0.
- Regression coverage spans all seven harnesses and project, local, and machine settings targets.
- No version, changelog, or README update in the implementation PR. Release metadata remains release-owned.

Related to #428. Overlaps with #1028, which removes provider defaults but predates the current first-use wizard and provider diagnostics.

leandrodamascena:
## What changed

This fixes the provider mismatch from #1098.

- `keep current` is now the first provider choice and the default
- Amazon Bedrock is the second choice, even when AWS credentials are detected
- Claude Code, Codex, and opencode no longer ship with project-level Bedrock model or provider pins
- `provider: current` removes old AI-DLC Bedrock overrides while keeping the provider configured by the user
- `provider: other` records a manual non-Bedrock setup and can stay pending until acknowledged
- provider checks now fail when AI-DLC-owned project settings disagree with the recorded choice
- the settings-layer step now says it controls the model preset, not every setup choice

For Codex, provider configuration stays in `~/.codex/config.toml`. The project config does not try to write a provider table that Codex ignores at project scope.

## Validation

- 126 focused provider, model, packaging, SDK, and first-run tests passed
- 69 final config and first-run tests passed after the provider choice change
- `t293-config-models` passed separately with 17 tests
- `bun scripts/package.ts --check` passed after rebasing on the latest `main`
- no version, changelog, or README change

Closes #1098


apackeer:
Reviewed `e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb`. I found two reproducible configuration gaps and some documentation that conflicts with the new behavior.

1. **[P2] Existing Codex Bedrock records do not acquire the new manual-setup action.**

   `codex-provider-configuration` is now required, but reconciliation runs only when changing provider answers. Existing records retain their old action list during an ordinary config refresh.

   Reproduction: start with a legacy Codex `amazon-bedrock` record whose only action is `bedrock-model-access: done`, run `aidlc config --yes`, then `aidlc config providers --check`. The record still contains only the old action, and the check exits **0** with `issues: []`. The user never receives the newly required instruction to configure the provider/model at user scope.

   Please reconcile required actions when upgrading/loading/checking existing records, as well as when mutating provider answers.

   References: [new required action](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-config-diagnostics.ts#L906-L917), [stored-list normalization](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-config-diagnostics.ts#L337-L339), [refresh carries existing records](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-init.ts#L3457-L3468).

2. **[P2] Reapplying the same provider choice cannot repair a detected override.**

   The new diagnostics detect stale project Bedrock overrides, but the unchanged-record shortcut prevents cleanup from running when the saved choice already matches the requested choice.

   Reproduced on Claude:

   ```text
   Record provider: current.
   Add CLAUDE_CODE_USE_BEDROCK=1 to the project settings.
   aidlc config providers --check
     -> exit 1: provider-claude-project-override
   aidlc config providers --provider current --yes
     -> exit 0: providers configuration unchanged
   aidlc config providers --check
     -> exit 1: the same override remains
   ```

   Please make the no-op decision account for provider-setting drift, or provide a repair path that respects file ownership and explains any conflicts. Re-entering the recorded answer currently does not perform the advertised cleanup.

   References: [unchanged-record shortcut](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-init.ts#L2143-L2147), [new cleanup branch](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-config-diagnostics.ts#L1131-L1137), [new override check](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/core/tools/aidlc-config-diagnostics.ts#L1598-L1617).

3. **[P2] Update documentation to match the provider picker and model inheritance.**

   The [getting-started guide](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/docs/guide/01-getting-started.md#L129-L132) says to run interactive `aidlc config providers` and select `other`, but the picker now offers only `keep current` and `amazon-bedrock`. Document `--provider other` explicitly or describe the new interactive path.

   The [README prose](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/README.md#L80-L83) still says Claude and Codex default to Bedrock. The [agent-system description](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/docs/reference/05-agent-system.md#L93) and [customization guide](https://github.com/awslabs/aidlc-workflows/blob/e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb/docs/guide/13-customization.md#L43) also still describe mid-size model pins for Codex/opencode reviewers, although these now inherit the session model.

Validation: packaging, `package.ts --check`, typecheck and lint exited 0. The focused smoke/unit run completed with **8 of 9 files passing**. `t294-config-diagnostics` had a provider-case timeout and an unnamed hook timeout; on isolated retry, **all 24 named cases passed**, but the unnamed hook timeout persisted, so verification is not fully passing. The two configuration findings above were reproduced separately through the CLI. No live-provider/e2e tests were run.

No release-metadata changes were found. I recommend addressing the findings before approval.

apackeer (changes requested):
Requesting changes on `e2c2909e43bf1d2956673d1dbb9c121ca90ff2cb` for the [findings and reproduction details already posted](https://github.com/awslabs/aidlc-workflows/pull/1101#issuecomment-5610070515):

- Existing Codex Bedrock records do not acquire the new manual-configuration action during refresh/check.
- Reapplying an unchanged provider choice skips cleanup even when provider diagnostics detect stale overrides.
- The provider-picker and reviewer-model documentation still describes the previous behavior.

Please address these findings before approval. Verification details, including the remaining provider-diagnostics hook timeout, are in the linked comment.

leandrodamascena:
Thanks, I fixed all three points in `858284cee`: legacy Codex records now add and persist the pending action, reapplying the same provider repairs stale overrides without removing unrelated settings, and the docs/tests match the provider-neutral behavior. Validation passed for t294 (26/26), focused parity tests (4/4), package drift, typecheck, and lint, and the new CI run is now starting.

wowzoo (changes requested):
Thanks for turning this around so quickly, and for the detail in #1098 — the acceptance
criteria made this verifiable rather than a matter of opinion.

I reviewed head `858284cee7e84dccd144d1821f9ea21d08d6050b` against `main`.
`a8f2dc6198c011343e7674bc82534c0adaf0b548` is both the recorded base and the merge base;
`main` has since moved one commit ahead with #1097, which touches
`harness/claude/settings.json` and `tests/smoke/t03-settings-json.test.ts` without
conflicting.

**All three earlier findings are fixed.** Two reproduce end to end:

- Seeding a legacy Codex record whose only action is `bedrock-model-access: done` and running
  an ordinary `aidlc config --yes` now persists `codex-provider-configuration: pending`, and
  `config providers --check` exits 1 naming it.
- With `provider: current` recorded and `CLAUDE_CODE_USE_BEDROCK=1` re-added by hand,
  `--check` exits 1, `--provider current --yes` repairs instead of printing
  `configuration unchanged`, `--check` returns to 0, and an unrelated key in the same `env`
  block survives.
- The documentation now matches the picker and the tier projection, and the generated
  artifacts are clean — `dist/claude/.claude/settings.json` ships
  `env: {AWS_AIDLC_DEFAULT_SCOPE}` only.

The new coverage is load-bearing: `t294` is 26 pass / 0 fail at this head, and 24 pass /
2 fail with `core/tools/aidlc-config-diagnostics.ts` and `core/tools/aidlc-init.ts` reverted
to `e2c2909e` and the packager re-run — the two failures are exactly the two new tests.
Upgrading a base install clears the stale `amazon-bedrock/global.anthropic.claude-sonnet-4-6`
and `openai.gpt-5.6-terra` agent pins, `bun run check` exits 0, and the tier change behaves:
an opencode reviewer resolves with `variant: medium` and no model override.

Two things I'd like resolved before this lands, and three smaller ones after.

---

### 1. `clearCodexProvider` deletes a project-scoped `model` that is in force

`core/tools/aidlc-config-diagnostics.ts#L1042-L1056` removes
`model|model_provider|model_context_window|model_reasoning_effort` with a line regex. A
project `.codex/config.toml` carrying

```toml
model = "gpt-5.5"
model_reasoning_effort = "low"
```

loses both on `config providers --provider current --yes` (exit 0). The probe at
`#L1633-L1644` matches equally broadly, so a user who re-adds their own `model` then gets
`provider-codex-project-override` and `--check` can never return 0.

These keys take effect, and it depends on trust. On `codex 0.154.0.444`, with a sandboxed
`CODEX_HOME` holding `model = "user-sentinel-model"` and a project `.codex/config.toml`
holding `model = "project-sentinel-model"`, `codex doctor --json` reports the user value for
an untrusted project and the project value once `[projects."<path>"] trust_level = "trusted"`
is set. `0.153.4.440` behaves the same. An AI-DLC install is in that trusted state by
construction: `docs/guide/harnesses/codex-cli.md#L97-L99` makes trusting the project step 3,
and `#L132-L134` says *"(or keep it project-level — trusted projects read it)"*. I could not
test the pinned floor `0.145.0` — worth confirming there.

Two obvious narrowings don't work, so it may save you the attempt:

- Scoping the regex to the region before the first table — TOML top-level keys must precede
  the first table, so that is exactly where a user's own settings live.
- Removing only `model_provider = "amazon-bedrock"` and its `[model_providers.amazon-bedrock…]`
  tables — `main` shipped `model = "openai.gpt-5.5"`, `model_context_window = 1000000` and
  `model_reasoning_effort = "high"` at top level, so those stale AI-DLC values would survive,
  against `docs/guide/18-install-and-lifecycle.md#L440-L442`.

The boundary is that a line regex cannot decide table scope and ownership at once, so the
cleanup needs an ownership signal of some kind. One candidate is the install baseline's
per-file hash (`core/tools/aidlc-init.ts#L179-L186`, written at `#L6305-L6312`): an untouched
`.codex/config.toml` is provably yours. That has a wrinkle I have not resolved — the first
`--provider current` rewrites the file, so unless the transaction also rewrites the baseline,
the next run reads AI-DLC's own edit as a user modification. `legacySignatures` may be the
better instrument since it already answers "is this an old shipped variant". Either way the
ask is just that the sweep stop deleting keys it cannot show it owns.

### 2. The codex Bedrock surface check was removed, so `--mark-done` alone certifies a state that doesn't exist

There is no codex writer for that answer
(`core/tools/aidlc-config-diagnostics.ts#L1144-L1160`), `--show` lists only the record file
(`#L1197-L1219`), and `--mark-done` verifies nothing external
(`core/tools/aidlc-init.ts#L1547-L1558`). With a sandboxed `HOME` containing **no**
`~/.codex/config.toml`:

```
--provider amazon-bedrock --region us-east-1 --yes  → exit 0, project config.toml unchanged
--check                                            → exit 1 (2 pending actions)
--mark-done bedrock-model-access --mark-done codex-provider-configuration --yes → exit 0
--check                                            → exit 0, "clean for codex"
```

Bedrock is configured nowhere — not in the project, not in the user config — and the tool
reports clean.

This is new. On `main`, `writeCodexProvider` wrote `region = "us-east-1"` into the project
`.codex/config.toml` and `providerValueIssues` asserted the recorded region and profile
against those bytes (`a8f2dc61:core/tools/aidlc-config-diagnostics.ts:1546-1553`). Same
sandboxed `HOME`, base arm: after opting in, tampering the region to `eu-west-1` makes
`--check` exit 1 with `provider-codex`, and `--mark-done bedrock-model-access` does not clear
it — the byte check keeps failing. So the change removes the only part of the codex provider
check that verified anything, leaving self-attested actions that `--mark-done` can always
satisfy.

The guidance compounds it: the only instruction is *"Configure the provider, credentials, and
model in ~/.codex/config.toml, then acknowledge the manual setup"* (`#L207-L212`) — no keys,
no example — against #1098's *"Bedrock opt-in remains functional from a provider-neutral
baseline"*.

I'd rather not push a specific mechanism: verifying `$CODEX_HOME/config.toml` may not be
sound, since `CODEX_HOME` can be relocated and the provider can be configured through other
channels, and a check that false-negatives lands in the same unsatisfiable state as finding 1.
The outcome I'd ask for is either that the check regains something it can verify, or that
`--show` and the receipt stop describing a self-attested state as "clean". Carrying the
concrete TOML — the block previously in `harness/codex/emit.ts` — would separately fix the
"nothing to act on" half. Attestation is clearly deliberate for `kiro-ide-chat-model` and
`copilot-byok-configuration`; the difference here is that something was removed to reach it.

---

### 3. `--check`, `--show` and doctor don't see higher-precedence Claude layers

The probe at `#L1614-L1633` reads only `<harnessDir>/settings.json`, which is the third layer
by this repo's own statement of precedence (`core/tools/aidlc-utility.ts#L2886-L2888`). With
`env.CLAUDE_CODE_USE_BEDROCK=1` in `.claude/settings.local.json`,
`--provider other --acknowledge --yes` and `--check` both exit 0 "clean for claude" while
Bedrock is in force, and neither `--show` nor doctor says otherwise. That is the documented
personal-override location (`docs/guide/01-getting-started.md#L126`), and the pre-change
`settings.local.json.example` shipped `AWS_REGION` with the comment *"Required for Bedrock"*.

What makes this worth raising rather than shrugging at is that the onboarding this PR writes
promises the validation and then points at the unread file, two bullets apart in the generated
`dist/claude/.claude/CLAUDE.md`:

- `:19` — *"Run `… config providers` to record and **validate** a project provider choice."*
- `:25` — *"Copy `.claude/settings.local.json.example` to `.claude/settings.local.json`
  (gitignored) to override the model or **set environment variables**…"*

Following both lands the user in the state above. A narrow reading of `validate` — the recorded
project-scoped choice checked against the project files — is defensible, but the clause right
before it is about what Claude Code keeps, so the frame is the effective provider. Worth noting
the codex bullet in this same commit claims only *"record the project choice and see any manual
provider action"*, which is exactly the accurate weaker phrasing.

A warning at exit 0 rather than a failure seems right — *"AI-DLC-owned settings are clean; the
effective provider is overridden by `<path>`"* — because `providerIssues` merges every
diagnostic into one failure channel (`#L1719-L1722`) with no acknowledgement state, so failing
would leave `--check` permanently non-zero for a user who legitimately keeps a personal
override. `DiagnosticDoctorCheck` already carries `severity?: "warn"` (`#L141-L146`), so the
doctor side needs no structural change; a warn tier in `--check` would be a field on
`DiagnosticIssue` plus a filter at `core/tools/aidlc-init.ts#L1447-L1462`, and is optional.

One thing I'd flag rather than decide: #1098 says two different things. Expected behavior
scopes `--check` to *"AI-DLC-owned provider settings"* (line 51), which this PR satisfies,
while the acceptance criteria say *"Provider records and effective harness settings cannot
disagree while `--check` exits 0"* (line 59), which isn't reachable without that
acknowledgement state. Whichever governs is your call — I'd just ask that `Closes #1098` not
take the broader clause with it. Also worth noting the same blind spot applies to
machine-owned Codex configuration and the instruct-only harnesses; I only reproduced Claude's
project-local path.

### 4. The Codex onboarding states precedence backwards

The generated `AGENTS.md` says *"Settings in `~/.codex/config.toml` merge over the project
`.codex/config.toml`. Put machine-specific overrides (model, …) there"*, and the new
`config.toml` header says the project inherits the model from the user's configuration. Per
the measurement in finding 1 that holds only for an untrusted project; in the trusted state
the guide requires, the project layer wins — and `codex-cli.md#L132-L134` already says
trusted projects read the project-level config. Stating the condition, or reversing the
sentence, would settle it.

### 5. `t140` has two assertions that depend on the ambient environment

`tests/harness/sdk-drive.ts#L393-L401` starts `env` from `processEnv()`, and Claude Code
documents that settings-file `env` values enter the process environment and are inherited by
subprocesses. Until this PR the shipped `settings.json` always supplied those keys, so the old
`toBe(...)` assertions held regardless; now both `toBeUndefined()` checks at
`tests/unit/t140-sdk-drive-model-resolution.test.ts#L40-L41` read the developer's environment.
With `CLAUDE_CODE_USE_BEDROCK=1` line 40 fails; with only `ANTHROPIC_DEFAULT_OPUS_MODEL` set
line 41 fails; with both unset 3 pass; on `main` with both set 3 pass. Nothing shipped changes
and CI runners carry neither variable, so CI stays green — but a maintainer with Bedrock in
`~/.claude/settings.json` sees a red suite CI cannot reproduce. Asserting on the settings
contribution, the way `tests/smoke/t03-settings-json.test.ts` does, would make it
environment-independent.

---

One non-code item: `Require Contributor Statement` is red because the description is missing
the verbatim sentence it looks for, and with the merge queue on `ALLGREEN` that blocks the
merge by itself. Everything else is green — all four unit shards, integration + e2e, smoke,
contract checks and build.

Separately, and not for this PR: the release channel's legacy `AGENTS.md` signatures never
match what ships. `scripts/package.ts` records the hash before the release path rewrites
native invocations, so at this head the codex descriptor holds `47ccebcd…` while the emitted
`dist-release/codex/AGENTS.md` is `c182c17d…` (opencode: `0f7fa481…` vs `0767fe4a…`), and
`main` mismatches the same way. That makes the documented copy-channel `AGENTS.md`
non-adoptable via legacy signatures. Happy to open a separate issue.

For whoever approves: the branch ruleset requires one approval from the `aidlc-admins` team in
addition to the count, so an approval from me satisfies the count only.

leandrodamascena:
@wowzoo @apackeer, I addressed the requested changes and rebased this PR onto the latest `main`.

- Codex cleanup now removes only the exact legacy AI-DLC Bedrock signature and preserves user-authored project model/provider settings.
- Codex Bedrock setup remains explicitly self-attested, includes concrete TOML guidance, and reports a warning instead of claiming the configuration is clean.
- Claude overrides from `.claude/settings.local.json` are surfaced as warnings.
- The Codex user/project configuration precedence wording is corrected.
- The SDK model-resolution test no longer depends on the ambient environment.

All CI checks are green on `65fa752eb`. Could you please take another look?

apackeer (changes requested):
Re-reviewed `65fa752ebc0cec273466c8bd1fc78a1eb1315ee8`. Two reproducible configuration-loss issues still need addressing before approval.

1. **[P2] Preserve user configuration across the next refresh, not only the provider command.**

   [The provider refresh preparation](https://github.com/awslabs/aidlc-workflows/blob/65fa752ebc0cec273466c8bd1fc78a1eb1315ee8/core/tools/aidlc-init.ts#L3520-L3533) copies the current config into the staged projection and marks the whole file regenerated. [That path records the customized file's hash as the ownership baseline](https://github.com/awslabs/aidlc-workflows/blob/65fa752ebc0cec273466c8bd1fc78a1eb1315ee8/core/tools/aidlc-init.ts#L5039-L5047). A subsequent ordinary refresh starts from the shipped defaults and [allows replacement because the current file matches that baseline](https://github.com/awslabs/aidlc-workflows/blob/65fa752ebc0cec273466c8bd1fc78a1eb1315ee8/core/tools/aidlc-init.ts#L5072-L5100).

   Reproduction through the source CLI against matching generated release files:

   - Install Codex, then add these top-level settings to `.codex/config.toml`:

     ```toml
     model = "gpt-5.5"
     model_reasoning_effort = "low"
     ```

   - Run `config providers --provider current --yes`: exit **0**, both settings preserved.
   - Run `config --from <matching release source> --harness codex --yes`: exit **0**, both settings deleted. No `--force` was used.

   Please preserve the distinction between user-authored settings and generated defaults across later refreshes. Add coverage for provider selection followed by a normal refresh; checking only the immediate provider update misses this loss.

2. **[P2] Preserve user-authored Claude model aliases and AWS settings.**

   [The Claude cleanup](https://github.com/awslabs/aidlc-workflows/blob/65fa752ebc0cec273466c8bd1fc78a1eb1315ee8/core/tools/aidlc-config-diagnostics.ts#L1033-L1040) treats the presence of any listed `ANTHROPIC_DEFAULT_*_MODEL` key as a Bedrock override, deletes those keys, and also deletes `AWS_REGION` and `AWS_PROFILE`. Presence alone does not establish that AI-DLC owns these values.

   Reproduction: install Claude, then add the following to `.claude/settings.json`'s `env`, leaving `CLAUDE_CODE_USE_BEDROCK` absent:

   ```json
   {
     "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-4-6",
     "AWS_REGION": "eu-west-1",
     "AWS_PROFILE": "team-profile",
     "MY_TEAM_SETTING": "preserved"
   }
   ```

   Run `config providers --provider current --yes`: exit **0**, the model alias, region, and profile are deleted; `MY_TEAM_SETTING` remains. “Keep current” changes this user's configuration despite there being no project Bedrock flag.

   Please restrict cleanup to values demonstrably owned by AI-DLC, preserve customized values or report a conflict, and apply the same distinction to provider diagnostics.

Validation: both cases reproduced twice in separate filtered unit runs after supplying the explicit refresh source in the Codex fixture. Packaging succeeded. The broader focused smoke/unit run passed **17 of 19 files**; `t293-config-models` and `t294-config-diagnostics` had five failures at generic refresh calls returning exit 4, including missing release-source diagnostics. Those remain unresolved local validation limitations, not additional established PR regressions. No live-provider or e2e tests were run.

rafaelgsr (changes requested):
[P1] Preserve user-owned provider files across later refreshes — core/tools/aidlc-init.ts:3529

Provider mutation copies the entire current Claude, Codex, or Kiro configuration into staging and records it as generated. The next ordinary refresh therefore considers user settings framework-owned and overwrites them. I reproduced loss of Codex model settings, a Claude environment key, and a Kiro MCP server. Preserve field-level ownership and add mutation-followed-by-refresh coverage.

[P1] Restrict Claude cleanup to settings demonstrably owned by AI-DLC — core/tools/aidlc-config-diagnostics.ts:1034

The cleanup treats any model alias as evidence of an AI-DLC Bedrock configuration, then deletes model aliases, `AWS_REGION`, and `AWS_PROFILE`. Running “keep current” can therefore destroy a user-managed Claude setup. Require an exact legacy signature or preserve conflicting custom values.

[P2] Check Claude local overrides for Bedrock records too — core/tools/aidlc-config-diagnostics.ts:1649

Local settings are inspected only for `current` and `other`. With a Bedrock record, `.claude/settings.local.json` can disable Bedrock or select another region while `--check` reports success. Inspect this higher-precedence layer for every provider record.

[P2] Reset manual acknowledgement when the provider changes — core/tools/aidlc-config-diagnostics.ts:949

The acknowledgement boolean survives provider changes. I switched an acknowledged `other` record to Bedrock without `--acknowledge`; the new Codex configuration action was still marked done. Reset acknowledgement on provider changes or scope it to the provider/action being attested.

[P2] Surface blocking provider drift in doctor — core/tools/aidlc-config-diagnostics.ts:2313

`providerDoctorCheck` filters surface issues to warnings, so blocking drift can produce a passing doctor result while `config providers --check` fails. Include blocking mismatches in the doctor result.

Thanks for the work to make provider defaults neutral while preserving manual setups.

I reviewed the current head `65fa752ebc0cec273466c8bd1fc78a1eb1315ee8` against current main `7998589725a932c43bb3bb870c665cc7c724d3df`, using merge base `aadc832354f2d8841489ed01b5e0d852b32b7962`. Project alignment is established by #1098 and the prior substantive maintainer reviews.

The first two findings overlap existing maintainer blockers; the remaining three are additional. No code changes or commits were pushed. Packaging, drift checks, typecheck, lint, and focused tests passed, and the material CI checks for this head are green. The branch currently conflicts with main and will need rebasing.

Requested event: **REQUEST_CHANGES**.

leandrodamascena:
Addressed the blocking review findings and pushed `f5762c8d9`.

Changes:
- merged current `main` and retained Kiro CLI/IDE provider handling as harness-managed;
- preserve user-owned Claude environment values, Codex model/provider fields, and Kiro MCP configuration across ordinary refreshes;
- remove Claude provider values only when they match the exact legacy AI-DLC Bedrock signature, while preserving custom aliases and AWS settings;
- inspect `.claude/settings.local.json` for Bedrock conflicts;
- reset provider acknowledgement and pending state when the provider changes;
- make `doctor` fail on blocking provider drift while retaining warning-only results;
- apply project flags after provider preservation so the recorded scope remains authoritative.

Validation:
- deterministic packaging, typecheck, and lint pass;
- focused provider, configuration, setup, and install tests pass;
- all four unit shards, smoke, integration/e2e, contract, security, documentation, changelog, and aggregate CI checks pass.

@rafaelgsr @apackeer @wowzoo, please re-review this revision.

rafaelgsr (dismissed):
[P1] Preserve unrelated settings when carrying provider fields — core/tools/aidlc-init.ts:3657

The new preservation copies only Claude `env` or Codex provider fields into staging, after which the entire file is marked runtime-generated and replaced without `--force`. I reproduced Claude project `model` disappearing when a custom environment key was preserved, and Codex `sandbox_mode = "read-only"` reverting while its provider fields survived. Merge into the current file or retain normal ownership-conflict handling for unpreserved fields.

[P1] Keep user-authored AWS_PROFILE during legacy Claude cleanup — core/tools/aidlc-config-diagnostics.ts:1138

The legacy signature does not include `AWS_PROFILE`, but cleanup deletes it. Adding `AWS_PROFILE: "team-profile"` to otherwise exact legacy settings and selecting `current` removes the user-authored profile. Preserve this key or require its absence before treating the configuration as the exact shipped signature.

[P2] Leave a manually configured OpenCode provider alone for current — core/tools/aidlc-config-diagnostics.ts:1322

Selecting `current` always removes `provider.amazon-bedrock`. With a user-configured OpenCode Bedrock block, the transaction exits with an `opencode.json (unowned whole file)` conflict instead of recording and preserving the current provider. Only remove settings proven to originate from a prior AI-DLC Bedrock answer.

[P2] Reset Bedrock attestations when region or profile changes — core/tools/aidlc-init.ts:1606

Acknowledgement is reset only when the provider name changes. Changing an acknowledged Codex record from `us-east-1` to `eu-west-1` retained both actions as done, and `--check` exited zero without revalidating model access or Codex configuration. Reset relevant actions whenever provider parameters change.

[P2] Remove the obsolete provider workflow from the guide — docs/guide/18-install-and-lifecycle.md:495

The guide still says Codex writes provider settings into project `config.toml`, that Bedrock leads, and that `unchanged` is the second provider answer. The troubleshooting guide also recommends the nonexistent `unchanged` choice. Update these references to match `current`-first behavior and user-level Codex setup.

Thanks for addressing the earlier review and rebasing the provider-neutral work onto current main.

I reviewed the current head `f5762c8d9c891ac05a526f75c46c76835d1c22ca` against `7998589725a932c43bb3bb870c665cc7c724d3df`, which is also the merge base. Project alignment remains established by #1098 and the prior substantive maintainer reviews. The local-override and doctor findings are fixed, but configuration preservation and acknowledgement remain incomplete.

No code changes or commits were pushed. Packaging, determinism, typecheck, and lint passed. Provider tests passed with an extended timeout; exact-head CI is fully green and the PR is mergeable. The remaining risk is user configuration loss or stale provider attestations in the paths described above.

Requested event: **REQUEST_CHANGES**.

leandrodamascena:
Thanks for the detailed reproductions. I addressed the latest blocking findings and pushed `1252de255`.

Changes:
- preserve unrelated user-owned Claude and Codex settings during provider mutation and later refreshes;
- preserve user-authored `AWS_PROFILE` during legacy Claude cleanup;
- keep manually configured OpenCode provider settings when selecting `current`;
- reset provider acknowledgements and pending actions when the Bedrock region or profile changes;
- update the outdated provider workflow documentation.

Validation:
- exact-head CI is fully green;
- all smoke, unit, integration, e2e, contract, security, and documentation checks passed;
- focused provider diagnostics passed 39/39 tests;
- deterministic packaging, typecheck, and lint passed.

The PR is mergeable and ready for re-review.

apackeer (changes requested):
Thanks for the continued work on this — the ownership-aware direction is the right one, and the opencode `previousProvider` removal and the attestation reset both check out end to end.

I reviewed the current head `1252de255722f89597f96a373aaa84230f2e7ac4` against `7998589725a932c43bb3bb870c665cc7c724d3df` (also the merge base). Alignment is established by #1098 and the prior maintainer reviews. No commits were pushed.

**[P0] Every Codex refresh writes an invalid `config.toml`** — `core/tools/aidlc-init.ts:3760-3763`. The section pattern's `(?=^\[|\s*$)` lookahead succeeds at every end-of-line under the `m` flag, so the lazy body stops after the first line of the staged section; the remainder (`sandbox_mode = …`) dangles after the substituted text and duplicates the key inside `[shell_environment_policy]`. A pristine install plus one `config --yes` (release source or project-copy source, no user edits) fails `tomllib` with `Cannot overwrite a value (at line 36)`, and `codex-cli 0.154.0` refuses to start: `Error loading config.toml … :36:1: duplicate key`. It compounds per run (`# writable_roots` lines: 2, 3, 4 across three refreshes). Because the merged bytes never equal the shipped bytes, `config.toml` is always in `regenerated`, so the planner (`:5502-5533`) also copies it without the ownership check and a `[tui]` edit is silently reverted on a `--from` refresh, where the equivalent `.claude/settings.json` edit correctly conflicts. Please match to the next header or EOF (for example `(?=^\[|(?![\s\S]))`) or split by lines, assert `merge(x, x) === x`, and parse the emitted TOML in the test — the focused suites pass at this head because none of them parses the result.

**[P1] Copy-channel Bedrock → `current`/`other`/`--reset` leaves Bedrock active and `--check` reports clean** — `core/tools/aidlc-init.ts:6596-6603` and `:3658-3673`. Without `--from`, `AIDLC_RUNTIME_ROOT`, or the compiled binary, the record-only sections stage from `copiedProjectSource`, so the staged `settings.json` is the project file; the keys `preserveClaudeProviderFields` removes from `currentEnv` are re-inherited from `stagedEnv`, and `clearClaudeProvider` (`aidlc-config-diagnostics.ts:1128-1140`) only knows the exact legacy signature. Running the command the generated `CLAUDE.md` gives — `bun .claude/tools/aidlc.ts config providers --provider current --yes` — after `--provider amazon-bedrock --region eu-west-1 --profile team` prints "current harness provider; project overrides removed" while `CLAUDE_CODE_USE_BEDROCK=1`, `AWS_REGION`, and `AWS_PROFILE` stay in `.claude/settings.json`, and `--check` exits 0 "clean". `--provider other --acknowledge` and `--reset` behave the same (`--reset` then reports "the shipped fallback is in use"). With `AIDLC_RUNTIME_ROOT` set the transition works, which is why t294 passes: its CLI cases all run with `runtimeEnv()`. This is the headline symptom of #1098. Please remove the previous record's values from the staged env the way `clearOpenCodeProvider(previousProvider)` does for opencode (`:1292-1310`), and add coverage for the copy-source path.

**[P1] Preserved Claude fields still bypass ownership for the rest of the file** — `core/tools/aidlc-init.ts:3643-3653`. Copying every top-level key except `companyAnnouncements|permissions|statusLine|env|hooks` means a user `env` addition marks the file `regenerated`, after which a `permissions.allow` entry or a user `PreToolUse` hook in the same file is deleted on an ordinary refresh with exit 0 and no `--force`. The same edit without the env addition conflicts correctly (exit 4). This is the previous round's finding narrowed rather than resolved. The comment on `preserveKiroMcpRegion` (`:3987-3996`) already states the fix shape: run preservation after the `regenerated` scan so the staged file equals the current one only when the preserved fields are the sole difference, and any other edit still meets the ownership check.

**[P2] `--check` misses a project-level Bedrock flag unless the full legacy signature is present** — `core/tools/aidlc-config-diagnostics.ts:1817-1828`. With record `other` and a hand-set `CLAUDE_CODE_USE_BEDROCK=1` plus `AWS_REGION` in `.claude/settings.json`, `--check` exits 0 "clean", while the same keys in `settings.local.json` produce `provider-claude-local-override`. This fired at `858284c`; the signature narrowing regressed it, and #1098's acceptance line "records and effective harness settings cannot disagree while `--check` exits 0" is not met for the AI-DLC-managed project file. A warning mirroring the local-override one would be enough.

**[P3]** `docs/guide/15-troubleshooting.md:43` still recommends the `unchanged` provider answer; `18-install-and-lifecycle.md` was updated but this row was missed.

Verified fixed from the last round: `AWS_PROFILE` and custom aliases survive legacy cleanup and a later refresh; a manually configured opencode Bedrock block is preserved under `current` and an AI-DLC-written one is removed; unrelated opencode and Kiro edits conflict instead of vanishing; a region or profile change resets both Codex attestations; a user `[mcp_servers.*]` table is carried intact (modulo the P0).

Validation: `bun scripts/package.ts` in a clean worktree at this head, 292 pass / 0 fail across t03, t140, t220, t239, t240, t243, t293, t294, t296, t299, t304, and t-active-space-includes, plus CLI reproductions for every finding above. Exact-head CI is green; `main` has moved 6 commits (`be94bde7`) with a clean trial merge.

Requested event: **REQUEST_CHANGES**.

leandrodamascena:
Thanks for the detailed review and concrete reproductions. I addressed all five findings and pushed the changes in `104421bc9`; the current merged head is `29f7da888`.

Changes:

- Codex refresh now preserves valid TOML table boundaries, remains byte-idempotent, and is parsed as TOML in the regression test.
- Copy-channel Claude transitions from Bedrock to `current`, `other`, or `--reset` remove the exact provider values associated with the previous record.
- User-owned Claude fields are preserved after regenerated-file classification, so an unrelated `permissions` or hooks edit produces an ownership conflict instead of being silently replaced.
- A project-level `CLAUDE_CODE_USE_BEDROCK=1` now produces a diagnostic warning when the recorded provider is `current` or `other`, even without the complete legacy signature.
- The troubleshooting guide now documents `current` instead of the obsolete `unchanged` provider answer.
- The branch was merged with the latest `main`.

Added regression coverage for:

- valid and byte-idempotent Codex refreshes;
- user-owned Codex provider fields;
- all three copy-channel Claude provider transitions;
- Claude environment preservation without bypassing ownership checks;
- partial project-level Bedrock overrides.

Validation:

- exact-head CI is fully green;
- all four unit shards passed;
- integration and e2e passed;
- deterministic packaging, typecheck, lint, build, security, and documentation checks passed;
- focused provider diagnostics passed 43/43;
- additional focused tests passed 167/167.

@apackeer @rafaelgsr @wowzoo, could you please re-review the current head?

apackeer (dismissed):
Reviewed `29f7da8882528d4123d3d370d621aadf1b9f779b` against merge base `be94bde7cae6f9e4d431d045e148c8d5d8eaa508`. Three functional issues still need changes.

1. **[P1] Preserve Claude permissions when Bedrock remains selected.** [core/tools/aidlc-init.ts:4014](https://github.com/awslabs/aidlc-workflows/blob/29f7da8882528d4123d3d370d621aadf1b9f779b/core/tools/aidlc-init.ts#L4014)

   Moving preservation after the regenerated-file scan fixes the `current` case, but a recorded Bedrock answer still changes the neutral staged settings before that scan. The whole file is therefore already marked regenerated. `preserveClaudeProviderFields` excludes `permissions` and `hooks`, and the planner copies the staged file without checking those user edits.

   Reproduction: install Claude, select `amazon-bedrock --region us-east-1`, add `permissions.deny = ["Bash(team-private-command:*)"]` and an unrelated environment key, then run an ordinary `config --from <matching release source> --yes`. The command exits 0, preserves the environment key, and deletes the deny rule without `--force`. This reproduces twice at the PR head; the same sequence at the merge base exits 4 and preserves the edit.

   Keep the ownership check for fields outside the provider change, including when an existing Bedrock record is reapplied. Extend the permissions regression to cover the Bedrock answer.

2. **[P1] Remove legacy Claude model aliases before discarding their ownership signature.** [core/tools/aidlc-init.ts:3663](https://github.com/awslabs/aidlc-workflows/blob/29f7da8882528d4123d3d370d621aadf1b9f779b/core/tools/aidlc-init.ts#L3663)

   For an existing recorded Bedrock installation, preservation removes `CLAUDE_CODE_USE_BEDROCK` and `AWS_REGION` from the current environment but carries forward all four shipped `ANTHROPIC_DEFAULT_*_MODEL` values. The subsequent `clearClaudeProvider` requires the complete legacy signature, including the two removed keys, so it no longer recognizes these model aliases.

   Reproduction using an actual merge-base installation: record Bedrock in `us-east-1`, then use the PR runtime to select `--provider current --yes`. The command exits 0 and removes the provider flag, but leaves values such as `ANTHROPIC_DEFAULT_OPUS_MODEL=global.anthropic.claude-opus-4-8[1m]`. `config providers --check` also exits 0 and reports clean. These project aliases still override the user's native model aliases with Bedrock-specific IDs after opting out.

   Identify and remove the exact legacy model values before removing the identifying provider fields, while preserving customized aliases. Add an upgrade case with an existing provider record; the current legacy-cleanup fixture has no such record.

3. **[P2] Clear the recorded OpenCode provider settings on reset.** [core/tools/aidlc-config-diagnostics.ts:1348](https://github.com/awslabs/aidlc-workflows/blob/29f7da8882528d4123d3d370d621aadf1b9f779b/core/tools/aidlc-config-diagnostics.ts#L1348)

   `preserveOpenCodeProviderFields` now carries the installed provider block into staging, but `applyConfigDiagnosticRecords` returns immediately when reset supplies a null provider record. Consequently `clearOpenCodeProvider(previousProvider)` never runs.

   Reproduction: install OpenCode, select `amazon-bedrock --region us-east-1 --opencode-default yes`, then run `config providers --reset --yes`. Reset exits 0 and deletes the record but leaves `provider.amazon-bedrock.options.region` in `opencode.json`. `--check` exits 0 and says the shipped fallback is in use. This reproduces twice at the PR head; the same reset removes the block at the merge base.

   Run ownership-aware cleanup using the previous record before dropping it on reset, and cover this transition for OpenCode.

One documentation correction: **[P3]** [harness/codex/onboarding.fills.ts:38](https://github.com/awslabs/aidlc-workflows/blob/29f7da8882528d4123d3d370d621aadf1b9f779b/harness/codex/onboarding.fills.ts#L38) tells users that project-level “model or provider keys” can intentionally override user configuration. Current [official Codex documentation](https://learn.chatgpt.com/docs/config-file/config-reference#configtoml) explicitly says project `model_provider` and `model_providers` are ignored. Keep the project-model guidance, but direct provider changes to user configuration as the rest of this PR does.

Validation: packaging and the determinism check passed. The existing focused smoke/unit suite passed 446 assertions across 19 files, including 43 provider-diagnostics cases. Three separate targeted regression cases failed identically on two runs at the PR head; the permission-preservation and OpenCode-reset controls passed at the merge base. An actual merge-base-to-PR upgrade independently confirmed the retained Claude aliases on two runs. No live-provider or e2e tests were run. No source changes or commits were made.

Requested event: **REQUEST_CHANGES**.

wowzoo (dismissed):
https://github.com/awslabs/aidlc-workflows/pull/1101

Thanks for working through all five of my earlier findings, and for naming the commit for each
one — checking them against the code was quick because of that.

I reviewed head `29f7da8882528d4123d3d370d621aadf1b9f779b` against merge base
`be94bde7cae6f9e4d431d045e148c8d5d8eaa508`. `main` is now `c8262851e`, six commits ahead of that
merge base; `git diff --name-only be94bde7 c8262851e` names none of
`core/tools/aidlc-config-diagnostics.ts`, `core/tools/aidlc-init.ts`,
`harness/*/onboarding.fills.ts` or the Claude/Codex/opencode harness trees.

**All five findings from my earlier review are resolved.** Each reproduces as fixed. Every
sequence below ran against the packaged head release with `AIDLC_RUNTIME_ROOT` pointed at it, a
sandboxed `HOME`, and a throwaway project directory.

- **The codex cleanup no longer removes values on a signature it cannot match.**
  `clearCodexProvider` (`core/tools/aidlc-config-diagnostics.ts#L1145-L1174`) is gated on the full
  shipped signature and each removal is pinned to the shipped value. On a merge-base install
  carrying the exact shipped block, `providers --provider current --yes` exits 0 and removes all
  five provider lines, and `--check` exits 0. An edit inside
  `[model_providers.amazon-bedrock.aws]` — `profile`, `region`, plus an added `max_attempts` —
  leaves the signature intact but reaches the ownership check instead: the same command exits 4
  with `1 config conflict(s): .codex/config.toml (locally modified or unowned)` and the user's
  values survive.
- **A codex Bedrock record is no longer described as clean.** With a sandboxed `HOME` containing
  no `~/.codex/config.toml`: opting in, then `--mark-done bedrock-model-access --mark-done
  codex-provider-configuration --yes`, then `--check` exits 0 with `providers configuration has
  1 warning(s) for codex: provider-codex-self-attested`. The warn tier at
  `core/tools/aidlc-init.ts#L1527-L1539` keeps the exit code while changing the claim.
- **The Claude project-local layer is read.** With `env.CLAUDE_CODE_USE_BEDROCK=1`, `AWS_REGION`
  and an OPUS pin in `.claude/settings.local.json`, `--provider other --acknowledge --yes` exits
  0 and `--check` then exits 0 reporting `provider-claude-local-override
  (CLAUDE_CODE_USE_BEDROCK, ANTHROPIC_DEFAULT_OPUS_MODEL, AWS_REGION)`. `--show` carries the same
  line and project `doctor` shows `warn Providers: 1 warning(s)`. One scope note, since my earlier
  review raised it and it is unchanged: the probe at `#L1817-L1858` reads project `settings.json`
  and `settings.local.json`, so the user and enterprise-managed files stay outside provider
  diagnostics.
- **The Codex precedence sentence now states its condition.**
  `harness/codex/onboarding.fills.ts#L38` scopes precedence to a trusted project. I re-measured
  `model` on `codex 0.155.0.514`: with a sandboxed `CODEX_HOME`, a trusted project's
  `.codex/config.toml` value is the effective one in `codex doctor --json`. @apackeer's narrower
  point about `model_provider` on that same line I could not settle here — see the end of the
  finding below.
- **`t140` no longer reads the ambient environment.**
  `tests/unit/t140-sdk-drive-model-resolution.test.ts#L44-L56` asserts on the shipped
  `harness/claude/settings.json` instead of `resolved.env`. `bun test <that file>`, single-file
  mode, in a shell exporting `CLAUDE_CODE_USE_BEDROCK=1` and `ANTHROPIC_DEFAULT_OPUS_MODEL`:
  3 pass / 0 fail, and the same with those values forced on the command line. That is the
  configuration that made it red before.

`bun test tests/unit/t294-config-diagnostics.test.ts` at this head is 43 pass / 0 fail with 381
expect() calls, single-file mode.

One new finding, which comes out of the first fix rather than being independent of it.

---

### [P2] `--show` reports a project config as provider-neutral when the opt-out deliberately left Bedrock values in it, and `--check` and `doctor` report no disagreement

`hasLegacyCodexProviderConfig` (`core/tools/aidlc-config-diagnostics.ts#L1177-L1185`) is an
all-or-nothing predicate over six elements — an anchored comment block, four exact key/value
lines, and the `[model_providers.amazon-bedrock.aws]` header — and it gates two things at once:
the cleanup returns early at `#L1151`, and the `current`/`other` probe in `providerSurfaceIssues`
reports only inside the same predicate at `#L1862`. Searching the authored and generated trees
(`rg -uuu -n "hasLegacyCodexProviderConfig|clearCodexProvider|provider-codex-project-override"`,
excluding `.git/` and `node_modules/`, exit 0) returns the definition, those two call sites, the
dispatch at `#L1352`, and verbatim copies under `dist/` and `dist-release/` — no third reporter.
`--check` (`core/tools/aidlc-init.ts#L1513-L1519`), `--show` (`#L1398-L1421`) and the doctor row
(`aidlc-config-diagnostics.ts#L2518-L2523`) all read the same `providerIssues`. So any divergence
from those six elements that still passes the ownership check is a config that is neither cleaned
nor mentioned.

The instance I could reproduce is a project that changed one of the four provider values — the
kind of edit the refresh preserves by design. From a merge-base install with a recorded
`amazon-bedrock` answer, changing only `model_reasoning_effort` from `"high"` to `"medium"`, then
the head runtime:

```text
providers --provider current --yes  → exit 0, record becomes {"provider":"current"}
.codex/config.toml still contains   → model = "openai.gpt-5.5"
                                      model_provider = "amazon-bedrock"
                                      model_context_window = 1000000
                                      [model_providers.amazon-bedrock.aws]
providers --check                   → exit 0, "providers configuration is clean for codex"
providers --show                    → lists that same file as
                                      "provider-neutral Codex project configuration"
aidlc doctor                        → "0 problems, 3 warnings", no provider row
config --from <head release> --yes  → exit 0, the lines are still there
```

Three separate runs of that sequence, same exit codes and same output each time.

I want to be precise about what I am and am not reporting, because the non-removal is your
documented behaviour: `docs/guide/18-install-and-lifecycle.md#L484-L485` says `--provider current`
"removes only exact legacy AI-DLC Bedrock defaults", and preserving an edited provider value is my
earlier finding's fix working as intended. What is wrong is the reporting on top of it, and the
strongest part is `--show`: the string "provider-neutral Codex project configuration" is a fixed
label chosen from `record.provider` alone at `core/tools/aidlc-config-diagnostics.ts#L1381-L1391`
and never from the file, so it asserts neutrality about a file containing
`model_provider = "amazon-bedrock"`. `--check` and `doctor` are the weaker half: they do not claim
neutrality in those words, they report the disagreement as absent.

Against #1098 the clause this misses is the acceptance criterion *"Provider records and effective
harness settings cannot disagree while `--check` exits 0"* (line 59). I flagged in my earlier
review that #1098's Expected-behavior scoping to *"AI-DLC-owned provider settings"* (line 51)
reads narrower than that criterion; this case sits in the gap, and the narrow reading does not
excuse it either, since the surviving bytes are AI-DLC-shipped. I am keeping this at the same
severity as the two disagreements you already closed in this PR with the warn tier, because it is
the same class: a recorded answer and a project file that disagree while the tool reports success.

The outcome I would ask for is that a config the cleanup declines to rewrite not be described as
provider-neutral or clean. I am deliberately not proposing a mechanism: the two obvious ones are
the pair this PR is already balanced between — widening the removal patterns is what my earlier
finding was about, and pinning them to the current shipped bytes is what produces this case.

On coverage, `tests/unit/t294-config-diagnostics.test.ts#L516-L556` pins the two endpoints — an
exact shipped signature (detected, cleaned, no issues afterwards) and a fully custom config
(preserved) — and nothing between them, which is why CI is green here. A case whose config
diverges from the signature in one provider value would pin it.

Two things I could not establish, in case they save you time:

- Whether the surviving `model_provider = "amazon-bedrock"` is honoured by Codex at project level.
  On this machine a managed policy makes `codex doctor --json` report
  `model provider = amazon-bedrock` regardless of what either the user or the project layer sets,
  so both arms of that control agree on a value neither layer supplied. That is also why this is
  P2 and not P1: what I can measure is the stale `model` and `model_reasoning_effort` pin, not a
  routing change. It bears on @apackeer's P3 on the same subject.
- Whether an install predating the provider machinery reaches this state without any user edit. I
  could not get there by any path: at `29a31f789` (the commit before #546) there is no
  `core/tools/aidlc-init.ts` and no `config` command to have recorded an answer; copying that
  release's distribution into a project and running the head's provider commands exits 2 with
  `harness.json: distribution must name a supported harness`; the head's install over that copied
  tree exits 4 on an unrelated stale stage file (`application-design.md: schema validation failed:
  reviewer requires review_artifact`); and dropping only that release's `config.toml` into a
  modern install exits 4 with the same `1 config conflict(s)`. So this finding needs an edited
  provider value, not merely an old install — but I would not treat that as settled if you know an
  upgrade path I did not try.

---

For scope: this is the codex sibling of the Claude alias survival in @apackeer's review of this
head, which I reproduced independently while controlling this one — breaking one
`ANTHROPIC_DEFAULT_*` value makes `preserveClaudeProviderFields`
(`core/tools/aidlc-init.ts#L3663-L3694`) drop `CLAUDE_CODE_USE_BEDROCK` and `AWS_REGION` while
carrying the four aliases forward, after which the `#L1826` warn branch no longer fires and
`--check` exits 0 reporting clean. opencode is not affected on this path: `--provider current
--yes` clears the `amazon-bedrock` provider block from `opencode.json` and `--check` exits 0.

The check rollup on this head reads 22 `SUCCESS` and 1 `SKIPPED` with no failures, and the
contributor-statement check is among the green ones. My earlier review was recorded against
`858284cee`; this replaces it.

Requested event: **REQUEST_CHANGES**. I will re-review the next head.

PR #1106, opened 2026-09-10
fix: discover stage-level questions file for per-unit stages under stage-level scopes (#1105)

### LINKED ISSUE #1105: [Bug]: `summaryQuestionFiles` ignores stage-level per-unit artifacts, blocking review/completion of per-unit stages under infra
### Description

Under a scope that skips `units-generation` (e.g. **infra**), a per-unit construction stage (e.g. `nfr-requirements`) cannot be reviewed or completed. The engine refuses every review request and every completion with `SUMMARY_QUESTIONS_MISSING`, even though the stage's questions/artifacts were correctly written at the stage-level path. No recovery action clears it — the `reconfirm-summary` / guard-recovery loop repeats the same refusal indefinitely (observed after 5+ attempts, triggering the guard-recovery `ask`).

Observed error:
  ```
  {"error":"Cannot start review for \"nfr-requirements\": its question flow has no
  nfr-requirements-questions.md file. Create and answer the stage questions, then
  record the consolidated summary checkpoint before generating artifacts. ...
  guard state for \"nfr-requirements\" has refused review-request 5 times ..."}
  ```
Root Cause

There is an asymmetry between the completion/summary guard and the question-file discovery helper it depends on, both in `core/tools/aidlc-lib.ts`:
  - `usesStageLevelPerUnitArtifacts(scope, stateContent)` (line **26501**) returns `true` exactly when the approved plan does **not** EXECUTE `units-generation`. For **infra** this is true, so a per-unit stage's artifacts live at the **stage level** — `construction/<slug>/` (e.g. `construction/nfr-requirements/`) - not under per-unit subdirs.
  - The completion guard **correctly** consults `usesStageLevelPerUnitArtifacts` (line **8277**) and skips the unit-set requirement for this case.
  - But `summaryQuestionFiles(...)` (line **7813**) — the helper that actually finds the `*-questions.md` file — does **not** consult `usesStageLevelPerUnitArtifacts`. For any per-unit stage it unconditionally iterates `construction/<unit>/<slug>/` (loop at line **7827**).
  
So under infra: artifacts are written at `construction/nfr-requirements/` and the guard's unit-set check is correctly skipped, but file discovery only looks under `construction/<unit>/nfr-requirements/`, finds nothing, and the guard falls through to `SUMMARY_QUESTIONS_MISSING` (line **8310**). Because the questions file
is never discoverable at the searched path, no summary/reconfirm recovery can satisfy the guard.

### Steps to Reproduce

1. Run an AI-DLC workflow under **infra** scope (any scope where the approved plan SKIPs `units-generation`, so `usesStageLevelPerUnitArtifacts` is true).
2. Reach a per-unit construction stage such as `nfr-requirements`; create and answer its questions and write artifacts (they land at the stage level, `construction/nfr-requirements/`, which is correct for this scope).
3. Request a review, e.g.: `aidlc engine log review --stage nfr-requirements --reviewer aidlc-architecture-reviewer-agent --iteration 1`
4. Observe the refusal. Attempt the `reconfirm-summary` guard-recovery action.

### Expected Behavior

The review/completion guard discovers the stage-level `nfr-requirements-questions.md` at `construction/nfr-requirements/` and allows the review to proceed, consistent with the guard already skipping the unit-set requirement for stage-level per-unit scopes.

### Actual Behavior

`summaryQuestionFiles` searches only `construction/<unit>/nfr-requirements/`, finds no questions file, and the guard returns `SUMMARY_QUESTIONS_MISSING`. Every review request and the `reconfirm-summary` recovery repeat the identical refusal; the stage cannot proceed.

Proposed fix
Mirror, in `summaryQuestionFiles`, the stage-level branch that already exists in the guard three functions up. 
In `core/tools/aidlc-lib.ts` (authored source; the per-harness `dist/<harness>/.../tools/aidlc-lib.ts` copies are regenerated by `bun scripts/package.ts`), after the `isPerUnitStage(stage)` check, resolve the scope from state and short-circuit to the stage-level directory when `usesStageLevelPerUnitArtifacts(scope, stateContent)` is true:
  ```ts
  // inside summaryQuestionFiles, per-unit branch
  if (usesStageLevelPerUnitArtifacts(scope, stateContent)) {
    return questionFilesInDir(join(rec, "construction", stage.slug), null);
  }
  ```
  
This is a one-branch change mirroring logic that already exists in the guard. Precedent for exactly this pattern lives in `core/tools/aidlc-state.ts` around lines 2687–2691, which already returns `[join(rec, "construction", stage.slug)]` when `usesStageLevelPerUnitArtifacts` is true — `summaryQuestionFiles` is simply missing the same branch. A regression test should cover a per-unit stage under a `units-generation`-SKIP scope resolving its stage-level questions file.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.8.2

### AI-DLC Phase

Construction

### Harness

Kiro CLI

### AI Model

Auto

### Environment

_No response_

### Additional Context

_No response_

chiliec:
# Summary

Closes #1105. Under a scope whose approved plan **SKIPs** `units-generation` (e.g. **infra**), a per-unit construction stage keeps its artifacts at the **stage level** (`construction/<slug>/`) rather than under per-unit subdirs. The summary-confirmation completion guard already handles this, but its question-file discovery helper did not — so review/completion of such a stage was refused forever with `SUMMARY_QUESTIONS_MISSING`.

## Changes

In `core/tools/aidlc-lib.ts`:

- `summaryQuestionFiles()` now takes the `stateContent` and, for a per-unit stage under a scope where `usesStageLevelPerUnitArtifacts(...)` is true, looks for the `*-questions.md` file at the stage level (`construction/<slug>/`) — mirroring the check the completion guard already applies at line ~8277. Every other lane is unchanged: non-per-unit stages and per-unit stages under scopes that run `units-generation` still iterate `construction/<unit>/<slug>/`.
- `checkSummaryConfirmationEvidence()` threads `options.stateContent` into that helper.

The asymmetry root cause is exactly as described in the issue: the guard consulted `usesStageLevelPerUnitArtifacts`, the discovery helper did not.

## User experience

**Before:** Running an AI-DLC workflow under `infra` scope and reaching a per-unit stage such as `nfr-requirements` — with the questions/artifacts correctly written at `construction/nfr-requirements/` — every review-request and completion is refused with `SUMMARY_QUESTIONS_MISSING`, and no recovery clears it.

**After:** The stage-level questions file is discovered, so the guard advances normally to the receipt/answer checks and the stage can be reviewed and completed.

## Checklist

* [x] Changes are documented
* [ ] If this change adds an input to any fingerprint, epoch, or receipt identity, the description names the human-visible change it detects — N/A (no identity input added)

## Test plan

New unit test `tests/unit/t338-summary-questions-stage-level.test.ts`:

- **infra scope (units-generation SKIP)** with a stage-level questions file present → the guard no longer returns the "its question flow has no …" refusal.
- **a scope that runs units-generation (EXECUTE)** with the same stage-level file present → still refuses (the per-unit lane is untouched), proving the fix is scoped.

Verified locally with bun 1.4.2:

```
$ bun test tests/unit/t338-summary-questions-stage-level.test.ts
 2 pass  0 fail

# RED→GREEN: revert the source fix (keep the test) →
#   FAIL: Received "Refusing to complete \"nfr-requirements\": its question flow has no
#   nfr-requirements-questions.md file. …"  (the exact issue #1105 refusal)
# restore the fix → 2 pass.

$ bun run typecheck        # tsc across the 3 project configs → clean
$ bun x @biomejs/biome check core/tools/aidlc-lib.ts tests/unit/t338-...test.ts  # No fixes
$ bun tests/gen-coverage-registry.ts   # registry updated for the new test; --check gate green
```

Related summary/guard suites stay green with the change (t310, t320, t332, t335, t280). Two suites (`t188-human-presence-gate`, `t186-foreach-per-unit-iteration`) have pre-existing failures on a clean `main` checkout on my machine, unrelated to this diff — I confirmed they fail identically with the change stashed.

Happy to adjust naming or the test shape if you'd prefer it folded into an existing summary-confirmation suite.

## Acknowledgment

### DIFF
--- a/core/tools/aidlc-lib.ts
+++ b/core/tools/aidlc-lib.ts
@@ -7813,10 +7813,19 @@ function questionFilesInDir(
 function summaryQuestionFiles(
   projectDir: string,
   stage: SummaryConfirmationStage,
+  stateContent?: string | null,
 ): SummaryQuestionFile[] {
   const rec = recordDir(projectDir);
   if (rec === null) return [];
   if (!isPerUnitStage(stage)) {
+    return questionFilesInDir(join(rec, stage.phase, stage.slug), null);
+  }
+
+  if (
+    stateContent !== undefined &&
+    stateContent !== null &&
+    usesStageLevelPerUnitArtifacts(getField(stateContent, "Scope"), stateContent)
+  ) {
     return questionFilesInDir(join(rec, stage.phase, stage.slug), null);
   }
 
@@ -8257,7 +8266,7 @@ export function checkSummaryConfirmationEvidence(
     return { ok: true, required: false };
   }
 
-  let questions = summaryQuestionFiles(projectDir, stage);
+  let questions = summaryQuestionFiles(projectDir, stage, options.stateContent);
   if (options.unit !== undefined) {
     questions = questions.filter(
       (question) => question.unit === options.unit,
--- a/tests/.coverage-registry.json
+++ b/tests/.coverage-registry.json
@@ -1721,6 +1721,10 @@
         {
           "file": "tests/unit/t335-change-control-review-summary.test.ts",
           "mechanism": "cli"
+        },
+        {
+          "file": "tests/unit/t338-summary-questions-stage-level.test.ts",
+          "mechanism": "none"
         }
       ],
       "status": "covered"
--- a/tests/unit/t338-summary-questions-stage-level.test.ts
+++ b/tests/unit/t338-summary-questions-stage-level.test.ts
@@ -1,0 +1,125 @@
+// covers: function:summaryQuestionFiles function:checkSummaryConfirmationEvidence
+//
+// t338 - under a scope whose approved plan SKIPs units-generation (e.g. infra),
+// a per-unit construction stage keeps its artifacts at the STAGE level
+// (construction/<slug>/), not under per-unit subdirs. The completion guard
+// already consults usesStageLevelPerUnitArtifacts and skips the unit-set
+// requirement for this case, but its question-file discovery helper
+// (summaryQuestionFiles) did not, so it only looked under
+// construction/<unit>/<slug>/, found nothing, and fell through to
+// SUMMARY_QUESTIONS_MISSING forever (issue #1105). This pins that the helper
+// discovers the stage-level questions file in that mode while leaving the
+// per-unit lane (a scope that DOES run units-generation) untouched.
+
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdirSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import {
+  checkSummaryConfirmationEvidence,
+  loadStageGraphAll,
+} from "../../dist/claude/.claude/tools/aidlc-lib.ts";
+import {
+  cleanupTestProject,
+  createTestProject,
+  seedAidlcMemory,
+  seededRecordDir,
+  seedStateFile,
+} from "../harness/fixtures.ts";
+
+const STAGE = "nfr-requirements";
+const tempDirs: string[] = [];
+
+afterEach(() => {
+  while (tempDirs.length > 0) cleanupTestProject(tempDirs.pop()!);
+});
+
+function withSummaryGuard<T>(fn: () => T): T {
+  const prior = process.env.AIDLC_SKIP_SUMMARY_CONFIRMATION_GUARD;
+  delete process.env.AIDLC_SKIP_SUMMARY_CONFIRMATION_GUARD;
+  try {
+    return fn();
+  } finally {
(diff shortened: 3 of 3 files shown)


apackeer:
Superseded by #1110, merged as 799858972. Your root cause was exactly right: the completion guard consulted `usesStageLevelPerUnitArtifacts` and the discovery helper did not. #1110 lands the same stage-level branch with one extra guard so isolated `--single` stage reviews keep per-unit discovery, and its t332 cases include the negative case your t338 test proposed (a stage-level file under a units-executing scope is ignored). #1105 is closed against that merge. Thank you for the independent diagnosis and the red-to-green test.

PR #1107, opened 2026-09-10
fix: preserve existing gitignore rules and clarify config filesystem errors

apackeer:
# Summary

Existing `.gitignore` files containing `aidlc/` rules or AI-DLC comments could block `aidlc config` as ambiguous legacy content. On filesystems that reject hard links, setup also failed late with raw JSON and an unrelated release-data remedy.

Config now preserves existing ignore rules and detects rejected transaction locking before the first-run setup questions.

## Changes

- Preserve unknown unmarked `.gitignore` content and append a separately owned AI-DLC block. Keep exact legacy adoption, marker validation, and other root-file ownership checks.
- Refuse lossy decoding of a non-UTF-8 `.gitignore` without changing its bytes.
- Probe hard-link creation using private temporary paths, preserve real transaction locks, and clean up probes even when closing a file fails.
- Report `EMLINK`, `ENOTSUP`, `EOPNOTSUPP`, and `ENOSYS` with storage remediation during both preflight and transaction application.
- Render the child command's error message and remediation in the wizard instead of its entire JSON plan.
- Document filesystem requirements and the existing-file behavior.

## User experience

A project with an ordinary unmarked `.gitignore`, including AI-DLC rules, can complete config without renaming that file. Repeated refreshes preserve its user-owned content.

If the destination rejects hard-link locking, first-run setup stops before collecting choices and identifies the required storage capability. For the reported S3-backed mount, the workaround remains keeping the workspace outside that mount, for example on ext4 or XFS on EBS. This PR does not add alternative locking or S3 filesystem support.

## Checklist

* [x] Changes are documented

## Test plan

- `bun scripts/package.ts`
- `bun run typecheck`
- `bun run lint`
- Targeted unit slices on the rebased commit, using `bash tests/run-tests.sh --debug -P 8 --unit --filter ...`: **182 passed, 0 failed** across `t243-install-mechanism`, `t299-first-run-wizard`, `t330-transaction-filesystem`, and `gen-coverage-registry`.

The focused suites cover dry-run/apply/refresh preservation, malformed markers, invalid encoding, concise human/JSON/quiet failures, live/dead lock ownership, and probe cleanup failures. Filesystem errors are simulated in isolated subprocesses around real file operations; no live S3 mount was exercised.

## Acknowledgment


wowzoo (changes requested):
https://github.com/awslabs/aidlc-workflows/pull/1107

Reviewed head `5b0c2e70` against base `a8f2dc61`. `main` has since moved to `77116239`, thirteen commits ahead of this base, which matters for the failing check.

## The failing check is not yours

`Tests (integration + e2e, deterministic)` fails on `tests/integration/t328-authority-rebinding.test.ts`. Single-file mode, three arms:

| tree | result |
|---|---|
| base `a8f2dc61`, without this PR | 19 pass / 3 fail |
| head `5b0c2e70` | 19 pass / 3 fail — the same three cases |
| `main` `77116239` | 22 pass / 0 fail |

The three are `the approval survives parking and resuming`, `an approval survives unit start and authorizes the developer dispatch`, and `after a Unit is paused or completed, the engine and the Stop hook agree it is not active`. `aadc8323` (#1097) adds the state-digest assertions in that file and is not an ancestor of this head, so a rebase onto `77116239` should clear it. #1108 will need rebasing behind it.

## P2, blocking — the ambiguity gate was the only thing reporting a blanket `aidlc/` rule

`core/tools/aidlc-init.ts` now reads `if (path !== ".gitignore" && /\baidlc\b|AI-DLC/i.test(current))`. Running the real command on both trees, in a fresh git repo whose `.gitignore` is the second preservation fixture from this PR (`"# AI-DLC output owned by this project\naidlc/\n!aidlc/README.md\n"`):

```
base a8f2dc61   exit 4   1 config conflict(s): .gitignore (legacy root integration ambiguous;
                         move or delete the unmarked AI-DLC content)
                         managed block not written

head 5b0c2e70   exit 0   configured … for Claude Code 2.8.2; next: … run `/aidlc --doctor`
                         managed block written, the aidlc/ line preserved
```

In both trees `git check-ignore` reports `intents.json`, `aidlc-state.md`, `audit/<host>-<clone>.md` and `memory/notes.md` as ignored — the user's own rule does that, not this PR. What changes is that the conflict was the only thing telling them. The managed block carries no rule re-including anything under `aidlc/`, and gitignore cannot re-include content beneath an excluded parent, so appending the block cannot undo it.

Nothing downstream reports it either. The success message points at `--doctor`, whose W1 row exists to "surface uncommitted changes under `aidlc/` so the shared records get committed and pushed to teammates" — and its comment notes that `git status --porcelain` omits gitignored paths so per-user runtime state never false-positives. That property also hides this. Measured in the configured fixture:

```
git -c status.showUntrackedFiles=all status --porcelain=v1 --untracked-files=all -- aidlc
  → 0 lines            (W1: "nothing to commit", green)

same tree, aidlc/ line removed
  → 12 lines
```

So config reports success, the doctor row it recommends reports clean, and the state, registry, audit shards and memory the block itself lists as `COMMITTED (NOT ignored … do NOT add ignore rules for these)` never reach a teammate. Nothing on disk is lost; the sharing contract breaks quietly, and only once work has accumulated.

The added tests do not cover it. Four preservation fixtures, two carrying a blanket `aidlc/`, and the assertions check bytes, marker count, manifest ownership and refresh behaviour — none checks how git resolves the resulting file.

The constraint, rather than a prescription: after config, either the committed-by-design record paths are trackable, or the user is told they are not. Where that belongs — the conflict list, an advisory row, a narrower exemption, the doctor — is your call.

## P3 — a probe-cleanup failure erases the storage remediation

`assertTransactionFilesystem` wraps a failed cleanup in an `AggregateError`, which drops the `TransactionLockError` type. `runFirstRunWizard` only recognises `TransactionLockError`, so it rethrows, and the top level prints a bare `aidlc config: <message>`; the `fix:` line with the storage guidance never appears. The PR's own test asserts that shape (`transactionLockError: false` for the aggregate), so it is deliberate as far as the type goes — but the case where the probe cannot be removed is most likely on exactly the misbehaving storage the remediation is for.

## On `EPERM`

Linux `link(2)` documents `EPERM` for a filesystem that does not support hard links, and both CI and the reported EC2 mount are Linux. This PR deliberately keeps it raw — `t330` asserts, for probe and apply alike, that `EACCES`, `EPERM`, `EIO` and `EXDEV` preserve the original error with no remediation — and #1108 classifies it only where it feeds the directory fallback rather than an abort. That reads like a considered split rather than an omission.

The question it leaves: if the reported mount answers `EPERM`, the preflight in this PR will not fire for it, and the confusing failure this change replaces is what that user still sees until #1108 lands. I have not measured what that mount returns, so this is a sequencing question for the stack rather than a defect here.

## Reads clean

Child-error envelope parsing, `renderFirstRunFailure`, the non-UTF-8 round-trip refusal, and the probe's close-before-cleanup ordering. `acquireLock`'s retry does not swallow the new error, since the code is never `EEXIST`. The changed conflict remediation does not disturb the `config conflict(s)` assertion in `t299`. The probe is created under the project root for an existing project, which the tests assert.

## Requesting changes

On the P2. A rebase onto `77116239` also clears the failing check. Happy to re-review on the new head.

apackeer:
Rebased onto `main` (`a0ee4415e`, clean) - that clears the `t328` check as you predicted - and addressed both points.

**P2.** New shared helper `core/tools/aidlc-gitignore.ts` (`committedRecordIgnoreConflicts`) runs `git check-ignore -v --no-index` on five representatives of the patterns the managed block documents as COMMITTED (`memory/**`, `codekb/**`, `intents.json`, `aidlc-state.md`, `audit/*.md`). Because the managed block re-includes nothing, checking the current file before the merge is equivalent to checking the merged result, so dry-run and apply agree. Two consumers:

- `aidlc config`: a user-owned rule that hides any probe is now a config conflict on the same exit-4 path the ambiguity gate used, naming the rule and the records. Your fixture (`# AI-DLC output owned by this project` / `aidlc/` / `!aidlc/README.md`) now yields:
  `error: 1 config conflict(s): .gitignore (.gitignore:2: aidlc/ hides committed workflow records (memory/**, codekb/**, intents.json, aidlc-state.md, audit/*.md); narrow the rule so teammates receive them)` - managed block not written, user bytes unchanged. An unmarked file mentioning AI-DLC that does *not* hide records (e.g. `aidlc/**/*.log`) still completes with its content preserved, which is what the PR set out to allow. Outside a git repo or without `git` the check is skipped, and:
- `/aidlc --doctor` gains a `Workspace record visibility` row beside W1 (advisory, same silent skip outside git), so a rule added after config is reported where W1 is blind.

Negative proof: with the helper short-circuited to `[]`, the fixture test fails `expect(refused.status).toBe(4) - Expected: 4, Received: 0` (block written); restored, exit 4 and bytes unchanged.

**P3.** A probe-cleanup failure now rethrows a `TransactionLockError` (same root/code, remediation intact) whose message carries both the hard-link sentence and `Could not remove temporary transaction lock probe …`, with `cause` the `AggregateError` of the lock, close, and cleanup errors. The wizard renders the `fix:` storage line for that case; t330's earlier `transactionLockError: false` pin is replaced by the preserved-type contract. `EPERM` stays raw per your note; that's #1108's seam.

Verified on the rebased head: t243 97, t256 13, t299 20, t330 32, gen-coverage-registry 37 - 199 pass / 0 fail; `bun scripts/package.ts --check`, `bun run typecheck`, `bun run lint`, `git diff --check` exit 0. #1108 will need rebasing onto this head.

Issue #1122, opened 2026-09-10
feat: warm re-scan / no re-interrogation on iteration (reuse CodeKB + prior answers)


SaRedfiche:
### Problem

On any iteration over an existing workspace/intent, AI-DLC re-runs a cold Initialization scan and re-interrogates the user with questions they already answered. A customer reported a 45-minute re-scan that re-asked answered questions for what was ultimately an 11-line fix. This is the "reuse prior workspace understanding" half of the last-30% iteration gap.

### What already exists (verified against source)

The code-understanding side is largely built and should be reused, not rebuilt:

- Durable space-level CodeKB store `aidlc/spaces/<space>/codekb/<repo>/` (`core/tools/aidlc-lib.ts` `codekbDir`).
- Git-tree-hash freshness fingerprint (`codekbScopeFingerprint` / `codekbSourceFingerprint`) and a `scope_version:1` freshness block in `reverse-engineering-timestamp.md`.
- A `CURRENT/STALE/...` verdict (`codekb-scope-diff`) + a reuse gate in the reverse-engineering stage (Reuse / Full rescan / Focused scan), so the `--rescan full` escape hatch effectively exists already as "Full rescan".
- `propagateStageInvalidation` (`core/tools/aidlc-validity.ts`) already marks downstream artifacts `stale`/`needs-revalidation` **without resetting completed stages** (#716 mechanism).

### The two real gaps

1. **Per-file delta, not store-wide.** The freshness verdict today is whole-scope `CURRENT/STALE` — one changed file flips the entire store STALE with no changed-file set, so there is no "re-scan only what changed". Add a delta computation (git diff of the store's recorded generation vs HEAD, on the existing fingerprint seam) that returns the changed-file set, and make the reverse-engineering developer brief accept it.
2. **Answered questions are not persisted as reusable Q→A pairs at all.** `QUESTION_ANSWERED` events are write-only audit rows read back only for within-attempt authorization — never to pre-fill a later run. A durable, provenance-stamped answer store + a pre-fill reader (invoked before a stage's question fences) is a from-scratch component. Surface "reused N prior answers"; a changed question or changed code must re-ask rather than silently reuse.

### Acceptance

- A second workflow over the same repo/intent completes Initialization by scanning only changed files and reports the cache hit + delta.
- Questions answered in the first run are not re-asked; the run reports how many prior answers were reused.
- `--rescan full` bypasses the cache and re-scans everything.

### Notes

[...]

PR #1125, opened 2026-09-11
feat: per-file changed_files delta on codekb-scope-diff STALE verdict

SaRedfiche:
## Why

Foundational piece of warm re-scan (#1122). Today `codekb-scope-diff`'s freshness verdict is whole-scope `CURRENT/STALE` — a single changed file flips the entire analyzed scope to `STALE` with no way to see *what* moved, so an iteration has to re-scan the whole tree. This adds the per-file delta an iteration needs to re-derive only what changed.

## What changed

- New exported helper `codekbScopeChangedFiles(repoDir, storedFingerprint, currentFingerprint)` in `core/tools/aidlc-lib.ts`. Both fingerprints are git tree objects (the existing temp-index `write-tree`), so the delta is a `git diff --name-only -z` between them. Returns `[]` when the two trees match, the changed-file list otherwise, and `null` when either token is absent/not a well-formed tree hash or git cannot diff them (caller falls back to a full rescan — never a silent empty delta).
- `handleCodekbScopeDiff` status mode surfaces `changed_files` on the `STALE` verdict only (JSON + human). The caller threads the `currentFingerprint` it already computed for the CURRENT/STALE decision into the helper — no second `write-tree`, and no window where a file reverts between two snapshots and yields `STALE` with an empty delta.
- Human output escapes each filename via `JSON.stringify`, so a repository filename containing ANSI/OSC control bytes cannot inject a terminal control sequence.
- Version `2.7.0` → `2.7.1` (`aidlc-version.ts` + CHANGELOG + README badge); `dist/` regenerated across all 7 harnesses.

Read-only and additive: the CURRENT/STALE/UNVERIFIED/UNKNOWN_SCOPE verdicts are unchanged; only the STALE payload gains the `changed_files` field.

## How verified

- `bun scripts/package.ts --check` (dist drift), `bun run typecheck`, `bun run lint` — all green.
- `tests/unit/t248-codekb-scope-diff.test.ts` extended and green: the delta contents, the empty-vs-null distinction, verbatim filenames, terminal-control-byte escaping, and the git-failure `null` fail-safe (a well-formed-but-absent stored tree hash — the cross-clone / rebased / gc'd-store case) at both helper and verb level.
- `tests/unit/t68-version-changelog-sync.test.ts` green (version/CHANGELOG/README lockstep); `scripts/ci-changelog-guard.ts` green (2.7.0 entry preserved).
- Pre-merge review gates: multimodel panel GO (GPT/Opus/Design PASS); adversarial crew GO across Security / AI-necessity / Correctness / Docs-honesty, with the Tests axis's `null`-fail-safe coverage gap fixed. Recorded in `docs/scan-log.md`.

## Upgrade notes

None for consumers of existing verdicts. A consumer that parses the STALE object gains an optional `changed_files` field (array, or `null` when the delta was not computable).


SaRedfiche:
**Review-ready — rebased onto current `main`, conflict cleared.**

This PR is now **MERGEABLE** (was CONFLICTING). Rebased onto the latest `main` and force-pushed; head is `d8b2af2b`, one clean commit, 0 behind.

**What the change does:** adds the warm CodeKB delta-rescan path — on a re-scan it compares the stored scope fingerprint against the current working tree and only re-ingests changed files, rather than a full cold rescan, using the existing `codekbScopeFingerprint` seam (which hashes on-disk content so rebases/amends don't spuriously invalidate it).

**Rebase resolution (5 conflicts, all lockstep/generated — no behavioral change):**
- Version files (`README.md`, `aidlc-version.ts`): took `main`'s `2.10.0`.
- `CHANGELOG.md`: folded the `changed_files` delta note into the current version section.
- Coverage baselines (`.coverage-ratchet.json`, `.coverage-registry.json`): took `main`'s baseline, then regenerated — function count `476 → 477` (this change adds one function on top of `main`), freshness guard green.

**Verification (deterministic subset, all green):**
- Delta's own tests (t248): 54 pass / 0 fail
- Scope + grid family (6 files): 87 pass / 0 fail
- Typecheck (all 3 tsconfigs): clean
- Coverage ratchet `--check`: OK (fresh, guards held)

Note: the full suite has pre-existing failures on pristine `main` (t208/t186/t118 and others), unrelated to this change — the deterministic subset above is scoped to what this PR touches.

CI lanes and the fork review workflows still need the maintainer's **"approve and run workflows"** click to dispatch (`UNSTABLE` reflects only that they haven't run yet, not a failure). Ready for review whenever you are.

PR #1133, opened 2026-09-11
fix: accept recommended option labels at stage gates

apackeer:
# Summary

Selecting `Approve (Recommended)` at a stage gate was refused even though the question-rendering guide instructs the conductor to add that label decoration. #1023 (@jisu-im-dev) landed the base fix on `main`: stage gates strip one trailing decorator before matching, and the received reply is preserved in approval audit records and refusal messages.

This PR carries the remaining delta over that merged tree: the shared Request Changes matcher accepts the decorator whether it sits inside or outside surrounding quotes and trailing punctuation, the regression tables cover the report and direct-state entrypoints and both composition directions, and the reference documents the engine-side contract.

Refs #1126.

## Changes

- Match the gate's Request Changes choice by shape (`REQUEST_CHANGES_CHOICE_RE`) instead of an ordered sequence of normalization passes, so the `(Recommended)` decorator composes with quotes and punctuation in both directions. No fixed ordering of the previous passes accepts both `"Request Changes (Recommended)"` and `"Request Changes" (Recommended)`.
- Preserve human-presence requirements and the three-revision threshold for `Accept as-is`.
- Cover report and direct-state entrypoints, raw reply preservation, invalid replies, revision limits, missing human turns, and the decorator inside and outside each wrapper. Document the stage-gate and Plan Approval matching contracts in `docs/reference/04-stage-protocol.md` and `docs/reference/12-state-machine.md`.

## User experience

Choosing `Approve (Recommended)` or `Request Changes (Recommended)` performs the selected gate action. `Accept as-is (Recommended)` works when that choice is available. A typed `Request Changes` reply is accepted whether the recommended label sits inside or outside the quotes and punctuation around it.

## Checklist

* [x] Changes are documented
* [ ] If this change adds an input to any fingerprint, epoch, or receipt identity, the description names the human-visible change it detects (not applicable)

## Test plan

- Reproduced the decorated-label refusal before applying the production fix, and the inside/outside composition regression before replacing the ordered passes.
- Passed all six files using `bash tests/run-tests.sh --unit` filtered to `t188-human-presence-gate`, `t205-gate-revision-backstop`, `t242-state-transition-guard`, `t324-team-unit-progress-gates`, `t328-plan-approval-runtime-authority`, and `t331-guard-deadlock-liveness`.
- Distribution regeneration, `bun scripts/package.ts --check`, `bun run typecheck`, and `bun run lint` passed on the merged head.
- Independent code review found no blocking defects. Coverage is deterministic CLI/unit testing; no live Codex session was run.

## Acknowledgment


wowzoo (changes requested):
https://github.com/awslabs/aidlc-workflows/pull/1133

Reviewed head `332b88b0` against base `main@9b8847b5`; `main` is now `a0ee4415`, so this is two commits behind.

This lands where #1126 pointed: `stripRecommendedDecorator` at the two gate-reply comparisons, and `isRequestChangesChoice` normalizing through the same helper instead of a bare `.trim()` — behaviour-preserving there, since the helper trims last. Two details worth crediting because they are easy to get wrong in the other direction: the orchestrate error now reports the **raw** reply rather than the stripped one, so a human sees what they actually sent, and `isNonAnswer` keeps the raw input so cancellation detection is untouched. The decorator instruction is also not Codex-specific — six harness question-rendering guides carry it — so the fix is worth more than the one reported harness.

## The one change I would not merge as written

`isRequestChangesChoice` normalizes in this order:

```
stripRecommendedDecorator(text)                  // anchored on $
  .replace(/^(?:[A-Za-z]|\d+)[.)]\s*/, "")       // option prefix
  .replace(/^["'`]+|["'`]+$/g, "")               // surrounding quotes
  .replace(/[.!]+$/, "")                         // trailing punctuation
  .replace(/\s+/g, " ")
```

The decorator strip is end-anchored and runs *before* the quote and punctuation removal, so anything trailing hides the anchor. Running the function as written:

```
"Request Changes (Recommended)"        true
"request changes (recommended)"        true
"B. Request Changes (Recommended)"     true
"2) Request Changes (Recommended)"     true
'"Request Changes (Recommended)"'      false
"'Request Changes (Recommended)'"      false
"Request Changes (Recommended)."       false
"Request Changes (Recommended)!"       false
```

The option prefix survives because it is at the start; quotes and punctuation sit where the anchor looks.

Two things to be fair about. The comment lists these tolerances individually and does not promise every combination — and non-composition predates this PR, since `'"B. Request Changes"'` was already rejected before it. And I found no documented harness that emits a quoted or punctuated decorated label: the gate path forwards the exact selected label, and the prose harnesses map a selection back to that label. So this is an edge for a manually typed or incorrectly forwarded reply rather than the picker path #1126 describes.

Still, the decorator is now one of the advertised tolerances and it behaves differently from the other three, which is the kind of asymmetry that gets rediscovered later. The constraint: the decorator should be removed in the same normalization pass as the quotes and punctuation, or after them, rather than ahead of the pass that strips what its anchor needs to see past — plus one combined regression case, since the table currently exercises each tolerance alone.

## Blast radius worth naming

`isRequestChangesChoice` is shared. Besides the gate-reply path it feeds guard-recovery selection in `core/tools/aidlc-lib.ts` and rejection validation in `core/tools/aidlc-state.ts`, so loosening it for a gate-reply reason also broadens what counts as Request Changes at those two sites. That may well be intended — they are the same human choice — but it is worth stating rather than discovering.

## The red checks

`Review lens (prompt-injection)` and `Review lens (security)` fail on this head, with annotations carrying only `Process completed with exit code 1` and no finding text. In that run `Build immutable review context` and `Start SHA-bound review check` succeeded, `Finalize interrupted review check` ran the interrupted path, and both `AIDLC adversarial review` and `Publish SHA-bound review` skipped — so no review was produced for you to answer. The lens step runs under `set -euo pipefail`, and the workflow is now `disabled_manually`.

The earlier green run on this branch is not a control either: it was on `2157cac6` rather than this head, and its own posted review says the read-only command sandbox failed before every command executed, so the diff, changed files, head snapshots and candidate lenses went uninspected and it reported "No findings" on that basis. Both the green and the red outcomes here are the workflow's own execution state rather than anything about this change.

## On the next push

Make the decorator stripping compose with the existing quote and punctuation normalization, and add a combined regression case. Everything else here reads correct, and the added coverage — the decorated approve commit, redundant decorated answers as no-ops on both the approve and reject paths, and the negative suffixes — is the right kind; it just does not cross the tolerances with each other.

wowzoo:
https://github.com/awslabs/aidlc-workflows/pull/1133

Bookkeeping I should have posted before reviewing this one: #1023 makes the same change and has been open since 2026-09-04, a week before this PR, authored by @jisu-im-dev — who also filed #1126 with the `file:line` citations both PRs act on.

The source edits are identical in both, including reporting the raw reply in the orchestrate error. This PR additionally carries the stage-protocol and state-machine documentation and a much wider test table, so we are consolidating here rather than carrying two PRs for one issue.

Recording it so the credit is not lost in the consolidation: the diagnosis and the first working fix are @jisu-im-dev's. A `Co-authored-by:` trailer on the merge commit would reflect that accurately if you are amending anyway for the normalization fix.

#1023 stays open until this merges, as a fallback.

rafaelgsr (dismissed):
Thank you for fixing the decorated stage-gate choices and for preserving the original reply in audit records and diagnostics.

I reviewed the current head 332b88b06227a875d1dc8076415ef624273fefb0 against 9b8847b5ba34840407d3f551b4e9dd1c7da0982d.

Project alignment is established by issue #1126, the internal ownership of this PR, and the prior substantive maintainer review accepting the direction.

## P2 — Make the decorator compose with existing normalization

In `core/tools/aidlc-lib.ts:7480-7483`, `stripRecommendedDecorator` runs before surrounding quotes and trailing punctuation are removed. Because its regex is end-anchored, `"Request Changes (Recommended)"` and `Request Changes (Recommended).` still fail matching. The failure reaches both rejection validation and guard-recovery selection.

Please move decorator stripping after the existing quote/punctuation normalization and add at least one combined regression case in `t331-guard-deadlock-liveness.test.ts`.

No fixes or commits were pushed because this is an internal PR.

Validation performed:

- Reproduced both failures on the exact head.
- Existing `t188` and `t331` tests pass, confirming the combined case is currently missing.
- Applied the minimal ordering correction diagnostically after integrating current `main@79985897`.
- All six stated blast-radius unit files passed.
- Deterministic packaging, typecheck, and lint passed.

The normal CI and security scanners are green. The two AI review-lens jobs produced “No candidates” but exited 1 because of workflow execution issues; they did not publish a review. The branch is also behind current `main`, so checks must rerun after the correction.

leandrodamascena (dismissed):
Thanks for carrying this fix forward from #1023 and expanding the regression coverage and documentation. I reviewed the current head `332b88b06227a875d1dc8076415ef624273fefb0` against current `main@7998589725a932c43bb3bb870c665cc7c724d3df`. Project alignment is established by #1126, the maintainer-confirmed reproduction, and the prior substantive review of this implementation.

**P2: The decorator normalization does not compose with the matcher’s existing quote and punctuation tolerance**

At `core/tools/aidlc-lib.ts:7480`, `stripRecommendedDecorator` is end-anchored but runs before surrounding quotes and trailing punctuation are removed. As a result:

```text
Request Changes (Recommended)        -> accepted
"Request Changes (Recommended)"      -> rejected
Request Changes (Recommended).       -> rejected
```

This contradicts the documented matcher contract directly above the function. Because `isRequestChangesChoice` is shared by stage rejection and guard-recovery selection, these forms are refused and the gate is presented again.

Please remove the prefix, surrounding quotes, and trailing punctuation before calling `stripRecommendedDecorator`, or normalize them in one equivalent pass, and add combined regression cases. I validated that ordering locally against both this PR head and the synthetic merge with current `main`.

No other blocking findings. The Approve and Accept as-is paths preserve the human-presence guard and revision threshold, and the audit/refusal paths retain the decorated label.

Validation performed with the diagnostic ordering fix, without pushing a commit:

- 58 focused unit tests passed.
- Package determinism passed for all seven harnesses.
- Typecheck passed on the combined `main + PR` tree.

The normal CI suites are green. The two review-lens jobs failed during workflow execution and reported no code finding. Since `main` is now 13 commits ahead, the corrected head must rerun CI against the current base.

apackeer:
Thanks @wowzoo, @rafaelgsr, @leandrodamascena — the ordering fix is in b06effd13f404cd2d744d1a6f8fa0fbd1b3dde7b: the decorator is now stripped after the option prefix, surrounding quotes, and trailing punctuation, and t331 crosses it with each tolerance plus pins that only one trailing decorator is removed. The branch is merged with current `main`; CI is green on this head. @jisu-im-dev is credited via `Co-authored-by` on the fix commit so it carries into the squash.

Could you take another look when you have a moment?

wowzoo (changes requested):
The four forms I blocked on now pass — and the new order refuses two forms the previous head accepted. That trade is the one thing left, and my own wording pointed you at it, so let me be precise about what I should have asked for.

Head `b06effd13`, base `be94bde7c`, which is the current default branch, so this head is up to date.

## What the change delivers

Calling `isRequestChangesChoice` directly, this head against the default branch:

| reply | `be94bde7` | this head |
|---|---|---|
| `Request Changes (Recommended)` | false | **true** |
| `request changes (recommended)` | false | **true** |
| `"Request Changes (Recommended)"` | false | **true** |
| `'Request Changes (Recommended)'` | false | **true** |
| `Request Changes (Recommended).` | false | **true** |
| `Request Changes (Recommended)!` | false | **true** |
| `B. Request Changes (Recommended)` | false | **true** |
| `B. "Request Changes (Recommended)"` | false | **true** |
| `2) Request Changes (Recommended)` | false | **true** |
| `Request Changes (Recommended) (Recommended)` | false | false |
| `please change it` | false | false |

`t331` crosses the decorator with each tolerance — both quote styles, a trailing period, a trailing exclamation mark, an alphabetic prefix, and `2) request changes (recommended).` which crosses prefix, case and punctuation at once — plus the double decorator and `"Approve (Recommended)"` as negatives. That is more than the single combined case I asked for, and `tests/unit/t331-guard-deadlock-liveness.test.ts` passes 25/25 here, with `bun run check` clean.

## P2 — the new order trades one set of compositions for another

Because each normalization runs exactly once in a fixed sequence, moving the decorator strip to the end fixes the forms where the decorator sits *inside* the quotes and punctuation and breaks the forms where it sits *outside* them. Measured on both heads of this branch:

| reply | `f62443b79` (before the fix commit) | `b06effd13` (this head) |
|---|---|---|
| `Request Changes. (Recommended)` | **true** | **false** |
| `"Request Changes" (Recommended)` | **true** | **false** |
| `Request Changes (Recommended).` | false | **true** |
| `"Request Changes (Recommended)"` | false | **true** |

So two replies that this branch accepted an hour ago are now refused, and because `isRequestChangesChoice` is shared, the refusal reaches stage rejection validation and guard-recovery selection as well as the gate reply. Nothing pins either regressed form, which is why CI stayed green through the change.

**This is partly on my wording.** My earlier review said the decorator should be stripped "in the same normalization pass as the quotes and punctuation, or after them". The second half of that is what produced the trade; only the first half is safe, and I should have said just that.

**The invariant:** these tolerances have to compose in both directions, because the decorator can appear inside or outside the quotes and punctuation, and no single fixed-order sequence covers both. Whatever shape achieves that, the two rows above need cases so the next reordering cannot silently swap them again. I have not implemented or measured a fix, so I am not proposing one beyond that constraint.

## Evidence scope on `t188`

I am resting nothing on `t188` locally. On the current default branch, before this change, it already fails **13 of 27** here, every failure carrying the same fixture error: `Stage feasibility is in state 'in-progress' but command requires one of: awaiting-approval`. On this head it is 19 of 33 — those same 11 shared failures, two cases this PR renamed, and the six it adds. Five of the six carry that same stage-state error; the sixth, `decorated Accept as-is respects the revision limit at count 3`, fails differently, on a guard-recovery ask carrying `SUMMARY_QUESTIONS_MISSING`, which I have not explained. CI is the adjudication for that file, and its four unit shards pass on this head along with everything else except the skipped `deploy`.

## Bookkeeping

Thank you for the `Co-authored-by:` trailer. #1023 is still open as the fallback, as agreed, so whoever merges this should close it pointing here rather than leaving it to go quiet.

Two other change requests from this head's predecessor are also still on the record, so this stays blocked on those independently of mine. Push the composition fix and the two cases and I will re-review the new head.

apackeer:
Thanks @wowzoo — you're right that moving the strip to the end just traded one composition direction for the other. `fd907c317` replaces the ordered passes with a single shape match that allows quotes and punctuation on both sides of at most one `(Recommended)`, so there is no ordering left to reorder:

```ts
/^(?:(?:[A-Za-z]|\d+)[.)])?[\s"'`]*request\s+changes[\s"'`.!]*(?:\(recommended\)[\s"'`.!]*)?$/i
```

Both regressed forms — `Request Changes. (Recommended)` and `"Request Changes" (Recommended)` — are pinned in t331 alongside a both-sides case (`"Request Changes (Recommended)".`); the double-decorator negative still holds. One side effect worth naming: `"Request Changes".` (punctuation outside the quotes, no decorator) now also matches, which closes the pre-existing quote/punctuation asymmetry you noted predated this PR. `253ab53b9` scopes the state-machine reference wording so the inside-or-outside tolerance is attributed to Request Changes only; Approve and Accept as-is still strip one trailing decorator and match exactly.

On t188: the local failures come from invoking the file with bare `bun test` — through `bash tests/run-tests.sh --unit --filter t188-human-presence-gate` it passes 33/33 here, matching CI. CI is green on `253ab53b9`.

wowzoo (approved):
Approving. The trade is gone, and it is gone in the way that cannot come back: there is no pass ordering left to reorder. I ran the control in both directions this time — every form the base and the two earlier heads accepted still passes, and the two I blocked on are pinned. One P3 on the wording of the new reference sentences, non-blocking.

Head `253ab53b9`, base `be94bde7c`, which is the current default branch, so this head is up to date. Everything below is measured on it.

## The regression is closed without losing anything

I extracted `isRequestChangesChoice` verbatim from four points and ran the same 65-form corpus through all four: `be94bde7c` (the default branch), `f62443b79` (this branch before the ordering fix), `b06effd13` (the head I blocked on), and this head.

| reply | `be94bde7c` | `f62443b79` | `b06effd13` | `253ab53b9` |
|---|---|---|---|---|
| `Request Changes. (Recommended)` | false | true | **false** | **true** |
| `"Request Changes" (Recommended)` | false | true | **false** | **true** |
| `Request Changes! (Recommended)` | false | true | **false** | **true** |
| `'Request Changes' (Recommended)` | false | true | **false** | **true** |
| `"Request Changes (Recommended)"` | false | false | true | true |
| `Request Changes (Recommended).` | false | false | true | true |
| `"Request Changes (Recommended)".` | false | false | false | **true** |
| `"Request Changes." (Recommended)` | false | true | false | **true** |
| `Request Changes (Recommended) (Recommended)` | false | false | false | false |
| `(Recommended) Request Changes` | false | false | false | false |
| `Request Changes (Recommended) please` | false | false | false | false |
| `Request Changes ( Recommended )` | false | false | false | false |
| `"B. Request Changes"` | false | false | false | false |
| `please change it` / `Approve` / `Request Change` / `requestchanges` | false | false | false | false |

**Forms accepted earlier on this branch but refused now: none.** Across all 65 forms the accept-set at this head is a strict superset of `be94bde7c`, of `f62443b79`, and of `b06effd13`. That is the check my last review should have asked for by name, and it now holds in all three directions.

Both regressed forms are pinned, along with a both-sides case, at `tests/unit/t331-guard-deadlock-liveness.test.ts:1676-1683`, with comments naming the direction a fixed-order normalization drops. The double-decorator and `extra`-suffix negatives survive at `:1688-1697`.

## The widening cannot break the caller

This is the half worth stating, because a matcher that gets more permissive can still lose at a call site that requires exactly one match. `resolveGuardRecoverySelection` (`core/tools/aidlc-lib.ts:6300-6302`) returns `null` unless `matchedOps.size === 1`, and it adds `request-changes` on top of two other matches — `normalized === normalizeGuardRecoveryText(remedy.action)` and `normalized === remedy.op` (`:6286-6289`) — plus a numeric index path (`:6294-6298`). So a reply that newly matches the widened regex *and* an offered remedy would make the selection ambiguous and refuse.

It cannot happen, and rather than hand-pick the remedy tables I tested a deliberate over-approximation of them: every double-quoted, single-quoted and backtick-delimited span in all 90 tracked `core/**/*.ts` files — 14 088 unique strings. That is knowingly larger than the set of real string literals, since it also picks up spans inside comments and fragments of template literals, and that is the point: a superset can add a false collision but cannot hide a real one. Against that set plus all sixteen `GUARD_REMEDY_OPS` values (`core/tools/aidlc-lib.ts:21103-21120`):

- **0 of the 16 ops match.** Every one contains a hyphen, which the regex does not admit.
- **3 of the 14 088 spans match**, and all three are the label itself: `"Request Changes"`, `"Request changes"`, `"request changes"`. None is a remedy `action`; they occur as the plan-approval and summary-confirmation option labels and as union type literals (`core/tools/aidlc-log.ts:645`, `:656`, `core/tools/aidlc-testing-posture.ts:1755`, `core/tools/aidlc-lib.ts:3062`).

The numeric index path cannot collide either, since a bare integer does not contain the words.

## `Approve` and `Accept as-is` were widened too, and that is also loss-free

The same control on the approval matcher (`core/tools/aidlc-state.ts:5408-5411`, `core/tools/aidlc-orchestrate.ts:8761-8764`), at `revisionCount = 3`:

| reply | `be94bde7c` | `253ab53b9` |
|---|---|---|
| `Approve` | true | true |
| `Accept as-is` | true | true |
| ` Approve ` | true | true |
| `Approve (Recommended)` | false | **true** |
| `Approve  (RECOMMENDED)` | false | **true** |
| ` Approve (Recommended) ` | false | **true** |
| `Accept as-is (Recommended)` | false | **true** |
| `approve` / `Approve.` / `"Approve"` / `A. Approve` | false | false |
| `Approve (Recommended) (Recommended)` / `(Recommended) Approve` | false | false |

Both columns model the caller's `userInput?.trim()` (`core/tools/aidlc-state.ts:5384`), which is where the whitespace tolerance in the second row comes from and which predates this PR.

Losses: none, and the case-sensitivity is real: `approve` stays refused. The scoping commit `253ab53b9` was the right instinct — without it the reference would have implied the inside-or-outside tolerance reached all three labels, and it reaches only `Request Changes`. See the P3 below for two places where that new sentence claims slightly more than the code does.

`core/tools/aidlc-orchestrate.ts:8768` also now formats `flags.userInput` in the refusal instead of the stripped value, so the human sees the reply they actually sent. That matches what `core/tools/aidlc-state.ts:5868` already did.

## The adjacent surfaces that stay exact are correctly out of scope

`aidlc-log.ts` is untouched, and both of its exact-label checks remain: `summary-confirmation` requires `Looks correct` or `Request changes` (`:642-652`) and `plan-approval` requires `Approve Plan` or `Request Changes` (`:653-662`). For the summary confirmation that is consistent rather than a gap — the question block at `core/aidlc-common/protocols/stage-protocol.md:404-413` declares no recommended option, and the rule that produces the decorator is conditional on one ("When a question has a recommended option, list it FIRST and append `(Recommended)` to its label", `harness/kiro/skills/aidlc/question-rendering.md:192-193`, identically in the other five). The `plan-approval` exactness is disclosed in the reference, and that sentence is preserved.

## One measured side effect, and why I am not blocking on it

Replacing the `.replace()` chain with a character class made the trailing tolerance broader than punctuation alone: `[\s"'\`.!]*` accepts any mix of whitespace, quotes, `.` and `!`, so `Request Changes ... !!! "" '''` and a reply whose second line is only `"..."` now match, where all three earlier points refused them. The author names the related `"Request Changes".` case in the thread.

I checked the direction of harm before accepting it. Both call sites use a `true` to *permit a rejection* — `core/tools/aidlc-state.ts:5862` gates permission to reject a stage, and `:6288` selects the `request-changes` remedy — so an over-admission never launders consent, and the two words must still both be present, in that order, with nothing but whitespace between them, so the human's intent is not in doubt in any admitted form. Nothing substantive can be swallowed: a single letter outside those two words refuses the whole reply, which is why `Request Changes - the plan omits the migration` and `Request Changes\nplease fix the schema` stay false.

I also checked the one risk a regex introduces that a `.replace()` chain does not. There is no nested quantifier and the two character classes are separated by a literal `(`, so it does not backtrack: 50k-character adversarial inputs (`Request Changes` + 50 000 `.` + a near-miss `(recommendedX)`, and 50 000 quotes before the words) all resolve in **under 0.2 ms**.

## P3 — the scoping sentence claims a little more than the code does

`253ab53b9` was added to narrow an overclaim, and it introduces two smaller ones in the same two sentences. Neither is a code defect; both are in `docs/reference/12-state-machine.md:800-802`, which is a normative reference chapter, so I would rather they were right.

**"Every stage gate choice accepts one trailing `(Recommended)` decorator, case-insensitively."** There is a third stage gate choice, and it does not. Ideation and Inception gates may offer `Add [Skipped Stage]` as a conditional third option — `docs/reference/04-stage-protocol.md:179-189` gives it its own heading, `docs/reference/03-orchestrator.md:861` renders it in the question block, and `core/aidlc-common/stages/inception/requirements-analysis.md:213-231` shows a live instance where the option is literally `Add User Stories` and selecting it runs `engine recompose --add user-stories` instead of reporting a gate result. So the conductor matches that label itself: the string appears nowhere in `core/tools/` or `core/hooks/`, there is no decorator tolerance for it anywhere, and if the renderer ever marks it as the recommended option the conductor has to strip the suffix on its own. Naming the three matched labels, as the next clause already does, would make the sentence exact.

**"Approve and Accept as-is are otherwise matched exactly."** Not quite: surrounding whitespace is tolerated as well, by `userInput?.trim()` at the caller (`core/tools/aidlc-state.ts:5384`, and `flags.userInput?.trim()` in the report path) and again by the `trim()` inside `stripRecommendedDecorator` (`core/tools/aidlc-lib.ts:7543-7545`). That tolerance is **pre-existing** — ` Approve ` is accepted on `be94bde7c` too, so this PR does not introduce it, it only inherits a sentence that now describes it as exact. "Matched exactly apart from surrounding whitespace" would cover it. Everything else in the sentence holds: `approve`, `Approve.`, `"Approve"` and `A. Approve` are all still refused.

## Pre-existing, and only worth a line because this change makes it reachable

`GATE_REJECTED` records `Feedback` and no received label (`core/tools/aidlc-state.ts:5969-5971`), while `GATE_APPROVED` records `"User Input": approvalInput` — the received reply, trimmed but with the decorator intact (`:5550-5551` and `:5704`). So the new paragraph at `docs/reference/04-stage-protocol.md:143-147` is accurate as scoped: approval records and refusal messages do retain the received label. The sentence it sits near, `:276`, is the one that is wrong — it says `report --result approved|rejected --user-input "<exact choice>"` records the user's choice as `GATE_APPROVED`/`GATE_REJECTED`, and for the rejection half it does not.

That is pre-existing and this PR did not cause it. It becomes reachable here only in this sense: `--user-input "Request Changes (Recommended)"` used to be refused with a visible error, and now succeeds and leaves no record of which label the human picked. The `Feedback` is still recorded, which is the substantive content, so I am not asking for anything in this PR — just noting it so it is not re-derived as new.

## Validation

| what | result |
|---|---|
| `bun run check` (packager, `--check`, three tsc projects, biome) | **rc=0**, 779 files, 3 biome infos |
| `tests/unit/t331-guard-deadlock-liveness.test.ts` | **25 pass / 0 fail**, 1580 expect() |
| `bash tests/run-tests.sh --unit --filter t188-human-presence-gate` | **33 pass / 0 fail** |
| `tests/unit/t265-plan-approval-guard.test.ts` via the runner | **55 pass / 0 fail** |
| `bash tests/run-tests.sh --unit` (whole tier) | **not usable as evidence — see below** |
| 65-form corpus × 4 branch points, both matchers | superset at this head, **0 losses** |
| 14 088 quoted spans across 90 `core/` files + all 16 `GUARD_REMEDY_OPS` vs the new regex | **0 op collisions**, 3 span matches, all the label itself |

I started a whole-tier unit run and then invalidated it myself by loading the machine, so I am reporting it as unusable rather than quoting its numbers: seven files failed and every single failure was a timeout, with elapsed times of 20 s to 16 minutes against budgets of 5 s to 30 s, plus `git commit` and `git symbolic-ref` subprocesses failing outright. Re-run unloaded, `t265-plan-approval-guard` is 55/0 and `t27` passes. The one that still fails is `t150-codex-packaging`, case 13 (`doctor enforces Codex 0.145.0 as the compact-session reload floor`) — and it fails identically on the base `be94bde7c` in a separate worktree, same case, same 5 s timeout, so it is pre-existing and environmental rather than anything on this branch. CI's four unit shards are green on this head, which is the adjudication for the tier.

On `t188` you are right and I was measuring the wrong thing: through the runner it is 33/33 here, matching CI. Bare `bun test tests/unit/t188-human-presence-gate.test.ts` fails 19 of 33, every failure on the same fixture refusal — `Stage feasibility is in state 'in-progress' but command requires one of: awaiting-approval` — which is the file's pre-existing dependence on runner-supplied setup, not anything this change did. I rested nothing on it last time and I rest nothing on it now.

## One note for the other two reviews

Both outstanding change requests asked for the ordering fix specifically — "move decorator stripping after the existing quote/punctuation normalization", and the alternative "or normalize them in one equivalent pass". What shipped is the second reading, and deliberately not the first: the first is exactly what `b06effd13` did, and it is what dropped the outside-the-wrapper direction. So a literal re-read of that prescription will not match this diff, and it should not. The three forms named across those two reviews — `Request Changes (Recommended)`, `"Request Changes (Recommended)"`, `Request Changes (Recommended).` — all pass here, along with the six other decorated forms in the table above.

This stays blocked on those two reviews independently of mine; my earlier `CHANGES_REQUESTED` is what this clears.

## Bookkeeping

`Co-authored-by: jisu-im` is on `b06effd13`, so it carries into the squash. #1023 is still open as the agreed fallback — whoever merges this should close it pointing here rather than leaving it to go quiet.

## Evidence scope

Measured with bun 1.4.0; all five `setup-bun` steps in `.github/workflows/ci.yml` pin `1.3.14`, so where a result depends on the runtime, CI is the adjudication rather than my run. CI is green on this head, including all four unit shards and the integration + e2e job.

I did not exercise the stage gate, the summary-confirmation checkpoint or plan-approval end to end — I read their matchers, their audit writes and the question blocks that feed them, and I exercised the two matchers directly. I did not review the reference chapters beyond the passages cited above. That is coverage I am not claiming.

wowzoo:
https://github.com/awslabs/aidlc-workflows/pull/1133

Base `main` is now `2d3ad2dfb`, and this PR's head is `253ab53b9`.

I owe you the explanation before anything else. On 13 September I said we would consolidate the `(Recommended)` fix here and close #1023. #1023 is what merged, as `2d3ad2dfb`, and you are finding that out from a merge conflict rather than from me. Sorry — that was mine to say first.

**This PR now conflicts.** Against `2d3ad2dfb` the conflicts are in `core/tools/aidlc-lib.ts`, `tests/unit/t188-human-presence-gate.test.ts` and `tests/unit/t331-guard-deadlock-liveness.test.ts`. Your `aidlc-orchestrate.ts` and `aidlc-state.ts` hunks auto-merge, and both documentation files merge clean.

**It is no longer a duplicate. It still carries two deltas over the merged tree.**

*The six forms.* I ran a 28-reply corpus by importing `isRequestChangesChoice` directly from each tree, `2d3ad2dfb` and `253ab53b9`. `main` accepts the twelve decorated replies where the decorator sits inside the quotes and punctuation, and refuses these six, where it sits outside them:

| refused on `2d3ad2dfb` | where the decorator sits |
|---|---|
| `Request Changes. (Recommended)` | outside the trailing punctuation |
| `Request Changes! (Recommended)` | outside the trailing punctuation |
| `"Request Changes" (Recommended)` | outside the quotes |
| `'Request Changes' (Recommended)` | outside the quotes |
| `"Request Changes (Recommended)".` | inside the quotes, punctuation outside |
| `"Request Changes." (Recommended)` | outside the quotes, punctuation inside |

Your head accepts all six, keeps every undecorated reply in the corpus, and still refuses `please change it`, `Approve`, a doubled `(Recommended) (Recommended)`, and an empty reply. `REQUEST_CHANGES_CHOICE_RE` permits wrapper noise on both sides of at most one decorator. That is not a property the merged implementation can reach by reordering: permuting its six post-trim normalization steps exhaustively, **0 of 720 orderings** accept all six, because `"Request Changes." (Recommended)` and `"Request Changes (Recommended)".` demand opposite positions for the decorator strip relative to the quote strip. So those six are a measured remaining delta, and shipped behaviour now rather than a review remark.

Both call sites reach it with the reply intact — `aidlc-lib.ts:6289` passes the response text unchanged and `aidlc-state.ts:5862` applies only `.trim()` — so the six rows are not merely theoretical.

*The documentation.* Your `docs/reference/04-stage-protocol.md` and `docs/reference/12-state-machine.md` changes state the engine-side contract: that one `(Recommended)` suffix is removed before an offered label is matched. `main` documents the two halves either side of that and not the middle — `core/aidlc-common/protocols/stage-protocol.md:206-207` tells the conductor to forward the exact selected label, and the harness question-rendering guides tell it to append the suffix, but no page on `main` says the engine strips one. Worth adding that the convention is not Codex-only, which the current comments imply: the same two instructions ship in all six guides — `codex:66`/`:75`, `copilot:130`/`:139`, `cursor:120`/`:130`, `kiro:193`/`:206`, `kiro-ide:193`/`:206`, `opencode:120`/`:130`.

**Two bookkeeping items with the rebase.** The body's `Closes #1126` no longer resolves to open work — the merge closed #1126 — so it wants to become `Refs #1126`, otherwise it promises a closure it cannot perform. And CI can say nothing about this head until the rebase lands.

**Where the review stands, precisely.** I approved `253ab53b9`. That does not unblock the merge: two change requests from other maintainers are standing on `332b88b06`, two heads back, and this repository does not dismiss reviews on push, so they persist until those reviewers read the current head. With `required_approving_review_count` at 0 there is no count for my approval to satisfy — those two reads are the review gate.

**One note on a sentence you already wrote.** Your `12-state-machine.md` change records that `answer --checkpoint plan-approval --details` still requires the undecorated label. That is accurate, and the same gap sits one checkpoint over: on `2d3ad2dfb`, `answer --checkpoint summary-confirmation --details "Looks correct (Recommended)"` is refused at `core/tools/aidlc-log.ts:641-651`, and unlike the plan-approval message, its refusal directs the conductor to re-present the choices and wait — which reproduces the same reply. Neither checkpoint's question spec names a recommended option, but neither does the approval gate spec at `stage-protocol.md:153-158`, and the decorator reached that gate anyway; that is what #1126 reported. Both sites are outside this diff and I am not asking you to widen the PR — only flagging it so your sentence is not read as bounding the problem.

Next from here: rebase onto `2d3ad2dfb`, resolve those three files, switch `Closes` to `Refs`, and push. I will review the new head.

Issue #1045, opened 2026-09-08
[Bug]: Construction `autonomous` mode is unreachable on skeleton-off runs — the ladder is suppressed by protocol while the orchestrator assumes it fires, and `set-autonomy` writes a state field the generated state file never contains


leeyoseph-dev:
## Summary

Construction `autonomous` mode is unreachable whenever the walking skeleton is off. Two independent causes stack, and they contradict each other in the shipped source:

1. The protocol tells the conductor **not to present** the ladder prompt for skeleton-off runs, so the mode is never granted.
2. `set-autonomy` cannot grant it manually either, because the state field it writes with `setFieldStrict` is **absent from the state file the engine itself generates** — even though the shipped state template declares it.

This affects four stock scopes out of eleven (`bugfix`, `express`, `refactor`, `security-patch`), plus any run where a team overrides `Skeleton Stance` to `off`.

All references are against `a277af218f0df7f325d3b8be7b6d90fce2c5bd40` (= upstream HEAD at the time of writing), Claude Code harness.

## 1. The ladder never fires, and the orchestrator assumes it did

`aidlc-common/protocols/stage-protocol-construction.md:52-55`

```
**Ladder prompt (fires once, immediately after walking skeleton gate)**

After an actual walking skeleton's gate approves, present exactly one ladder
prompt. Do not present it for skeleton-off or zero-Unit execution:
```

`tools/aidlc-orchestrate.ts:2286-2295` reasons from the opposite premise:

> skeleton-off runs Bolt 1 "as a regular Bolt with the standard batch-gate path", and since `Construction Autonomy Mode` is `unset` (treated as `gated`) **until the post-Bolt-1 ladder prompt sets it**, that batch gate IS presented (it is only skipped when `autonomous`, which cannot be true before Bolt 1 ships).

The orchestrator's comment treats the ladder as something that arrives after Bolt 1 in every stance. The protocol tells the conductor to suppress it in exactly the stance the comment is describing. So `Construction Autonomy Mode` stays `unset` forever on a skeleton-off run, and the session-resume clause (`:73`) does not help either — it re-fires the ladder only when "the walking skeleton is already `[x]` complete", which never happens when the skeleton is skipped.

## 2. The manual path is blocked by a missing state field

`tools/aidlc-bolt.ts:1121` writes the grant with `setFieldStrict`:

```ts
updated = setFieldStrict(content, "Construction Autonomy Mode", flags.mode);
```

[...]

PR #1140, opened 2026-09-11
feat: grant Construction autonomy on demand

### LINKED ISSUE #1142: [Feature]: grant Construction autonomy on demand, not only at the walking-skeleton ladder
### Description

Let the human grant (or revoke) `Construction Autonomy Mode` **whenever they say so** during Construction, instead of only at the one ladder prompt that fires after a walking-skeleton gate.

The mechanism already exists — `aidlc-bolt.ts set-autonomy --mode <autonomous|gated>`, its `AUTONOMY_MODE_SET` audit row, and the human-presence guard that refuses an escalation without a fresh human turn. What is missing is that the protocol sanctions exactly one moment for using it. `stage-protocol-construction.md` presents the ladder "immediately after walking skeleton gate" and tells the conductor not to present it otherwise, so there is no defined path for a human who says "run the rest of Construction autonomously" at any other point.

Concretely I would like the construction protocol to describe two paths to the same recorded grant:

- **The ladder prompt** — unchanged. The prompted path under skeleton-on, once per intent, after the skeleton gate, because the skeleton is the evidence the path works end to end.
- **On demand** — at any point in Construction, on an explicit typed request. The only path available under `skeleton: off`, and also available under skeleton-on to someone who wants the grant earlier than the skeleton gate or wants to revoke it later.

The safety property should not change: escalation to `autonomous` still needs a fresh human turn (so an unattended run cannot grant itself more autonomy), de-escalation to `gated` stays presence-free, and everything that halts under `autonomous` today keeps halting — failure halt-and-ask, the Build-and-Test loop-back's rung 4, and the outright refusals in `park`, `scope-change`, `recompose`, and blocking-sensor override.

### Use Case

My team runs one AI-DLC workflow per unit of work, in scopes that deliberately declare `skeleton: off` — the unit plugs into a system that already runs end to end, so there is nothing to bootstrap and the skeleton ceremony buys nothing. We want autonomy in effect **from the moment the first unit reaches code generation**: a developer picks a unit, reviews its design at the per-unit design gate, then says "go" and lets Construction run unattended, with failures still stopping to ask.

Today that is not expressible. The ladder is the only sanctioned grant point and it is suppressed in exactly the stance we run, so the mode stays `unset` (treated as `gated`) and every stage gate is presented. Adopting `skeleton: on` purely to unlock the prompt would mean adding a ceremony we affirmed off, and it would still not let the grant precede the skeleton gate.

Generalising the grant also helps skeleton-on users: today the choice is frozen at one moment, so someone who watched two Bolts and now trusts the run has no sanctioned way to say so, and someone who granted autonomy and changed their mind has no documented way to take it back mid-Construction.

### Version

v2 (current on main)

### Area

Construction

### Additional Context

There is a prerequisite bug: **#1045** — `set-autonomy` writes the grant with `setFieldStrict`, and the field is absent from the state file the generator emits even though `knowledge/aidlc-shared/state-template.md` declares it under `## Current Status`, so the write throws `Field not found in state file`. That makes *both* grant paths unrecordable, which is presumably why the ladder-only restriction has not been felt as a limitation before now. Any implementation of this request has to fix or depend on that.

I have a draft PR up that implements this feature and carries the #1045 fix it needs: **#1140**. Happy to reshape it if you would rather see the bug fixed on its own first, or if the on-demand path should be spelled differently (a dedicated verb, or an explicit prompt the conductor may offer rather than a request it responds to).

jstrunk:
# Summary

Users can explicitly grant or revoke Construction autonomy during Construction, including workflows with `skeleton: off`. An existing choice survives resume and is not asked again at the ladder.

## Changes

- Keep the first actual Construction approval human after conditional skips or approved stage overrides; completing that stage keeps its anchor fixed.

- Keep the on-demand path and existing ladder path on the same recorded grant.
- Preserve the first Construction-stage human approval after an early grant.
- Share completion-approval policy between report and state mutation; existing unit-major per-unit-stage gates remain human-owned.
- Document actual first-stage review behavior and update coverage metadata.

Depends on #1143. Together they address #1045; closes #1142. Original on-demand change by @jstrunk, followed by maintainer integration fixes.

## User experience

A user can say “run the rest autonomously” or “gate every stage from here.” The first stage's review, Code Generation Plan Approval, and failure stops retain their own requirements. A recorded autonomy answer is not repeated.

## Checklist

* [x] Changes are documented

## Test plan

The smoke/unit slice completed 304 files. Runtime regressions passed; the coverage registry was refreshed and its checks passed. Two fixture-discovery failures passed with Git discovery bounded to the fixture directory. Fifteen Windows-only cases were skipped.

Focused t261, t339, t120-classify-roundtrip, and t47-construction-bolts passed, as did the distribution build, determinism check, and TypeScript checks. No live-harness merge gate is claimed.

## Acknowledgment

Local follow-up validation: the conditional-skip regression passed 8 cases; the final stacked routing/coverage slice passed 58.

### DIFF
--- a/core/aidlc-common/protocols/stage-protocol-construction.md
+++ b/core/aidlc-common/protocols/stage-protocol-construction.md
@@ -42,20 +42,38 @@ EXECUTE stage (`isSkeletonGateStage`) always presents a stage-level approval
 gate regardless of autonomy mode. That gate covers that stage's artifacts
 across the Units that have settled — not a Bolt's combined design artifacts
 and generated code. Audit: emit `GATE_APPROVED` as usual. `BOLT_COMPLETED`
-is not emitted on this gate. Skeleton-off uses the ordinary first-stage gate;
+is not emitted on this gate. Skeleton-off uses the ordinary first-stage gate,
+which follows `Construction Autonomy Mode` like every other Construction stage
+gate — so an on-demand `autonomous` grant made before that stage runs skips it;
 a zero-Unit stage has no skeleton or Bolt ceremony at all.
 
 > **Planned (non-executable).** A later Bolt-major walk would present a
 > Bolt-level gate covering that Bolt's design artifacts and generated code
 > together, with the enclosing `BOLT_COMPLETED` tying the gate to the Bolt.
 
-**Ladder prompt (fires once, immediately after walking skeleton gate)**
+**Autonomy grant — the ladder prompt, and the on-demand path**
 
-After an actual walking skeleton's gate approves, present exactly one ladder
-prompt. Do not present it for skeleton-off or zero-Unit execution:
+`Construction Autonomy Mode` is the human's grant and governs the remaining
+Construction stage gates. Two paths set it, and both record it exactly the same
+way, through `aidlc-bolt.ts set-autonomy --mode <choice>`:
+
+- **The ladder prompt** — the prompted path, under skeleton-on. Present exactly
+  one per intent, after the walking-skeleton gate approves: the skeleton is the
+  evidence that the path works end to end, so the prompted grant follows it. Do
+  not present it for zero-Unit execution.
+- **On demand** — at ANY point during Construction, whenever the human asks in
+  a typed message ("run the rest autonomously", "gate every stage from here").
+  This is the only path available under `skeleton: off`, where there is no
+  skeleton gate to hang a prompt on, and it is equally available under
+  skeleton-on to a human who wants the grant earlier than the skeleton gate or
+  wants to revoke it later. Never infer it: act only on an explicit request.
+
+Either way the escalation to `autonomous` still requires a fresh human turn —
+`set-autonomy` refuses otherwise — so an unattended run cannot grant itself more
+autonomy. De-escalation to `gated` carries no presence requirement.
 
 ```question
-prompt: "The walking skeleton shipped. How should the remaining Bolts run?"
+prompt: "How should the remaining Construction stages run?"
 header: Autonomy
 multiSelect: false
 options:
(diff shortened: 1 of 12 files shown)


apackeer:
Thank you, @jstrunk — the two-path protocol design is the right shape and we've kept it. We reviewed it end to end and are taking it to `main` now.

What we did on your branch: rebased your commit onto current `main` (authorship preserved) and appended maintainer commits that (1) keep the first in-scope Construction stage's own human approval after an early on-demand grant and keep the existing unit-major per-stage human gates, with one shared policy across `report`/`approve`/`reject`; (2) honour conditional skips when locating that first stage; (3) add t339 coverage and the `12-cli-commands.md` / `03-orchestrator.md` docs. Marking it ready for review since #1143 merges alongside it.

apackeer (approved):
I reviewed the current head `45f6e104433dc2d50cd34b8d95e57d13f8771496` against `main` (`7711623994cfac2853c81581c6119ee94eca020d`).

Alignment: implements #1142 as the two-path grant (ladder prompt under skeleton-on; explicit on-demand grant/revoke at any point in Construction, the only path under `skeleton: off`), recorded through the existing `set-autonomy` receipt with the fresh-human-turn requirement on escalation. Maintainer commits on top keep the first actual Construction stage's approval human-owned after an early grant — including after conditional skips — keep the existing unit-major per-stage human gates, share that policy across `report`/`approve`/`reject` (`isAutonomousConstructionGate`), add t339 coverage and the command/orchestrator docs, and fix the docs-site anchor that `Deploy Documentation --strict` rejected.

Verified locally: t339, t261, t120, t47, the smoke/unit tier (304 files), distribution build, two-build packaging determinism, typecheck. All checks are green on this head. Approving.

PR #1143, opened 2026-09-11
fix: persist Construction autonomy choices

jstrunk:
# Summary

`set-autonomy` can now save the user's choice when the state file lacks `Construction Autonomy Mode`. The missing field is inserted under `Current Status`, so existing workflows work without a migration.

## Changes

- Replace the strict field update with insert-or-update behavior.
- Exercise the production-shaped state fixture, missing-field insertion, updates without duplicates, and refusal when the required insertion heading is absent.
- Preserve the fresh-human-turn check and audit-first ordering.

Addresses the persistence portion of #1045. #1140 supplies the on-demand protocol path. Original implementation and regression tests by @jstrunk; original author credit is preserved.

## User experience

Before, choosing autonomy could fail with “Field not found in state file.” After, the choice is saved and existing Construction behavior can honor it.

## Checklist

* [ ] Changes are documented

## Test plan

t261: 69 passing cases. Distribution build, independent-build determinism, and all TypeScript projects passed. This is focused local validation, not a completed live-harness merge gate.

## Acknowledgment

### DIFF
--- a/core/tools/aidlc-bolt.ts
+++ b/core/tools/aidlc-bolt.ts
@@ -61,7 +61,6 @@ import {
   requireLiveClaimForTeamUnit,
   resolveBoltDag,
   resolveProjectDir,
-  setFieldStrict,
   setOrInsertField,
   validateUnitName,
   slugify,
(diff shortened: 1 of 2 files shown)


apackeer:
Thank you, @jstrunk — this is exactly the right fix, and the fixture analysis (the helper manufacturing the one precondition the production path lacks) was the most valuable part of the PR. We reviewed it end to end and are taking it to `main` now.

What we did on your branch: rebased your commit onto current `main` (authorship preserved; only the SHA changed), no functional changes. The red `Require Contributor Statement` check was a stale run against an earlier description and clears on this push.

apackeer (approved):
I reviewed the current head `e841723b6db04dff784b6f65c61d700378969a20` against `main` (`7711623994cfac2853c81581c6119ee94eca020d`).

Alignment: fixes the second cause of #1045 with the same insert-if-absent pattern its sibling runtime fields (`Skeleton Stance`, `Construction Iteration`) already use. The fail-closed shape check (`appendUnderHeading` throws on a missing `## Current Status`) and the fresh-human-turn guard on escalation are preserved, and the t261 fixture now matches the generated state shape so the regression is actually caught.

Verified locally: t261 69/0 on the production-shaped fixture, distribution build, two-build packaging determinism, typecheck. Required checks are green on this head. Approving.

PR #1149, opened 2026-09-11
fix: make guard recovery executable across workflow states

apackeer:
# Summary

Guard refusals can prescribe recovery that the next tool call rejects, or lose the human's selection while delivering that recovery. Summary reconfirmation can also invalidate the review it is meant to unblock. This change makes recovery operations explicit and preserves progress while keeping plan approval, reviewed outputs, and human feedback separately enforced.

Follow-up to the Plan Approval and Change Control work in #1000 and native Kiro IDE recovery fix in #1111.

## Changes

- Render and validate structured recovery operations for source and native installs. Admit a selected native restart's returned reset command only for the current target, scope, and direction; retain that selection through Copilot delivery.
- Distinguish command selections from interactions requiring a separate human answer. Request Changes still needs the human's exact subsequent feedback.
- Keep summary-owned questions editable while reviewed outputs remain frozen. Recheck summary authorization at terminal verdicts, use the actual Unit lifecycle for remedies, and preserve Change Control acceptance records and notices even when a verdict fails.
- Add a production-guards test profile, a required CI job, and regressions that execute the offered recovery sequence. Explicitly selected tests that execute no cases now fail instead of silently passing.

## User experience

Before: a user can choose a restart and hit another guard, or reconfirm an unchanged summary and enter another review loop. After: the selected recovery can proceed through its returned commands; unchanged confirmation bookkeeping preserves review binding. Substantive question edits, missing inputs, and withdrawn authorization still prevent an invalid terminal verdict.

The question fingerprint preserves substantive content, file presence/type, comments, and examples while normalizing only the canonical confirmation answer. The stored recovery operation and interaction detect a changed remedy or target; Copilot's selection digest detects a replaced selection before delivery completes.

On upgrade, an older in-flight recovery ask may require one repeated selection. Reviews created under the earlier question fingerprint may need a fresh review through normal recovery. Abort retains existing source/native trusted-tool admission: the conductor must obtain consent, and the Plan Approval hook does not authenticate abort consent.

## Checklist

* [x] Changes are documented

## Test plan

After rebasing onto main (`aa1c76c92`), 532 test cases passed across 21 selected files, with zero failures or skips:

- Production-guards unit/integration slice: 123 cases across seven files, including ten production journeys, 35 source/native remedy cases, and six Unit lifecycle scenarios.
- Smoke/unit regression slice: 409 cases across 14 files, including all 58 Copilot adapter cases with a freshly compiled dispatcher and `AIDLC_REQUIRE_COMPILED_COVERAGE=1`.
- Packaging determinism, all TypeScript configurations, lint, coverage-registry freshness, and `git diff --check` passed.

Both test slices used `bash tests/run-tests.sh --debug -P 8` with full per-file logs and runner summaries. The production slice also used `--production-guards --no-llm`; no live-model gates were enabled.

Coverage includes production-guard journeys, real Unit lifecycle transitions, source/native remedies, compiled Copilot adapter execution, stale or unrelated selection rejection, commented confirmation answers, and explicit all-skipped runner failures. The missing-profile negative check fails as intended.

Execution is on Linux with deterministic fixtures. Live model/UI journeys and Windows/macOS execution are not claimed. Release metadata is unchanged.

## Acknowledgment


wowzoo (commented):
Reviewed in full; not putting a review event on it yet. Most of this holds up under attack, and one thing does not belong to me to wave through: the change newly admits a destructive Bolt operation past the Plan Approval guard, with consent carried by conductor prose. That is disclosed and deliberate, and it is the one decision here I would rather see a second maintainer sign off on than approve alone.

Head `288ffb944`, base `be94bde7c` — the current default branch, so this head is up to date. Everything below is measured on it.

## P2 — the abort admission is new authority over a destructive operation

`isPlanApprovalPrerequisite` on `be94bde7c` admitted three things: `engine orchestrate next|continue`, `engine testing-posture resolve|render|fingerprint|verify`, and `engine log decision|answer` at the plan-approval checkpoint. `engine bolt` was not among them. This change adds it, for one exact argv, at `core/hooks/aidlc-plan-approval-guard.ts:575`.

That argv is not inert. `--discard` spawns `aidlc-worktree.ts discard` (`core/tools/aidlc-bolt.ts:829-834`), which removes the Bolt worktree and clears the slug from the Bolt refs — and the new coverage says so in its own title: `direct abort admission discards the worktree without minting Plan Approval` (`tests/integration/t-guard-native-remedies.test.ts:716`). So the guard now lets a worktree deletion through on a route that previously required approval, and the only thing standing between the offered remedy and that deletion is the conductor obtaining the human's selection in prose. The hook does not authenticate it, which the description and the comment at `core/tools/aidlc-guard-operation.ts:150-158` both say plainly.

I am not asking you to change it. The comment's reasoning is sound as far as it goes — a direct log refusal publishes no active directive, so requiring a consumed marker would strand the recovery it just offered — and `t265b` pins the behaviour deliberately. What I am unwilling to do is approve a new prose-consented path to a destructive operation as the only reviewer on a six-day-old change. The invariant worth stating is that a destructive admission either carries mechanical selection evidence or has the trade recorded where the next person auditing this hook will find it, not only in a PR description that stops being visible after the squash.

## P3 — the restart selection is neither one-shot nor bound to the turn that made it

`isSelectedGuardRestartContinuation` (`core/hooks/aidlc-plan-approval-guard.ts:601-653`) checks `marker.state_present`, but not `marker.state_sha256` against the current state digest, and nothing consumes the selection on execution. Its other conditions are state-derived — scope equality, graph order, unique checkboxes, effective `EXECUTE` — so a materially changed state does fail them. An unchanged state does not: a selection made earlier stays admissible in a later turn, including after the human has said in prose that they no longer want the restart, and it can be executed more than once.

`t265b`'s case is titled `native reset requires current selection rather than an earlier state or another remedy` (`tests/unit/t265-plan-approval-guard.test.ts:1118`), and it does cover those two — an earlier state, and a different remedy. "Current" in the sense of "the human's latest turn" is not what is enforced, so the title claims a little more than the code does.

## P3 — one new test cannot pass under a documented invocation

`tests/unit/t-runner-production-guards.test.ts:422-424` reads `process.env[GUARD_PROFILE_ENV]` and asserts it matches `/^(fixture|production)$/`. The runner supplies that marker, so the case passes through `run-tests.ts` and fails under `bun test tests/unit/t-runner-production-guards.test.ts` with `Received value must be a string: undefined`. The workflow always goes through the runner, so nothing surfaces it there.

`docs/reference/09-testing.md:335` documents the direct form — "Run with `bun tests/run-tests.ts --integration` or directly: `bun test tests/integration/tNN-stage-SLUG.test.ts`" — and this is the only file in the repository that hard-asserts the marker. The tolerant pattern is already in this change two files over: `tests/integration/t-guard-recovery-production.test.ts:62` uses `process.env.AIDLC_TEST_GUARD_PROFILE === "production" ? test : test.skip`, which degrades to a skip.

## What held up, so it does not get re-derived

**The admitted command shapes are exact and fail closed.** `parseGuardRestartContinuationCommand` (`core/tools/aidlc-guard-operation.ts:131-148`) admits two spellings and rejects everything adjacent:

| command | result |
|---|---|
| `aidlc engine jump execute --target T --direction redo --scope S` | admitted |
| `aidlc.exe …` (same shape) | admitted |
| `PATH=/tmp/evil aidlc …` | rejected |
| `… && rm -rf /tmp/x` / `… ; touch /tmp/x` | rejected |
| `--target $(id -u)` / `--target \`id\`` | rejected |
| `aidlc  engine …` (double space) | rejected |
| `env aidlc …` | rejected |
| `/tmp/evil/aidlc …` | rejected |
| `--direction forward` | rejected |

That matters because `mutationIntent` returns early for an admitted string (`core/hooks/aidlc-plan-approval-guard.ts:853-855`), skipping `shellUsesDynamicEvaluation`, `shellCommandAltersExecutableResolution`, `shellInvocationNeedsApproval` and `shellWriteTargets` (`:856-867`). The table is why that is safe: every input that could make those checks fire is rejected by the parse before the return is reached.

Trusting the bare name `aidlc` or `aidlc.exe` rather than authenticating the resolved binary is **pre-existing**, not introduced here: `isNativePlanApprovalPrerequisite` (`:561-567`) matches on the lowercased name, and `be94bde7c` already returned `!executableResolutionChanged && !dataDriven` for it at its own line 613-615.

**The fixture guard set is unchanged and the production profile closes the settings route too.** `FIXTURE_GUARD_ENV` (`tests/harness/runner-profile.ts:140-146`) is exactly the five keys the previous `tests/run-tests.ts` set. The production profile forces every inherited `AIDLC_(SKIP|DISABLE|ALLOW_DIRECT)_*` key to `"0"` (`:181-183`) plus the sixteen known names (`:184`), and the rationale at `:148-150` is correct: `resolveProjectFlag` (`core/tools/aidlc-lib.ts:553`) returns the environment value whenever `Object.hasOwn` is true, so writing `"0"` — not deleting — is what also masks a settings-file bypass. Inherited copies of the profile marker are removed before the real value is set (`:172-175`). And the coupling is test-enforced rather than a comment: `tests/unit/t-runner-production-guards.test.ts:357-371` walks `RECORDABLE_PROJECT_BYPASSES` imported from `core/tools/aidlc-settings.ts` and asserts each key is present with `"0"`.

**The zero-cases rule cannot fail an ordinary run.** It is scoped to `--filter` (`tests/run-tests.ts:702-707`, `:1036-1040`); without a filter a fully skipped file still reports `SKIP`, so a tier run or a shard is unaffected, including files that skip everything on a platform.

**The new job becomes required without touching branch protection.** `test_guards` joins `needs` on `Tests (smoke + unit)` and its result is asserted explicitly (`.github/workflows/ci.yml:115-157`). Its bun pin matches every other job in that file.

**The fingerprint change is bounded.** `summaryInputReviewFingerprint` (`core/tools/aidlc-lib.ts:10069-10109`) normalizes line endings and masks one unambiguous visible confirmation answer — one summary section and one answer line, otherwise it hashes the whole normalized content. Invalid UTF-8 hashes the raw bytes rather than decoding through replacement characters. The loop indexes `visible[index]` and `source[index]` together, which is only safe if the visible projection is line-for-line; it is, with lengths matching across plain text, backtick and tilde fences, block and inline comments, CRLF input and a leading BOM. The reviewer protocol describes all of this clause by clause, including the fail-closed ambiguity case, fences and trailing comments staying bound, a `review_artifact` naming a questions artifact staying fully byte-bound, and the upgrade consequence. All seven harness `SKILL.md` files receive an identical guard-recovery block — normalizing the invocation token, tool names and picker name, the block hashes to one value across all seven.

## Validation

| what | result |
|---|---|
| `bun run check` (packager, `--check`, three tsc projects, biome) | **rc=0**, 787 files |
| `t-guard-recovery-production` + `t-guard-native-remedies` + `t-review-verdict-unit-state` | **41 pass / 10 skip / 0 fail**, 2153 expect() — the skips are the production-profile-gated cases at `t-guard-recovery-production.test.ts:62` |
| `t265-plan-approval-guard` with `core/hooks/aidlc-plan-approval-guard.ts` reverted to `be94bde7c` | **56 pass / 5 fail** — exactly the five new `t265b hook lifecycle` cases |
| the same file restored | **61 pass / 0 fail**, 531 expect() |
| new unit files | 96 pass / **1 fail** — the P3 above |

## Evidence scope

Measured with bun 1.4.0; every job in this workflow pins `1.3.14`, so where a result depends on the runtime, the workflow is the adjudication rather than this run.

Not examined: `core/tools/aidlc-directive.ts` (+45/−2), `core/tools/aidlc-orchestrate.ts` (+17/−12), `core/tools/aidlc-runtime-paths.ts` (+15), `tests/integration/t112.serial.test.ts` (+48/−23), and the reference documents other than the reviewer protocol and the testing guide. That is coverage this review does not claim.

apackeer:
@wowzoo — thank you for the careful pass, and for not waving the abort admission through alone.

Since your review the head has moved (currently `ceb7c521a`, with two small fixes described below still to land):

- **Your two P3s are fixed.** A later human response now supersedes a recorded restart selection (`consumeSharedDirectiveAsk`), exercised through the real Plan Approval hook in `t265b`: same pick again → admitted, "cancel" → refused, re-pick → admitted. The runner-marker assertion in `t-runner-production-guards` skips when the file is launched outside the runner.
- **For the P2 we chose to remove the harm rather than add a consent gate.** `worktree discard` (the owner behind `bolt abort --discard`) now parks the attempt before removing anything: raw working-tree bytes (tracked and untracked, bypassing clean filters) plus the reviewed-source refs land under `refs/aidlc/parked/<slug>/<stamp>/`, then `WORKTREE_DISCARDED` records `Parked ref`, then teardown. `worktree restore --slug <slug>` recovers it into `.aidlc/restored/…` on a `restore/` branch, so the recovery continuation that recreates the same slug still succeeds and both attempts survive; `worktree purge` removes parked refs explicitly. A mistaken abort now costs one command instead of the work. The abort argv and the hook are unchanged.
- **The trade is recorded where you asked**: hook comment (`aidlc-plan-approval-guard.ts`), `aidlc-guard-operation.ts`, and `06-hooks-and-tools.md` state that consent for the admitted abort is conductor-obtained and not authenticated, that discard is recoverable, and that a mechanical selection receipt remains a later candidate. A separate threat-model review concluded a receipt would add no user turn where recorders exist but cannot be made human-only by argv and is enforceable only where the hook governs; its failure mode is a guard refusing the remedy it just offered, which is the class this PR exists to remove.

The AI review on the current head raised two additional items that are legitimate and small, landing next: `persist-credentials: false` on the new `test_guards` checkout (it uploads test logs), and restoring process-identity gating in `spawnState` so a project-set `AIDLC_COMPILED_EXECUTABLE` cannot redirect a source orchestrator's child. Its remaining finding is the abort-consent point above, which will be dismissed with the rationale recorded in the docs.

Would you re-review once the final head is up? If you still want a second maintainer on the admission, name who you'd like me to loop in.

apackeer:
Final head is up: `52d47047c`. CI is green on it (contract checks, smoke, 4 unit shards, deterministic integration + e2e, production guards).

Since my comment above:

- `ceb7c521a` — parking now stores **raw bytes** for paths under configured clean filters (same `filteredRawIndexEntries` + `hash-object --no-filters` path reviewed-source binding uses), so a lossy filter can no longer alter what `restore` gives back. Regression in `t78` (a lossy filter uppercased parked bytes before the fix). Ignored files remain the one thing not parked — by the repository's own ignore rules — and that limit is documented in `12-state-machine.md`.
- `adf5db70e` — the new `test_guards` CI job checks out with `persist-credentials: false` (it uploads test logs); the orchestrator's state child is chosen by process identity again rather than an environment-supplied executable.
- `52d47047c` — the compiled orchestrator spawns by `process.execPath` only; `worktree restore/purge --slug` infer the repository holding the parked refs, so the documented selector-free command works in multi-repo intents (regressions in `t166`).

The AI review's remaining item on this head was the abort-consent receipt; it is dismissed with the rationale recorded in that dismissal and in the hook/docs (prose consent is the boundary, discard is recoverable, receipt deferred with liveness tests as the bar).

@wowzoo — ready for your re-review whenever you have time; if you still want a second maintainer on the admission, name them and I'll loop them in.

leandrodamascena (approved):

PR #1153, opened 2026-09-12
fix(ci): simplify AI PR review execution

leandrodamascena:
## Summary

- Replace the multi-job matrix and artifact handoffs with one sequential review job.
- Run prompt-injection, security, and final AIDLC review passes in order.
- Suppress Codex transcripts and publish only the validated final review.
- Validate model output as the isolated review user before copying it into the workspace.
- Keep incomplete inspection blocking while deriving inspected paths from the immutable manifest.

## Root cause

Run 34685071576 printed the full Codex transcript into the Actions log. Both model calls then failed after successful inference because the runner tried to inspect output inside a `0700` temporary directory owned by `ai-pr-review`.

## Review fixes

- Require `inspection.status: "complete"`; missing or failed inspection stops publication. The publisher derives changed paths from the immutable manifest.
- Redact each model response as `ai-pr-review` before installing it as `runner:ai-pr-review`, then verify that the review user retains read access.
- Detect reviewer-control changes with the same merge-base diff used by context construction, preventing upstream-only changes from skipping unrelated reviews.
- Preserve a failing result when an existing SHA-bound review requests changes.
- Keep the three 15-minute model budgets within the 60-minute job timeout.
- Load reviewer controls from the trusted base and skip self-review when the PR changes those controls.

## Validation

- `bun run check`
- `bun test --timeout 20000 tests/unit/t300-ai-pr-review.test.ts` — 14 passed, 149 assertions
- Workflow YAML parse
- All six workflow `run` blocks pass `bash -n`
- Coverage registry freshness and ratchet guard
- `git diff --check`

No version, README badge, or changelog change. This PR changes internal CI behavior.

Follow-up to #1152.

## Acknowledgment


apackeer (dismissed):
Reviewed PR #1153 at `a7e9e22a13cbf20f5ee808dec16b7faccb96daf1` against `9b8847b5ba34840407d3f551b4e9dd1c7da0982d`. Recommendation: request changes.

1. **P1 — Preserve a blocking outcome for incomplete inspection.** [`.github/scripts/ai-pr-review.ts:295`](https://github.com/awslabs/aidlc-workflows/blob/a7e9e22a13cbf20f5ee808dec16b7faccb96daf1/.github/scripts/ai-pr-review.ts#L295) unconditionally creates `inspection.status: "complete"` from the manifest, including when the response explicitly reports failed inspection or required evidence remaining inaccessible after fallback. The actual validator/renderer accepts that response and produces `COMMENT`, “No findings.”, and the reviewed marker. The identical failed-inspection response is rejected by the base revision. The revised assertion at `tests/unit/t300-ai-pr-review.test.ts:129` now expects this failure to become complete. Keep manifest-derived accounting, but retain a separate blocked inspection outcome that prevents publication as a completed review. Recovered convenience-command failures can remain non-blocking.

2. **P2 — Preserve reviewer access after redacting lens files.** [`.github/workflows/ai-pr-review.yml:324`](https://github.com/awslabs/aidlc-workflows/blob/a7e9e22a13cbf20f5ee808dec16b7faccb96daf1/.github/workflows/ai-pr-review.yml#L324) runs `perl -i` as `runner` after installing the lens files as `runner:ai-pr-review` with mode `0640`. In-place Perl editing creates replacement files; the containing directory has no setgid bit, and the runtime setup never adds `runner` to the `ai-pr-review` group. The replacement therefore gets the runner group, and Perl silently ignores the failed attempt to restore a group the writer does not belong to. Mode `0640` then prevents the final `ai-pr-review` process from reading either specialist report, despite the prompt requiring both. Redact before the final privileged install, or explicitly restore and verify group/read access after redaction.

3. **P2 — Detect control changes from the PR diff, not the branch tips.** [`.github/workflows/ai-pr-review.yml:126`](https://github.com/awslabs/aidlc-workflows/blob/a7e9e22a13cbf20f5ee808dec16b7faccb96daf1/.github/workflows/ai-pr-review.yml#L126) uses `git diff "$base" "$head"`, which includes changes made only on `main` after the PR branched. Once this PR merges, an unrelated branch still based on today's main will be classified as changing reviewer controls and skip all review passes successfully. Using the proposed base tree and an existing unrelated five-file doctor change produces three control paths in this comparison and zero in its actual PR diff. Use the same merge-base comparison as `buildContext` (`base...head`) so upstream-only changes do not trigger self-review exclusion.

Validation: the targeted t300 unit file completed with 14 passing tests and 143 assertions; workflow YAML parsed and all six shell blocks passed `bash -n`; `git diff --check` passed. Local diagnostic probes compare validator behavior on base/head, compare endpoint/merge-base Git diffs, and verify that Perl replaces the file inode. The cross-user permission conclusion also uses the runtime setup and Perl 5.38.2 source; this sandbox could not reproduce multiple Unix groups. No live model or Actions workflow was launched. The inspected Actions run [34686991620](https://github.com/awslabs/aidlc-workflows/actions/runs/34686991620) skipped self-review and therefore did not exercise the model passes.

PR #885, opened 2026-08-24
fix: unwedge reviewer recovery after post-review edits + summary re-confirmation (2.6.26)

apackeer:
A spent stale-receipt recovery pass could dead-end an interactive stage: retry review refused, approval refused on the stale receipt, and the Request Changes reset refused on human-turn freshness because the approve-time revision backstop and the re-recorded summary confirmation silently consumed the turn. A fresh human-backed consolidated-summary confirmation now re-arms exactly one recovery pass, backfilled Recovered GATE_REJECTED rows no longer consume the human turn, and the refusal guidance names the working reset paths.

Note: this branch carries version slot 2.6.26 as committed; origin/v2 CHANGELOG head has since moved past it, so the slot re-bumps at landing per the changelog conflict policy (rebase + re-bump by the second-to-merge).

Closes #851

PR #1154, opened 2026-09-12
fix: measure the review budget against the engine's own ordinal, not the caller's

wowzoo:
## Problem

An `advisory` stage cannot absorb a Request Changes at its approval gate. The only way forward
discards the attempt the human was mid-revision on.

A gate rejection is an attempt boundary — `reviewAttemptAccounting` moves the floor on
`GATE_REJECTED` — so the next review request is ordinal 1 again. `stage-protocol-reviewer.md` says
exactly that should happen on the Part 0 revision path: *"an `advisory` review re-runs as one fresh
advisory pass"*. But a conductor that keeps counting its own passes asks for `--iteration 2`, and the
budget is measured against that claim as well as against the engine's own `expected`:

```ts
if (!recoveryEligible && budget !== null && iteration > budget) { ... }  // the caller's claim
if (!recoveryEligible && budget !== null && expected > budget) { ... }   // the engine's count
...
if (iteration !== expected) { /* "the next iteration is N. Retry with --iteration N." */ }
```

On an advisory stage the budget is 1, so the wrong claim alone raises `REVIEW_BUDGET_EXHAUSTED`
twenty lines before the check that would have named the right ordinal. From there the run has no
forward move:

- the budget refusal says *"do not ask the reviewer again; include the findings in the approval
  summary for the human"*, so the conductor reports `revised`;
- `report --result revised` refuses with `REVIEW_EVIDENCE_MISSING`, because the revision path needs
  the fresh receipt the first refusal just forbade;
- the guard-recovery ask offers a single remedy for a stage in `revising` — a redo jump, which
  restarts the stage.

Two guards, each correct on its own, prescribe what the other refuses. That is the shape #978 named
and #1000 (`d7521b19`, "Supersedes #978") built the machinery for; #885 addressed the neighbouring
case where the stale-receipt recovery pass is already spent. This is a surviving instance, and its
surviving cause is narrow: one of the two guards reads an argument where the other reads state.

The stale-receipt recovery from #758 does not apply here. After a gate rejection the earlier receipt
is not stale — it is outside the attempt window entirely, because the same boundary moved the receipt
scan's floor. `scopeStale` is therefore false and `recoveryEligible` never opens.

## Evidence

Reproduced independently on both Kiro surfaces in one end-to-end run of `rough-mockups`
(`review_class: advisory`), from the audit ledger:

```
10:37:43Z  REVIEW_COMPLETED       iteration 1, READY — terminal receipt
10:48:44Z  GATE_REJECTED          the human asked for changes
10:51–52Z  ARTIFACT_UPDATED ×3    the revision the rejection unfroze
10:52:19Z  ERROR_LOGGED           --iteration 2 → REVIEW_BUDGET_EXHAUSTED, remedies: [redo-jump] only
                                  (report --result revised → REVIEW_EVIDENCE_MISSING)
11:06:31Z  STAGE_JUMPED           the human took the redo; the attempt is gone
```

The refusal names *"review pass 2"*, which is `reviewBudgetMessage(stage, iteration, budget)` — the
caller's number. Had the engine's own count been the subject it would have said pass 1, because the
rejection had already reset it.

## Fix

Delete the `iteration`-based branch.

Admission does not widen: `iteration === expected && iteration > budget && expected <= budget` is
unsatisfiable, so whenever only the deleted branch matched, `iteration !== expected` still refuses —
now with the ordinal to retry. Genuine exhaustion is unchanged, measured by `expected` as before.

What changes is the diagnosis for a wrong ordinal: a retryable error naming the next ordinal, instead
of a typed budget refusal whose only remedy discards the stage.

## Tests

- **New**: an advisory stage where request 1 is recorded READY, the gate is rejected, and
  `--iteration 2` is then refused with *"the next iteration is 1"* rather than *"allows 1 review
  pass"* — and the corrected ordinal is accepted. Fails without this change.
- **Two existing cap cases updated.** Both reached `--iteration 2` while the attempt's next ordinal
  was still 1 — one with no prior request at all, one after an appended `BOLT_STARTED` boundary reset
  it — so they asserted the budget message for what was really an ordinal error. They now spend the
  attempt's single pass first, which is what "the cap is 1" is about.

### DIFF
--- a/core/tools/aidlc-log.ts
+++ b/core/tools/aidlc-log.ts
@@ -2076,13 +2076,20 @@ function handleReview(args: string[]): void {
             message,
           );
         }
-        if (!recoveryEligible && budget !== null && iteration > budget) {
-          refuseAttemptGuard(
-            "REVIEW_BUDGET_EXHAUSTED",
-            "Review requests do not exceed the configured attempt budget.",
-            reviewBudgetMessage(flags.stage, iteration, budget),
-          );
-        }
+        // The budget is measured against `expected` ONLY. `iteration` is the
+        // caller's claim about which pass this is, and it is validated against
+        // `expected` further down with a message that names the right ordinal.
+        // Measuring the budget against the claim instead turned a recoverable
+        // off-by-one into an unrecoverable refusal: after a gate rejection the
+        // accounting floor moves (reviewAttemptAccounting treats GATE_REJECTED as
+        // an attempt boundary), so `expected` is 1 again while a conductor that
+        // kept counting passes `--iteration 2`. On an `advisory` stage, whose
+        // budget is 1, that claim alone produced REVIEW_BUDGET_EXHAUSTED - and
+        // its guidance ("do not ask the reviewer again; include the findings in
+        // the approval summary") then routes to a gate that refuses for
+        // REVIEW_EVIDENCE_MISSING, because the revision path needs the fresh
+        // receipt the refusal just forbade. The only remedy left is a redo jump,
+        // which discards the attempt the human was mid-revision on.
         if (!recoveryEligible && budget !== null && expected > budget) {
           refuseAttemptGuard(
             "REVIEW_BUDGET_EXHAUSTED",
--- a/tests/unit/t271-review-iteration-ceiling.test.ts
+++ b/tests/unit/t271-review-iteration-ceiling.test.ts
@@ -448,6 +448,46 @@ describe("t271 review iteration ceiling", () => {
     expect(rows.length).toBe(1);
   });
 
+  test("a gate rejection resets the ordinal, so a stale --iteration is corrected rather than refused as over budget", () => {
+    // Regression: the budget was measured against the CALLER's `--iteration` as
+    // well as against the engine's own `expected`. A gate rejection is an attempt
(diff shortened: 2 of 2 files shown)


apackeer (approved):
Reviewed commit `5e06b8cb9ee91c87b227c32121de3d6df4296c3f`. No blocking findings.

The change correctly lets an incorrect caller ordinal reach the retry guidance after a gate rejection. The remaining `expected > budget` check still enforces exhaustion, and `iteration !== expected` still rejects incorrect ordinals before a request is recorded. Pending retries and the single-use stale-receipt recovery path remain guarded.

The regression covers the rejected gate, stale ordinal, and successful corrected request. The two adjusted cap tests now consume the current attempt's allowance before asserting exhaustion.

Validation: 75 tests passed across `t271-review-iteration-ceiling`, `t291-review-receipt-recovery`, and `t331-guard-deadlock-liveness`, with no failures or skips. Ran `bash tests/run-tests.sh --debug -P 8 --unit --filter '^(t271-review-iteration-ceiling|t291-review-receipt-recovery|t331-guard-deadlock-liveness)\.test\.ts$'`. Model-driven live harness tests were not run.

PR #1151, opened 2026-09-12
feat!: classic scope v1 parity with scope-owned ceremony switches

apackeer:
# Summary

Restores the AI-DLC v1 experience for the default `classic` scope (Inception + Construction through Build and Test, one approval per stage, one advisory review pass as decision support, sensors and the learnings ritual kept, no summary-confirmation question) by making ceremony scope-owned: reviewers, sensors, the learnings ritual, and summary confirmation can be switched off per scope, per intent, or by environment. Field feedback from the first classic engagement runs is folded in: a readable review copy for people, an engine `wait` verb so the conductor never writes its own shell loop, and one `.aidlc-engine/` folder for the engine's record-level files. Consolidates every per-intent setting into one `config-change` transaction, unifies default-scope resolution, and sets the `balanced` preset to medium effort across all agent groups.

Commits are one per slice, reviewable independently: scope + lib resolver; sensors; learnings module + prose cutover; engine/CLI settings transaction; default-scope resolver + presets + wizard; then the four field-feedback slices (classic advisory + 18 stages, readable review copy, wait verb, `.aidlc-engine/`).

## Changes

- **`classic`**: 18 of 33 stages (the seven Operation stages and CI Pipeline leave its grid; Build and Test stays as the one integrated cross-unit build), `skeleton: off`, `review_cap: advisory`, `sensors: on`, `learnings: on`, `summary_confirmation: off`, `change_control: relaxed`. Every reviewer-bearing stage runs one advisory pass whose findings reach the human at the gate; `--review adversarial` cannot raise the cap and `--review none` still silences the reviewer. Explicit autonomy keeps the swarm's single pre-merge review.

  `core/scopes/aidlc-classic.md` frontmatter after this change:

  ```yaml
  skeleton: off
  review_cap: advisory
  change_control: relaxed
  sensors: on
  learnings: on
  summary_confirmation: off
  ```
- **Scope keys** `sensors`, `learnings`, `summary_confirmation` (`on|off`). Per-intent `/aidlc --sensors|--learnings|--summary-confirmation on|off`. Kill switches `AIDLC_DISABLE_SENSORS`, `AIDLC_DISABLE_LEARNINGS`, `AIDLC_DISABLE_SUMMARY_CONFIRMATION` (recordable via `aidlc config flags --bypass`). Precedence: env → intent state line → scope default → on. Creation writes the three lines `(from scope <name>)`; `--status` shows value and source.
[...]

PR #1159, opened 2026-09-13
fix: make /aidlc compose scopes durable across an engine reinstall

### LINKED ISSUE #963: [Bug]: composed /aidlc compose scopes live only in the harness tree — lost on a copy-channel reinstall, invisible to doctor, unusable from another harness
**Re-verified against v2 @ 2.8.2** (originally filed at 2.6.122; see the comment below for what changed).

## Summary

A `/aidlc compose` CUSTOM scope is stored as two artifacts, both in the **harness tree**: `<harness>/scopes/aidlc-<name>.md` (identity) and an appended column in `<harness>/tools/data/scope-grid.json` (the EXECUTE/SKIP plan). The shared, tool-neutral `aidlc/` tree records only the scope **name** (`intents.json`), never the definition.

Three consequences follow from that placement:

1. **Silent loss.** A copy-channel reinstall overwrites `scope-grid.json`, dropping the composed column. The `.md` survives, so the scope name stays "valid" and resolves as an **all-SKIP phantom** — an emptied plan with no error.
2. **No detection.** Nothing flags this. `plugin sync` and #910's "Composed plugin surface" check are plugin-scoped; a composer scope is not a plugin. Doctor's `Scope validation` check passes an all-SKIP grid.
3. **No portability.** A teammate who pulls the shared `aidlc/` tree but works in a different harness (`.kiro/`, `.codex/`, …) has no scope definition at all, so a composed-scope intent shared via `aidlc/` cannot be resolved or re-shaped on their tool.

This sits just outside #875/#910, which hardened **plugin** composition against a reinstall. An ad-hoc composer scope is a different artifact and is covered by none of it.

## Code-grounded, v2 @ 2.8.2

| Claim | Evidence |
|---|---|
| The shared tree records only the scope name | `core/tools/aidlc-lib.ts:2685` — `IntentRegistryEntry.scope?: string` |
| Composer scopes have no stage `scopes:` frontmatter, so a bare re-transpose cannot rebuild them | `core/tools/aidlc-graph.ts:1075` |
| The fold-back that protects them reads the **on-disk** `scope-grid.json` — the very file a reinstall overwrites | `core/tools/aidlc-graph.ts:1483` `mergeComposedScopes`, called at `:2040` |
| A missing grid column silently degrades to all-SKIP rather than erroring | `core/tools/aidlc-lib.ts:26237` — `stages: grid[name]?.stages ?? {}` |
| Doctor does not flag it: `validateScope` over an all-SKIP grid walks no `consumes`, so it returns 0 errors and the check **passes** | `core/tools/aidlc-utility.ts:4732` |
| `plugin sync` and the composed-surface check are plugin-only | `core/tools/aidlc-utility.ts:3413` |

Scope of the blast radius is small, which is what makes this tractable: a composed scope is written with `keywords: []` and no `runner: true`, so it generates **no scope runner** (`core/tools/aidlc-runner-gen.ts:15-21`). The only two artifacts are the `.md` and the grid column.

## Which upgrade channels lose it

- **Copy channel** — a manual copy of a release tarball's `runtime/<harness>/` over the project (`docs/guide/glossary.md:22`). This overwrites `scope-grid.json` and **loses the composed column**. An `rm`-then-copy loses the `.md` too.
- **Hand-edits / partial restores** of the harness tree — same outcome.
- **Native `aidlc update` / `aidlc init` refresh** — this path **already preserves** composed scopes and is *not* affected: `generatedOverlayCandidate` carries project-only `<harness>/scopes/*.md` into the staged tree (`core/tools/aidlc-init.ts:3413`), `cpSync(currentGrid, stagedGrid)` seeds the existing grid (`:3557`), and the subsequent `compileStageGraph()` folds the composed column back (`:3665`).

So the loss window is narrower than a blanket "every reinstall" — but the **detection** gap and the **portability** gap are unconditional, and the copy channel is a documented, supported install path.

## Field observation

A composed scope existed only under `.claude/scopes/` + `scope-grid.json`. Surviving a copy-channel reinstall required a **manual** pre-install backup of the harness tree — nothing in the framework preserved or flagged it.

## Root cause

Composed scopes are **workflow data** but they live on the **tool-specific** side of the split. Per the project's own "one core, many harnesses" model, the harness tree is the tool-specific shell and `aidlc/` is the shared truth. Stock scopes belong in the harness tree because they *are* engine surface, regenerated from `core/`. A composed scope has no producer in `core/` — it is authored by the user's composer run, which makes it project data that happens to be stored in a generated tree. That is the whole defect; the three symptoms above are its projections.

The two-file split (`.md` here, grid column there) is the mechanism: one of the two files is a generated artifact that any regeneration or copy-over replaces, and losing it degrades silently instead of failing.

## Proposed fix

Make `aidlc/` the **source of record** for composed scopes and the harness tree a **projection** that `graph compile` regenerates — the same relationship the project already uses for `core/` → `dist/` and for stage frontmatter → `scope-grid.json`. Do not change the runtime read path.

**1. A durable record, one file, in the shared tree**

`aidlc/scopes/<name>.md` — project-level, alongside `aidlc/active-space`, above `spaces/`. Project-level rather than space-level because scope resolution is already global today (`/aidlc --scope <name>` is not space-qualified), so this matches current semantics exactly instead of layering a behaviour change onto a bug fix.

One file carries both halves, so the split that causes the silent degradation stops existing:

````markdown
---
name: my-lean-feature
depth: standard
keywords: []
description: ...
---

## Stage Grid

```json
{ "stages": { "intent-capture": "EXECUTE", "units-generation": "SKIP" } }
```
````

The frontmatter reuses the exact keys `loadScopeMetadataAll` already parses, so the identity parser is shared rather than duplicated.

**2. Compile projects it into the harness tree**

`aidlc-graph compile` gains `aidlc/scopes/` as the fold-back source in place of the on-disk `scope-grid.json`, and additionally materializes `<harness>/scopes/aidlc-<name>.md` when it is absent. Both symptoms then resolve through one mechanism:

- **Durability** — a copy-channel reinstall wipes the harness projection; the next compile rebuilds it from `aidlc/`. Self-healing, regardless of which channel did the wiping.
- **Portability** — a teammate on `.kiro/` runs `graph compile` and the scope materializes in *their* harness tree. Cross-harness collaboration works with no extra mechanism.

**3. Detection, as the safety net**

Extend doctor (a sibling to #910's check) to fail on:

- a harness scope `.md` with no grid column (the all-SKIP phantom),
- an `aidlc/scopes/` record with no harness projection,
- a scope referenced by `intents.json` or a state file with no resolvable definition anywhere,

each with `graph compile` named as the remedy — so the loss can never again be silent, including for any path that bypasses compile.

**4. Migration**

Existing installs have composed scopes only in the harness tree. The first `graph compile` back-fills `aidlc/scopes/` from the current `.md` + grid column pair (one-way adoption), so no existing composed scope is lost on upgrade.

## Related

- #875 / #910 — plugin composition erased by a reinstall. This gap sits just outside that fix: ad-hoc composer scopes vs. plugin composition.

### PR DESCRIPTION
**Pull request by DrakeRHEE**

Closes #963.

## Problem

A `/aidlc compose` custom scope was stored only as a harness pair — `<harness>/scopes/aidlc-<name>.md` plus an appended `scope-grid.json` column — while the shared `aidlc/` tree recorded only the scope *name* (`intents.json`). Both halves live in a generated tree, and one of them is a shipped generated file.

That placement caused three failures, all silent:

1. **Loss.** A copy-channel reinstall replaces `scope-grid.json`, dropping the composed column. `loadScopeMapping` falls back to `{}` for a missing column (`aidlc-lib.ts`), so the scope stays "valid" and resolves as an **all-SKIP phantom** — an emptied plan, no error.
2. **No detection.** `plugin sync` and #910's "Composed plugin surface" check are plugin-scoped; a composer scope is not a plugin. `Scope validation` cannot see it either: `validateScope` over an all-SKIP grid walks no `consumes`, so it returns zero errors and the check **passes**.
3. **No portability.** A teammate who pulls the shared `aidlc/` tree but works in a different harness (`.kiro/`, `.codex/`, …) has no definition at all, so a composed-scope intent shared via `aidlc/` cannot be resolved on their tool.

The recompile guard that was supposed to protect the scope folded it back from the **on-disk** `scope-grid.json` (`mergeComposedScopes`) — the very file the reinstall had just overwritten. There was no durable source to fold back from.

## Approach

A stock scope belongs in the harness tree because it *is* engine surface, regenerated from `core/` on every build. A composed scope has no producer in `core/` — it is authored by a user's composer run — so it is project data that merely happened to be stored in a generated tree.

This makes `aidlc/scopes/<name>.md` the **source of record** and the harness pair a **projection** that `graph compile` regenerates. That is the same relationship the project already uses for `core/` → `dist/` and for stage frontmatter → `scope-grid.json`, and it puts the definition under the `aidlc/` roof where the user's committed work already lives. **No runtime read path changes** — the engine still reads the harness grid.

One record carries both halves, so the two-file split that caused the silent degradation stops existing:

````markdown
---
name: my-lean-feature
depth: Standard
keywords: []
description: ...
---

# my-lean-feature

Whatever the author wants to write, including a "## Stage Grid" heading of
their own and fenced code of their own.

<!-- BEGIN aidlc composed-scope-grid: generated by `aidlc engine graph compile` — reshape the plan through /aidlc compose, not by editing here -->

```json
{ "stages": { "intent-capture": "EXECUTE", "units-generation": "SKIP" } }
```

<!-- END aidlc composed-scope-grid -->
````

A record the writer produced is exactly the harness identity `.md` plus that one appended region, so projecting is a pure strip and back-filling is a pure append — neither direction re-renders frontmatter, so no authored identity or prose is reformatted. The identity is everything **above** the region: content appended below the END sentinel stays in the record (nothing rewrites a record that already exists) but is not carried into the harness scope file, which the guide and the parser's own docstring now state rather than leave to be inferred.

The boundary is a sentinel pair rather than a Markdown heading because the identity half is **prose the user wrote**. A heading is something a scope author can legitimately write while documenting their own plan, and anchoring on one let an authored heading capture the split: the parser adopted the first ```json fence in the user's prose as the grid — resolving the scope to a different set of executed stages — and dropped every authored line after the collision on the next write-back. The parser now requires exactly one sentinel pair (refusing anything else rather than picking an end) and looks for the fence only inside the region, and the emitter refuses an identity that already carries a sentinel. This repo already fences generated regions inside authored files the same way (SKILL.md's `<!-- BEGIN: compiled ... -->` blocks, the guide's scope-stage matrix, plugin prose fragments).

Both symptoms then resolve through one mechanism: a reinstall wipes the projection and the next compile rebuilds it from the record; a collaborator on another harness runs `graph compile` and the scope materializes in *their* tree.

Placement is project-level rather than space-level deliberately: scope resolution is already global (`/aidlc --scope <name>` is not space-qualified), so this matches today's semantics instead of layering a behaviour change onto a durability fix.

## What changed

**`core/tools/aidlc-graph.ts`**
- `composedScopesDir()` — the record directory, resolved on the same ladder as `rulesDir()` (env seam → project workspace root → module → packaged distribution), with `AIDLC_COMPOSED_SCOPES_DIR` as the seam.
- `parseComposedScopeRecord` / `renderComposedScopeRecord` / `loadComposedScopeRecords` — the record contract. Malformed input **throws with the path named**: a record's whole purpose is to survive, so a silent skip would reintroduce the quiet loss it exists to prevent.
- `composedFoldBack()` — replaces the on-disk grid as compile's fold-back source. Records win; the on-disk grid remains a **one-time migration fallback** for an install composed before records existed. Both are still gated on an installed identity file, so an orphan column is not resurrected, and a record named after a stock scope can never shadow it.
- `materializeComposedScopeIdentities()` / `backfillComposedScopeRecords()` — the write side, wired into the `compile` CLI inside the existing audit lock so a reader never observes a scope half-restored. Materialize runs *before* the transpose (the fold-back needs the identity file present); back-fill runs *after* (the emitted grid is what names the composed columns).
- `composedScopeWritesEnabled()` — the record and scopes directories have independent env seams, so a caller isolating one and not the other would make compile derive a record from a fixture scope and write it into the real workspace. Writes now require the seams to be set together or not at all. Reads are unaffected, so the grid fallback still works under a half-isolated fixture.

**`core/tools/aidlc-utility.ts`** — a new **Composed scope durability** doctor check, failing on the three ways the pairing breaks: a scope file with no grid column (the phantom), a record not yet projected, and a **runnable** workflow whose recorded `Scope` has no resolvable definition. `graph compile` is named where compile can reach the cause; the third also says to restore the missing record. A missing column with **no record behind it** is counted apart, because the transpose emits a column only for a scope some stage declares in its `scopes:` frontmatter — so compile would exit 0 and leave that row red. That arm names the three routes that do reach it: restore the record, finish tagging the stages, or delete the scope file. Plugin-owned scopes are excluded (a disabled plugin's scope legitimately has no column). All three **fail** rather than advise — every one silently changes which stages a workflow runs. Completed workflows are skipped by the same enumeration `activeWorkflowDependencyViolations` already uses in this file: a finished workflow needs no scope definition, and `intent archive` refuses a completed intent, so flagging one would be unrecoverable.

**`core/tools/aidlc-init.ts`** — the staged refresh reads the project's records **and materializes a missing identity file into the staged root before compiling**, exactly as the `compile` CLI does. Reading the records is not sufficient on its own: the fold-back only resurrects a column whose identity file exists, so without the materialize step a record whose projection was already gone — the reinstall this change exists for — was filtered out and its column silently dropped into the tree about to be installed.

**Docs** — `docs/guide/05-scopes-and-depth.md` gains a "Where a composed scope is stored" section: record vs. projection, the record's shape and which half is hand-editable, that `compile` writes the harness scope file only when it is missing (so a record edit needs the projection deleted to re-project), and how to recover from an unreadable record. The doctor check is added to `12-cli-commands.md` and `15-troubleshooting.md` — describing the three things it actually fails on, and stating that it does **not** compare a record's descriptive frontmatter against its projection, since the grid always comes from the record and compile will not clobber a hand-edit. `docs/reference/03-orchestrator.md`, `docs/reference/17-skill-system.md`, and `docs/harness-engineering/04-scopes.md` point at the new section.

## Migration

None required. The first `graph compile` after upgrading back-fills a record for every composed scope already on disk, one way and once, then keeps it as the source. A project with no composed scope is untouched and no `aidlc/scopes/` directory is created.

Per the Release Metadata Policy in `AGENTS.md`, this PR leaves `core/tools/aidlc-version.ts`, the README badge, and `CHANGELOG.md` release entries alone — the release-preparation PR carries them.

## Testing

Clean: `bun scripts/package.ts --check` (two independent builds byte-identical across all seven harnesses), `bun run typecheck` (all three projects), `biome check --error-on-warnings`. Tested on macOS.

**Targeted verification** — everything that touches the changed surfaces, 337 tests / 4,156 assertions, all passing:

- scope + grid + doctor: `t344`, `t341`, `t124-scope-transpose`, `t125-scope-files`, `t191-composed-scope-write`, `t198-compose-surfaces`, `t204-compose-marker-doctor`, `t184-stage-graph-drift`, `t257-default-scope-resolver`, `t223-naming-enforcement`, `t222-plugin-runner-naming`, `t313-plugin-doctor-checks`, `t314-plugin-reinstall-doctor`, `t224-plugin-selection`
- the scope surfaces #1151 introduced, since this branch now sits on top of it: `t337-ceremony-policy`, `t338-ceremony-verb`, `t340-default-scope-resolver`, `t341-orchestrate-wait`
- compile golden-masters: `t110`, `t88`, `t89`
- drift guards for what this PR edits: `t244-scope-matrix-doc-sync`, `t174-docs-legacy-refs-gate`, `t239-documentation-parity`, `t68-version-changelog-sync`, `gen-coverage-registry`, `t134-mechanism-honesty`

`t243-install-mechanism` — the projection-channel and legacy-signature inventory guard for the `aidlc-init.ts` surface this touches — runs in the CI unit shards. This change adds no entry to any harness signature inventory and leaves every `dot-gitignore` alone.

The five **live-Claude composer journeys** also pass, re-run on the rebased tree — `t189-compose-dispatch`, `t192-compose-front-journey`, `t193-compose-report-journey`, `t196-compose-inflight`, `t197-compose-chat-inflight` — which is what confirms the real composer write path still works end to end on top of #1151.

### New tests

Both start from **the verbatim output of a live composer run**, captured into `tests/fixtures/composed-scope-live/` by driving `/aidlc compose` on t192's task and approving at the gate. That matters: a hand-written imitation of a composed scope missed things the real composer actually emits — `skeleton:` and `change_control:` frontmatter, `depth: Standard`, and a 40-line prose body carrying its own `##` headings, which is exactly why the grid boundary is a sentinel and not a heading. Using the real bytes is what proves the projection round-trips the artifact users actually have. Each file guards the fixture's premise, so a refreshed capture cannot silently stop proving anything; refreshing means re-driving t192 and copying the composed `.md` out. No model runs in the suite — the fixture is the seam.

- **`tests/unit/t344-composed-scope-record.test.ts`** — the record contract in-process: render/parse round-trip is byte-stable and the projection carries no generated region, over both a synthetic case and the real fixture (all 33 grid cells, every frontmatter key, both prose headings recovered exactly). A dedicated block pins that **authored prose cannot capture the boundary** — a record whose prose contains both a `## Stage Grid` heading and a decoy ```json fence still yields the real grid, keeps every authored line including that heading, round-trips byte-stable, refuses duplicated sentinels, and refuses an identity that already carries one. A further case pins the three ceremony switches #1151 gave a scope — `sensors`, `learnings`, `summary_confirmation` — by key name through a round trip, because a projection that re-rendered frontmatter would silently turn a scope's disabled ceremony back on. Then each malformed shape throws with the offending path named (including a fence *outside* the region), and `composedFoldBack`'s source priority — record beats a conflicting column, a grid-only column survives as the migration fallback and is reported for back-fill, a stock name never shadows, a name with no identity file is dropped from both sources, a malformed grid contributes nothing while records still fold. Three cases cover the record's edges: the sentinel refusal names its file, the missing-region error prints both sentinels byte-exact (so the hand-restore it offers is actually reachable), and content appended below the END sentinel stays in the record while staying out of the projection — the one place the round trip is deliberately asymmetric.
- **`tests/integration/t341-composed-scope-durability.test.ts`** — t314's composed-scope twin, driving the shipped tools through a full reinstall: back-fill → destroy the harness pair the way a copy-channel reinstall does → doctor **fails** naming the scope → one compile restores both halves, the identity file byte-identical to the composer's original and the grid carrying the **approved** cells → doctor goes green → a new intent on the recovered scope carries the authored plan (the property that actually broke for users), asserted on a `units-generation: SKIP` + `observability-setup: EXECUTE` pairing that no stock scope and no all-SKIP phantom can produce. Plus the merge-copy phantom shape on its own, the no-composed-scope case leaving no trace, and **the same recovery through the real `aidlc config` install path** rather than a simulated one — which is the route that was broken. Two more cases drive the doctor row's honesty through the shipped CLI: a scope nobody stage-tagged is reported as unbuildable and compiling again is shown not to move the row, and a record hand-copied into the harness scopes tree makes compile refuse with that file named.

Every new guard was verified by mutation: reverting the sentinel anchor to a heading fails exactly the two collision tests, removing the materialize step from the staged refresh fails exactly the update-path test, and reverting each of the message and doctor-arm fixes fails exactly the case that covers it. Notably, dropping the path argument the writer's refusal now takes is caught **only** by the CLI-driven case and not by calling the writer directly — which is why that one is an integration case rather than a unit one. Nothing here passes vacuously.

Running the wider suite is also what caught the cross-seam write bug described above: it surfaced in three unrelated tests, as a scope file leaked into the shipped `scopes/` tree that the new durability check then correctly flagged as a phantom.

## Remaining work (deliberately not in this PR)

The composer still writes the harness pair directly; the record is minted by the compile that the stage-graph rebuild hook fires immediately afterwards. That leaves a brief window — compose, then reinstall before any compile — in which the scope is still lost. The new doctor check catches exactly that state.

Closing the window means moving the composer's sanctioned write target to `aidlc/scopes/<name>.md`, which touches prompt prose in seven harness `SKILL.md` files plus the composer agent, its knowledge doc, and `detect --json`'s advertised write path. That is a prose-surface change with a different review shape, so it is better as a follow-up on top of a landed record format than bundled here.

Tracked as #1202, filed before this merges so the remaining window does not leave with the issue this PR closes — the new doctor row detects that state but does not close it.

Also left out: adding `aidlc/scopes/**` to the "COMMITTED (NOT ignored, listed here for the record)" comment in `harness/*/dot-gitignore`. It is worth having — it would stop a future contributor from ignoring the records and silently breaking portability — but editing those files changes the shipped `.gitignore` bytes, which drops the current release's hash out of each harness's `legacySignatures.wholeFileHashes` adoption inventory. Doing it correctly means adding the superseded hash to seven manifests (three of which have no `.gitignore` legacy block yet) and re-pinning t243's inventory: a lot of opaque-hash churn to carry in a bug fix, for a comment. Nothing is ignored under `aidlc/scopes/` today, so records are committed by default and the fix works without it. The user-facing instruction to commit them is in the new User Guide section instead.

### CODE REVIEW

**Review by wowzoo — CHANGES_REQUESTED** [review]
https://github.com/awslabs/aidlc-workflows/pull/1159

Reviewed head `7d2e5166` against base `main` @ `9b8847b5` — up to date, one commit ahead. All PR checks currently pass, with `deploy` skipped.

The design argument carried a long way: making `aidlc/scopes/<name>.md` the record and the harness pair a projection puts the definition on the same footing as `core/` → `dist/`, and #963 named its evidence per claim. The record contract is where the problems are, and two of them break the property the format is built on.

## Blocking — an authored `## Stage Grid` heading silently truncates the record and swaps the grid

`parseComposedScopeRecord` splits on the **first** `\n## Stage Grid` and keeps everything before it as `identity`; `renderComposedScopeRecord` re-appends that heading. So the boundary is a heading a user can legitimately write in their own prose. Round-tripping a record whose description documents its own grid:

```
input record                371 bytes, real grid {intent-capture: EXECUTE, units-generation: SKIP}
parsed.stages               {"DECOY": "EXECUTE"}        ← the prose example fence became the grid
parsed.identity             frontmatter + "Notes for humans." only
lost                        "do not edit by hand", "More authored prose that must survive."
re-rendered                 164 bytes, byte-stable: false
```

Two consequences, and the second is the one this PR exists to prevent:

- Authored prose after the collision is dropped the next time compile writes the record back — against the PR's own statement that "no authored identity or prose is ever lost or reformatted" and the `renderComposedScopeRecord` comment that a round trip is byte-stable.
- The grid adopted is the first ```json fence after the collision point, so the scope resolves to a **different set of executed stages** than the record holds. That is a silent change to which stages a workflow runs.

`t340` only ensures its fixture lacks the heading, so nothing covers the collision. The format needs to reject a colliding heading, anchor on the last occurrence, or use a delimiter a user cannot author by accident.

## Blocking — `aidlc update` does not restore a missing identity, so the two recovery paths disagree

The `compile` CLI calls `materializeComposedScopeIdentities` **before** `compileStageGraph`, and its comment gives the reason: "the fold-back only resurrects a grid column whose identity file exists." `composedFoldBack` bears that out — `recordNames` filters on `installedScopeNames.has(name)`, so a record whose identity file is absent is excluded.

The staged refresh in `core/tools/aidlc-init.ts` sets `AIDLC_SCOPES_DIR` to the staged harness scopes directory and `AIDLC_COMPOSED_SCOPES_DIR` to the project's real records, then calls `compileStageGraph()` directly — with no materialize step. A record whose identity is missing from the staged projection is therefore filtered out, its column is not restored, and that staged tree becomes the installed one.

That contradicts the comment added just above it, which says pointing at the real records means "the record is the source of record on both paths". It is the source on both paths only when the identity file already exists. Worth noting the exposure is narrower than it looks: #963 records that the native `aidlc update` path already carries project-only `<harness>/scopes/*.md` through the overlay, so the case that bites is the one where the identity is already gone — which is the reinstall this PR is fixing.

## The doctor row and the docs claim agreement that nothing checks

`docs/guide/12-cli-commands.md` describes the new row as "Every composer-authored scope's durable `aidlc/scopes/<name>.md` record and its harness projection agree." The check tests three things — a missing or empty grid column, a record with no scope metadata, and a workflow scope that does not resolve. None compares the record's identity against the projected file.

`materializeComposedScopeIdentities` skips a scope whose identity file already exists, so nothing reconciles them either. A record and its projection can hold different `depth`, `description` or `keywords` indefinitely while the row reports agreement. Either the check should compare them or the sentence should describe what it actually verifies.

## The dangling arm is inconsistent with this file's own policy on completed work

The loop skips only archived intents. The comparable check earlier in the same file enumerates "every non-complete workflow" and skips `intent.status === "complete"` as well as `Status: Completed`/`Archived`, with `t224` pinning that completion removes the block. So the new arm holds completed workflows to a standard the established one deliberately does not.

That matters because the state is unrecoverable: a finished workflow whose composed scope was deleted afterwards fails `--doctor` forever, and `intent archive` refuses a completed intent outright — "A completed workflow is already terminal and is not archived." The only supported recovery is recreating a scope the user deliberately removed.

## Blocking — release metadata in a `fix:` PR

This bumps `core/tools/aidlc-version.ts` to `2.8.3`, the README badge to `version-2.8.3-blue`, and adds a `## [2.8.3] - 2026-09-13` entry. `AGENTS.md` assigns those three to the release-preparation PR, and the pre-submit list in `CONTRIBUTING.md` repeats it for ordinary fix PRs.

Nothing catches it: the changelog guard fails only when a PR *removes* an existing heading, and `t68` checks that the three surfaces agree — which they do here, so the bump is internally consistent and fully green. #1040 removed the same three edits after maintainer direction.

I found no public release-preparation PR, branch, or request for 2.8.3. If this was coordinated somewhere I cannot see, say so and I will drop the item.

## On the malformed-record throw

A record that is missing frontmatter or `name`, has no `## Stage Grid`, has no json fence, or collides with another record's name aborts compilation, and through the staged refresh it aborts `aidlc update` too. Every message names the path and the required edit, and this matches the pattern `aidlc-rule-schema.ts` describes for `compileStageGraph` — rules are hand-editable too, so the precedent does apply. What the PR does not add is any recovery beyond manually repairing or deleting the record, which is worth a sentence in the docs given the record is user-authored and committed.

## Coverage

`t340` passes 19/0 and `t341` 5/0 here, on `7d2e5166` after the packager, and the ratchet check reports "OK (fresh, guards green, ratchet held)" at 391 → 396. `t341` does prove the approved EXECUTE and SKIP cells survive, comparing the recovered grid and then the stage decisions in fresh intent state.

Two qualifiers on what that covers. `simulateReinstall` deletes the identity file and the grid column directly rather than running the copy installer or the update path, so the end-to-end channel is not exercised. And the registry is claim-based rather than execution coverage, so the five new entries record intent: there is no test for the heading collision, the staged-update path, or record-versus-projection divergence.

## Requesting changes

On the author's next push: drop the three release-metadata edits, and address the record-format collision, the staged-update recovery gap, and the agreement the doctor row and docs claim. The placement decision and the record/projection relationship are right — these are defects inside an approach worth keeping. Happy to re-review as soon as there is a new head.

**Comment by DrakeRHEE** [comment]
Thanks for the thorough review. All five findings were real. New head is `8d8c1f24`.

I reproduced both blocking bugs before fixing them, and I checked each new test by breaking the fix again to make sure the test actually fails.

## 1. Record format: the grid boundary is now a sentinel

I reproduced your case exactly: 328 bytes in, the grid came out as `{"DECOY": "EXECUTE"}`, two authored lines vanished, 136 bytes back, byte-stable false.

You were also right about why `t340` missed it. I knew the heading could collide, and I wrote a test asserting the fixture didn't contain it. That guards the fixture, not the code. Wrong half.

I went with your third option, a delimiter users can't type by accident:

````markdown
<!-- BEGIN aidlc composed-scope-grid: generated by `aidlc engine graph compile` — ... -->

```json
{ "stages": { ... } }
```

<!-- END aidlc composed-scope-grid -->
````

Why not the other two. Anchoring on the last occurrence works for records this code writes, since the generated region is always appended last. But a hand-edited record can put an authored region last, and then we're back to picking an end and hoping. Rejecting a colliding heading would break a legitimate document for a reason the author had no way to predict.

The sentinel also matches how this repo already fences generated content inside authored files: `SKILL.md`'s `<!-- BEGIN: compiled ... -->` blocks, the scope-stage matrix in the guide, plugin prose fragments.

Three checks now, not one:

- The parser counts sentinels and refuses anything but exactly one pair. It never picks a side.
- It looks for the `json` fence only inside the region, so an authored fence can't be mistaken for the grid.
- The writer refuses an identity that already has a sentinel, so you get the error where the bad input came in rather than one step later.

`t340` has a block for this now: real grid wins over the decoy, every authored line survives including the colliding heading, round trip is byte-stable, duplicate sentinels are refused, writer refuses a sentinel-bearing identity. If I put the heading anchor back, exactly two of those fail.

## 2. Staged update: materialize runs there too now

Confirmed. `prepareRefreshSource` called `compileStageGraph()` with no materialize step, and `recordNames` filters on `installedScopeNames.has(name)`. So a record whose identity file was already gone got filtered out, and its column went missing from the tree about to be installed.

Your point that the case which bites is the one where the identity is already gone is the whole thing. That's the reinstall this PR is for.

It materializes into the staged root first now. I also rewrote the comment that said the record was "the source of record on both paths", since that wasn't true. Reading the records is necessary but not enough; the fold-back needs the identity file present too.

`t341` has a test that runs the real `aidlc config` install path: install, compose the old way, compile to mint the record, delete the projection, refresh from the same release, then check both halves came back with the approved cells. Remove the materialize step and that test fails.

That also closes part of the coverage gap you flagged. The update channel is exercised end to end now. The copy installer still isn't.

## 3. Doctor row and docs: fixed the sentence

You're right that nothing compares a record's identity to its projection, and that `materializeComposedScopeIdentities` skipping an existing file means nothing reconciles them either.

I took your second option and fixed the wording rather than adding the comparison. Reconciling means deciding to overwrite someone's hand-edit, and that's a behaviour change, not a durability fix.

The CLI reference and the troubleshooting list now say what the check actually fails on, and say plainly that it does not compare descriptive frontmatter. The guide gives the remedy if a projection has drifted: delete the harness file and recompile. The code comment records the reasoning, which is that the grid always comes from the record, so a divergence can only affect `depth`, `description`, `keywords` — never which stages run.

## 4. Dangling arm: matches the existing enumeration

Also right, and the consequence you described is what settles it. `intent archive` refuses a completed intent, so a completed workflow with a deleted scope had no way out except recreating the scope.

It now skips completed workflows the same way `activeWorkflowDependencyViolations` does, with `intent.status === "complete"` plus `Status: Completed`/`Archived`. Completion releases this check just like it releases the plugin-selection block that `t224` pins.

## 5. Release metadata: dropped, and this one was my mistake

All three are gone and byte-identical to `main` now. Nothing was coordinated anywhere, so please keep the finding rather than withdrawing it.

Here's what happened. At the start of the session I read `AGENTS.md` from a local checkout that was 39 commits behind `main`, where it still said "IMPORTANT: Every user-visible PR bumps `core/tools/aidlc-version.ts`…". #1096 replaced that with the current Release Metadata Policy. I created my worktree from `origin/main` afterwards and never re-read the file, so I was writing against current code while following rules that had been reversed. The last eight merged `fix:`/`feat:` commits all leave `aidlc-version.ts` alone, which I should have looked at.

## Malformed-record recovery

Added, as you suggested. The guide has a "When a record is unreadable" paragraph now: repair the file the error names, or delete it — if the harness projection is still there, the next compile back-fills a fresh record from it. The error message itself names that path, and `t340` checks that it does.

## What I didn't change

The composer still writes the harness pair directly. The record gets minted by the compile that the rebuild hook fires right after, so there's a short window where a reinstall still loses the scope. It's in Remaining work, and the new doctor check catches that state.

Happy to fold it into this PR if you'd rather see it as one change. It's prompt prose across seven `SKILL.md` files plus the composer agent, its knowledge doc, and the write target `detect --json` advertises.

## Verification on the new head

278 tests / 3,700 assertions across the scope, grid, doctor, compile and docs-drift surfaces. `t243` at 84/84 for the `aidlc-init.ts` change. `package.ts --check` deterministic across all seven harnesses, typecheck on all three projects, biome clean.

**Comment by DrakeRHEE** [comment]
Rebased onto `main` at 2.9.0 (`a8c477c2`) — the branch was six commits behind and had gone `CONFLICTING`.

The textual conflicts were small: one import line in `aidlc-graph.ts` and one in `aidlc-init.ts`, both just the union of an import each side had added at the same spot, plus the two generated coverage artifacts. I regenerated `tests/.coverage-registry.json` and `.coverage-ratchet.json` with `bun tests/gen-coverage-registry.ts` rather than hand-merging them; the delta against `main` is additive (`function` 404 → 409). The three release-metadata files are byte-identical to `main` — worth restating since an earlier revision of this PR got that wrong.

The part worth a second look is #1151, which landed in the same area: a scope now owns three ceremony switches (`sensors`, `learnings`, `summary_confirmation`) as frontmatter keys. The record contract needed no change to carry them — projection is a pure strip and back-fill a pure append, so nothing re-renders frontmatter and the new keys ride along. I'd rather that were pinned than inferred, so `t340` gains a case asserting those keys by name through a round trip: a projection that re-rendered frontmatter would silently turn a scope's disabled ceremony back on, which is the same class of silent behaviour change this record exists to prevent.

One thing to flag: #1151 also added `tests/unit/t340-default-scope-resolver.test.ts` and `tests/unit/t341-orchestrate-wait.test.ts`, so my two new files now share their numbers. I left them as they are because the repo already does this (`t314` spans four files), but I'm happy to renumber if you'd rather keep the IDs distinct.

Re-verified on the rebased tree: `bun scripts/package.ts --check` byte-identical across all seven harnesses, `bun run typecheck`, `biome check --error-on-warnings`, and the targeted set at 323 tests / 4,113 assertions — now including #1151's own `t337-ceremony-policy`, `t338-ceremony-verb`, `t340-default-scope-resolver`, and `t341-orchestrate-wait`. The five live-Claude composer journeys were re-run too, since #1151 sits underneath them: the real composer write path still works end to end.

**Review by wowzoo — COMMENTED** [review]
https://github.com/awslabs/aidlc-workflows/pull/1159

Re-reviewed head `cda58188` against base `main` @ `22f5d1b1` — up to date, four commits ahead. 22 checks pass on this head and `deploy` is skipped, with `Require Contributor Statement` among the passing ones.

The heading collision, the staged-update gap, the doctor arm, the docs wording and the release metadata are all addressed at this head, and the collision fix in particular is stronger than the two alternatives I offered.

## What I verified at this head

**An ordinary Markdown heading can no longer capture the grid boundary.** The sentinel pair at `aidlc-graph.ts:465-467` replaces the heading, and the parser does three separate things with it: it counts both sentinels and refuses anything but exactly one pair rather than picking an end (`:497-513`), it searches for the ```json fence only inside the region so an authored fence cannot be adopted (`:521-524`), and the writer refuses an identity that already carries a sentinel (`:570-575`). Counting rather than searching from one end is the part that matters — it means a hand-edited record cannot resolve to whichever side the implementation happened to prefer. An ordinary authored heading can no longer capture the split, and the writer guard holds: a record whose prose carries the sentinel pair *and* a real generated region is refused rather than written with two.

**The staged update path restores an identity before it compiles.** `materializeComposedScopeIdentities(root)` now runs ahead of `compileStageGraph()` at `aidlc-init.ts:3876`, with `AIDLC_COMPOSED_SCOPES_DIR` pointed at the project's real records (`:3865`) and `root` as the write target, so the restore lands in the staged tree and not in the live project. The comment that claimed the record was the source of record on both paths is corrected rather than left to be read as true.

**The doctor arm matches the enumeration the rest of the file uses.** `aidlc-utility.ts:4936-4941` now skips archived intents, `intent.status === "complete"`, and `Status: Completed`/`Archived`, so completion releases this check the way it releases the plugin-selection block.

**The docs describe what the check actually fails on.** `docs/guide/12-cli-commands.md:666` and `15-troubleshooting.md:384` now state the three failing conditions and say plainly that descriptive frontmatter is not compared, and `aidlc-utility.ts:4900-4905` records why comparing it would be a behavior change rather than a durability fix. Fixing the sentence was the right call over adding the comparison.

**Release metadata is gone.** `git diff 22f5d1b1...cda58188` touches none of `core/tools/aidlc-version.ts`, `README.md`, or `CHANGELOG.md`. Thank you for tracing where the stale rule came from — that is more useful than a fix on its own, because the 39-commits-behind checkout is a trap anyone can walk into.

## The new tests fail without the fixes

I ran this head's two test files against `f84b0063`, the earlier revision of this same branch, because `main` has no `parseComposedScopeRecord` and a base-branch arm would measure nothing. Single-file mode in both trees, after `bun install --frozen-lockfile --ignore-scripts` and `bun scripts/package.ts`:

| file | head `cda58188` | earlier revision `f84b0063` |
|---|---|---|
| `tests/unit/t340-composed-scope-record.test.ts` | 30 pass / 0 fail | 13 pass / **17 fail** |
| `tests/integration/t341-composed-scope-durability.test.ts` | 6 pass / 0 fail | 4 pass / **2 fail** |

The cases that fail without the fix are the ones that should: the real grid winning over the decoy fence, every authored line surviving including the colliding heading, the round trip staying byte-stable, a duplicated sentinel being refused, and the writer refusing a sentinel-bearing identity. Worth being exact about the totals, since they are larger than that list — the rest of `t340`'s 17 are sentinel-format and changed-diagnostic assertions, and of `t341`'s two, the staged-refresh case is the one that exercises the update defect while the full journey stops earlier, on a sentinel-presence assertion.

The ceremony case added in `d8059ac4` passes on the earlier revision too, which is what a pin should do — the record contract genuinely needed no change to carry #1151's keys, and asserting them by name is the difference between knowing that and assuming it. `bun run check` is rc=0 here (777 files, 3 biome infos).

## Four non-blocking items

### P3 — content below the END sentinel is carried nowhere, and the docstring says otherwise

`parseComposedScopeRecord` returns `identity: body.slice(0, beginAt)` (`aidlc-graph.ts:556`), so anything a user writes *after* the END sentinel is outside the identity. It stays in the record — I traced every write into the record directory and there is exactly one, `aidlc-graph.ts:719`, guarded by `if (existsSync(recordPath)) continue;` at `:717`, so no shipped path ever re-writes an existing record and nothing is lost from disk. What it does mean is that `materializeComposedScopeIdentities` projects only the part above the region (`:683`), so a restored harness scope file silently lacks that trailing content while the record keeps it.

That is small, and I raise it only because the docstring at `:449-453` describes the pair as "a pure strip" in which "no authored identity or prose is ever lost or reformatted", and the guide's "The record's shape" section claims the area above the sentinel and calls the region generated without saying who owns the area below. Appending to the end of one's own Markdown file is the ordinary way to add to it, so what a user needs stated is which of the two areas the tool carries. The behaviour itself is defensible; the sentence describing it is what does not match.

Two smaller things share that shape and are worth the same sentence. A record whose grid fence someone compacted to one line, or which carries extra blank lines before the region, re-renders to different bytes, so the byte-stability the docstring claims holds for records the writer produced rather than for records generally. And a hand-authored record carrying only an authored sentinel pair parses that pair's fence as its grid — reasonable, since a hand-authored region is indistinguishable from a generated one, but it is the one case where the reserved sentinel is doing all the work. Nothing covers any of this: no case in `t340` or `t341` presents a record with content below the END sentinel — the hand-built bodies in the parse-failure block end at it (`:259-260`) and every other input is produced by the renderer — and all four identity assertions compare with `trimEnd()` (`:90`, `:133`, `:173`, `:227`).

### P3 — the phantom row's remedy does not clear the phantom it names

A scope file that no stage tags gets no grid column, because the transpose only emits columns for names a stage declares and `enabledScopeNames()` is `null` on a default install — `harness.json` carries no `plugins` key in any of the eight trees the packager emits at this head (`aidlc-graph.ts:1940-1943`, `aidlc-lib.ts:525-527` and `:347-350`, `transposeScopeGrid` at `:1753-1780`). The durability row then reports it as a phantom, which is right, and names `graph compile` as the fix, which does not work for that cause.

Measured in a temp project with the shipped tools, authoring `.claude/scopes/aidlc-hotfix-probe.md` with no stage `scopes:` tag:

| step | result |
|---|---|
| `aidlc-graph.ts compile` | status 0, no grid column for the scope |
| `aidlc-utility.ts doctor --verbose` | status 1 — `Composed scope durability: 1 problem(s) - 1 with no grid column (resolves as an empty all-SKIP plan) [hotfix-probe]`, `fix: run bun .claude/tools/aidlc-graph.ts compile` |
| that fix applied, then doctor again | compile status 0, doctor status 1, same row |

The detection is correct — an untagged scope file really does resolve all-SKIP, and that is worth failing on. What the row does not do is name the cause or a remedy that reaches it, and the way in is step 1 of the add-a-scope workflow in `docs/harness-engineering/04-scopes.md:127-140`, whose step 4 is "Run `/aidlc --doctor`". A half-finished authoring session is exactly when someone runs it.

The constraint is that the row distinguishes a column compile can rebuild — one whose name has a record — from one it cannot, and names the second cause. Both signals are already in hand at that point: the records are loaded in the same block (`aidlc-utility.ts:4913`) and `stockScopeNames` is built from `loadGraph()` at `:4908-4911`. There is no false positive on a clean install: all eleven stock scopes are stage-tagged, and the arm skips them regardless.

### P3 — the one record error that does not name its file

`renderComposedScopeRecord` refuses a sentinel-bearing identity (`aidlc-graph.ts:570-575`) with a message that carries no path, and its only caller is reading the file the message is about (`backfillComposedScopeRecords`, `:721-725`, `readFileSync(identityPath)`).

I reached it by constructing the state directly — a grid column present, no record, and the harness identity file carrying a sentinel region — and calling the function with both directory seams set:

```
Composed scope identity already contains an aidlc composed-scope-grid sentinel;
it must hold only the authored scope file, not a generated grid region.
```

I did not find a path the docs recommend that arrives there; the way in is a hand-restore that copies `aidlc/scopes/<name>.md` into the harness scopes directory and then loses the record. So this is narrow, and I would not raise it at all except that the throw sits at `aidlc-graph.ts:3273`, after `stage-graph.json` and `scope-grid.json` are already written at `:3271-3272` — so every subsequent `graph compile` fails the same way, and the message is all the user gets. Every other error in this feature names its file, `t340` pins that for parse failures, and the guide states it as the contract in "When a record is unreadable". The constraint is just that this message name the offending file, which the caller already holds.

One sentence in the same area: the missing-region error offers "or by restoring the region by hand" while printing `COMPOSED_GRID_BEGIN.slice(0, 42)` (`:502`), a sentinel truncated mid-word. The record will not parse unless the sentinel is byte-exact, so hand-restoring is the one route that message does not actually enable. The delete-and-recompile route it names first is fine, and the guide carries the full sentinel.

### P3 — two header comments describe the format this PR replaced

`tests/unit/t340-composed-scope-record.test.ts:14-15` says a record is "exactly the harness identity `.md` plus one appended `## Stage Grid` section", and `tests/integration/t341-composed-scope-durability.test.ts:39` refers to "the record's `## Stage Grid` section". Both are in the two files that document the replacement. The other occurrences at this head read as deliberate — the collision fixture, the history note at `aidlc-graph.ts:456`, and the guide's example of a heading a user may write.

## Your two questions

**Numbering.** There is a written rule and it points the other way from what I would have guessed, so here are both halves and the call is yours. `tests/README.md:74` says that when two changes pick the same number, "the second to merge rebases and renumbers to the next free integer in the same level" — which is this case for the new **unit** `t340`, since #1151 merged first. The integration `t341` has no same-level collision; `t341-orchestrate-wait.test.ts` is in the unit tier.

Against that: nothing enforces it. Nothing keys on the numeric prefix — `tests/gen-coverage-registry.ts` keys entries by path — and the same-level duplicates are already everywhere on `main`, with three unit files each for `t328`, `t317`, `t314`, `t304` and `t248`, and three integration files for `t327`. So renumbering the one unit file follows the documented rule and costs a rename; leaving it matches what the tier already looks like.

**The composer's direct write.** Keep it out of this PR. It is prompt prose across seven `SKILL.md` files plus the composer agent, its knowledge doc and the `detect --json` write target: a different surface with a different failure mode, against a change that is currently green and reviewed. The one thing I would ask is that the remaining window get its own tracked item before this merges. `Closes #963` will close the issue whose first consequence is that exact silent loss, so without a separate item the window leaves with the issue — and the new durability row detects the state but does not close it.

## Where this stands

Nothing here is blocking. Every item from my earlier review is resolved, the implementation reads clean, and the four P3s are message, comment and documentation text — the record-as-source-of-record decision and the sentinel are both right.

Nothing is needed from you meanwhile; the P3s are yours to take or leave.

On the ruleset, since it bears on what happens next: `required_approving_review_count` is `0` and `required_reviewers` is empty at this head, so there is no team approval to wait for. I am not claiming what that leaves once the change request clears — `require_extra_approval_for_unattributed_changes` is enabled and I cannot tell from outside whether it is another gate on this branch, so treat that as unmeasured. I am not enabling auto-merge.

DrakeRHEE:
Thank you for re-running the mutation arm against `f84b0063` rather than against `main` — you are right that a base-branch arm would have measured nothing there, and your totals are more honest than the list I gave: I named the five cases that motivated the fix, not the 17 that move.

I took all four P3s. Three of them are text, one turned out to be a real dead end.

**The phantom row's remedy (P3-2).** This one I would have wanted to fix even if you had marked it P1, because the row was telling users to run something that cannot work. Your reproduction is the shape I confirmed: an untagged scope file gets no column, compile exits 0, and the row stays red with `graph compile` as its only advice — while step 4 of the hand-authoring workflow in `04-scopes.md` is running doctor. The two causes are now counted apart, using the two signals you pointed at as already in hand. A missing column with a record behind it still names compile, because compile does fix that one. Without a record the row says compile cannot reach it and names the three routes that do: restore the record, finish tagging the stages, or delete the scope file. `t341` pins it by compiling a second time and asserting the row does not move — a test that would have failed the old message.

**The unnamed error (P3-3).** `renderComposedScopeRecord` now takes the path from its caller. Worth reporting one thing here: my first test for it called the writer directly and **passed with the fix reverted** — because the whole point of the item is that the *shipped* path names the file, and calling the writer with a path proves nothing about the wiring. So the case now drives it through the compile CLI, with a record hand-copied into the harness scopes tree — the way in you described. That version does fail when the caller's argument is dropped. Mutation-checked in both directions.

The truncated sentinel: both sentinels now print in full. You are right that hand-restoring was the one route that message did not enable, and printing 42 characters of a byte-exact string is worse than not offering the route at all.

**The trailing content and the docstring (P3-1).** The behaviour stays as it is — carrying the area below the region would be a format change, not a text fix, and you traced correctly that nothing on disk is lost. What was wrong was the sentence. The docstring now says the identity is what sits *above* the region and that content below the END sentinel stays in the record without reaching the projection, and it scopes byte-stability to records the writer produced, since a hand-compacted fence or extra blank lines before the region re-render differently. The guide's "record's shape" section says to treat the region as the end of the file and names which side is carried. `t344` pins the asymmetry, which also fills the coverage hole you found — every other identity assertion compares with `trimEnd()` and would not have noticed.

**Stale header comments (P3-4).** Both replaced. You were right that the other occurrences are deliberate: the collision fixture, the history note, and the guide's example of a heading a user may legitimately write all need to say `## Stage Grid`.

**Numbering.** I followed `tests/README.md`. The unit file is now `t344-composed-scope-record.test.ts`; the integration file keeps `t341`, since its same-level neighbour is in the unit tier and renumbering it would only gap-fill. Thank you for finding the written rule and then arguing the other side of it — I would have left the collision alone on the strength of `t314`.

**The composer's direct write.** Filed as its own item before this merges, as you asked: #1202. Your reasoning is the part I had missed — `Closes #963` would take the description of the remaining window with it, and a check that detects a state is not the same as closing it.

Re-verified on this head: the targeted set at 337 tests / 4,156 assertions including `t55-test-suite-drift` for the rename, `bun scripts/package.ts --check` byte-identical across all seven harnesses, `typecheck`, and `biome check --error-on-warnings` (the same three pre-existing infos, all in files this branch does not touch).

wowzoo (approved):
Approving. All four items from my last review are closed, and the four new cases fail without the fixes.

Base `9288f80f4c289891e362864c0f3055309648927b`, head `1abe456f93556331938a9f7af9cec40ac3188be8`. Reviewed the delta `2f4e27b8..1abe456f` — `core/tools/aidlc-graph.ts`, `core/tools/aidlc-utility.ts`, three doc lines, `t341`, and the renamed `t344`.

## What closed each item

**The phantom row's remedy now reaches its cause.** A missing column is counted in two arms (`core/tools/aidlc-utility.ts:4950-4956`), and the recordless arm no longer offers `graph compile` as the remedy on its own: it says compile cannot rebuild a column with no record behind it, and asks for the record restored or the stage tagging finished *first*, with compile after — or the scope file deleted. `docs/guide/12-cli-commands.md` and `docs/guide/15-troubleshooting.md` carry the same split, so the table and the prose no longer promise a remedy the row cannot deliver.

**The sentinel refusal names its file.** `renderComposedScopeRecord` takes `identityPath` (`core/tools/aidlc-graph.ts:579-582`) and the only production caller passes it (`:740-743`). `grep -rn 'renderComposedScopeRecord(' core/ tests/` finds the definition, that one production call site, and test calls — and the tests exercise both the named and the unnamed form (`tests/unit/t344-composed-scope-record.test.ts:289-294`), so the optional parameter is not hiding an unfixed path.

**The missing-region error is hand-restorable again.** Both sentinels print in full (`core/tools/aidlc-graph.ts:508-517`), so the route that message names is now reachable.

**The docstring matches the behaviour, and the guide says who owns which half.** `core/tools/aidlc-graph.ts:449-462` now scopes byte-stability to records the writer produced and states that content below the END sentinel stays in the record without reaching the projection; `docs/guide/05-scopes-and-depth.md:296` says to treat the region as the end of the file and why. The coverage gap I named is closed in the part that was missing: `t344` now presents a record with content below the END sentinel and pins the asymmetry — the grid and identity are unaffected, and neither the projection nor a re-render carries the appended prose. That it also stays on disk rests on the write guard rather than on that assertion, which the case says itself.

**The two stale header comments are corrected**, in `t344` and in `t341`. Five `## Stage Grid` mentions survive and each is deliberate: the history note at `core/tools/aidlc-graph.ts:466`, the collision fixture at `t344:210`, `:225`, `:256`, and the guide's example of a heading a user may write at `docs/guide/05-scopes-and-depth.md:307`.

**The renumber lands on the next free integer in its level.** `tests/unit/` holds `t340-default-scope-resolver` through `t343-intent-create-positionals`, so `t344` is next, which is what `tests/README.md:70-74` asks for.

## Control

Both arms in the same mode, single-file runs, packager run before each, with `core/tools/aidlc-graph.ts` and `core/tools/aidlc-utility.ts` reverted to `2f4e27b8` and the new tests kept:

| file | sources reverted | sources restored |
|---|---|---|
| `t341-composed-scope-durability` | **6 pass / 2 fail** | 8 pass / 0 fail |
| `t344-composed-scope-record` | **31 pass / 2 fail** | 33 pass / 0 fail |

The four failures are exactly the four cases the delta adds for these fixes: the two doctor-arm cases in `t341`, and in `t344` the sentinel refusal naming its file and the missing-region message printing both sentinels byte-exact.

Also on this head: `bun run check` (packager, `--check`, three tsc projects, biome) clean, and `bun tests/gen-coverage-registry.ts --check` reports the committed registry fresh with the ratchet held. Measured with bun 1.4.0 on macOS; CI is the full-suite adjudication and it is green here — 22 checks pass, `deploy` skipping.

## One P3, non-blocking, nothing needed before merge

`loadComposedScopeRecords` (`core/tools/aidlc-graph.ts:602-609`) turns **every** `readdirSync` failure into `{}`, while its docstring describes only the absent-directory case. The file-level read a few lines below is unguarded, so an unreadable *file* throws with its path — but an unreadable *directory* is silently indistinguishable from "nothing has been composed yet". That now feeds the new doctor arm: with records present but the directory unreadable, every composed scope classifies as recordless, and the row asks the user to restore records that are already there. The invariant worth holding is that a listing failure stays distinguishable from an absent directory, so the row's remedy cannot invert. Narrow, and no worse than what it replaces.

## What this unblocks

The pull request reads `BLOCKED` only because of my earlier change request on the long-superseded head `7d2e5166`; `dismiss_stale_reviews_on_push` is false, so it stayed on the record. There are no review threads and CI is green, so this approval clears the last gate. Thank you for the merge from `main` as well — that removed the conflict without a force-push.

I am not enabling auto-merge.

PR #1157, opened 2026-09-13
refactor(kiro): serve Kiro IDE and Kiro CLI from one distribution

wowzoo:
Replaces #1063, closed unintentionally: a brief visibility change on the head fork detached it
from this repository's fork network, which closed the pull request and made reopening impossible.
Nothing about the change itself was reconsidered, and the head commit is the one #1063 carried
(711e4ded7783715f8866ea18919c5bbf37ea02bf).

---

## What this does

`kiro` and `kiro-ide` maintain the same `.kiro/` shell twice. Kiro IDE and Kiro CLI run the same agent runtime, so this consolidates them into **one row, `kiro`, serving both surfaces**, and removes `kiro-ide`.

**Breaking:** `aidlc config --harness kiro-ide` no longer resolves — use `--harness kiro`.

## Why now

The two adapters implement the same contract twice and have never converged:

| | `harness/kiro` | `harness/kiro-ide` | gap |
|---|---|---|---|
| v2.1.7 | 415 | 275 | −140 |
| v2.3.0 | 424 | 374 | −50 |
| v2.6.18 | 935 | 743 | −192 |
| v2.7.0 | 1047 | 2029 | +982 |
| this base | 1075 | 2029 | +954 |

690 → 3,104 lines. The point is not that the gap grows monotonically — it narrowed at 2.3.0 and flipped sign at 2.7.0 — but that **the same contract is implemented twice and pays maintenance twice every generation.** Five call sites already normalized `kiro-ide` to `kiro` before this change (`aidlc-model-policy.ts`, `aidlc-plugin.ts`, `aidlc-plugin-emit.ts`, `scripts/plugin-hooks-template/compose.ts`).

The merged adapter is 3,010 lines — the union, not a sum, because the duplicated halves collapse.

## What the merge had to solve

**The engine.** `settings/cli.json` gains `"chat.agentEngine": "v3"`, so a plain `kiro-cli` in the project reaches the agent runtime this row targets with no flag at the call site. Measured both ways on kiro-cli 2.21.1: with the key the session enters that engine, without it it does not.

**Hook registration.** The row registers through standalone manifests (14 files, 16 registrations) that both surfaces read. Two triggers are surface-specific per `docs/features/hooks.md` — Session Start is IDE-only, Agent Spawn is CLI-only — so lifecycle responsibilities register on **both** or they are dead on one surface. Only Prompt Submit, Pre Tool Use and Pre Task Execution can refuse, so every guard that must refuse sits on `PreToolUse`.

[...]

PR #1161, opened 2026-09-13
fix(doctor): report Kiro IDE ignore sources that hide .kiro/

### LINKED ISSUE #1146: [Bug]: Kiro IDE fs_read guard denies .kiro/ reads when global gitignore excludes it — repo !.kiro/ negation not honored, blocks all stages; --doctor misses it
## Description

On the **Kiro IDE** harness, the agent's `fs_read` permission guard denies reads of any path matching a pattern in the **global** git excludes file (`~/.config/git/ignore`). When that global file contains `.kiro/` (a common personal-harness setup — mine is installed by a separate `~/.agents` config), every read under `.kiro/` is denied.

The framework already ships the intended mitigation in the project `.gitignore` (`!.kiro/`, inside the `# AI-DLC:gitignore` block's project-scoped override), but Kiro's guard does **not** honor that repo-level negation, nor does it consult git's actual ignore resolution. `git check-ignore` correctly reports the path as **not** ignored, yet the guard still denies it and cites the global file as the source.

Because every stage body, agent persona, protocol module, and knowledge file lives under `.kiro/`, this blocks the workflow completely — no stage can run.

Separately, `/aidlc --doctor` does not catch this: it reports all project checks green because it never verifies that the active harness can actually **read** a known `.kiro/` file.

## Steps to Reproduce

1. On macOS with Kiro IDE, have a global git excludes file that ignores `.kiro/`:
   ```
   # ~/.config/git/ignore
   .kiro/
   ```
   (Confirm it is the active excludesfile; here it is git's default `~/.config/git/ignore`.)
2. Open an AI-DLC project whose committed `.gitignore` carries the shipped project-scoped override `!.kiro/`.
3. Start a workflow (`/aidlc <scope>`) so the orchestrator emits the first `run-stage` directive (e.g. Intent Capture).
4. Observe the agent attempt to read a stage file, e.g. `.kiro/aidlc-common/stages/ideation/intent-capture.md`.

## Expected Behavior

The read succeeds. The project's `!.kiro/` negation (and/or git's own ignore resolution, which reports the path as not-ignored) should un-block reads of tracked `.kiro/` files, so stages can run. The guard should not deny a path that `git check-ignore` reports as not-ignored.

Additionally, `/aidlc --doctor` should detect and report the condition (a `.kiro/` file is not readable by the active harness) with a concrete remedy, rather than passing all checks.

## Actual Behavior

The read is denied:

```
Tool call denied by user's permissions.
Rule: deny fs_read matching ".kiro/"
Source: /Users/omaraws/.config/git/ignore.
```

Meanwhile git itself un-ignores the path (exit `1` = not ignored):

```
$ git check-ignore -v .kiro/aidlc-common/stages/ideation/intent-capture.md
$ echo $?
1
```

So the guard's gitignore evaluation diverges from git's: it matches the global `.kiro/` pattern and ignores both the repo-level `!.kiro/` negation and git's resolved state. Every subsequent stage read fails the same way, and the workflow cannot proceed past the first `run-stage` directive.

`/aidlc --doctor` reported `0 problems, 2 warnings` (36 project checks passed) and did not surface the read block.

## AI-DLC Version

v2 (current on main)

## Release / Commit

Installed runtime `aidlc 2.8.2`; repo clone at `v2.8.1-11-g0865300b` (`0865300b`).

## AI-DLC Phase

Not phase-specific (blocks all stages; first observed entering Ideation / Intent Capture).

## Harness

Kiro IDE

## AI Model

Claude Opus 4.8

## Environment

- OS: macOS 26.6.2
- Harness: Kiro IDE
- Global excludesfile: `~/.config/git/ignore` containing `.kiro/`
- Project `.gitignore`: ships the AI-DLC block plus the project-scoped `!.kiro/` override
- git: repo initialized; `git check-ignore` confirms `.kiro/` and nested files are not ignored

## Additional Context

Root cause lives in Kiro IDE's `fs_read` guard (a naive gitignore matcher that reads the global excludes file, does not reconcile cross-file negations, and does not consult git's index/resolution) — which AI-DLC cannot patch directly. But the impact is AI-DLC-specific and total, so two framework-side improvements would prevent a silent, total block:

1. **Doctor readability probe** — add a check that attempts to read a known-tracked `.kiro/` file under the active harness and fails loudly with a remedy when it is blocked. This is the fast, deterministic signal that was missing here.
2. **Known-issue doc + remedy** — document that a global `.kiro/` ignore breaks Kiro IDE despite the shipped `!.kiro/` negation, with the workaround: add an `fs_read` allow for the workspace `.kiro/**` in `~/.kiro/settings/permissions.yaml` (an explicit allow overrides the gitignore-derived deny). Adjusting the global excludes file is broader and re-introduces the personal-config leak that file exists to prevent, so it is not the recommended fix.

Workaround confirmed to be needed: `git init` + `!.kiro/` makes git un-ignore `.kiro/`, but the Kiro guard still denies, so the `permissions.yaml` allow (or removing `.kiro/` from the global excludes) is required to proceed.

Adjacent but distinct existing issues (different guard — plan-approval mutation classification, not gitignore-based denial): #1039, #965.

apackeer:
# Summary

Fixes #1146.

On Kiro IDE, a `.kiro/` rule in git's global excludes file makes the IDE's `fs_read` guard deny every stage, agent, and protocol read, so no stage can run. The IDE compiles each ignore file into its own matcher, so a `!.kiro/` in the project `.gitignore` cannot undo it even though `git check-ignore` reports the path as not ignored; and Kiro applies deny-overrides across scopes, so a `permissions.yaml` `fs_read` allow cannot clear it either. `/aidlc --doctor` reported `0 problems` because its reads run in a subprocess, not through `fs_read`.

The guard itself is a Kiro IDE defect and is not patchable here; this change gives `--doctor` a deterministic detection with a remedy, plus the troubleshooting entry.

## Changes

- `core/tools/aidlc-utility.ts`: new exported `kiroIdeIgnoreSourceChecks(projectDir, harness, env)`, wired into `collectDoctorReport` for the Kiro harness when the IDE conductor (`agents/aidlc.md`) is present. It evaluates each ignore source Kiro IDE honours **on its own** - git's global excludes file (only inside a git repo, as Kiro does), `~/.kiro/settings/kiroignore`, the project `.gitignore`, and `.kiroignore` - by running `git -c core.excludesFile=<source> check-ignore -v --no-index` against a stage path in a scratch repository. Only that one file applies, and the gitignore semantics are git's own rather than a hand-rolled matcher. Verbose `check-ignore` exits 0 for a directly matched negation, so a leading `!` in the reported pattern counts as not hidden.
- Global-source matches **fail**, naming `file:line` and the pattern, with the remedy: remove or narrow the rule; a negation in another file or a permissions allow does not help; per-repo personal ignores belong in that repo's `.git/info/exclude`. Workspace-source matches **warn**, because whether those files apply is governed by the IDE's `kiroAgent.agentIgnoreFiles` setting, which doctor cannot read. Missing git degrades to a passing advisory row.
- Docs: doctor check table (`docs/guide/12-cli-commands.md`), a new **Kiro IDE Read Denials** troubleshooting section with the exact denial text and a quick-fix row (`docs/guide/15-troubleshooting.md`), doctor summary in `docs/reference/06-hooks-and-tools.md`.
- `tests/unit/t340-kiro-ide-ignore-sources-doctor.test.ts`: global rule detected and named despite a repo `!.kiro/`; same-file negation clears it (directory and direct-file forms); global-fail/workspace-warn split; row emitted only for the IDE conductor surface.

## User experience

Before: on the reporter's setup doctor reports `0 problems, 2 warnings` while every stage read is denied. After, against the packaged `dist/kiro-ide` with a repo `!.kiro/` and a global `.kiro/` (git's own `check-ignore` exits 1 = not ignored):

```
  fail  Kiro IDE ignore sources: ~/.config/git/ignore:1 ".kiro/" hides .kiro/ - the IDE's fs_read guard denies every stage, agent, and protocol read
        fix: remove or narrow the ".kiro/" rule in ~/.config/git/ignore; Kiro IDE evaluates each ignore file on its own, so a "!.kiro/" in another file and a permissions.yaml fs_read allow do not override it (Kiro applies deny-overrides across scopes); keep per-repo ignores in that repo's .git/info/exclude, which git honours and Kiro does not list as an ignore source; then re-run `bun .kiro/tools/aidlc.ts doctor`
1 problem, 4 warnings.
```

After removing the rule: `ok    Kiro IDE ignore sources: none hide .kiro/ (1 file(s) checked)`.

No release metadata changes (version, badge, changelog) per policy.

## Checklist

* [x] Changes are documented
* [x] If this change adds an input to any fingerprint, epoch, or receipt identity, the description names the human-visible change it detects (none added)

## Test plan

- `bun test tests/unit/t340-kiro-ide-ignore-sources-doctor.test.ts` - 5 pass
- `bun test tests/smoke/t148-kiro-file-structure.test.ts` - 20 pass (existing Kiro doctor shape test)
- `bun scripts/package.ts --check` - deterministic across all 7 harnesses
- Manual reproduction above against `dist/kiro-ide` with an isolated `HOME`/`XDG_CONFIG_HOME`/`GIT_CONFIG_GLOBAL`

Not verified: whether Kiro IDE reads `.git/info/exclude`. Kiro documents its sources as named workspace files plus the two global files, so the remedy text treats it as invisible to the IDE.


wowzoo:
https://github.com/awslabs/aidlc-workflows/pull/1161

Thank you for picking this up — it is the detection half of #1146, and evaluating each ignore source on its own is the right shape for it.

I am working on #1157 at the moment, which serves Kiro IDE and Kiro CLI from one distribution. It rewrites `core/tools/aidlc-utility.ts` where this check lands, and it removes the agent-form distinction this check gates on, so merging this onto it conflicts in that file and the gate itself needs rethinking rather than re-applying. I would rather review this on top of #1157 than against a tree that is about to change underneath it.

So: once #1157 is merged and this is rebased on it, I will review this properly.

PR #1163, opened 2026-09-13
fix: refuse a review whose findings table cannot be read, instead of recording no findings

wowzoo:
A reviewer that writes its `### Findings` table under different column names has its findings
silently dropped, while its `NOT-READY` verdict is recorded as final. The gate then renders
`No findings`, and `--reject-finding` has nothing to select.

I hit this on a live run. The review record carried `"verdict": "NOT-READY"` and `"findings": []`,
while its `body` held a five-row findings table whose first row was `Critical`. The table's header
was `| ID | Severity | Location | Finding | Evidence | Recommendation |` — every finding named, none
of them recorded.

### Why the current code allows it

`parseReviewSection` addresses each cell by column name, so it checks the header first, and returns
an empty findings list when a required name is absent (`core/tools/aidlc-lib.ts:11077-11079`). That
return is silent: the verdict survives, the rows do not.

Nothing downstream notices.

- `validateReviewAppendix` checks encoding, ownership fields, verdict, iteration and heading
  boundaries, and never looks at the findings table (`core/tools/aidlc-lib.ts:10442-10534`).
- `aidlc-log review --verdict` catches only a thrown parse error; an empty list proceeds into the
  record and `REVIEW_COMPLETED` (`core/tools/aidlc-log.ts:2413-2423`, `:2451-2467`).
- The record validator requires `findings` to be an array of well-formed entries. Empty is valid for
  either verdict, and nothing relates it to `body` (`core/tools/aidlc-lib.ts:11234-11268`).
- Record/completion matching compares stage, unit, iteration, reviewer, verdict, request id and
  fingerprints — not findings (`core/tools/aidlc-lib.ts:11434-11449`).
- The gate reads `record.findings` and never reparses `record.body`
  (`core/tools/aidlc-review-brief.ts:119-140`), so a NOT-READY review with zero findings renders the
  fallback row `| - | - | - | No findings | No action required | Resolved |`
  (`core/tools/aidlc-review-brief.ts:404-425`). `--reject-finding` searches the structured findings
  only, so those rows cannot be dispositioned (`core/tools/aidlc-review-brief.ts:359-383`).

The reviewer protocol already classifies this as an incomplete attempt: "a malformed findings table"
is listed among the forms that are "not a verdict", and it says the logger refuses them
(`core/aidlc-common/protocols/stage-protocol-reviewer.md:150`). Both reviewer knowledge files
already mandate the exact header (`core/knowledge/aidlc-architecture-reviewer-agent/reviewing.md:76`,
`core/knowledge/aidlc-product-lead-agent/reviewing.md:60`). The code is what does not hold that line.

### What this changes

1. `parseReviewSection` throws when the findings table's header is missing a required column, naming
   the declared header, the expected columns, and what is missing. A malformed *row* already throws;
   a malformed *header* now does too.
2. `aidlc-log review --verdict` refuses a `NOT-READY` completion whose findings TABLE yields no
   rows — the shape the gate would otherwise render "No findings" over. `parseReviewSection` reports
   `tablePresent` so the caller can tell that from a body that carried prose instead of a table; the
   latter is left alone, because the reviewer protocol already classifies it as an incomplete review
   and bodies written before the table contract would otherwise stop recording. The bounded
   retried-incomplete fallback (`core/tools/aidlc-log.ts:2306-2310`), which deliberately records
   `NOT-READY` with an empty body and no findings, is untouched.

Both refusals happen before the record is written and before `REVIEW_COMPLETED` is emitted, and the
reviewer's draft is only deleted after a successful completion, so the review text survives. The
request stays unmatched, which is the state the protocol already handles with one `--retry-pending`
redispatch that consumes no review iteration.

I did not add header synonyms or a positional fallback. The header I observed has no `Status`
column, and `Status` drives disposition and subsequent-review semantics; `Evidence` has no field in
the record schema. Mapping those columns would mean inventing data rather than tolerating
formatting.

Two limits are deliberate. Records already written with `findings: []` are not repaired — reads go
through the record rather than its body, and the records are digest-bound, so they are left alone. And
a `NOT-READY` body that carries no findings table at all still records: the reviewer protocol covers
that shape as an incomplete review, and refusing it here would also reject bodies written before the
table contract.

### Tests

`tests/unit/t304-review-brief.test.ts`:

- the parser refuses the measured header and names expected, received and missing columns;
- review completion refuses a `NOT-READY` review for that header, and for a canonical header with no
  rows: nonzero exit, no `REVIEW_COMPLETED` in the audit, and the draft preserved.

All three fail at the base and pass with the change.

No test fixture is modified. An earlier revision of this change refused every `NOT-READY`
completion that recorded no finding, which reached past this defect: thirteen suites write
`### Findings` followed by prose rather than a table, and one of them
(`t115` — "a NOT-READY verdict still satisfies the precondition") failed on CI because of it. The
refusal now requires a table that yields no rows, and `t304` pins the accepted shape as well as the
refused ones.


apackeer (changes requested):
Thanks for tracking this down from a live run and for the careful write-up. The write-side refusal is the right seam, the scope limits are stated and reasoned, and the tests failing at base make the regression concrete.

I reviewed the current head 1da8137a5 against a0ee4415e (merge base; `main` has since moved to 0e54914f3 with no overlapping changes to `parseReviewSection`, `handleReview`, `parseReviewArtifact`, or t304). Alignment basis: internal bug fix restoring the contract already stated in `stage-protocol-reviewer.md:140`.

One blocking item, introduced by this change:

**P2 — the legacy embedded-review reader now throws uncaught.** `parseReviewSection` has a second caller that the PR body does not account for: `parseReviewArtifact` (`core/tools/aidlc-review-brief.ts:74`), reached from `readReviewArtifactContexts:148` for a scope that has no review record and whose artifact carries an embedded `## Review`. Before this change a non-canonical header there parsed as `findings: []`; now the throw at `core/tools/aidlc-lib.ts:11090` escapes, so `aidlc-review-brief review` / `context` exit 1 with no stdout, and `aidlc-state approve` exits 1 through `acceptedRiskDispositionField` (`core/tools/aidlc-state.ts:5658`, `:5512`); `--reject-finding` reaches it at `aidlc-review-brief.ts:362`. Only `aidlc-jump.ts:85` catches. Both reviewers reproduced this with a legacy fixture. It contradicts the protocol's migration promise (L173-175: legacy embedded reviews stay readable at the brief and redispatch context) and the PR body's "reads go through the record" claim, which holds for record-era scopes only. A plain `reject` still succeeds, so the state is recoverable, but nothing tells the human that; the error text says to fix columns in a receipt-frozen artifact.

Suggested correction: keep the strict parse for admission, and handle the legacy read path explicitly — either catch in the legacy loop of `readReviewArtifactContexts` and surface an Unresolved fallback finding naming the unreadable table (the incomplete-review path already does this), or fail closed at the brief/approve seam with a message that names the recovery (Request Changes, then a record-era re-review). Either way, a t304 case with a persisted legacy artifact would pin it.

Two non-blocking notes on pre-existing behavior the diff does not change; both fall under the "no findings table" limit you scoped out, so they are candidates for a follow-up rather than this PR:

- The heading scan (`aidlc-lib.ts:11069-11082`) only recognizes `### Findings` exactly. A canonical table under `### Findings (1)` or `### Findings:` reads as `tablePresent: false` and a NOT-READY review still records `findings: []`. Same class as the prose case; same off-template reviewer; noting it so the follow-up scope is visible.
- `aidlc-lib.ts:11098` skips `table[1]` as the separator without checking it. A canonical header followed directly by two finding rows records only the second and deletes the draft (reproduced). Validating the separator before slicing would make "no finding is silently dropped" hold end to end.

Validation: `bun test tests/unit/t304-review-brief.test.ts` 36/36 at head; with the base `aidlc-lib.ts` / `aidlc-log.ts` swapped in and dist repackaged, exactly the three new tests fail. CI is green on this head. Requesting changes for the legacy-reader item only.

Issue #1116, opened 2026-09-10
[Bug]: an unrelated session in a fresh clone is blocked and logged into a teammate's intent — hooks read the lone-intent fallback as participation


kubode:
# AI-DLC Version

Observed on 2.7.1 (`a277af218f0df7f325d3b8be7b6d90fce2c5bd40`), not re-run on 2.8.2. The relevant logic is present on `main` (`aadc8323`): `plan-approval-guard.ts` still resolves through bare `stateFilePath(projectDir)`, `record-human-turn.ts` still gates on `existsSync(stateFilePath(projectDir))`, and `session-start.ts` still calls `writeSessionBinding(projectDir, sessionId, selection.space, selection.intent)`. Three of those files have changed since my pin — `plan-approval-guard.ts` by about a third — so I would rather re-run it on 2.8.2 than infer it.

# AI-DLC Phase

Not phase-specific — the hooks fire in sessions running no workflow at all. The blocking symptom additionally requires the *teammate's* workflow to be in `code-generation`.

# Scope

Narrow, and worth stating up front: `activeIntent()` falls back only at `records.length === 1` (`aidlc-lib.ts:1648-1652`), counting directories that hold an `aidlc-state.md`. So this needs a space holding exactly **one** committed intent record, in a clone that has never run `/aidlc`. Measured with two records and no cursor: every hook exits 0, no shard is written, and the binding records `intent: null` — nothing fires at all. The window is a team's *first* shared intent in a given space, and it closes once a second one exists there.

That is also what makes it worth fixing: it fires during adoption, on people who never opted into the framework.

# Affected surface

Harness-neutral by inspection — the resolution and hook logic are shared — though I have only tested Claude Code (macOS 26.6 / Darwin 25.6.0, bun, hooks driven directly).

---

## Description

The `active-intent` cursor is gitignored (per-user) while the record and `intents.json` are committed by design, so a teammate's in-flight record arrives in a fresh clone with no cursor beside it. Resolution falls through to that record, and hooks then act on it as though the session were a participant in the workflow.

Bare `stateFilePath(projectDir)` — resolution with no session identity — appears at seven call sites across six hooks:

| call site | reachable from an ordinary session? |
|---|---|
| `record-human-turn.ts:104` | yes |
| `plan-approval-guard.ts:676`, `:780` | yes |
| `validate-state.ts:28` | yes, on PreCompact |
| `run-sensors.ts:111` | yes, once a shard exists |
[...]

PR #1171, opened 2026-09-14
fix: read the lone-record fallback as resolution, not as workflow participation

wowzoo:
A session in a fresh clone that has never run the framework is blocked by, and logged into, a
teammate's intent.

The `active-intent` cursor is per-user and gitignored while the intent record and `intents.json` are
committed by design, so a teammate's in-flight record arrives in a clone with no cursor beside it.
`activeIntent` falls through to that record (`core/tools/aidlc-lib.ts:2003`), and the hooks then act
on it as though this session were a participant. Driving the reporter's recipe at `a0ee4415` (the
base this branch was written on, before the rebase; the three files involved are byte-identical at
`572c8265`): the
plan-approval guard refuses an unrelated `Write` with exit 2, and `SESSION_STARTED`,
`PLAN_APPROVAL_BLOCKED`, `HUMAN_TURN`, `SUBAGENT_COMPLETED` and `SESSION_ENDED` all land in the
teammate's audit shard.

It reaches approval authority, not only the ledger. `humanActedSinceGate`
(`core/tools/aidlc-lib.ts:7373`) collects `HUMAN_TURN` across every shard in the record —
`auditShards(projectDir)` at `:7379`, with no narrowing — so a turn minted by a clone that never
joined the workflow can satisfy the human-presence check at the teammate's gate.

The window is narrow and that is what makes it worth fixing: the fallback applies only at
`records.length === 1`, so it needs a space holding exactly one committed record in a clone with no
cursor. It is a team's first shared intent, and it closes once a second record exists — which means
it fires during adoption, on people who have not opted into the framework.

### The fallback stays; what changes is what follows from it

The fallback is load-bearing. A git worktree does not inherit the cursor, and
`tests/e2e/t07-audit-fork-merge.test.ts` and `tests/integration/t49-bolt-sensor-failures.test.ts`
both depend on resolving the committed record that way. So resolution is unchanged, and
`activeIntent`, `resolveWorkflowSelection` and the path helpers keep their behaviour.

`isWorkflowParticipant(projectDir, selection)` answers the separate question, from evidence only a
checkout that actually joined can hold — local, machine-specific artefacts rather than committed
ones:

1. an existing session binding for that target;
2. the active-intent cursor read alone — `cursorIntent`, split out of `activeIntent` so the fallback
   underneath it cannot answer for the cursor;
3. the record named by this worktree's own `.aidlc/worktree-meta.json` `intentRecord`, written by
   worktree creation at `core/tools/aidlc-worktree.ts:518`, so the worktree case needs a reader
   rather than a new marker;
4. the Unit participant marker, which the orchestrator already consults as participation evidence.

The first three name the exact `{space, intent}`. The fourth does not — the Unit marker is
checkout-wide, so it admits the resolved record whenever this checkout claimed any Unit; it is
consulted last for that reason. A committed record on its own satisfies none of them, and no field is
added to `WorkflowSelection`.

### Applied at the write sites, not only where paths resolve

`aidlc-log-subagent.ts:92` appended with intent and space omitted, so it re-resolved through the same
fallback; in the reporter's data that path produced most of the entries. A change limited to the bare
`stateFilePath(projectDir)` sites would have missed it.

- `aidlc-session-start.ts`: a non-participant still records its session identity, but the binding it
  writes carries `intent: null`, so the hook takes the existing no-workflow path — no intent stamp,
  heartbeat, workflow context or `SESSION_STARTED` follows. A cold session behaves as before, and a
  worktree whose metadata names the record proceeds normally.
- `aidlc-record-human-turn.ts`: participation is checked before both the ledger mint and the turn
  marker — the marker resolves through the same fallback — and the resolved target is handed to the
  append instead of being re-resolved.
- `aidlc-log-subagent.ts`, `aidlc-session-end.ts`: resolve, then decline to write. The
  session-local in-flight ledger still settles.
- `aidlc-plan-approval-guard.ts`: a non-participant has no workflow from this hook's perspective and
  fails open, exactly as a pre-workflow session does. The `GUARD_DISABLED` row follows the same rule.
- Kiro and Kiro IDE presence floors: participation joins the existing carve-outs (autonomous
  Construction, the deterministic off-switch, no open gate), so a non-participant is permitted rather
  than blocked for lacking a human turn.
- Not covered here: `aidlc-validate-state.ts` still writes its heartbeat and recovery breadcrumb
  against the resolved record on PreCompact, and appends `SESSION_COMPACTED` when a shard exists. It
  is outside the five surfaces the report's reproduction exercises, so I left it for a separate look
  rather than widening this change.
- `humanActedSinceGate`: requires participation, then reads the shards of the resolved record rather
  than re-resolving. Cross-shard presence is unchanged for participants; `t261`'s strictly-later
  cross-shard allow still holds. Narrowing presence to one shard or one session would reverse that
  encoded contract, so it is not part of this change.

**What this does not close.** Entries already written are not rewritten — the audit log is
append-only and digest-bound. That matters beyond tidiness for one clause of the report: a
`HUMAN_TURN` already recorded by a clone that never joined stays eligible under the existing
cross-shard presence policy, because the new check asks whether the checkout doing the ASKING
participates and then reads the record's shards as before. So this change stops such a turn from
being minted; it does not retire one that is already on disk. Closing that would mean a participant
no longer trusting a sibling shard, which reverses the contract `t261` encodes, and I have left that
decision to you rather than folding a policy change into a defect fix.

### Tests

`tests/integration/t342-workflow-participation.test.ts` drives the reported lifecycle in a clone that
only received the record — five hooks exit 0, no audit directory is created, and the binding records
`intent: null` — then proves a foreign `HUMAN_TURN` does not answer the gate, checks each of the four
signals independently, rejects worktree metadata naming another record or malformed, and keeps the
existing two-record no-op.

`t160` covers `cursorIntent` and the resolution-versus-participation split; `t311` covers the binding
a lone committed record must not produce.

I ran the fallback and presence suites against a second worktree at the unmodified base to separate
new failures from existing ones. The base fails 52 of 348 in that set (`t188` 13, `t261` 39) and this
branch fails the same 52; `t342`'s 8 cases pass. Two further failures appeared only when `t311` and
`t326` ran beside `t342`, which spawns five hook subprocesses per case; each passes alone (11/0 and
27/0), and `t311` documents its own wall-clock budget as the reason it is `.serial.`.

`typecheck`, `biome check` on the changed files, and `package.ts --check` (deterministic across two
builds, seven harnesses) all pass.

Reported-by: @kubode

Refs #1116


rafaelgsr (changes requested):
Thanks for tackling the important distinction between workflow resolution and participation.

I reviewed the current head `cdf849815e97330f245d893877e6693251c2695c` against `572c82659cbac0737467cf6d0703f9a27e58b2bf`. The linked #1116 report and repository-member confirmation establish alignment. The change also cherry-picks cleanly onto current `main` (`7998589`).

## Findings

- **[P1] Migrate bindings created by the old lone-record fallback** — `core/tools/aidlc-lib.ts:4344`

  If an affected session upgrades with a binding created by the old lone-record fallback, `isWorkflowParticipant` trusts that binding unconditionally. I reproduced `SESSION_STARTED`, `HUMAN_TURN`, `SUBAGENT_COMPLETED`, and `SESSION_ENDED` being written to the teammate's record after upgrading. This preserves the authority contamination #1116 is intended to fix.

  Please version or record provenance for trustworthy bindings and migrate or revalidate legacy bindings whose only evidence may be the fallback. Add an upgrade regression covering an already-polluted session.

- **[P2] Count valid legacy UUID stamps as participation** — `core/hooks/aidlc-session-start.ts:169`

  A valid legacy session stamp resolves `stampedTarget`, but the resulting selection has no binding and is rejected as a non-participant. Existing test `t169` consequently loses the expected rebind offer. A successfully resolved local session UUID stamp should count as participation evidence before the selection is nulled.

- **[P2] Gate PreCompact before fallback-derived writes** — `core/hooks/aidlc-validate-state.ts:30`

  In a teammate-only clone, PreCompact still writes `validate-state.last` and `.aidlc-recovery.md` under the foreign record. If a shard exists from earlier contamination, it also appends `SESSION_COMPACTED`.

  Resolve the payload session, establish participation, and pin the selected paths before any heartbeat, recovery, invalidation, or audit writes. Please add this case to the participation regression suite.

- **[P2] Complete the repository test-contract migration** — `tests/integration/t342-workflow-participation.test.ts:1`

  Three unit shards remain red. The new test is absent from the coverage registries, and the t238/t265/t312 fixtures do not establish a real participation signal under the new contract. Regenerate the coverage registry and update those fixtures with an exact cursor, binding, metadata, or participant marker rather than weakening the production check.

- **[P2] Add the required release metadata**

  Rebasing leaves version `2.9.0` unchanged. Under current project policy, this user-visible fix needs the next available patch bump, matching README badge, and dated CHANGELOG entry.

## Validation

- Packaging drift check and typecheck pass.
- New `t342`, `t160`, authority tests `t188`/`t261`, and the focused e2e audit suite pass.
- Existing `t169` and both targeted upgrade/PreCompact diagnostics reproduce the findings.
- Current GitHub CI has unit shards 1–3 and aggregate smoke+unit failing.
- PR #1157 overlaps the Kiro adapter surfaces; whichever lands second must preserve these changes during rebase.

No fixes were pushed; diagnostics were local only.

**Review event: `REQUEST_CHANGES`.**

apackeer (changes requested):
Thank you for the careful split between resolution and participation — keeping the lone-record fallback for worktrees while refusing to read it as membership is the right shape, and `t342` is a faithful encoding of the report. The five-hook recipe from #1116 is fixed on this head.

I reviewed the current head `cdf849815e97330f245d893877e6693251c2695c` against `572c82659cbac0737467cf6d0703f9a27e58b2bf` (cherry-picks cleanly onto `main` at `be94bde7`). Alignment: #1116 with maintainer-confirmed reproduction and @rafaelgsr's review. A second independent reviewer (Codex) reproduced the findings below.

**P1 — Plan Approval guard fails open for a legitimately bound participant.** `core/hooks/aidlc-plan-approval-guard.ts:925` resolves participation with `resolveWorkflowSelection(projectDir)`, ignoring the payload `session_id`. With ordinary commands only — session A `/aidlc intent A`; session B `/aidlc intent B`, then `intent archive B` (clears the shared cursor, `core/tools/aidlc-utility.ts:6868`) — A's binding is its only signal, and on Windows (no ancestry by design, `core/tools/aidlc-lib.ts:4217`) the guard now permits A's unapproved code-generation write (exit 2 on base → exit 0 on head; identity via env restores exit 2). The Copilot and Cursor adapters forward `session_id` and still fail the same way because the core ignores it; the Kiro CLI (`harness/kiro/hooks/aidlc-kiro-adapter.ts:659`), Kiro IDE (`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:1604`) and opencode (`harness/opencode/plugin/aidlc-opencode-adapter.ts:523`) adapters additionally drop the session id from their normalized guard calls; the Codex adapter exports it (`harness/codex/hooks/aidlc-codex-adapter.ts:199`) and still rejects. Please thread the payload identity into the guard the way `aidlc-record-human-turn.ts:114-124` and `aidlc-session-end.ts:86-89` do, pin the whole guard evaluation to that selection, and fix the three adapters' normalized calls.

**P1/P2 — Legacy bindings** (reproduced, agree with the prior round): a binding written by the old fallback is trusted by space/intent match alone (`core/tools/aidlc-lib.ts:4344`). **P2 — Legacy UUID stamp** (`t169`): the `resume` selection has `binding:null`, is rejected, and session-start then persists `intent:null` for that session (`core/hooks/aidlc-session-start.ts:152,169,177`). These two need one migration/rejoin policy: revalidating bindings must not blindly accept every stamp, since session-start writes stamps automatically (`:286`).

**P2 — Kiro IDE argument-less Plan Approval path still blocks non-participants.** `harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:1422` reads bare `stateFilePath(projectDir)` and returns the local block at `:1438` before the gated core hook runs (exit 2 measured after SessionStart bound `intent:null`; populated-write control exits 0).

**P2 — Stop hook** (`core/hooks/aidlc-continue-workflow.ts:1315`) checks state existence, not participation. For a session with no binding and no transcript it consults `next` against the teammate's record, emits `decision: block`, and writes `.aidlc-stop-hook/block-count.json` there. Bounded by the no-progress cap and protected when SessionStart ran with the same id, so conditional — but the same class as the fixed hooks.

**P2 — PreCompact** (`core/hooks/aidlc-validate-state.ts:28,64,72`) writes `.aidlc-recovery.md` and the heartbeat under the teammate's record; `SESSION_COMPACTED` appends once this clone's shard exists. The guard also writes its heartbeat before the participation check (`aidlc-plan-approval-guard.ts:876`).

**P2 — Test contract.** `t265` seeds state at the bare intents root (`tests/unit/t265-plan-approval-guard.test.ts:464`), which `intent===null` can never make a participant — move it into a real record with a cursor; `scripts/build-binaries.ts:1277` uses the same fixture (this is the `t238` red, so native release validation breaks too). `t312` writes `.active-intent` (`tests/unit/t312-attest-resolve-anchor.test.ts:106`), not `active-intent`. Regenerate the coverage registry for `t342` (`t134`).

**P3** — `core/tools/aidlc-utility.ts:6932` (`/aidlc space`) still binds via `activeIntent(...)`, which includes the fallback. — The fresh-clone behaviour change (first scoped `/aidlc` on a lone record now routes to the intent-pick prompt) deserves a line in `docs/guide/03-spaces-and-intents.md` and `docs/reference/01-architecture.md`. — The Unit marker is checkout-wide (`core/tools/aidlc-unit.ts:3550`), so it cannot give the exact-target property of the other signals; worth stating as a known limitation.

One correction to the prior round: per `AGENTS.md` "Release Metadata Policy", fix PRs do not bump the version, badge, or changelog — no release metadata is needed here.

Validation: issue recipe, legacy-binding, archive-sequence, Stop (Kiro/Claude payloads), PreCompact and Kiro IDE legacy probes against an installed `dist/claude`; `t342`/`t160`/`t311` pass; `t265`/`t312`/`t134`/`t169`/`t238` fail as in CI (the same slice passes on base). Nothing pushed.

**Review event: `REQUEST_CHANGES`.**

Issue #1047, opened 2026-09-08
Release process change: native distribution and project configuration from v2.8.0


leandrodamascena:
## Summary

Starting with AI-DLC `2.8.0`, we are changing how releases are built, installed, configured, and updated.

The repository no longer commits `dist/` as the user installation channel. `dist/` and `dist-release/` are now ignored, disposable build outputs used by maintainers, tests, and release assembly. Repository source archives therefore do not contain a ready-to-copy harness distribution.

The supported release channel is now a native, per-user `aidlc` installation plus versioned runtime assets. This gives users one consistent lifecycle across Claude Code, Kiro CLI, Kiro IDE, Codex CLI, Cursor, opencode, and GitHub Copilot without requiring Bun or Node.js.

> [!IMPORTANT]
> The GitHub Releases page is the source of truth for availability. A version tag starts the release workflow; installers and runtime archives are usable after the corresponding GitHub Release has been published. Do not use GitHub source archives as a substitute for release assets.

## What a release now contains

A native release includes:

- a platform-specific `aidlc` executable for supported macOS, Linux, and Windows targets
- all seven harness runtimes in the installed version
- `install.sh` and `install.ps1`
- the versioned manual-copy archive `aidlc-runtime-X.Y.Z.tar.gz`
- `version.json`, SHA-256 checksums, byte-length metadata, and release provenance

Installation is per-user and does not require `sudo`. The installer places versioned runtimes under the user install root and exposes one stable `aidlc` command. It verifies source identity and checksums before activation. A compatible GitHub CLI also verifies the signed attestation.

## Recommended installation

macOS and Linux:

```bash
tmp="$(mktemp -d)"
curl -fsSL \
  https://github.com/awslabs/aidlc-workflows/releases/latest/download/install.sh \
  -o "$tmp/install.sh"
sh "$tmp/install.sh"
rm -rf "$tmp"
export PATH="$HOME/.local/bin:$PATH"
```

Windows PowerShell:

```powershell
$download = Join-Path $env:TEMP "aidlc-install-$PID"
New-Item -ItemType Directory -Force $download | Out-Null
$installer = Join-Path $download install.ps1
Invoke-WebRequest `
  -Uri https://github.com/awslabs/aidlc-workflows/releases/latest/download/install.ps1 `
  -OutFile $installer
& $installer
Remove-Item -Recurse -Force $download
```

[...]

PR #1174, opened 2026-09-14
fix: separate Bun copy and native installer runtimes

### LINKED ISSUE #1173: Restore Bun-only manual-copy runtime and separate native installer runtime assets
## Problem

The `aidlc-runtime-2.8.2.tar.gz` release asset is generated from
`dist-release/<harness>`. That projection rewrites hooks and tool invocations
to:

```text
aidlc engine ...
```

As a result, copying `runtime/<harness>/` from the archive still requires the
native `aidlc` executable in `PATH`. This breaks the manual-copy channel for
customers who cannot install native executables and must run AI-DLC through an
existing Bun installation.

This also corrects the manual-copy contract currently described in #1047.

## Reproduction

1. Download and extract `aidlc-runtime-2.8.2.tar.gz`.
2. Copy `runtime/claude/` into a project.
3. Ensure Bun is available but `aidlc` is not in `PATH`.
4. Start Claude Code.

The generated hooks and status line execute `aidlc engine ...` and fail with
`aidlc: command not found`.

## Expected behavior

Release the two runtime projections as distinct assets:

- `aidlc-runtime-X.Y.Z.tar.gz`
  - Generated from `dist/<harness>`.
  - Intended for manual-copy installations.
  - Invokes `bun .../tools/aidlc.ts`.
  - Works without a native `aidlc` executable.
- `aidlc-native-runtime-X.Y.Z.tar.gz`
  - Generated from `dist-release/<harness>`.
  - Consumed by `install.sh`, `install.ps1`, and native lifecycle commands.
  - Invokes `aidlc engine ...`.
  - Preserves the Bun-free native installation experience.

The platform-specific `aidlc-*` executables remain separate release assets.

## Acceptance criteria

- The manual runtime archive is generated from `dist/`.
- The native runtime archive is generated from `dist-release/`.
- Shell and PowerShell installers download the native runtime archive.
- Native install, update, rollback, and offline flows use the native runtime
  archive.
- Both assets are represented in `version.json`, `checksums.txt`, and release
  provenance.
- Release verification validates both runtime assets.
- Tests prove the manual archive invokes Bun and has no required `aidlc engine`
  entrypoints.
- Tests prove the native archive invokes `aidlc engine` and does not require
  Bun.
- Packaging coverage includes every supported harness.
- README, installation, troubleshooting, architecture, and release
  documentation clearly distinguish both channels.
- Native installations remain Bun-free.
- Manual-copy installations work with Bun and no native executable.

leandrodamascena:
## Summary

- preserve the historical `aidlc-runtime-X.Y.Z.tar.gz` name and native `dist-release/` contents used by existing installers and every shipped 2.8.x updater
- publish the Bun/manual-copy projection as `aidlc-copy-runtime-X.Y.Z.tar.gz`, outside `version.json` and `checksums.txt`, with a dedicated `.sha256` sidecar covered by release provenance
- keep release readers forward-compatible with structurally valid unselected asset kinds while strictly validating selected runtime, installer, binary, target, and verification metadata
- align Unix and PowerShell installers, lifecycle code, release verification, fixtures, preview inventory, and documentation with the compatible asset contract
- add regression coverage for the frozen 2.8.x manifest parser, sidecar integrity, exact fixture metadata, all seven harness projections, native installation, and Bun-only execution with no `aidlc` binary on `PATH`

## Compatibility

Existing 2.8.x clients continue to parse `version.json` and self-update because the manifest still contains exactly one runtime under the historical native asset name. Manual-copy users explicitly select the out-of-band copy archive and verify its versioned checksum sidecar; both files are included in the release provenance subjects.

## Validation

- `bun run typecheck`
- `bun run lint`
- `bash -n scripts/install.sh`
- `git diff --check`
- `bun test tests/unit/t238-build-binaries.test.ts -t "native build compiles, gates, and runs version plus a delegate"`
- `bun test tests/unit/t243-install-mechanism.test.ts -t "release readers tolerate unselected future assets but validate selected metadata"`
- `bun test tests/unit/t243-install-mechanism.test.ts -t "copy projections stay Bun-invoked while release projections are native"`
- `bun test tests/unit/t244-install-management.test.ts -t "all harness runtimes install together|release candidate verification rejects mutable metadata"`
- `bun test tests/unit/t331-preview-channel-lifecycle.test.ts`
- `bun test tests/unit/t332-preview-release-pipeline.test.ts`

Closes #1173


apackeer (changes requested):
Thanks Leandro — separating the Bun-shaped copy runtime from the native runtime is the right fix for the #1173 users who cannot run a native executable, and the t238 proof that the copy archive's statusline runs under `bun` with `PATH=""` (`tests/unit/t238-build-binaries.test.ts:333-391`) is exactly the kind of evidence this change needs. Docs are consistently updated across README, guide, and reference.

I reviewed the current head 0932066cab9104584b07a9ffb8a493f0c48de7fb against main 572c82659cbac0737467cf6d0703f9a27e58b2bf (merge-base = base; no drift, no file overlap). Alignment basis: maintainer-owned issue and PR. Because #1173 explicitly reverses the manual-copy contract announced in #1047 ("the archive … uses the matching native command"), I reviewed this as a contract change rather than a pure bug fix, with particular attention to mixed-version operation.

### P1 — every shipped 2.8.x install loses `aidlc update` against the first release built from this branch

**Trigger.** Any 2.8.0–2.8.2 `aidlc` binary runs `aidlc update`, `aidlc update --check`, `aidlc update --dry-run`, or `aidlc update --from <dir>` against a release whose `version.json` was produced by this branch (`scripts/package-release.ts:324-331` adds a second `kind: "runtime"` asset).

**Path.** In v2.8.2, `readReleaseManifest` (`core/tools/aidlc-release.ts:304`) rejects any `kind: "runtime"` asset whose name is not `aidlc-runtime-<version>.tar.gz`. Both `fetchReleaseMetadata` and `acquireRelease` call it on the full downloaded `version.json`, before any asset is selected, so the old binary never reaches the point where it would pick the legacy name. The old validator also rejects unknown `kind` values and unknown installer names, so no rename of the new asset avoids this.

**Observed** (real v2.8.2 `aidlc-linux-arm64`, checksum-verified, `AIDLC_INSTALL_ROOT`/`AIDLC_BIN_DIR` in a temp root, release fixture built by this branch's `writeReleaseFixture({ version: "2.8.3" })` and served with `serveReleaseFixture`):

```
update --check                 exit 3  update refresh failed: aidlc-native-runtime-2.8.3.tar.gz: invalid asset metadata
update --check --json          exit 3  {"ok":false,"code":3,"status":"failed","message":"update refresh failed: ..."}
update --dry-run               exit 1  aidlc-native-runtime-2.8.3.tar.gz: invalid asset metadata
update --from <dir> --offline  exit 4  aidlc-native-runtime-2.8.3.tar.gz: invalid asset metadata
```

The in-project update check takes the same path (`core/tools/aidlc-update.ts` → `fetchReleaseMetadata`) and lands in the catch at v2.8.2 `:279`, reporting `update refresh unavailable`. Existing installs therefore get no signal that a newer version exists, and a user who tries anyway gets an error that does not mention re-running the installer.

**Contract violated.** `docs/guide/18-install-and-lifecycle.md` and #1047 promise `aidlc update` as the upgrade path for native installs; that promise breaks for the entire installed base at once.

**Correction — a decision is needed, and I'd like your view:**

- **(a) Keep `aidlc-runtime-<v>.tar.gz` native.** Its bytes and semantics stay what every 2.8.x binary and cached installer expects. Publish the Bun archive under a new name (e.g. `aidlc-copy-runtime-<v>.tar.gz`) **outside** `version.json` and `checksums.txt`, with its own `.sha256` sidecar; `attest-build-provenance` already covers `build/release/*`. Cost: `scripts/verify-release.ts:185-187` pins the exact asset inventory and `:200-217` rejects any extra file in the release directory, and `release.yml:590` runs that verifier after attestation — so this option also needs a verifier allowance for the unmanifested archive plus its sidecar. It deviates from #1173's "both assets in `version.json`/`checksums.txt`" criterion, which I think is worth trading for a non-breaking upgrade.
- **(b) Ship as designed and accept the break.** Then the release notes, README, `18-install-and-lifecycle.md`, and #1047 must state prominently that 2.8.x installs cannot self-update and must re-run `install.sh`/`install.ps1` once, and the copy-channel sentence in #1047 needs correcting. This still leaves the silent `update refresh unavailable` on every 2.8.x machine until the user acts.

### P2 — the legacy-name fallback is untested

`installRuntimeAsset` (`core/tools/aidlc-lifecycle.ts:280-287`, used at `:1260` and `:1267`) is what lets the new binary acquire a pre-PR release — `aidlc use <2.8.x>` for a version that is not retained (`useCommand:1906`), `aidlc update --version <2.8.x>` (`:1695`), `versions install` (`:1456`), and `install-apply --from <2.8.x directory>` (`:2130`) — and the PR description lists that compatibility as a goal. (`rollbackCommand:1765-1802` only re-activates retained files and never selects an archive, so it is not affected.) No fixture exercises the fallback branch. Note that omitting the native asset from today's `writeReleaseFixture` is not enough: `release-fixture.ts:233-242` fills `aidlc-runtime-*` from `dist/`, so the legacy-named archive would carry Bun projections and the test would pass for the wrong reason. A genuinely legacy fixture needs `dist-release/` bytes under `aidlc-runtime-<v>.tar.gz` with no native asset in the manifest, and the assertion should be that the installed runtime's hooks invoke `aidlc engine` (Bun-free) after the fallback install.

### P3 — fixture manifest shape drifts from `package-release`

`tests/harness/release-fixture.ts:314-321` excludes only `runtimeAsset` from the `{ target }` spread, so the native runtime asset carries `target: "<host-triple>"` in every fixture manifest (confirmed by dumping a generated fixture), while `package-release.ts:324-331` emits no `target` for runtime assets. `readReleaseManifest` happens not to check `target` on runtime kinds, so nothing fails today, but the lifecycle tests are validating a manifest shape the release never produces.

### P3 — per-harness coverage

The new t238 block asserts the copy/native invariant only for `runtime/claude/.claude/settings.json`; #1173's acceptance criteria call for every harness. I extracted both archives from a fixture built from this branch and grepped all seven `runtime/<harness>/` roots: the copy archive has no `aidlc engine` entrypoint and the native archive has no `bun …ts` invocation (the only file mentioning both is `cursor/install.ts`, byte-identical dual-mode in both archives; `plugins/` is identical in both and its hooks fall back to bun). No defect, but a per-harness assertion would make the guarantee durable.

### P3 — non-blocking note on forward compatibility

The new `readReleaseManifest` (`core/tools/aidlc-release.ts:303-316`) still enumerates the exact allowed runtime names. That is the property that produced the P1 for 2.8.x readers; tolerating unknown *manifest* entries (while still strictly validating the assets actually selected and keeping the installed-runtime integrity trigger in `aidlc-install-paths.ts:388-394` untouched) would spare the next asset addition from repeating it. Not a defect at this head — flagging it because the same decision is being made again.

### CI

All checks green except `Require Contributor Statement`: the PR body needs the license sentence the workflow expects (`.github/workflows/pull-request-lint.yml:188`). Merge state is `UNSTABLE` until then.

### Validation performed

CI results at 0932066; `bun scripts/package.ts` on the head; fixture built with the branch's `writeReleaseFixture`; empirical probes with the shipped v2.8.2 linux-arm64 binary (offline `--from` and online via `serveReleaseFixture` with `--release-base-url`); both archives extracted and grepped across all seven harness roots and `plugins/`. No changes pushed.

**Event: REQUEST_CHANGES** — blocking on P1 (decision + implementation) and the contributor statement; P2 should land with it; P3s at your discretion.

leandrodamascena:
Thanks for the thorough compatibility review. I addressed all feedback in commit `735fdba60`.

I reviewed the current head `735fdba60f35b72756d0f8d7782ca2326124db49` against base `572c82659cbac0737467cf6d0703f9a27e58b2bf`. This remains aligned with maintainer-owned issue #1173.

Changes made:

- Preserved `aidlc-runtime-X.Y.Z.tar.gz` as the native `dist-release/` archive, so existing 2.8.x clients can parse the manifest and self-update.
- Added `aidlc-copy-runtime-X.Y.Z.tar.gz` for Bun/manual-copy users outside `version.json` and `checksums.txt`.
- Added a dedicated `.sha256` sidecar for the copy archive; both files remain covered by release provenance.
- Removed the native-name fallback because the historical runtime name and semantics no longer change.
- Corrected fixtures so runtime assets do not carry a binary `target`.
- Added archive assertions for all seven harnesses.
- Allowed structurally valid future manifest entries when unselected, while retaining strict validation for selected runtime, installer, binary, target, and verification metadata.
- Updated installers, lifecycle code, release verification, preview asset count, and documentation.
- Added the contributor license statement to the PR body.

Validation completed:

- `bun run typecheck`
- `bun run lint`
- `bash -n scripts/install.sh`
- `git diff --check`
- Release build test: 1 passed, including 3,204 assertions
- Focused lifecycle/install tests: 4 passed
- Preview lifecycle: 5 passed
- Preview pipeline: 19 passed

CI currently has contract, security, documentation, smoke, changelog, and contributor-statement checks passing. Unit shards and integration/e2e are still running. The remaining merge-status failure reflects the existing `CHANGES_REQUESTED` review.

Could you please re-review the updated head?

apackeer (approved):
Re-reviewed `735fdba60f35b72756d0f8d7782ca2326124db49` against `572c82659cbac0737467cf6d0703f9a27e58b2bf`. The previous blocking compatibility issue is resolved.

The historical `aidlc-runtime-<version>.tar.gz` remains native and manifest-listed; the Bun copy archive has its own name and checksum sidecar outside the legacy inventories ([packaging](https://github.com/awslabs/aidlc-workflows/blob/735fdba60f35b72756d0f8d7782ca2326124db49/scripts/package-release.ts#L290-L328)). The release verifier includes both additional files and checks their checksum correspondence. Both publication workflows attest them through the existing complete-directory glob.

I repeated the compatibility probe with the checksum-verified, shipped v2.8.2 Linux arm64 binary against a locally built 2.8.3 fixture. `update --check` and its JSON form correctly report the newer release (exit 5/action-needed); `update --dry-run`, online `update`, and offline `update --from` complete successfully. Installed hooks remain native. This probe used isolated install directories, no Bun in the child PATH, and an unavailable optional provenance verifier; the next release's binary and signature were fixtures.

The alternate native archive name and fallback are gone. The native install assertion, corrected fixture metadata, all-seven-harness invocation checks, and selected-asset validation address the other prior findings. The contributor statement is now present.

One nonblocking P3 documentation follow-up: the [README quick start](https://github.com/awslabs/aidlc-workflows/blob/735fdba60f35b72756d0f8d7782ca2326124db49/README.md#L33-L48) and [Getting Started](https://github.com/awslabs/aidlc-workflows/blob/735fdba60f35b72756d0f8d7782ca2326124db49/docs/guide/01-getting-started.md#L27-L42) introduce Bun-only copying, then show native `aidlc config` commands without identifying the step copy users should skip. The [reference prerequisite](https://github.com/awslabs/aidlc-workflows/blob/735fdba60f35b72756d0f8d7782ca2326124db49/docs/reference/06-hooks-and-tools.md#L1012) also still describes Bun as source-only. The complete Copy Channel chapter already documents the correct procedure.

Validation: `bash tests/run-tests.sh --debug -P 8 --unit --filter '^t(238|243|244|331|332)-'` completed with 9 files, 242 passed cases, zero failures, and 15 Windows-only skips on Linux arm64. No live-model variables were enabled. The native build test also exercised release packaging and its determinism guard. The latest CI test, contract, security, and contributor-statement checks succeeded at the reviewed head.

No blocking findings remain. Approving.

PR #1176, opened 2026-09-14
fix: join FR-keyed story-map rows when user-stories is skipped

### LINKED ISSUE #1117: [Bug]: traceability sensor reports every FR id as a phantom gap when user-stories is skipped (storyAssignments US-only parse)
### Description

When a scope skips `user-stories`, the `traceability` sensor reports **every `FR` id as a gap** at `units-generation`, even when coverage is complete. The source-id fallback is correct; the failure is one layer down, in the unit-assignment parse.

`resolveUpstream()`'s `units-generation` branch resolves **source** ids correctly, falling back from `US` to `FR` when `stories.md` is absent:

```ts
const source = existsSync(stories)
  ? idsFromFile(stories, [ID_PATTERNS.US], "stories.md")
  : idsFromFile(requirements, [ID_PATTERNS.FR], "requirements.md");
```

It then resolves unit **assignments** through `storyAssignments()`, which parses `unit-of-work-story-map.md` with a US-only pattern — `core/tools/aidlc-sensor-traceability.ts` **line 233**:

```ts
const stories = extractIds(line, [ID_PATTERNS.US]);
```

With no stories on disk that map comes back empty, so every `FR` id falls through **line 324**:

```ts
if (!mapped.assignments.has(id)) result.extraGaps.push(id);
```

Two consequences: every source id is reported as a gap, and `storyAssignments()` additionally returns the reason `unit-of-work-story-map.md contains no story-to-unit mappings`.

`ID_PATTERNS.FR` and `ID_PATTERNS.NFR` already exist in the same file (lines 58–59), so this may be a small change at the assignment parse rather than anything structural — for example threading the caller's pattern list through, or accepting `[US, FR, NFR]` when `stories.md` is absent. The `no story-to-unit mappings` reason string probably also wants to be conditional on stories being in scope.

Related but distinct: #1011 is a different failure in the same sensor (zero-Unit `code-generation`, `extractUnitName()`'s three-segment regex). This one is the `units-generation` stage with units present and no stories.

### Steps to Reproduce

1. Run a scope that skips `user-stories`. Ours also skipped `rough-mockups`, `refined-mockups` and `functional-design`; the relevant part is only that `stories.md` and a populated `unit-of-work-story-map.md` do not exist.
2. Trace `units-generation` coverage by `FR` id in `traceability.json` — the documented convention when there are no stories.
3. Fire the sensor the stage declares:

```bash
bun .claude/tools/aidlc-sensor.ts fire traceability \
  --stage units-generation \
  --output-path "$PWD/aidlc/spaces/default/intents/<record>/inception/units-generation/traceability.json"
```

### Expected Behavior

The sensor validates the `FR`-keyed `traceability.json` and passes when every `upstream_ids` entry is covered with a valid status and each `OK` entry names a target — the same checks it already implements, none of which require a story map when stories are out of scope.

### Actual Behavior

On an increment with 37 coverage rows across 8 units, every in-scope FR and NFR1–NFR5 `OK`, **0 gaps, 0 orphans**, and 32/32 target files present, the sensor returns:

```json
{
  "pass": false,
  "gaps": ["FR1","FR1.1","FR1.2","FR1.3","FR2","FR2.1","FR2.2","FR2.3",
           "FR3","FR3.1","FR3.2","FR3.3","FR4","FR4.1","FR5","FR5.1",
           "FR5.2","FR5.3","FR6","FR6.1","FR6.2","FR7","FR7.1"],
  "invalid_targets": [
    "FR1.1: target \"U6\" is not mapped in unit-of-work-story-map.md",
    "FR1: target \"U6\" is not mapped in unit-of-work-story-map.md",
    "..."
  ]
}
```

23 phantom gaps — the complete FR set — plus one `invalid_targets` entry per target.

`default_severity` is `advisory`, so nothing was blocked. The cost is subtler: on a legitimate no-stories path the sensor cannot be satisfied at all, which trains teams to discount sensor output generally — including the sensors that are telling the truth.

### Environment

- Found on **2.6.9**; the code path was re-verified against `main` before filing (line numbers above are from `main`).
- Harness: Claude Code. Model: Claude Opus 4.8 (1M) via Bedrock.
- Phase: Inception (`units-generation`), and the same parse feeds per-unit `code-generation`.
- Interim workaround on our side: a small plugin sensor that validates FR-level coverage without a story map. Happy to share it if useful.

vnlebaoduy:
# Summary

When a scope skips `user-stories`, the `traceability` sensor at `units-generation` resolves its source ids correctly - `resolveUpstream()` falls back from `US` ids in `stories.md` to `FR` ids in `requirements.md` - but then resolves unit assignments through `storyAssignments()`, which parsed `unit-of-work-story-map.md` with the `US` pattern only (`core/tools/aidlc-sensor-traceability.ts:233`). A map whose rows carry `FR1 | U1 | u1-auth` therefore yields zero assignments, and every `FR` id is reported as a gap even when each is mapped, so the sensor can never pass for that scope shape.

Closes #1117.

## Changes

* `core/tools/aidlc-sensor-traceability.ts` - `storyAssignments()` takes the id pattern to join on. The `units-generation` branch passes the same pattern it used for the source (`US` when `stories.md` exists, otherwise `FR`), so the two halves of the join can no longer disagree. The two per-unit call sites (`functional-design`, `code-generation`) are already inside `existsSync(stories)` guards and pass `US` explicitly; their behaviour is unchanged.
* `tests/unit/t281-sensor-traceability.test.ts` - new case "units-generation joins FR rows when user-stories is skipped": requirements only, units seeded, an FR-keyed story map, and an FR-keyed `traceability.json` pass; a row that maps `FR2` to the wrong unit is still refused through `invalid_targets` (`FR2: target "u2-profile" is not mapped in unit-of-work-story-map.md`).

## User experience

Before: in any scope without `user-stories` (the fallback the `units-generation` stage prose describes: "otherwise enumerate every FR"), the sensor reported every FR as a phantom gap at `units-generation` and the stage recorded `SENSOR_FAILED` regardless of the artifact's content.

After: FR-keyed rows join to units the same way US-keyed rows do; the sensor passes on complete coverage and still names a wrong-unit target.

## Checklist

* [ ] Changes are documented

## Test plan

On `572c8265` (main):

```
bun test tests/unit/t281-sensor-traceability.test.ts                      # 14 pass
bun tests/run-tests.ts --unit --filter "<the 31 unit files mentioning traceability>"   # 31 / 31 PASS
bun tests/gen-coverage-registry.ts --check                                # fresh, guards green, ratchet held
bun scripts/package.ts --check                                            # deterministic, 7 harnesses
```

The new case fails without the sensor change: `git stash push -- core/ && bun test tests/unit/t281-sensor-traceability.test.ts` reports 1 failure (`pass` is `false` because every FR is a gap).

## Acknowledgment


vnlebaoduy:
Rebased onto `main` @ `891669aa` (after #1179 landed). The one conflict was the guarded `storyAssignments` call in the code-generation branch; kept both changes (`unit && existsSync(storyMap)` from #1179, the explicit `ID_PATTERNS.US` from this PR). t281 now 16 / 16, matching the combined result noted in the #1179 review; lint, typecheck, coverage registry, and packaging determinism green on the rebased head `51bdd2cf`.

apackeer (approved):
Thanks for this fix, and for the precise root-cause write-up — threading the source fallback through `storyAssignments()` so the two halves of the join can't disagree is exactly the right shape, and the regression case was well targeted.

I reviewed the current head `02a393f1` against `main` at `0e54914f`. The change restores the FR fallback documented in `units-generation.md:129-132`, and Leandro's #1179 review already scoped this PR as the complementary follow-up.

During review I found that #1117's own reproduction still failed at `51bdd2cf`: the reporter traces `NFR1–NFR5 OK` alongside the FRs with matching story-map rows, but the FR-only join dropped those rows, so `verifyTargets` reported each as `not mapped in unit-of-work-story-map.md` — five false `invalid_targets`, `pass=false`. I pushed `02a393f1` on top of your commit: `storyAssignments()` takes the pattern list and the no-stories join accepts `[FR, NFR]`, while the required upstream set stays FR-only so NFR coverage remains optional. The t281 case now also covers FR-only coverage passing with NFRs in `requirements.md`, an NFR row joining, and a wrong-unit NFR row being refused. Your `sourcePattern` became `hasStories`, which also removes the repeated `existsSync`.

Validation on `02a393f1`: t281 16/16 (each half fails on the prior sensor as claimed), lint, typecheck, coverage registry, packaging determinism across seven harnesses; CI, Markdownlint, and Security Scanners are green on this head.

Approving. One non-blocking follow-up for a separate PR: `units-generation.md:124` still says story-map rows are keyed `USx.y` — worth adding "or FR/NFR when `stories.md` is not produced" so authors in a no-stories scope know to write those rows.

PR #1179, opened 2026-09-14
fix: resolve the zero-Unit code-generation location in the traceability sensor

### LINKED ISSUE #1011: [Bug]: traceability sensor can never pass on zero-Unit code-generation (poc/bugfix/refactor/security-patch/express)
### Description

The `traceability` sensor can never pass for **zero-Unit Code Generation**, so every workflow in a scope that skips Units Generation records a `SENSOR_FAILED` for `traceability` on stage 3.5 no matter what the artifact contains.

The stage definition and the sensor disagree about where `traceability.json` lives:

- `core/aidlc-common/stages/construction/code-generation.md` (lines 99–111) states that for a zero-Unit directive the stage writes its artifacts under `<record>/construction/code-generation/` **"with no synthetic Unit segment"**.
- `core/tools/aidlc-sensor-traceability.ts` `extractUnitName()` (lines 195–199) derives the Unit from the output path with a **three-segment** regex:

  ```ts
  const match = normalized.match(/\/construction\/([^/]+)\/[^/]+\/traceability\.json$/);
  ```

  The zero-Unit path has only two segments after `construction/`, so `extractUnitName` returns `null`.

`code-generation` is not one of the stages with an early return in `resolveUpstream()` (`user-stories`, `domain-design`, `units-generation` are), so it falls through to `resolveUnitContext()`, which returns `cannot derive the construction unit from output path: …` and short-circuits before the unit-independent target-existence check for `stage === "code-generation"` (lines ~512–529) can run.

Meanwhile `code-generation.md` frontmatter still declares the sensor:

```yaml
sensors:
  - required-sections
  - linter
  - type-check
  - traceability
```

**Affected scopes** — every scope whose `code-generation.scopes` includes it but whose `units-generation.scopes` does not, derived from the compiled `stage-graph.json`:

`poc`, `bugfix`, `refactor`, `security-patch`, `express` — 5 of the 11 shipped scopes.

`default_severity: advisory`, so the approval gate is not blocked. The practical effect is that the one deterministic traceability check the framework ships is silently inert in the five lightest scopes — exactly the scopes where there is no design stage and no per-Unit structure to fall back on.

### Steps to Reproduce

```bash
# 1. Install the Claude Code distribution into an empty git repo
mkdir demo && cd demo && git init
cp -r /path/to/aidlc-workflows/dist/claude/.claude/ .claude/
cp -r /path/to/aidlc-workflows/dist/claude/aidlc/   aidlc/

# 2. Create a poc-scope intent
bun .claude/tools/aidlc-utility.ts intent-create \
  --scope poc --label "demo" --arguments "any small CLI"

# 3. Reach Code Generation and let it write the zero-Unit artifacts to
#    aidlc/spaces/default/intents/<record>/construction/code-generation/
#    (no Unit segment, per the stage definition)

# 4. Fire the sensor the stage declares, on the artifact the stage produces
bun .claude/tools/aidlc-sensor.ts fire traceability \
  --stage code-generation \
  --output-path "$PWD/aidlc/spaces/default/intents/<record>/construction/code-generation/traceability.json"
```

### Expected Behavior

The sensor validates the zero-Unit `traceability.json` and passes when every `upstream_ids` entry is covered with status `OK` and each `coverage[].target` resolves to an existing workspace-relative file — the checks it already implements for `stage === "code-generation"`, none of which need a Unit name.

Alternatively, if zero-Unit traceability is intentionally out of scope for this sensor, the stage should not declare `traceability` in its `sensors:` list for zero-Unit runs, so the run does not record a spurious `SENSOR_FAILED`.

### Actual Behavior

`result: "failed"` on every invocation, with a `reason` that is independent of the artifact's contents:

```json
{
  "pass": false,
  "gaps": [],
  "orphans": [],
  "missing_from_table": [],
  "missing_from_upstream_ids": [],
  "invalid_entries": [],
  "invalid_targets": [],
  "findings_count": 1,
  "reason": "cannot derive the construction unit from output path: /…/construction/code-generation/traceability.json"
}
```

To confirm the artifact itself is not the problem: the same file passed every other check on the way to this one. Working backwards through the sensor's earlier validations I first got `upstream_ids must be an array of non-empty strings` (schema wrong), then `target file does not exist: …` for a comma-joined multi-path `target` (one target per entry is required), and only after both were correct did the message become the unit-derivation failure — i.e. the artifact is schema-valid with 39 `upstream_ids` and 129 `coverage` entries whose targets all exist, and it still cannot pass.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.7.0 / `96b11d39028955d4f92375e783525db5275cdfd8`.

Also verified unfixed on current `main`: `extractUnitName`'s regex and the zero-Unit output-path paragraph in `code-generation.md` are both unchanged there.

### AI-DLC Phase

Construction

### Harness

Claude Code

### AI Model

Claude Opus 5

### Environment

- OS: macOS (darwin 25.6.0)
- bun 1.4.0
- Scope: `poc`, Depth: Minimal, Test Strategy: Minimal
- Greenfield, zero-Unit (no `unit-of-work.md`, no Unit DAG)

### Additional Context

**Possible fix directions** (in rough order of least surprise):

1. Let `extractUnitName` recognise the stage-level path and return a sentinel, and have the `code-generation` branch of `resolveUpstream()` skip the Unit-scoped resolution when that sentinel is present. The `stage === "code-generation"` target-existence loop is already Unit-independent, and `resolveBoltDag()` returning `none` is the normal, expected state in these scopes rather than an error.
2. Give `code-generation` its own early-return branch in `resolveUpstream()` alongside `user-stories` / `domain-design` / `units-generation`, sourcing `upstream_ids` from `requirements.md` when no Unit DAG exists.
3. Make the sensor binding conditional on the directive carrying a Unit, so a zero-Unit run does not fire a check that cannot apply.

**Docs note:** `.claude/knowledge/aidlc-shared/verification.md` documents the `traceability.json` shape as `{stage, unit, upstream_ids, coverage[], reverse[]}`, and `unit` is already optional there — so the schema anticipates the zero-Unit case that the path-derivation does not.

**Related observation, same area:** for `stage === "code-generation"` each `coverage[].target` must be a single existing workspace-relative path (lines ~512–529), which means a requirement covered by several files needs one `coverage` entry per file. `verification.md`'s example shows one entry per ID with a single `target`, so the "repeat the ID" pattern is not obvious from the docs. A sentence in `verification.md` would save the next person the round trip.

This was found while running a full `poc` workflow end to end (7 stages, 96 sensor firings). Happy to open a PR for whichever fix direction you prefer.

vnlebaoduy:
# Summary

For a zero-Unit directive (no Unit DAG: `poc`, `bugfix`, `refactor`, `security-patch`, `express`) the `code-generation` stage writes its artifacts under `<record>/construction/code-generation/` "with no synthetic Unit segment" (stage prose, Step 1). The `traceability` sensor's `extractUnitName()` derives the Unit from a three-segment path (`construction/<unit>/<stage>/traceability.json`), so the two-segment zero-Unit path yields `null`, `resolveUnitContext()` returns "cannot derive the construction unit from output path", and the sensor records `SENSOR_FAILED` on every zero-Unit code-generation run no matter what the artifact contains.

Closes #1011.

## Changes

* `core/tools/aidlc-sensor-traceability.ts`
  * `isZeroUnitOutput()` recognises the stage-level `construction/code-generation/traceability.json` location; `resolveUnitContext()` returns an empty-Unit context for it instead of a refusal. Restricted to `code-generation`, the only stage that runs without a Unit today; every other per-Unit stage keeps deriving a Unit exactly as before.
  * `constructionDir()` is the one place that maps a (possibly empty) Unit to its record directory: `construction/<unit>/<stage>/` per Unit, `construction/<stage>/` for the zero-Unit run. The code-generation branch reads NFR requirements and `rules.md` through it, so a zero-Unit run picks up the stage-level equivalents.
  * The story-map join runs only when a Unit exists; with no Unit, every AC in `stories.md` is upstream, matching the stage-level scope.
  * The empty-upstream reason names the zero-Unit run instead of `unit ""`.
* `tests/unit/t281-sensor-traceability.test.ts` - new case "code-generation resolves the zero-Unit stage-level location": requirements plus stage-level `construction/functional-design/rules.md`, a `construction/code-generation/traceability.json` covering `FR1`, `NFR1`, `BR1.1` with an existing target file passes; a file that omits `BR1.1` is refused through `missing_from_upstream_ids`, which proves the stage-level rules file was resolved.

This is deliberately confined to the sensor's path resolution so it rebases cleanly onto #1005 (which moves per-Unit artifacts to `construction/units/<unit>/` and states that zero-Unit fallbacks stay stage-level): `constructionDir()` is the single line to update there. No version bump, badge, or CHANGELOG entry, per the Release Metadata Policy in `AGENTS.md`.

## User experience

Before: every zero-Unit workflow recorded `SENSOR_FAILED` for `traceability` at code generation with "cannot derive the construction unit from output path", regardless of coverage.

After: the sensor evaluates the zero-Unit artifact against the stage-level upstream set and passes on complete coverage; per-Unit runs are unchanged.

## Checklist

* [ ] Changes are documented

## Test plan

On `572c8265` (main):

```
bun test tests/unit/t281-sensor-traceability.test.ts                      # 14 pass
bun tests/run-tests.ts --unit --filter "<the 31 unit files mentioning traceability>"   # 31 / 31 PASS
bun tests/gen-coverage-registry.ts --check                                # fresh, guards green, ratchet held
bun scripts/package.ts --check                                            # deterministic, 7 harnesses
```

The new case fails without the sensor change: `git stash push -- core/ && bun test tests/unit/t281-sensor-traceability.test.ts` reports 1 failure ("cannot derive the construction unit from output path").

Independent of #1176 (same file, different branch of `resolveUpstream`); whichever merges second rebases trivially.

## Acknowledgment


leandrodamascena (approved):
Thanks for the focused fix.

I reviewed the current head `de9756eb` against `main` at `a517a061`. The change aligns with #1011 and restores the existing zero-Unit traceability contract.

During review, I found that a stage-level path was also accepted when a Unit DAG existed. I pushed `de9756eb` to require the DAG to be absent and added regression coverage.

Validation passed: 15 traceability tests, typecheck, lint, coverage registry, packaging determinism across seven harnesses, and integration on the current `main`.

Non-blocking follow-up: #1176 conflicts in the same code block. The combined result passes all 16 traceability tests after retaining both changes.

PR #1182, opened 2026-09-15
fix: improve Windows installation and preserve unowned files during uninstall

apackeer:
# Summary

Windows installation currently refuses elevated sessions and, when the command directory is missing from PATH, requires users to update PATH in each new session. Install for the account running PowerShell, register User PATH automatically, and show the installed command and next step. Uninstall and old-version pruning now select verified installer-owned files so user-added or modified content is preserved.

## Changes

* Add `-NoModifyPath` to skip persistent and current-process PATH changes. Report installed files separately from command readiness, with direct-command guidance for PATH conflicts or failures.
* Record per-version `installed-files.json` inventories, with the inventory hash in `version.json`. File hashes identify unchanged installer-owned files and detect user edits, so uninstall preserves changed or unlisted files and removes directories only when empty. Pruning refuses versions containing unowned or changed paths.
* Record the account and exact User PATH entry appended by this install in `windows-path.json`. This receipt identifies only the installer-added entry for removal; pre-existing entries and later unrelated PATH edits are preserved.
* Bind Windows cleanup journals to the selected files, expected hashes, directories, and purge/preservation options. These inputs detect edited cleanup targets or options so recovery stops instead of deleting a different set. Files are checked before deletion; older journals without a bound file plan are refused.
* Add ownership, preservation, PATH, and recovery regression coverage; extend stable/preview release lifecycle checks; update installation and troubleshooting guidance.

## User experience

Installation targets the current account and registers `aidlc` for new terminals. Users can opt out or invoke the printed command directly when another command takes precedence on PATH. Uninstall reports unowned or changed files it preserves, including with `--purge`; old installations retain files that lack ownership evidence.

## Checklist

* [x] Changes are documented

## Test plan

* **After rebase:** `bun run check` passed packaging, determinism, both typechecks, and lint; `git diff --check` passed. Focused `t243-install-mechanism`, `t244-install-management`, and `t-uninstall-plan` suites finished with **145 passed, 0 failed, and 15 Windows-only cases skipped on Linux**. A five-case read-only PowerShell parameter/version-selection probe passed; it did not execute the installer body.
* **Before rebase:** full native install/uninstall fixture checks passed on Ubuntu ARM64 and Windows, covering regular uninstall, purge, PATH opt-out, pre-existing PATH entries, later PATH additions, and command conflicts. Every case preserved all 13 sentinels, with no unlisted file loss observed. Windows also passed 34 installer/ownership/output tests and 92 uninstall tests; native builds, PowerShell 5.1 parsing, and PSScriptAnalyzer checks passed.
* **Limits:** full native install/uninstall checks were not repeated after rebase. Native fixtures were unsigned local builds: SHA-256 verification remained enabled, but release-signature verification was not exercised. These focused checks do not represent the full suite or live model/IDE gates.
* [ ] Repeat native Windows installation and regular/purge uninstall on disposable fixtures, including PATH opt-out, pre-existing entries, later edits, changed/unlisted files, and interrupted cleanup. Confirm only planned files and the owned PATH entry are removed.

## Acknowledgment


apackeer:
Pushed the three review fixes in d479a9f88cb8b06b2fc60986a3a33817a4136889:

- Apply the Windows uninstall root guard only when a pending continuation will be relaunched, so it does not block unrelated commands on an idle install.
- Preserve a version with malformed or non-object `version.json` metadata instead of aborting uninstall. Its files have no usable ownership evidence.
- Check the PATH receipt binding even when the journal's receipt was removed after scheduling. Both recovery and the emitted Windows worker now refuse that change before deleting files or touching User PATH.

Added four regression tests and extended the existing journal-rejection coverage. The committed diff exactly matches the tested patch.

| Verification | Passed | Failed | Platform skips |
| --- | ---: | ---: | ---: |
| Linux: planner, install mechanism, portable Windows cleanup cases | 137 | 0 | 61 Windows-only |
| Native Windows: planner, installer PATH, emitted cleanup worker | 141 | 0 | 4 POSIX-only |

Packaging, determinism, typecheck, lint, and `git diff --check` also passed. These counts exclude skips. The Windows worker tests use registry adapters; full native install/uninstall acceptance scenarios and live model/IDE gates were not rerun for this follow-up.

File-by-file deletion, ownership/hash checks, and nonrecursive removal of empty directories remain intact. Clearer legacy-preservation and error messages remain suggested improvements; legacy files without ownership evidence continue to be preserved.

leandrodamascena (approved):
Thanks Arden, this looks good overall. I reviewed the current head `d479a9f88` against current main `dca4b4c9b`.

The Windows PATH ownership and bounded uninstall flow look consistent. I did not find a broader behavior problem.

Small follow-up before merge: README still says there are 70 `aidlc-*.ts` tools, but this PR adds the 71st. Updating it to 71 should fix the only failing assertion in `t239-documentation-parity`.

The rest looks good to me.

PR #1186, opened 2026-09-15
fix: keep a JSON-scalar gate reply instead of parsing it away

wowzoo:
The Plan Approval matcher accepts a bare `1` or `2`, and the reply never reaches it.

`offeredPlanApprovalChoice` maps `1` to `Approve Plan` and `2` to `Request Changes` whenever the
challenge does not require exact option labels (`core/tools/aidlc-testing-posture.ts`). Exactly one
production caller passes `--exact-option-labels true`: the kiro-ide adapter, and only on its legacy
Plan Approval write mediation path (`harness/kiro-ide/hooks/aidlc-kiro-adapter.ts:520`, inside
`processLegacyPlanApprovalWrite`). A challenge issued the ordinary way — the conductor running
`aidlc-log decision` — leaves the flag unset and stores `requireExactOptionLabels: false`, on
kiro-ide as much as anywhere else, so the shortcut is live there too. It is also unreachable.

`extractResponseText` in `core/hooks/aidlc-record-human-turn.ts` JSON-parses the reply so a picker
that delivers its selection as JSON can be unwrapped, and then reports no text for anything that is
not an object, an array or a string. A reply that is itself a JSON scalar parses to a number or a
boolean and falls out of every branch. Measured on this base: `1`, `2`, `true` and `null` extract to
the empty string, while `1.`, `01`, `1)`, `#1` and `Approve Plan` survive — `01` survives only
because JSON rejects a leading zero.

A dropped reply leaves the human's presence recorded and the reply itself gone: the hook still appends
`HUMAN_TURN` and still touches the conversational marker, and only the guarded step is skipped,
because it runs under `sessionId && humanResponseText`. So no Plan Approval response is written, and
the later `answer` refuses through the message `certifyPlanApprovalReceipt` raises whenever challenge,
response, choice or identity do not line up: `Plan Approval requires the actual offered choice from
this prompt and session`. The human sees that after typing a choice the matcher would have accepted.

### Observed in a real gated run

This is not only a unit-test finding. On a gated kiro-ide Code Generation run whose challenge was
issued the ordinary way — the conductor's own `aidlc-log decision`, no `--exact-option-labels`, so
the stored challenge carries `requireExactOptionLabels: false` and the numeric shortcut is live — a
human answered the numbered prompt with a bare `1`. The hook appended `HUMAN_TURN` and recorded no
response. Twenty-two seconds later the conductor, having read the reply correctly, called
`answer --details "Approve Plan"` itself — the right label — and the engine refused with the message
above, because the authority is the recorded response and not the argument. Thirty-three minutes of
source-reading later the same human retyped `Approve Plan` verbatim, and the receipt was written ten
seconds after that. So a conductor that behaves correctly cannot rescue the dropped reply, and no
`PLAN_APPROVAL_BLOCKED` row is written in between: the only trace is one `ERROR_LOGGED` carrying the
message that names neither the reply nor the choices that would have worked.

### The change

The parse hands over only for the shapes an envelope can take — an object, an array, or a quoted
string — and a JSON scalar keeps the text the human typed. The key preference order, the array scan
and the object walk are untouched and still run for a real envelope.

The quoted-string case matters and is easy to lose. `"\"Approve Plan\""` must still unwrap to
`Approve Plan`; if the parse handed over for objects and arrays only, that reply would arrive with
its quotes and match no offered choice. A second test pins it.

Two consequences beyond the top-level reply, both intended:

- A scalar-looking string **inside** an envelope now survives too — `{"answer": "1"}` extracts `1`
  rather than nothing — and a container that holds one first stops there, so `["1", "Approve Plan"]`
  now yields `1` where it previously fell through to `Approve Plan`. A genuine JSON number, boolean or
  null value extracts to the empty string before and after.
- The extracted text also reaches `consumeSharedDirectiveAsk`, whose remedy matcher already accepts a
  number (`^([1-9]\d*)[.)]?$`), so a bare `1` now selects the first remedy of a guard-recovery ask.
  The decorated forms `1.` and `1)` already reached that matcher, since neither is valid JSON; the
  bare form joins them.

### Tests

`tests/unit/t265-plan-approval-guard.test.ts` gains two cases, both driving
`aidlc-record-human-turn.ts` with a real `UserPromptSubmit` payload against a seeded challenge:

- a bare `1` reaches the offered-choice match, so `answer --details "Approve Plan"` records the
  receipt. On this base the same case fails with the answer refused — exit 1 where 0 is expected.
- a JSON-quoted `"Approve Plan"` still unwraps to the label and pairs.

I ran 12 suites that reference this hook against a second worktree at the unmodified base to separate
new failures from existing ones. The base fails 14 of 380 in that set (`t188` 13, `t276` 1); with the
change the same 14 fail. Five further names appeared only in the batch run, and each of them is an
assertion on a subprocess exit code that came back as a killed child (`exitCode: null` or `code: -1`)
at the 5-second test timeout, at three different spawn sites — the git baseline, the decision command
and the human-turn hook. Each passes when re-run alone. The three in `t328` also issue their challenge
with `--exact-option-labels true`, which disables the branch this change touches.

`typecheck`, `biome check` on the changed files, and `package.ts` all pass.


apackeer (approved):
Thanks for the thorough writeup and the real-run trace — the distinction between an envelope and a scalar is exactly the right framing, and pinning the JSON-quoted label alongside the bare `1` keeps the fix from over-correcting.

I reviewed the current head fb8ec7c5 against main (891669aa). Rebased locally onto current main with zero conflicts; the patch is identical after rebase (same patch-id). Alignment: internal bug fix restoring a contract `offeredPlanApprovalChoice` already honored — a bare `1`/`2` when `requireExactOptionLabels` is false — that the extraction step was discarding before the matcher ever saw it. No new authority path opens: the challenge lookup, exact-label handling, and break-glass detection (`typedPrompt` reads the raw string) are untouched.

- The new `bare numeric reply` case fails against the pre-fix hook (`answer` exits 1) and passes with the fix; the JSON-quoted `"Approve Plan"` case still unwraps to the label.
- `bash tests/run-tests.sh --unit --filter '^t(188|228|265|276|328|334)-'`: 8 files, 0 failures.
- `typecheck`, `biome check` on both files, and `package.ts --check` (deterministic across all seven harnesses) pass.

One note for the record, not a change request: a genuine JSON number inside an envelope (`{"answer": 1}`) still extracts to nothing, as you disclosed; that is pre-existing and outside this fix.

Approving.

PR #1191, opened 2026-09-15
fix(doctor): read the Linux login-independent PATH sources for the hook PATH baseline

### LINKED ISSUE #1168: [Bug]: doctor: 'Runtime hook PATH' check probes getconf PATH, making its own fix text unactionable
## Summary

The machine check `Runtime hook PATH` in `aidlc-config-diagnostics.ts`
(`deriveNonInteractivePath` / `binaryProbe`) reports:

warn  Runtime hook PATH: aidlc is interactive-only at /home//.local/bin/aidlc
      fix: Add ~/.local/bin to the login-independent environment used by the harness,
           not only an interactive shell rc file.

The suggested fix cannot satisfy the check. The probe's baseline PATH is `getconf PATH`,
which returns glibc's compile-time `_CS_PATH` constant (`/bin:/usr/bin` on Ubuntu 24.04).
It is not the process environment and is unaffected by `/etc/environment`,
`~/.config/environment.d/`, PAM, systemd user sessions, or any shell rc file.

### Steps to Reproduce

1. Install `aidlc` into `~/.local/bin`.
2. Add `~/.local/bin` and `/usr/local/bin` to the `PATH` line in `/etc/environment`; log out
   and back in. Verified present.
3. Create `/usr/local/bin/aidlc -> ~/.local/bin/aidlc`.
4. Run `aidlc --doctor`.

Observed: the warning persists after every step.

Evidence that the condition the check describes is already met:

$ env -i bash -c 'echo "$PATH"; command -v aidlc'
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
/usr/local/bin/aidlc

A bare non-login, non-interactive shell — the same class of environment hooks run in —
resolves `aidlc` successfully. The check still warns, because `resolveExecutableOnPath`
searches only the `getconf PATH` directories.

### Expected Behavior

After `aidlc` is reachable from a login-independent, non-interactive environment — the
environment hooks actually run in — `aidlc --doctor` reports the `Runtime hook PATH` check
as `ok`.

Concretely, given that the following succeeds:

$ env -i bash -c 'command -v aidlc'
/usr/local/bin/aidlc

the check should pass, because the condition it exists to detect ("hooks will not be able to
invoke aidlc") is demonstrably false.

More generally, at least one of these should clear the check:

- adding `~/.local/bin` to the `PATH` line in `/etc/environment`
- adding a `PATH` entry via `~/.config/environment.d/*.conf`
- symlinking `aidlc` into `/usr/local/bin`

These are the actions a user would take in response to the fix text, and the first two are
literally what the fix text asks for.

### Actual Behavior

The check reports `warn` in all of the above cases, unchanged:

warn  Runtime hook PATH: aidlc is interactive-only at /home//.local/bin/aidlc
      fix: Add ~/.local/bin to the login-independent environment used by the harness,
           not only an interactive shell rc file.

`deriveNonInteractivePath` derives its search path from `getconf PATH`
(`/bin:/usr/bin` on this system), so:

- `/usr/local/bin` is outside the searched set — the symlink is invisible to the check
- `/etc/environment` and `~/.config/environment.d/` are never read at all
- the message still resolves and prints the realpath `~/.local/bin/aidlc`, ignoring the
  login-independent alias through which `aidlc` is in fact reachable

The verdict is therefore independent of every environment change the fix text requests. The
only mutation observed to flip it to `ok` is symlinking `aidlc` into `/bin` or `/usr/bin`.

### AI-DLC Version

v2 (current on main)

### Release / Commit

_No response_

### AI-DLC Phase

Initialization

### Harness

GitHub Copilot

### AI Model

Claude Opus 5

### Environment

- Ubuntu 24.04, GitHub Copilot CLI, AI-DLC Copilot plugin
- `aidlc` installed at `~/.local/bin/aidlc`
- `getconf PATH` -> `/bin:/usr/bin`

## Analysis

- `/usr/local/bin` is never searched, so the conventional remedy (symlink into
  `/usr/local/bin`) does not clear the check.
- `/etc/environment` is never read, although it is the canonical login-independent PATH
  source on Debian/Ubuntu and in this case already contained both directories.
- The only configuration found that clears the check is symlinking `aidlc` into `/bin` or
  `/usr/bin` — writing into distro-package-owned directories, which is worse practice than
  the setup being flagged.

## Suggested fix

In order of preference:

1. Probe behaviourally rather than by path-string analysis, e.g. resolve `aidlc` via
   `env -i bash -c 'command -v aidlc'`, which measures what hooks actually experience.
2. Build the probe PATH as the union of `getconf PATH`, the `PATH` in `/etc/environment`,
   `~/.config/environment.d/*.conf`, and the current process PATH; warn only if `aidlc`
   resolves in none of them.
3. If the present semantics are intentional, rewrite the fix text to name the only action
   that satisfies it and explain why `/usr/local/bin` and `/etc/environment` do not count.

Severity is `warn` and hooks function correctly in practice, so impact is limited to
misleading guidance — but the current text sends users to edit files the check never reads.

vnlebaoduy:
# Summary

`aidlc doctor`'s `Runtime hook PATH` check builds its Linux baseline from `getconf PATH` alone. That value is glibc's compile-time `_CS_PATH` (`/bin:/usr/bin` on the Debian family), not the environment a login session or a systemd user service receives, so it is unaffected by `/etc/environment`, `/etc/login.defs`, or `environment.d`. The check's own remediation ("add ~/.local/bin to the login-independent environment used by the harness") therefore cannot satisfy it: a user who adds the directory to `/etc/environment`, logs out and in, and even symlinks `aidlc` into `/usr/local/bin` still sees `aidlc is interactive-only`, while `env -i bash -c 'command -v aidlc'` finds it.

Closes #1168.

## Changes

* `core/tools/aidlc-config-diagnostics.ts`
  * On Linux, `deriveNonInteractivePath()` appends the `PATH` entries of the login-independent sources to the `getconf` baseline: `PATH=` lines in `/etc/environment` (pam_env), `ENV_PATH` in `/etc/login.defs`, and `PATH=` lines in `/etc/environment.d/*.conf` and `$XDG_CONFIG_HOME/environment.d/*.conf` (systemd user environment). This mirrors what the macOS branch already does with `/etc/paths` and `/etc/paths.d`. Values are treated as the literal assignments those files hold: surrounding quotes are stripped and `$PATH` / `${PATH}` references are dropped, since the entries they would splice in are already accumulated. `ENV_SUPATH` is not read; it is root's PATH.
  * `RuntimeProbeOptions.systemRoot` roots the `/etc/...` reads so tests can point the probe at a fixture; the macOS branch reads its registries under the same root. Default `/`, so callers are unchanged.
  * The `aidlc` remediation now names the surfaces the probe reads (`/etc/environment`, `ENV_PATH` in `/etc/login.defs`, `~/.config/environment.d/*.conf`) instead of "the login-independent environment".
* `tests/unit/t294-config-diagnostics.test.ts` - new case: with a fake `getconf` returning `/bin:/usr/bin` and a fixture system root, the baseline contains the entries from all four sources in order after the `getconf` ones, drops `$PATH` references and quotes, excludes `ENV_SUPATH`, deduplicates, and a directory that exists only in `/etc/environment` is found on the baseline; with an empty system root and home the baseline is exactly the `getconf` value.
* `docs/reference/01-architecture.md` - the runtime-probe paragraph names the per-platform PATH sources.

On this machine (Ubuntu, `getconf PATH` = `/bin:/usr/bin`, `/etc/environment` carrying the distribution default) the baseline goes from `/bin:/usr/bin` to `/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/sbin:/usr/games:/usr/local/games:/snap/bin`, so the issue's `/usr/local/bin/aidlc` symlink resolves. No version bump, badge, or CHANGELOG entry, per the Release Metadata Policy in `AGENTS.md`.

## User experience

Before: on Linux, `aidlc doctor` warned `Runtime hook PATH: aidlc is interactive-only` for any install outside `/bin` or `/usr/bin`, and following the printed fix did not clear it.

After: directories declared in `/etc/environment`, `login.defs`, or `environment.d` count as the login-independent PATH; the warning clears once the directory is declared there, and the fix text says exactly where to declare it. Installs that really are only on an interactive shell rc PATH still warn.

## Checklist

* [x] Changes are documented

## Test plan

On `891669aa` (main):

```
bun test tests/unit/t294-config-diagnostics.test.ts                         # 29 pass
bun test tests/unit/t174-docs-legacy-refs-gate.test.ts                      # 3 pass
bun tests/run-tests.ts --unit --filter "<the 50 unit files mentioning doctor / diagnostics>"   # 50 / 50 PASS
bun tests/gen-coverage-registry.ts --check                                   # fresh, guards green, ratchet held
bun scripts/package.ts --check                                               # deterministic, 7 harnesses
```

The new case fails without the probe change (`git stash push -- core/ && bun test tests/unit/t294-config-diagnostics.test.ts -t "Linux baseline"`).

## Acknowledgment


apackeer (approved):
Thanks for tracking this down to glibc's `_CS_PATH` and for mirroring the macOS registry approach — the `systemRoot` seam made the probe genuinely testable, and the `/etc/environment` case from #1168 flips to `ok` exactly as described.

I reviewed the current head 9d1abf08 against main (merge-base 891669aa; merges cleanly onto current main 261083ce). Alignment: this is a bug fix restoring the `Runtime hook PATH` check's own contract (its fix text must name an action the probe can detect), following option 2 from the issue and the existing macOS `/etc/paths(.d)` precedent.

While reviewing I found a few places where the remediation still named a surface the probe couldn't see, so I pushed two focused commits on top of yours rather than round-tripping:

- `8510efce`: environment.d values written as systemd documents them (`PATH=$HOME/.local/bin:$PATH`) were dropped because every `$`-prefixed entry was discarded; the probe now expands `$HOME`/`${HOME}`, so `~/.local/bin` declared there is found. macOS received the Linux file list; the remediation is now per platform (`/etc/paths.d` on darwin) and the `bun` text shares the same surface list. login.defs `ENV_PATH` without the optional `PATH=` prefix is accepted, and pam_env's first-`#` comment termination is honoured. The new test asserts the user-visible outcome through `probeRuntime` (`found` / `interactive-only`, empty issues), and the User Guide paragraphs are updated.
- `9d1abf08`: unresolved environment.d variables are now discarded rather than blanked — blanking turned `/opt/$TOOLCHAIN/bin` into `/opt//bin` → `/opt/bin`, which an unrelated executable could satisfy. Only the `${PATH:+:…}` / `${PATH:+…:}` splice idioms are removed; a real-resolver negative regression covers it.

Validation on 9d1abf08: typecheck, lint, `bun scripts/package.ts --check` (deterministic, 7 harnesses), coverage registry, full t294 (31 pass), and CI/Markdownlint/Security all green; the docs gate and runtime probe tests also pass on the tree merged with current main.

Not changed, by design: the baseline is a union of declared sources, so a directory only some session types inherit can still read as `found` — the same approximation macOS already makes and the trade the issue proposed. Approving.

Issue #22, opened 2025-12-11
Feature Request: AI-DLC CLI Tool to manage aidlc project initialization, rules and context


ankitsri-aws:
AI-DLC workflows have proven successful for intelligent software development. This proposal introduces a complementary CLI tool (aidlc) that streamlines project initialization while preserving existing manual setup flexibility. The tool reduces setup time from 15-30 minutes to under 2 minutes for developers adopting AI-DLC workflows.

A lightweight CLI tool providing optional rapid project initialization alongside existing manual configuration for advanced use cases.

**Technical Requirements**

1. Cross-platform execution on Linux, macOS, and Windows (AMD64 & ARM64)
2. Zero external dependencies beyond Go standard library
3. Project initialization completes in under 30 seconds
4. Backward compatibility with existing manual setup procedures
5. Generated structures identical to manually configured projects

raj-jain-aws:
Agree with @dreamorosi comments. I think we'll get to an AI-DLC CLI as this expands, but at the moment this does not seem to be solving a real problem and would introduce more complexity.

PR #1194, opened 2026-09-15
fix: name the accepted values in review-path refusals (#1082)

### LINKED ISSUE #1082: [Bug]: several review-path refusals print the rejected value but not the valid values that would satisfy them
### Description

Several human-facing refusals across the review path reject an input by echoing the **offending value** but never printing **what would have been accepted** — neither the expected *form* nor the *current valid set*. The valid set is present in the same scope at each site (an enum predicate, the live findings list, the next legal ordinal), so printing it is a one-line change. The tightest instance is the review-finding status enum: `parseReviewSection` in `core/tools/aidlc-lib.ts` throws `invalid finding status "Fixed"` while the four legal statuses sit in `validReviewFindingStatus` a few lines above, unprinted. `Fixed` is the word a reviewer reaches for after a human has just fixed something, so this is a high-traffic mistake that a two-word addition to the message would end at the source.

The same anti-pattern recurs at three more sites (listed under Additional Context), each with a different accepted set to name. One of them (`#27`) sits on code that PR #833 rewrote for exactly this readability goal — the strings were made plainer, but the accepted set still is not named, so the class survived the touch.

### Steps to Reproduce

1. In a review artifact table, set a finding's **Status** cell to `Fixed` (or any value outside the enum).
2. Run any command that parses the review section — e.g. reach the reviewer-backed human gate for that stage, or invoke the disposition/brief path that calls `parseReviewSection`.
3. Observe the thrown error.

### Expected Behavior

The refusal names what would satisfy it. For the status enum:

```
core/user-stories/…#R-02: invalid finding status "Fixed".
Valid statuses: New, Unresolved, Resolved, Accepted risk, or "Rejected: <reason>".
```

The valid set is already computable at the throw site (`validReviewFindingStatus`), so the reader fixes the typo in one step instead of opening the source to reverse-engineer the enum.

### Actual Behavior

`core/tools/aidlc-lib.ts:10653-10656` throws with the rejected value only:

```ts
if (!validReviewFindingStatus(status)) {
  throw new Error(
    `${artifact}#${id}: invalid finding status ${JSON.stringify(status)}`,
  );
}
```

The accepted values live immediately above at `core/tools/aidlc-lib.ts:10585-10592` (`New`, `Unresolved`, `Resolved`, `Accepted risk`, `Rejected: <reason>`) and are never surfaced. The reader is told what is wrong, not what is right.

### AI-DLC Version

### Phase
Construction (review / gate path; also reached during Inception summary confirmation — see #22 below)

### Harness / Model
Reproduced originally on Claude Code; defect is harness-neutral (`core/`).

### Additional Context

Same anti-pattern, three more sites — related in shape, though the accepted set and the exact fix differ at each:

- **#27 — `--reject-finding` rejects a selector without listing the current selectors.** `core/tools/aidlc-review-brief.ts:380-383` throws `Cannot reject <artifact>#<id>: it is not a current review finding for this gate.` The full set of current findings is in hand at that point (`findings`, built just above at `:359-365`). A stem-vs-full-path mismatch in the identifier is invisible until the accepted identifiers are printed. Fix: list the current `artifact#id` selectors. (This is the closest sibling to #30 — an accepted set sitting in local scope, unprinted.)
- **#22 — summary-confirmation refusals name the file but not the offending sub-section.** `core/tools/aidlc-lib.ts:8500-8501` instructs the reader to *"remove or repair every invalid or duplicate post-summary section named by the validation error"* — but the `SUMMARY_CONTENT_STALE` (`:8597-8599`) and `SUMMARY_HASH_SCOPE_INVALID` (`:8536-8539`) refusals that carry this recovery text **do name the relevant file** (`${question.path}` / the offending `Hash Scope` value) — they just do not name the specific section, heading, or line that changed. `doctor` was clean at the time; the exact construct had to be reasoned out of the digest-scope spec. Softer than the others (the file and scope value ARE printed); fix: also name the offending heading/line and what about it broke the digest.
- **#31 — `--why stale` leads with a hard-coded "changed upstream" one-liner.** `core/tools/aidlc-review-brief.ts:824-828` maps `stale → "Re-check required after upstream work changed."` and prints it verbatim in the human-facing `**Why now:**` line (`:833`). When the re-check is triggered by a *conductor edit that self-invalidated the receipt* (nothing upstream changed), that opening one-liner is misleading. It is **not** the whole story, though: the `stale` branch below (`:847-873`) appends the real invalidation details from `reviewInvalidationDetails` (`**Changed upstream:**`, `**Downstream artifacts requiring re-check:**`, `**Downstream reviews requiring re-check:**`), so the accurate cause IS present under the misleading lead. Fix: soften the fixed `stale` string (or derive its wording from the invalidation details) so `**Why now:**` does not lead with a cause that may not apply. (The `--why` validator at `:945-946` correctly prints `<first|revision|stale>` — the defect is the *content* of the `stale` lead line, not a missing valid-values list.)

**Why group these, and why #30 leads:** the sites share a shape — a refusal that echoes the rejected input while an accepted set (or a more accurate cause) sits in local scope — but the concrete fix differs per site: #30 and #27 print an accepted set already in hand; #22 adds a sub-section locator to a message that already names the file; #31 rewords a misleading lead line whose accurate detail is already appended below it. #30 is the cleanest single-line repro and the honest anchor; the rest are variations, not copies of one fix.

**Prior touch that missed the class:** `core/tools/aidlc-review-brief.ts` (#27, #31) was rewritten by **#833** ("actionable Review brief at every reviewer-backed human gate") — the strings were made plainer without naming the accepted set / correcting the misleading lead, so the class survived the touch.

**Proposed fix (PR offered against `main`):** address each site with the change it needs — append the status enum from `validReviewFindingStatus` (#30); list the current `artifact#id` selectors (#27); add the offending heading/line to the summary refusals (#22); soften/derive the `stale` lead line (#31). Small and localized, with a regression test per site. Happy to open it; #30 alone is a clean first commit if you'd prefer to land the pattern incrementally.

vivibui:
Addresses #1082. The remaining `SUMMARY_CONTENT_STALE` locator work stays tracked in that issue.

## What

Four refusals on the review path echoed the rejected value without naming the
accepted form. Each now names it.

| Site | Before | After |
|---|---|---|
| finding status enum (`aidlc-lib.ts`) | `invalid finding status "Fixed"` | adds `Valid statuses: New, Unresolved, Resolved, Accepted risk, or "Rejected: <reason>".` |
| `--reject-finding` selector (`aidlc-review-brief.ts`) | `…it is not a current review finding for this gate.` | adds `Current rejectable findings: <artifact>#<id>, …` |
| summary-confirmation hash scope (`aidlc-lib.ts`) | names the rejected scope only | adds `Supported: "confirmed-content-v1".` |
| `--why stale` lead line (`aidlc-review-brief.ts`) | `Re-check required after upstream work changed.` | `Re-check required: the previous review receipt is no longer valid.` |

The fixed statuses now live in one tuple that derives the TypeScript union, runtime
predicate, and rendered refusal. The `Rejected: ` prefix is also shared by the type,
validator, and displayed form. Tests cover the fixed values and the rejected-reason
format.

The `stale` change corrects the lead line. That reason also covers a conductor edit
that self-invalidated its own receipt when nothing upstream changed, so the old line
asserted the wrong cause for that path. The accurate detail was
already appended below it as `**Changed upstream:**` / `**Downstream …:**`, which is
why the fix is to stop leading with the guess.

## Not included, and why

The issue's site #22 has two halves. The hash-scope half is fixed above. The other —
`SUMMARY_CONTENT_STALE` naming *which* section changed — is **not** done, because it is
not a message change: the receipt stores a single opaque SHA over
`confirmed-content-v1` with no per-section hashes and no prior content, so the changed
heading is not derivable at the throw. Locating it needs a receipt-format change
(per-section hashes or retained content), which requires a separate PR. This PR keeps
#1082 open for that remaining work.

## Verification

- `bun run check` — packaging deterministic across all seven harnesses, three typecheck
  projects, lint over 775 files: clean. Coverage registry regenerated; `--check` green,
  ratchet held.
- Behavioural coverage for all four changed sites, each asserted through its emitting
  code path:
  - status source agreement + rejected-reason validation — `tests/unit/t342-refusal-valid-values.test.ts` (5 tests)
  - `--reject-finding` selector list — `tests/unit/t304-review-brief.test.ts`, including
    mixed resolved and open findings so only rejectable selectors are advertised
  - `stale` lead line — `tests/unit/t304-review-brief.test.ts`, on a rendered brief
  - hash scope — `tests/integration/t185-stage-artifact-guard.test.ts`, through the CLI
- t342 negative-controlled: reverting the status message fails it.
- t304 33/33, t342 5/5, t185's hash-scope case green.

`t185` shows 16 pre-existing failures in this environment; **identical on clean `main`**
(94 pass / 16 fail both sides), so they are not from this change. Verified on macOS;
this machine cannot complete a full local tier, so CI is the arbiter for the whole
suite.

---


leandrodamascena (approved):
Addressed the follow-up findings and pushed `675f3a390`.

Changes:
- `--reject-finding` now lists only `New` and `Unresolved` findings as valid selectors;
- the fixed status tuple now drives the TypeScript type, runtime predicate, and rendered refusal message;
- tests cover mixed resolved/open findings and rejected-status reason validation;
- merged current `main` without conflicts;
- changed the PR description from `Closes #1082` to `Addresses #1082`, keeping the remaining `SUMMARY_CONTENT_STALE` locator work open.

Validation:
- `bun run check` passes;
- 38 focused review tests pass;
- the targeted hash-scope integration test passes;
- the coverage registry guard passes.

The PR is mergeable and the PR validation checks are green. Full CI has not appeared for the new SHA yet. Approved based on the completed local validation and current PR checks.

leandrodamascena (approved):
Thanks @vivibui approved!

PR #1195, opened 2026-09-15
fix: refuse orphaned positionals at intent-create instead of storing a truncated description (#1114)

### LINKED ISSUE #1114: [Bug]: intent-create silently truncates --arguments=<value> at the first space, losing part of the project description permanently
### Description

Passing the project description in the attached form `--arguments=<value>` to
`aidlc engine intent create` stores only the text before the value's first ASCII space.
The remainder is discarded with no error and no warning, and because
project-description.json is written once at intent creation there is no supported way to
repair it afterwards.

In my case a 160-character description was stored as its first 115 characters, cutting off
three of the five things I had asked for. The truncation point is exactly the single ASCII
space the description contained.

The loss is not cosmetic. project-description.json is the authoritative initial request:
`aidlc engine workspace project-description` returns it verbatim, the intent-capture stage
uses it as the `[desc]` source register, and the claim-sources sensor derives `[desc]` from
it. So every requirement past the first space is absent from the workflow's source of truth
for the rest of the run.

### Steps to Reproduce

1. Create an intent passing a description that contains at least one ASCII space, using the
   attached form:

   aidlc engine intent create --scope <scope> \
     '--arguments=deploy this to AWS and also tidy the directory layout' \
     --label demo

2. Read the stored description:

   aidlc engine workspace project-description

3. Observe that only "deploy" was stored.

### Expected Behavior

The whole value is stored, as it is when the same text is passed positionally or after the
`--` delimiter. Alternatively, the attached form is rejected with a clear error rather than
silently keeping a prefix.

### Actual Behavior

Only the text before the first ASCII space is stored in
<record>/project-description.json. No error, no warning, exit code 0. The `Project` preview
field in aidlc-state.md and the WORKFLOW_STARTED audit `Request` field carry the same
truncated value, so nothing in the record reveals that anything was lost.

Observed value in my run (115 of 160 characters, cut immediately before the space in
"media convert"):

  "今までこのローカル端末で動くようにシステムを作っていたが、社内のみんなが使えるようにAWSにデプロイできるように変更したい。
   アプリケーションはamplifyを使うので、アプリは書き換えてもいい。動画のエンコードはawsのmedia"

The lost tail contained three separate requirements (use MediaConvert for encoding,
reorganise the directory layout, produce documentation).

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.8.1 (native install)

### AI-DLC Phase

Initialization

### Harness

Kiro IDE

### AI Model

Claude Opus 5

### Environment

- OS: Windows (win32), PowerShell 5.1
- aidlc: 2.8.1 (runtime 2.8.1), native install; command resolves to
  %LOCALAPPDATA%\aidlc\bin\aidlc.cmd
- Kiro IDE 1.0.437

### Additional Context

The shell is not responsible. I measured the argument-passing path with a script that dumps
process.argv, over the real invocation route:

  PowerShell -> exe directly,   '--arguments=alpha beta gamma'
    -> ["--arguments=alpha beta gamma"]              (intact)

  PowerShell -> .cmd shim,      '--arguments=alpha beta gamma'
    -> ["--arguments=alpha beta gamma"]              (intact)

  PowerShell -> .cmd shim,      unquoted
    -> ["--arguments=alpha","beta","gamma"]          (splits, as expected)

So a single-quoted value with spaces reaches the executable whole through both PowerShell
and the Windows .cmd wrapper.

Corroborating evidence from the same session: the identical description passed positionally
and after the `--` delimiter reached the engine complete - the full text, including every
clause after the space, appeared in the engine's own print directives for
`orchestrate next compose '<text>'` and `next --scope <name> -- '<text>'`. Only the
`--arguments=<value>` call truncated it.

What I could not isolate: where inside the engine the value is cut. I did not want to create
throwaway intents in a real workspace to bisect it, so I am reporting the observable
behaviour and the ruled-out cause rather than naming a line. The separated form
`--arguments "<description>"` shown in `intent create --help` is the obvious thing to compare
against.

Suggested direction, in order of value:
1. Do not lose the tail silently. Even if the attached form is unsupported, storing a prefix
   without a diagnostic is the damaging part - the run continues on an incomplete request and
   nobody notices until a later stage reads it.
2. Accept the attached form, since `--arguments=` is a natural spelling and the docs' example
   for the same flag is the separated form only.
3. Consider a repair path. project-description.json is fixed at creation and there is no
   verb to correct it, so a truncated description currently means either living with a wrong
   record or recreating the intent.

Workaround for anyone hitting this: use the separated form,
`--arguments "<description>"`, and verify with
`aidlc engine workspace project-description` before proceeding.

vivibui:
Closes #1114.

## Diagnosis

On current `main`, both parsing layers preserve an attached value when it arrives
as one argv token:

- `parseArgs` round-trips `--arguments=deploy this to AWS` intact.
- `shellArg` in `aidlc-orchestrate.ts` quotes values containing spaces.

The reproducible loss occurs earlier, when an unquoted value is split into
`--arguments=deploy` plus orphaned positional words. `intent-create` previously
ignored those words, stored only `deploy`, and exited successfully.

## Fix

The two supported creation forms, `intent-create` and `intent create`, now refuse
positional arguments. The error names the orphaned words, the prefix that would
have been stored, and the quoting remedy:

```text
intent-create does not accept positional arguments, but received "this", "to",
"AWS". This usually means an unquoted --arguments=... was split by the shell:
only "deploy" would have been kept. Quote the whole value, e.g.
--arguments="<full description>".
```

The refusal happens before creation, so no partial intent is minted.

The two-token alias counts both `intent` and `create` as verb tokens. The retired
`init` command is deliberately excluded from the positional guard and retains its
dedicated transition refusal.

## Why refusal is required

`project-description.json` is written once at creation, returned verbatim by
`workspace project-description`, used as the `[desc]` source by Intent Capture,
and consumed by the claim-sources sensor. Silently storing a prefix corrupts the
workflow's source description with no repair path.

## Verification

- `bun run check` — deterministic packaging across all seven harnesses,
  typecheck, and lint passed.
- `bun tests/gen-coverage-registry.ts --check` — fresh; guards and ratchet passed.
- `t343-intent-create-positionals.test.ts` — 7/7, covering split input,
  attached and detached intact values, both creation forms, alias orphans,
  generic stray positionals, and the retired `init` refusal.
- Regression sweep (`t118`, `t162`, `t297`, `t343`) — 65/65.

---


leandrodamascena (approved):
Looks good to me. The guard correctly refuses split positional arguments before creating the intent, covers both command forms, and preserves the legacy `init` error.

One follow-up: I could not reproduce the exact quoted PowerShell case from #1114. A single attached argument works on v2.8.1, the PR base, and this branch. This PR fixes the unquoted shell-split case, so I think we should keep #1114 open until the reporter confirms the Windows behavior.

vivibui:
Sounds good!

PR #1199, opened 2026-09-16
feat(config): allow non-colliding harnesses to coexist in one project

### LINKED ISSUE #1196: [Bug]: aidlc config refuses to add a second harness — the documented "multi-harness project" state is unreachable through the supported path
## Description

The docs explicitly describe a project holding more than one harness tree at once ("multi-harness project" / "more than one harness tree is present") as a real, supported state, and doctor even specifies behavior for it ("selects the invoking harness"). Yet `aidlc config --harness <Y>` refuses with `project uses <X>; refusing <Y>` whenever a project already has a different harness `X` — even for distinct, non-colliding directories such as `.kiro/` and `.claude/`.

As a result, the multi-harness state the docs acknowledge is **unreachable through the supported path (`aidlc config`)**. doctor is built to handle multiple harnesses, but the only entry point to that state is closed — an internal inconsistency. No doc states a "one harness per project" constraint.

## Steps to Reproduce

1. Configure a project with Kiro: `aidlc config --harness kiro` (creates `.kiro/`).
2. Try to add Claude Code to the same project: `aidlc config --harness claude`.
3. It is refused:

```
$ aidlc config --harness claude
error: project uses kiro; refusing claude
fix: aidlc config --harness <name>
```

The same refusal occurs for any differing harness (e.g. `--harness kiro-ide`).

## Expected Behavior

Because `.claude/` and `.kiro/` are distinct, non-colliding directories, an intentional addition should project `.claude/` alongside the existing `.kiro/` so both harnesses coexist — consistent with the docs' "multi-harness project". At minimum, a supported path to reach the documented state should exist.

## Actual Behavior

Requesting a different `--harness` on an already-configured project is refused unconditionally with `project uses X; refusing Y` even when directories do not collide, and no new harness directory is created.

## AI-DLC Version

v2 (current on main)

## Release / Commit

2.9.0 / `a8c477c`

## AI-DLC Phase

Initialization

## Harness

Kiro CLI, Claude Code (adding `claude` to a `kiro`-configured project)

## Environment

- OS: macOS
- Running from source (commit `a8c477c`, version 2.9.0)

## Additional Context

- **Where the inconsistency lives:**
  - Docs acknowledge multi-harness: `docs/reference/01-architecture.md` (~L430, doctor selects the invoking harness in a "multi-harness project"); `docs/guide/18-install-and-lifecycle.md` (~L563-566, the row follows the invoking harness "when more than one harness tree is present").
  - `discoverProjectHarnesses()` in `core/tools/aidlc-runtime-paths.ts` already returns multiple harnesses via `HARNESS_PRECEDENCE`.
  - But `existingProject()` in `core/tools/aidlc-init.ts` (~L5199-5211) refuses the second addition. The same `project uses X; refusing Y` message also appears at ~L1298 / L4131 / L5780 / L6394. The guard was introduced in #756.
- **Healthy invariant a fix must preserve:** `tests/unit/t243-install-mechanism.test.ts` — "an explicit installed harness never silently yields to the existing project harness." An explicit `--harness Y` must not be silently redirected to existing `X`; that is correct. The bug is that the *mechanism* for that invariant is too broad and also blocks an **intentional addition**.
- **Not a one-line fix (care needed):** `kiro` and `kiro-ide` both project into `.kiro/`, sharing one directory, so they cannot coexist as separate trees the way `.kiro/` + `.claude/` can. The checkout-wide shared harness-native rule include (Claude `@`-import, Kiro resources/steering, etc.) has a contention that `docs/guide/03-spaces-and-intents.md` flags as future work.
- Related: #984 (no clean uninstall because AI-DLC shares `.kiro/`/`.claude/` — same shared-directory root cause), #1047 (native per-project config).

fsatsuki:
# Summary

Allow more than one AI-DLC harness to coexist in a single project when their engine directories do not collide (for example `.claude` alongside `.kiro`),
while still refusing pairs that share a directory and preserving the guard that an explicit `--harness` never silently yields to the existing project harness.

Closes #1196.

## Changes

`core/tools/aidlc-init.ts`:

- `existingProject` no longer refuses when an explicit `--harness` names a distribution the project does not yet have. It returns "no existing match" so the flow scaffolds the requested harness in its own directory, treating it as an intentional addition rather than a mistaken refresh.
- The main scaffold path now guards the real hazard directly: when adding a harness the project did not have, it refuses only if the new engine directory collides with an already-present harness. `kiro`/`kiro-ide` (both `.kiro`) and
  `opencode`/`copilot` (both `.aidlc`) are refused with a clear message (`harness <Y> shares directory <dir> with installed <X>; they cannot coexist in one project`).
- `planRootIntegrations` treats a shared root managed block (`.gitignore`, `AGENTS.md`) written by a sibling harness as a first-writer contribution: instead of failing with "managed block has no ownership baseline", it preserves the existing block and records it as owned, so a second harness does not fight over the shared bytes.

The silent-yield invariant is unchanged: a resolved source whose distribution disagrees with a matched existing harness is still refused (`project uses X; refusing Y`).

## User experience

Before: on a project already configured for one harness, `aidlc config --harness <other>` always failed with `project uses <X>; refusing <other>`, even for a non-colliding directory. There was no supported way to reach the "multi-harness project" state the docs already describe.

After: adding a non-colliding harness succeeds and both engine directories coexist, sharing the neutral `aidlc/` workspace and a single AI-DLC block in `.gitignore`/`AGENTS.md`. Adding a harness that shares a directory
(`kiro`+`kiro-ide`, `opencode`+`copilot`) is refused with an explicit, actionable message.

## Checklist

* [ ] Changes are documented

## Test plan

Automated (all green):

- `bun test tests/unit/t243-install-mechanism.test.ts` — 87 pass / 0 fail,   including the new/updated cases:
  - a non-colliding harness is added alongside an existing one (both dirs present, a single `# BEGIN AI-DLC:gitignore` block)
  - a harness that shares an engine directory cannot coexist (refused)
  - a refresh source that disagrees with the sole project harness is refused(silent-yield guard)
  - an explicit installed harness never silently yields (added opencode is stamped `opencode`, `.claude` untouched)
- `bun test tests/unit/t230-dispatcher-routes.test.ts` — 121 pass / 0 fail (harness discovery unaffected).

Manual, via a locally built native `aidlc` (build binary → package release → `install.sh --offline --from`):

- `aidlc config --harness kiro` then `aidlc config --harness claude` on the same project → both succeed; `.kiro` and `.claude` coexist; the shared `.gitignore`/`AGENTS.md` block stays single (claude's `.gitignore` action is `preserve, shared with another harness`).
- `aidlc config --harness kiro-ide` on a `.kiro` project → exit 4 with the shared-directory message.

Known limitation (out of scope, noted on the issue): a project whose `.gitignore` carries the AI-DLC content WITHOUT the managed-block markers (an older copy-channel install) still hits the pre-existing `legacy root integration ambiguous` guard when a second harness is added.

## Acknowledgment


apackeer (approved):
Thanks @fsatsuki — the engine-directory collision guard is exactly right, and opening a supported path to the multi-harness state the docs already describe (#1196) is the correct direction. Your t243 cases gave a solid base to extend.

I reviewed the current head 6af267fc7 against `main` (`7998589` at PR time; `main` moved since and CI ran on the merge). Alignment basis: #1196 plus the architecture/install docs that already define the multi-harness state.

**What I found and fixed** (three follow-up commits, your commit untouched):

- The shared-block rule broke one refresh later: adding claude preserved kiro's `.gitignore` block, but claude's next refresh silently rewrote it because the sibling's hash had been recorded as claude's own baseline; kiro then failed with `managed block was locally modified` and `--force` just flipped it back. The same branch never checked that a sibling existed, so a stale unowned block was adopted instead of refused.
- `.gitignore` is now declared `shared: "union"` by every harness. The packager ships each harness's block at `tools/data/root-blocks/gitignore`, and `aidlc config` writes one deterministic union of every installed harness's entries (extra lines under `# <harness> harness`), so both install orders converge on the same bytes and the second harness's ignores actually land. Adding a harness combines a sibling-owned block when the sibling's copy is available (`merge (combined with kiro)`); an older install keeps ownership (`preserve (owned by kiro)`) until it refreshes; a sibling that declares the union but has lost its copy produces a named conflict instead of being silently dropped. `managed block has no ownership baseline` fails closed again.
- `AGENTS.md` is the added CLI's own instruction file (kiro → codex left Codex reading the Kiro onboarding), so an addition sharing an exclusive managed block is refused before planning, based on the sibling's `aidlc-projection.json` rather than its baseline — `--force` and a missing `aidlc-manifest.json` cannot bypass it. Coexistence is therefore Claude Code + one other harness until we have a harness-neutral `AGENTS.md`.
- `aidlc config` without `--harness` on a multi-harness project now asks for `--harness` (exit 2), like the interactive walk.
- Eight t243 regression tests, troubleshooting rows, install-guide and porting-guide docs.

**Validation:** t243 95/95, t304, t230, `bun scripts/package.ts --check` deterministic across all seven harnesses, typecheck and lint clean; CI 23/23 on 6af267fc7. The earlier `t311` failure never touched the changed code and has not recurred.

Approving; squash-merging.

PR #1201, opened 2026-09-16
fix: match record-root paths case-insensitively on Windows so Kiro IDE artifact writes are audited

apackeer:
# Summary

On Windows, Kiro IDE (VS Code `Uri.fsPath`) reports a written file with a lower-case drive letter (`c:\...`) while the project directory carries `C:\...`. The write-audit PostToolUse hook checked "is this write under the record root" with an exact string-prefix compare. It judged every artifact write to be outside the record and exited without emitting `ARTIFACT_CREATED` or `ARTIFACT_UPDATED`. Every artifact-producing stage then blocked at its gate with `SUMMARY_ARTIFACT_UNAUTHORIZED`, while `aidlc doctor` stayed green because the drop is an advisory fail-open.

## Changes

- `core/tools/aidlc-lib.ts`: new exported `normalizeDriveLetter(p)` upper-cases a leading `<letter>:` and leaves every other character unchanged. The drive letter is the one path component Windows never compares case-sensitively (per-directory case sensitivity applies to names, not the volume designator), so folding it cannot conflate two distinct directories. A POSIX absolute path never starts with `<letter>:`, so the function is a no-op there. It has no platform parameter and no environment override.
- `core/hooks/aidlc-write-audit-log.ts`: the file path, record root, and codekb root are drive-letter-normalized before the existing exact prefix checks. The `File` field carries the normalized spelling, so it matches the project-derived artifact path exactly.
- `redactProjectDirPrefix` accepts both drive-letter spellings of the project dir (nothing else is folded), so the recorded `File` always redacts to `<project-dir>/...` whichever drive case the configured project dir carries.
- `core/hooks/aidlc-run-sensors.ts`: the sensors-dir recursion guard had the same exact-prefix shape. It now normalizes the drive letter on both sides.
- `docs/reference/06-hooks-and-tools.md`: one sentence on the drive-letter comparison.
- Tests: `t348` covers the helper on every host, plus drive-letter redaction on Windows CI. `t07` adds Windows-CI cases (a drive letter that differs from an upper- or lower-case project dir logs Write/Edit with the record breadcrumb, `File: <project-dir>/...`, and `Summary Authorization Id`, and a codekb write keeps its breadcrumb) and an every-host case showing that a case-distinct sibling of the record root is neither logged nor authorized. `t94` adds a Windows-CI recursion-guard case. `t318` simulates process ancestry so its session-conflict cases are reachable on Windows.
- The coverage registry is regenerated for the new export.

Earlier revisions of this PR tried case-insensitive containment and then realpath canonicalization plus Unit-name aliasing. This revision drops both. Summary-scope lookup, review-freeze, receipt matching, and the #863 moved-workspace suffix fallback are unchanged from `main`.

## User experience

Before: on Windows with Kiro IDE, stage artifact writes produced no `ARTIFACT_*` audit rows, and the summary-confirmation gate reported `SUMMARY_ARTIFACT_UNAUTHORIZED` for every artifact-producing stage.

After: those writes are audited with the correct record-relative `Context`, and the gate passes. No behavior change on Linux or macOS.

## Checklist

* [x] Changes are documented

## Testing performed

On macOS: `bun scripts/package.ts --check` (deterministic), `tests/gen-coverage-registry.ts --check` (OK), typecheck, and lint all passed. t07, t348, t94, t318, t332, t264, t310, and t-review-summary-input-boundary all passed; the Windows-only cases were skipped as expected. t305 had one failure (`a post-authority branch cleanup failure is idempotently reconcilable`), which fails identically on unmodified `main` on the same machine and does not touch the changed code. The Windows-only cases run in the Windows CI job.

## Acknowledgment


leandrodamascena:
/aida accept F7 its a false positive.

PR #1204, opened 2026-09-16
fix(kiro-ide): let approved Code Generation delegate instead of refusing it as an opaque mutation (#1175)

### LINKED ISSUE #1175: [Bug]: Kiro IDE Plan Approval guard blocks all subagent dispatch, making mode: subagent Code Generation unrunnable
### Description

On the `kiro-ide` harness, the Plan Approval PreToolUse guard rejects every tool call whose payload carries no extractable file path and is not a shell tool, for as long as a Code Generation window is active. Kiro IDE's delegation tool (`invoke_sub_agent`) has exactly that payload shape, so Code Generation — which ships as `mode: subagent` with `lead_agent: aidlc-developer-agent` — can never delegate on this harness.

The adapter *does* contain dedicated handling for `invoke_sub_agent` (it translates the call into a synthetic `Task` payload for the core guard), but that code sits below the rejecting `return`, so it is unreachable in the only situation where delegation happens.

The rejection does not consult approval state: a valid, human-owned `PLAN_APPROVAL_RECORDED` receipt does not unblock it.

### Root cause

All references are to `~/.local/share/aidlc/versions/2.8.2/runtime/kiro-ide/.kiro/hooks/aidlc-kiro-adapter.ts` (byte-identical to the project's `.kiro/hooks/aidlc-kiro-adapter.ts`, verified with `diff`), inside `case "plan-approval-guard"`.

| Line | Code | Effect on `invoke_sub_agent` |
|---|---|---|
| 171 | `PLAN_APPROVAL_SAFE_READ_TOOLS` | not listed |
| 1136 | `mutationCapableTool()` — anything not in that set | `true` |
| 1139 | `inputPaths()` reads only `path`, `file_path`, `filePath`, `paths[]`, `operations[].path` | args are `name` + `prompt` → 0 paths |
| 1337 | `opaqueMutation = mutationCapable && (no args \|\| (!shell && paths.length === 0))` | `true` |
| 1571 | `if (state.active && Object.keys(toolArgs).length > 0 && !isKiroShellTool(toolName))` → return block (1576) | **rejected; approval not consulted** |
| 1617-1657 | `invoke_sub_agent` → synthetic `Task` translation | **unreachable** |

Two intentions collide here. The comment above line 1571 states that an active Code Generation window should refuse unattributable mutations, which is a deliberate defence. But the delegation handling 30 lines below becomes dead code exactly when delegation is supposed to happen.

Channel detection is not involved: within this `case`, `ide.channel` only drives a host-marking side effect and does not gate the rejecting region.

### Evidence

- **Not specific to delegation.** Calling the unrelated `memory` tool (`command: list`; args present, no path, not a shell) returns the *identical* message. The rule is tool-agnostic.
- **Not specific to the developer agent or to writing.** Dispatching `aidlc-architecture-reviewer-agent` for a read-only review returns the same message.
- **Boundary behaves as the code implies.** `str_replace` (path present) and `execute_bash` (explicitly exempted via `isKiroShellTool`) both pass while the same window is active.
- **Approval is genuinely recorded.** The receipt under `aidlc/.aidlc-sessions/plan-approval/` carries `"choice": "Approve Plan"`, `"status": "generation"`.

### Suggested fixes

Any one of these should restore delegation:

1. Move the `invoke_sub_agent` / `subagent_<agent>` translation (1617-1657) above the `opaqueMutation` computation at 1337, so recognized dispatch shapes are attributed before the opaque check runs.
2. Exempt recognized dispatch tool names from `opaqueMutation` so they fall through to that translation.
3. Add a `matcher` to `.kiro/hooks/aidlc-plan-approval-guard.json`, limiting the hook to write, shell, and dispatch tools. It currently ships with no matcher, so it fires for every tool, including ones it has no correct verdict for.

(3) is the smallest change; (1) or (2) address the ordering itself. Additionally, consider whether line 1571 should consult `state.approved` at all — as written, a fully approved stage is still refused.

### Steps to Reproduce

1. `aidlc config --harness kiro-ide` in an empty git repository
2. `/aidlc --scope express <any project description>`
3. Complete Requirements Analysis and approve it
4. At Code Generation, write the plan and unit-test instructions, record the fingerprint, and obtain a Plan Approval receipt (`{"emitted":"PLAN_APPROVAL_RECORDED",...}`)
5. Follow the stage's Step 4 and delegate to `aidlc-developer-agent` via `invoke_sub_agent`

### Expected Behavior

The dispatch reaches the `invoke_sub_agent` handling in the adapter, is translated into a `Task` payload, and is evaluated by the core Plan Approval guard against the recorded receipt.

### Actual Behavior

The dispatch is intercepted by PreToolUse with:

```
```

Code Generation cannot be delegated at all. The workaround is the `Subagent failure recovery` path in `stage-protocol-ensemble.md` §11: abandon delegation and run the stage inline in the conductor's own session. Path-bearing writes and shell commands are not blocked, so the approved plan can still be executed — but the shipped `mode: subagent` topology is unreachable.

Collateral impact: while a Code Generation window is active, every path-less non-shell tool is unusable, not just delegation (`memory` is one confirmed example). This silently reduces the conductor's capabilities mid-stage.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.8.2

### AI-DLC Phase

Construction

### Harness

Kiro IDE

### AI Model

Claude Opus 5

### Environment

- OS: macOS (darwin), zsh
- Harness: Kiro IDE (version not captured)
- Project: greenfield, `express` scope, zero-Unit Code Generation
- `aidlc --doctor`: 0 problems, 4 warnings (interactive-only `aidlc` on PATH, absent update cache, uncommitted `aidlc/` records, 1 plugin needing attention) — none related to this failure

### Additional Context

Four further problems surfaced in the same session. They are independent of the issue above and can be split into their own issues if preferred.

### A. Plan Approval receipt cannot be satisfied by Kiro IDE's numbered-prose answers

`skills/aidlc/question-rendering.md` mandates that Kiro IDE renders every structured question as numbered prose, and that the user answers with a number. But `aidlc engine log answer --checkpoint plan-approval` refused the recorded answer after the human replied `1`:

```
Refusing to record Plan Approval: Plan Approval requires the actual offered choice from this prompt and session
```

The challenge file has `"requireExactOptionLabels": false`, yet the receipt was only accepted after the human typed the literal string `Approve Plan`. Asking a user to retype an option label defeats the numbered-prose contract. The `human-turn` record appears to store only a timestamp, so it is unclear how the reply is matched.

### B. `aidlc engine review-brief summary` is not callable

`stage-protocol.md` §3 requires this command at the consolidated-summary checkpoint and says to print its output verbatim. It fails:

```
$ aidlc engine review-brief summary --stage requirements-analysis --questions-file <path>
{"error":"aidlc-review-brief.ts does not export main(argv)"}
```

`aidlc engine review-brief --help` also reports `unknown verb '--help' for engine noun 'review-brief'`. The checkpoint had to be presented without the mandated decision brief.

### C. `aidlc config --harness kiro-ide` does not return in a non-interactive shell

The command produced no output and did not exit within 120s when run non-interactively, although it had in fact completed the configuration (`.kiro/`, `aidlc/`, `AGENTS.md` were all created correctly). A non-interactive run should either finish or say what it is waiting for.

### D. `log answer` refuses feedback the human volunteered without being asked

At the consolidated-summary checkpoint the human chose `Request changes` *and* supplied the change in the same message. Recording that feedback was refused:

```
Cannot record this answer because no new human reply has arrived for the question. Wait for the human to type an answer, then try again.
```

The protocol's flow assumes feedback always arrives in a separate turn after `What should change?`. Re-asking a question the human has already answered is forbidden elsewhere in the same protocol, so the two rules conflict.

vivibui:
Closes #1175. **Workshop blocker** — without this, `mode: subagent` Code Generation
cannot run on `kiro-ide` at all.

## What was happening

While a Code Generation window is active, the adapter's `opaqueMutation` test treats
any mutation-capable tool with arguments, no file path, and not a shell as
unattributable. Kiro IDE's delegation tools carry `name` + `prompt` and no path, so
they matched — and Code Generation ships as `mode: subagent`.

The adapter already knows how to handle delegation: further down it translates the call
into a synthetic `Task` payload and forwards it to the core
`aidlc-plan-approval-guard.ts`, which consults approval state properly. That forward was
simply unreachable, because the opaque branch returned first.

Worth being precise about the shape of the bug, since it changes the fix: excluding
delegation only from the *refusal* inside that branch would be wrong too — the branch's
other exit is `return null`, i.e. **no mediation at all**, which is worse than blocking.
The fix has to keep delegation out of the opaque branch entirely so control reaches the
forward.

## The fix

One predicate plus one clause:

```ts
function isKiroDelegationTool(toolName: string): boolean {
  return toolName === "invoke_sub_agent" ||
    (toolName.startsWith("subagent_") && toolName !== "subagent_response");
}
```

…excluded in `opaqueMutation`. `subagent_response` stays out because it is the
completion shell rather than a dispatch — the same exclusion, for the same reason, that
the `SUBAGENT_COMPLETED` matcher already makes.

**This does not weaken the guard.** It moves the decision to the authority that can make
it. Approval state now decides, and the existing test
*"populated 1.x PreToolUse write and dispatch payloads are blocked **before**
approval"* still passes unchanged — because the core guard refuses an unapproved
dispatch just as the adapter used to.

## Verification

- t218 **86/86**, t265 (plan-approval guard) **55/55**, `bun run check` clean
  (deterministic across all seven harnesses, typecheck, lint).
- New assertion extends the existing *valid human approval* flow: once approval is
  recorded, both `invoke_sub_agent` and `subagent_aidlc-developer-agent` are allowed.
  Negative-controlled — removing the one clause returns exit 2 and fails the test.
- The assertion supplies the exact `AIDLC-TESTING-CONTRACT` hash on purpose. Without it
  the core guard refuses for a *different* reason, and the test would have passed for
  the wrong one. That is how I confirmed the refusal had genuinely moved from the
  adapter to the core guard rather than merely disappearing.

Verified on macOS; this machine cannot finish a full local tier, so CI is the arbiter
for the whole suite.

## Two things left deliberately out of scope

**The rule is tool-agnostic, and this fixes only delegation.** The reporter showed the
unrelated `memory` tool (`command: list` — args, no path, not a shell) drawing the
identical refusal, and `subagent_response` has the same shape. Each would be fixed by
adding it to `PLAN_APPROVAL_SAFE_READ_TOOLS`, but every addition to that set is a
security judgment about what may run during an approval window, so they belong in their
own change rather than riding along here.

**Coordination with #1157.** That refactor (serve Kiro IDE and Kiro CLI from one
distribution) deletes this whole region — `PLAN_APPROVAL_SAFE_READ_TOOLS`,
`isKiroShellTool`, `mutationCapableTool`, `opaqueMutation` — so this will conflict with
it. I raised it against `main` anyway because #1157 is 195 files at
`CHANGES_REQUESTED`, and this bug blocks a customer workshop now. Its added lines
mention neither `invoke_sub_agent` nor `opaqueMutation`, so I could not confirm the
relocated logic fixes the ordering; worth checking when it rebases. Happy to port this
into whatever shape #1157 lands with.

---


vivibui:
@wowzoo @leandrodamascena @apackeer @harmjeff Wonder if you can help me review this? Need for a WS next week

leandrodamascena (approved):
Thank you for fixing this workshop blocker with a focused change. I reviewed head `2e50346f` against current `main` at `87cc7f05`, including a clean local merge of the ten intervening commits.

No blocking findings.

The ordering fix is correct: recognized Kiro IDE delegation now bypasses only the adapter's opaque-mutation classification and still reaches the core Plan Approval guard. It does not directly authorize the dispatch. The existing negative case continues to block developer dispatch before approval, while the new coverage allows both `invoke_sub_agent` and `subagent_aidlc-developer-agent` only after the approval receipt and testing contract are valid. Other pathless mutation-capable tools remain blocked, read aliases remain available, and `subagent_response` remains excluded from dispatch handling.

Validation on the PR head:

- `bun scripts/package.ts`: passed.
- `t218` and `t265`: 141 passed, 0 failed.
- `git diff --check`: passed.

Validation after merging current `main` locally:

- Merge completed without conflicts.
- `bun run check`: passed, including deterministic packaging for all seven harnesses, typecheck, and lint.
- `t218` and `t265`: 147 passed, 0 failed.
- I also checked generic and named non-developer delegation after approval; both receive the same allowed verdict.

The branch is ten commits behind current `main`. The visible GitHub checks cover PR metadata only and were run against the older head/base combination. Repository automation has not evaluated the current base; the local merged validation above covers that compatibility gap.

PR #1205, opened 2026-09-16
fix: resolve repo identity from the main checkout so a git worktree finds its codekb

### LINKED ISSUE #1203: [Bug]: repo identity is derived from directory location, so a git worktree breaks the codekb gate and the repos workaround breaks Construction
## What happens

If the project directory is a git worktree, Reverse Engineering and the gate that approves it
disagree about where the code knowledge base belongs, and there is no supported setting that makes
them agree.

The stage writes its nine artifacts. The gate then reports that they do not exist, and the workflow
cannot move on. Two settings look like the fix, and neither is:

- `--repo <name>` puts the store in the right directory, but the gate does not accept that flag and
  keeps looking in the wrong one. Placing the store correctly is what makes it invisible.
- `repos` in `intents.json` makes the store and the gate agree — and breaks Construction, because
  the same name is also used as a directory to run git in, and that directory does not exist. No
  error names either cause.

The only thing that works is creating a symlink by hand inside the framework's own data directory,
pointing the worktree's name at the repository's.

If instead you let the store be written under the worktree's name, the gate passes, but the store
is now named after a branch: the next worktree of the same repository does not find it, re-analyses
the whole repository from scratch, and overwrites what the previous run learned.

This needs **one** worktree and **one** intent. Nothing concurrent, no second worktree, no parallel
sessions — the most ordinary use of `git worktree add` reproduces it.

## Impact on developers

- **One worktree per branch — a common setup — does not work out of the box.** Reverse Engineering
  cannot be approved without a manual symlink, and you create a new one for every worktree.
- **The workaround cannot be shared.** The symlink is named after a branch, so it is untracked and
  meaningless in anyone else's checkout. There is nothing to commit that fixes this for a
  teammate.
- **Nothing explains the failure.** The gate reports missing artifacts, not a name mismatch, so
  the first encounter is spent hunting for a file that was written successfully.
- **The knowledge base is rebuilt instead of reused.** Because the freshness check looks under the
  wrong name, it always answers "no current store", so every intent in a new worktree re-analyses
  the entire repository even when a current analysis is sitting there.
- **Earlier analysis is silently lost.** The cumulative-merge path has no baseline to merge
  against under the wrong name, so each rescan replaces the store rather than adding to it. Deep
  coverage produced by an earlier intent disappears with no warning.
- **The one documented lever makes things worse.** A developer who finds `repos` and sets it will
  fix the naming and break Construction, with no message connecting the two.

## Why it happens

AI-DLC has no notion of a repository's identity separate from its directory location. Two functions
conflate them: `codekbRepoName` names the store after `basename(projectDir)`, and `repoDir` turns a
repository name into `join(projectDir, repoName)` without checking that it exists. Mechanism and
line references below.

## Proposed fix

Two changes, both in `core/tools/aidlc-lib.ts`:

1. **`codekbRepoName`** — in the 0-recorded-repos case only, derive the repository name from the
   main worktree, `basename(dirname(resolve(<show-toplevel>, <git-common-dir>)))`, falling back to
   `basename(projectDir)` whenever git cannot answer. Leave the `repos.length === 1` branch and
   the 2-or-more fallback untouched. Backwards compatible by construction — see below.
2. **`resolveConstructionRepo`** — stop returning a `cwd` that was never checked to exist. Fail
   with an error naming the resolved path and both plausible causes (a wrong `repos` entry, or a
   lone-repo layout with no sibling directory).

The first unblocks worktree users; the second closes the trap the first removes the motive for.
They are independent, if you would rather take them separately.

## Mechanism 1 — identity from the directory name

`core/tools/aidlc-lib.ts:2091-2104`:

```ts
export function codekbRepoName(
  projectDir: string,
  space?: string,
  intent?: string,
): string {
  const selection = resolveWorkflowSelection(projectDir, { space, intent });
  const repos = intentRepos(projectDir, selection.intent ?? undefined, selection.space);
  return repos.length === 1 ? repos[0] : basename(projectDir);
}
```

For repository `myapp` with a worktree at `../myapp-feature-x`, identity becomes
`myapp-feature-x`.

Everything downstream routes through that one value, including the gate that must approve the
stage. `producesDirsForStage` (`core/tools/aidlc-state.ts:2721-2749`):

```ts
if (KNOWN_CODEKB_STAGES.has(stage.slug)) {
  const repos = intentRepos(pd);
  const resolved = repos.length > 0 ? repos : [codekbRepoName(pd)];
  return resolved.map((repo) => codekbDir(pd, repo));
}
```

and `producesArtifactsExist` (`:2762-2796`) stats that directory:

```ts
const dirs = producesDirsForStage(pd, stage);
if (KNOWN_CODEKB_STAGES.has(stage.slug)) {
  return dirs.length > 0 && dirs.every((dir) =>
    produces.every((name) => isRegularFile(join(dir, artifactFilename(name))))
  );
}
```

Writer and reader share the resolver, so they agree with each other and both disagree with the
repository. The store is named after a branch, and a second worktree of the same repository
resolves to a third name and rescans from scratch.

`--repo` does not resolve this. It exists on the `codekb-path` query
(`core/tools/aidlc-utility.ts:7093-7099`), so the store can be placed correctly — but
`producesDirsForStage` / `producesArtifactsExist` accept no override and run inside the transition
that must approve the stage. Placing the store correctly is exactly what makes it invisible to the
gate.

### The same wrong name also defeats store reuse and cumulative merge

Two more consumers resolve the store through `codekbRepoName`, which is why a worktree does not
merely mis-name its store but rebuilds it from scratch every time:

- `codekbStoreIsCurrent` (`core/tools/aidlc-lib.ts:2435-2450`) — the programmatic form of
  `codekb-scope-diff`'s CURRENT verdict, used by reuse receipts:

  ```ts
  const repo = requestedRepo ?? codekbRepoName(projectDir, sp);
  const timestamp = join(codekbDir(projectDir, repo, sp), "reverse-engineering-timestamp.md");
  if (!existsSync(timestamp)) return false;
  ```

  Under the wrong name the timestamp is absent, so the verdict is "not current" and the stage
  rescans even when a current store exists.

- `codekbStoreGeneration` (`:2427-2434`) returns the literal `"none"` for a directory that does
  not exist. That token is the compare-and-swap generation for **cumulative merges**, so a merge
  has no baseline to merge against and publishes a fresh store — replacing prior coverage instead
  of accumulating it.

So the reuse and cumulative-merge machinery is present and working; it is simply pointed at a path
that is never there. Correcting the resolver re-enables both without touching either.

## Mechanism 2 — the workaround resolves to a directory that does not exist

`core/tools/aidlc-lib.ts:14013-14016`:

```ts
// The on-disk dir for a sibling repo: an immediate child of the workspace root.
export function repoDir(projectDir: string, repoName: string): string {
  return join(projectDir, repoName);
}
```

`resolveConstructionRepo` (`:18531-18562`) uses it on both routes that carry a name:

```ts
// repos.length === 0 (legacy) AND an explicit --repo: honour it as a sibling
// anchor …
return { repo: requestedRepo, cwd: repoDir(projectDir, requestedRepo) };
…
if (repos.length === 1) {
  return { repo: repos[0], cwd: repoDir(projectDir, repos[0]) };
}
```

Only the *name* is validated (`isValidRepoName`); the resolved cwd never is. Consumers are
`core/tools/aidlc-swarm.ts:1735` and `core/tools/aidlc-worktree.ts:327`. The sibling-repo
assumption in that comment is right for a multi-repo workspace root and wrong whenever the project
root is itself the repository — and nothing distinguishes the two.

#663 is the precedent: the same resolution failing silently, accepted and fixed as a bug.

## Reproduce

```bash
git clone <repo> myapp && cd myapp
aidlc config --harness kiro          # tool path varies by harness

git worktree add ../myapp-feature-x -b feature-x
cd ../myapp-feature-x
# start an intent, reach Reverse Engineering

bun .kiro/tools/aidlc-utility.ts codekb-path
#   aidlc/spaces/default/codekb/myapp-feature-x     <- worktree directory name
bun .kiro/tools/aidlc-utility.ts codekb-path --repo myapp
#   aidlc/spaces/default/codekb/myapp               <- the repository
```

Then, in order:

- artifacts under `codekb/myapp` (correct) → the approval gate refuses; it stats
  `codekb/myapp-feature-x`.
- artifacts under `codekb/myapp-feature-x` → gate passes, but the store is named after a branch
  and the next worktree starts over.
- add `repos: ["myapp"]` to fix the name → Construction resolves cwd to `<worktree>/myapp`, which
  does not exist, with no error naming either cause.

## Why the identity change is backwards compatible

For any project root that is **not** a linked worktree, `dirname(--git-common-dir)` *is*
`projectDir`, so the derived name is byte-identical to `basename(projectDir)`. Where git is
unavailable or the root is not a repository, the fallback is today's behaviour. The only inputs
whose output changes are linked worktrees, which are broken today.

This matters because `tests/unit/t182-codekb-placement.test.ts` deliberately pins the
`0 recorded → basename` contract, and `t72`, `t183`, `t248`, `t-tui-t50-bugfix-scope` and
`t-exec-codex-journey-workspace` rely on it in their sandboxes. All keep passing under the rule
above. (Deriving identity from the `origin` remote instead would break them, which is why I am not
proposing that.)

The resolution itself is not new code: `core/tools/aidlc-worktree.ts:232-263` already does exactly
this, including the relative-`--git-common-dir` and `realpathSync` handling.

## What this does not cover

The main-worktree derivation fixes linked worktrees and nothing else. A **clone** whose directory
name differs from the repository still diverges, because for a clone the main worktree *is* the
clone — `dirname(--git-common-dir)` is the clone directory, so the derived name equals
`basename(projectDir)` and the behaviour is unchanged. Measured on two checkouts of the same
repositories:

| checkout | `basename(projectDir)` | main-worktree derivation | `origin` |
|---|---|---|---|
| clone at `…/aidlc-workflows-jstrunk` | `aidlc-workflows-jstrunk` | `aidlc-workflows-jstrunk` | `aidlc-workflows` |
| worktree at `…/proj-feature-x` | `proj-feature-x` | `proj` | `proj` |

Only the remote gives the canonical name in the clone case. I am **not** proposing that here,
because preferring the remote over the location is a much broader behaviour change: it alters
resolution for every checkout that has a remote whose name differs from its directory, and remote
names are mutable (a rename or a re-point would move a workspace's store) where a location is not.

So this is a question for the maintainers rather than something I want to decide in a bug fix: is
repository identity meant to be **derived from location** (this PR, minimal, worktree-correct),
**derived from the remote** (topology-independent, but mutable and a wider change), or **recorded
once** in the space/intent record and thereafter stable? If it is the third, `repos` is close to it
already and the second change below is the start of making it trustworthy.

The narrow fix is still worth landing on its own: it converts the reported failure from
"unworkable without a hand-made symlink" into "works", and it is a no-op for every checkout that is
not a linked worktree.

## Acceptance

- With the project root in a linked git worktree whose directory name differs from the
  repository's, `codekb-path` prints the repository-named directory with no `--repo`.
- Reverse-engineering writes there and its approval gate opens, with no symlink and no `repos`
  edit.
- A second worktree of the same repository resolves to the same store name.
- A worktree whose store is current **reuses** it instead of rescanning
  (`codekbStoreIsCurrent` returns true), and a rescan **merges** cumulatively against the existing
  store rather than replacing it (`codekbStoreGeneration` returns a real token, not `"none"`).
- A non-worktree checkout, and a project root that is not a git repository, resolve exactly as
  today (t182 unchanged).
- `resolveConstructionRepo` never returns a `cwd` that does not exist; a `repos` entry or `--repo`
  resolving to a missing directory errors with the resolved path, not a downstream git failure.
- Multi-repo workspaces with real sibling checkouts are unaffected.

## Related

- #567 — the other half of running AI-DLC from a worktree. Independent: different file, different
  fix, separate PR.
- #663 — precedent for `repoDir`-based resolution failing silently.
- #896, #514, #441 — other codekb bugs, all distinct mechanisms.

_Observed on v2.9.0, upstream `main` at `9288f80`. Files: `core/tools/aidlc-lib.ts`; symptoms
visible via `core/tools/aidlc-state.ts`, `core/tools/aidlc-utility.ts`, `core/tools/aidlc-swarm.ts`,
`core/tools/aidlc-worktree.ts`._

jstrunk:
# Summary

Fixes #1203.

Running AI-DLC from a git worktree makes Reverse Engineering and the gate that approves it disagree
about where the code knowledge base belongs, and no supported setting makes them agree. One worktree
and one intent is enough — nothing concurrent is involved.

Repository identity was derived from directory location in two places. This corrects both.

## Changes

**1. `codekbRepoName` resolves the repository name from the main checkout.**

With nothing recorded in `repos` it returned `basename(projectDir)`. `git worktree add` names its
directory after a branch, so a project root in a linked worktree keyed the codekb store on a branch
name. The store and the approval gate read the same resolver, so both agreed with each other and
both disagreed with the repository.

Now the 0-recorded case resolves `basename(dirname(--git-common-dir))`, falling back to
`basename(projectDir)` whenever git cannot answer. The 1-recorded and 2-or-more branches are
untouched.

Not derived from the `origin` remote. That would also cover a clone whose directory name differs
from the repository, but remote names are mutable and it would change resolution for every checkout
whose remote and directory names differ — a wider decision than this fix, and one I left as an open
question in the issue.

**2. `resolveConstructionRepo` verifies the resolved cwd.**

It turned a repo name into a directory with `repoDir` (`join(projectDir, repoName)`) and returned it
unchecked; only the name was validated. That is right for a multi-repo workspace and wrong when the
workspace root IS the repository, where no such child exists. Construction git operations then ran
against a missing directory with no message naming either cause.

Recording `repos` is what a developer reaches for when the store is named after a worktree, so this
path was reachable by following the only documented lever. It now throws a message naming the
resolved path. Both call sites already wrap the call in try/catch, so it surfaces as a clean CLI
error.

The two changes are independent and sit in separate commits if you would rather take one.

## User experience

**Before.** In a linked worktree, Reverse Engineering writes its nine artifacts and the gate reports
that they do not exist. `--repo` places the store correctly but the gate accepts no override, so
placing it correctly is what makes it invisible. `repos` makes the two agree and breaks Construction
against a directory that is not there. The only workaround that breaks nothing else is a hand-made
symlink under `aidlc/spaces/<space>/codekb/` — untracked, named after a branch, so it cannot be
committed for a teammate and is recreated per worktree. Because the freshness check also looked
under the wrong name, every intent in a new worktree re-analysed the whole repository, and each
rescan replaced prior coverage rather than adding to it.

**After.** The store and the gate resolve to the same repository-named directory with no flag, no
`repos` edit and no symlink. A second worktree of the same repository resolves to the same store, so
a current analysis is reused and a rescan merges into it. A recorded repo that does not exist on disk
fails immediately with the path in the message.

Unchanged for every project root that is not a linked worktree: there,
`dirname(--git-common-dir)` *is* `projectDir`, so the resolved name is byte-identical to today's.

## Checklist

* [x] Changes are documented

On the last item: no fingerprint, epoch or receipt identity gains an input. `codekbStoreGeneration`
and `codekbStoreIsCurrent` are unchanged; they simply now receive the directory that exists, so a
worktree stops reporting "no current store" and stops replacing prior coverage.

## Test plan

New coverage, both with existing owners so the generated coverage registry is unchanged:

* `tests/unit/t182-codekb-placement.test.ts` — a real `git worktree add` off the fixture repo, then
  `codekbRepoName` from inside the worktree. Asserts it resolves the repository name, that the main
  checkout is unchanged (the backwards-compatibility claim), and that both checkouts resolve to one
  store. Mutation-checked: reverting the resolver to `basename(projectDir)` fails 2 of these.
* `tests/integration/t166-multi-repo-construction.test.ts` — a recorded repo whose directory is
  absent dead-ends through the CLI with the resolved path and "does not exist" in the message.

Verified locally against `main` at `9288f80`:

* `bun scripts/package.ts` and `bun scripts/package.ts --check` — deterministic across two builds
  for all seven harnesses.
* `bun run typecheck` clean; `bun run lint` clean and identical to pristine (3 pre-existing infos).
* `bun test tests/unit/t182-codekb-placement.test.ts` — 16 pass.
* `bun test tests/integration/t166-multi-repo-construction.test.ts` — 38 pass / 4 fail, where the
  same 4 fail on a pristine checkout of `9288f80` (confirmed by re-running with these changes
  stashed). My addition is the +1 pass.
* Full unit tier — the only failures are 3 in `t100`, also confirmed failing on pristine `9288f80`.
* First commit verified independently: it typechecks and passes `t182` on its own, so neither commit
  is broken in isolation.

Not run here: the SDK-driven integration tests (`t183`, `t72`) need live model credentials and stop
with `Could not load credentials from any providers` in my environment. `t183` covers codekb
placement, so it is the one I would most want a CI run to confirm.

## Acknowledgment


wowzoo (dismissed):
Requesting changes on one P2. The worktree case this fixes does work — I verified it — but the new resolver returns a **different repository's** name when the project root is not the git toplevel, which is the same failure class the change sets out to remove.

Head `faaed4940`, recorded base `9288f80f4`. The default branch has moved to `d378d0df` since, so everything below was measured on `d378d0df` merged into this head, which merged with no conflicts.

## What works

`codekbRepoName` from a linked worktree root now returns the repository name rather than the branch-shaped directory name. Measured in-process on the merge result, with a real `git worktree add`:

| project root | before | after |
|---|---|---|
| linked worktree | `pr-1205` (the worktree directory) | **`aidlc-workflows-fork`** (the main checkout) |

Both halves are pinned by tests that fail without them. Reverting only `core/tools/aidlc-lib.ts` to `9288f80f4` and keeping the tests: **56 pass / 3 fail** — the two `t182` worktree cases plus `t166`'s `worktree create names the missing directory instead of failing inside it`, which is the `repoCwdOrThrow` half. Restored: **59 pass / 0 fail**, 245 expect(). `bun run check` is clean on the merge result. bun 1.4.0, macOS.

## P2 — the common dir is resolved against the wrong base, so a nested repository wins

`mainCheckoutRepoName` measures `--git-common-dir` with `cwd: projectDir` and then resolves it against `--show-toplevel`:

```ts
const name = basename(dirname(realpathSync(resolvePath(topPath, commonRaw))));
```

Git returns that path **relative to the directory the command ran in** — projectDir — not relative to the toplevel. The two are the same only when the project root *is* the toplevel. When they differ, the resolved path lands one or more levels too high; if a real git dir happens to sit there, the function returns that repository's name and nothing signals it.

Reproduction, two nested repositories, project root one level below the inner repository's root:

```
/tmp/probe/outer          git init          (outer repository)
/tmp/probe/outer/inner    git init          (inner repository)
/tmp/probe/outer/inner/sub                  (the project root)
```

From `…/inner/sub`: `--show-toplevel` is `…/outer/inner`, `--git-common-dir` is `../.git` — which is `…/inner/.git` relative to the cwd it was measured in. Resolved against the toplevel it becomes `…/outer/.git`, which exists, so:

| tree | `codekbRepoName("…/outer/inner/sub")` |
|---|---|
| `9288f80f4` (base) | `"sub"` |
| this head | **`"outer"`** |
| correct | `"inner"` |

I called the shipped function directly on the merge result for that row, not just the path arithmetic.

The consequence is the one the change exists to prevent: the store is keyed on a name that does not identify this repository, so two different repositories can collide on one store, and a store written under the old key becomes invisible. It is silent — there is no error, and the fallback never triggers because `realpathSync` succeeds on the wrong path.

**How reachable:** a project root below its repository root is a supported topology, not a contrivance. `resolveProjectDir` returns an explicit `--project-dir` (or `AIDLC_PROJECT_DIR` / `CLAUDE_PROJECT_DIR`) as given, with no normalisation to a git root (`core/tools/aidlc-lib.ts:593-611`), and the suite configures exactly that shape: `tests/unit/t243-install-mechanism.test.ts:1598-1618` creates `monorepo/.git` and configures the project at `monorepo/packages/service`. In that plain case the mis-resolved path does not exist, so `realpathSync` throws and the fallback keeps today's answer. The wrong-repository answer needs a git dir at the mis-resolved level — a home directory that is itself a git repository is the everyday instance, since `resolve("~/work/proj", "../../.git")` is `~/.git`; a repository vendored inside another is the other.

**The invariant a fix has to satisfy:** the value from `--git-common-dir` must be resolved against the directory it was measured in. With that base the toplevel probe becomes only a "is this a repository at all" check. That correction alone removes the wrong-repository answer, but it does not settle the whole behaviour — resolving correctly also means an ordinary subdirectory project starts resolving to the repository-root name instead of its own, which is the scope question below. I have not implemented or measured a fix, so I am not proposing a shape beyond that constraint.

## The test gap that let it through

`t182`'s new block covers the linked-worktree root and the main-checkout no-op, and asserts both resolve to one store. There is no case where the project root sits **below** the git toplevel — which is exactly the topology above. A case with the project root in a subdirectory, and one with a nested repository at the mis-resolved location, would fail today.

## Two things for you and the maintainers to settle, not defects

1. **Scope.** The description frames this as worktree-only, and the comment says it is "identical to `basename(projectDir)` for every root that is not a linked worktree". With the resolve base corrected that stops being true: a project root in a subdirectory would begin resolving to the **repository** name instead of the subdirectory name, which moves an existing store. Whether that is the wanted semantics — and whether it needs a fallback that finds a store under the old key — is a product call worth making explicitly rather than as a side effect.
2. **Grounding.** #1203 is your own report and nothing on it has been answered by a maintainer yet, so the problem statement has not been accepted by anyone but you. Your reasoning for keeping identity locational rather than moving it to the `origin` remote reads well to me, and I would rather that decision be confirmed by whoever owns this area than assumed by the review.

The second half of the change — verifying the resolved cwd instead of returning it unchecked, with a message naming the path — I have no findings on; `t166` pins it and reverting the source breaks that case.

Once the resolve base is corrected and the subdirectory topology has a case, ping me and I will re-review the new head.

jstrunk:
Thanks — the P2 is real and I have fixed it. New head `332d6fbc`, rebased onto `2d3ad2df` (the
default branch has moved again since your measurement).

## The P2

Your diagnosis was exact: `--git-common-dir` is answered relative to the directory it was measured
in, and I resolved it against `--show-toplevel`. I confirmed the mechanism across five topologies
before changing anything:

| project root | `--show-toplevel` | `--git-dir` | `--git-common-dir` |
|---|---|---|---|
| main checkout root | `solo` | `.git` | `.git` |
| main checkout subdir | `solo` | `solo/.git` | `../../.git` |
| linked worktree root | `solo-feature` | `solo/.git/worktrees/solo-feature` | `solo/.git` |
| linked worktree subdir | `solo-feature` | `solo/.git/worktrees/solo-feature` | `solo/.git` |
| nested repo, root below inner | `outer/inner` | `outer/inner/.git` | `../.git` |

The last row is your reproduction. `../.git` is `outer/inner/.git` relative to the project root, and
`outer/.git` relative to the toplevel — which exists, so the old code answered `outer` with nothing
to signal it.

## Two corrections

1. **Resolve against `projectDir`**, the directory both values were measured in.
2. **Return null unless `--git-dir` and `--git-common-dir` differ.** That difference identifies a
   linked worktree and nothing else: `--git-dir` is `<common>/worktrees/<name>` there, and the two
   are equal for a main checkout, a subdirectory of one, and a repository nested inside another. All
   of those now keep `basename(projectDir)` byte-for-byte.

The second point is what answers your scope question, so I want to be explicit about the choice.
You noted that correcting the resolve base alone would make a subdirectory project start resolving
to the repository-root name, moving an existing store. I did not want that as a side effect of a
worktree fix, so the gate keeps the change worktree-only and the description's claim — "identical to
`basename(projectDir)` for every root that is not a linked worktree" — is now actually guaranteed
rather than merely intended. Whether identity *should* become repository-wide for subdirectory
projects is a real question, but it moves stores and belongs to whoever owns this area, not to this
PR.

## The test gap

Both missing topologies now have cases, plus a linked-worktree subdirectory:

* subdirectory of a plain checkout → resolves to the subdirectory name (unchanged).
* repository nested in another, project root below the inner one → resolves to its own project-root
  name, never the outer repository's. **Fails against the previous resolver** with
  `Expected: not "outer"`, so it pins your finding rather than merely covering the area.
* subdirectory of a linked worktree → resolves the repository name.

Worth flagging honestly: the plain-subdirectory case passes against the old resolver too, because
the mis-resolved path does not exist there and `realpathSync` throws into the fallback — exactly as
you described. It is a regression guard, not a second bug-prover. The nested case is the one that
demonstrates the defect.

`t182` is 19 pass / 0 fail. `bun run typecheck`, `bun run lint` and
`bun scripts/package.ts --check` are clean on the new head; the unit tier and `t166` show no new
failures against this base.

## On grounding

Agreed, and I would rather have that confirmed than assumed too. #1203 is my own report and no
maintainer has weighed in on it, so the locational-vs-remote decision is still mine alone — which is
why I left it as an explicit open question in the issue rather than settling it in code. If whoever
owns this area prefers identity to come from the remote, or to be recorded once and pinned, this
change is compatible with either: it only narrows what the 0-recorded case answers for a linked
worktree, and both alternatives would replace that branch wholesale.

Re-review whenever suits you.

leandrodamascena (approved):
Thanks for working on this @jstrunk! Approved

Issue #868, opened 2026-08-22
Source attribution: multi-repo swarm worktrees are not bound to their creating repository


apackeer:
## Context

Deferred by design from #813's review rounds (triaged as MEDIUM, out of scope for the feature PR). On a multi-repo intent, `aidlc-worktree` records intent, base, and swarm provenance at creation, but **no repository identity**; the merge selects its target repository from the caller's `--repo` at merge time, and `SWARM_SOURCE_MERGED` emits no repository field.

## Failure shape

1. A multi-repo Unit's worktree is prepared from `repo-a`.
2. Its reviewed commit object is also reachable in a compatible `repo-b`.
3. Invoking the merge with `--repo repo-b` lands the source and emits authority for `repo-b`, while the worktree belongs to `repo-a` (cleanup may then fail, but the merge and authority row already landed).

A related inherited scope gap, recorded during round-3 review: on a legacy intent with no recorded repos, an explicit `--repo X` is honored as a sibling anchor while `workspaceSourceState` still fingerprints `projectDir`, so the aggregate never observes the repo the merge landed in and the post-merge footprint check passes vacuously.

## Suggested direction

Record the creating repo / Git common-dir identity in `worktree-meta.json` and the `WORKTREE_CREATED` row, verify it through merge and cleanup, and emit a repository field on `SWARM_SOURCE_MERGED`. Pre-upgrade worktrees without the field keep the documented migration behavior; modern metadata makes it monotonic (consistent with the modernity rules #813 shipped).

Related: the shell-exclusion asymmetry in #869 is the other multi-repo edge in the same surface.

apackeer:
Addressed in #813.

apackeer:
Correction to this issue's residual note, from the final review rounds of #813: the "passes vacuously on an empty change set" symptom described above does not occur in either layout — the workspace-roof shape refuses as `unbindable`, and the legacy shape moves the fingerprint. The underlying gap is real but narrower than described: on a legacy intent with **no recorded repos**, an explicit `--repo` sibling anchor is honored while the workspace aggregate still fingerprints the project directory, so the aggregate never observes the repository the merge landed in. The creating-repository binding that #813 ships (selector + hashed common-dir provenance, verified at merge and discard) prevents cross-repository mutation, but the zero-repo-intent aggregate blindness remains open here.

PR #1206, opened 2026-09-16
fix: allow Construction from a git worktree outside the main checkout

### LINKED ISSUE #567: Construction worktree tool locks out developers who work from git worktrees
## Summary

`aidlc-worktree.ts` (the Construction-phase primitive that all Bolt/Unit worktrees flow through) hard-refuses to run when invoked from inside any linked git worktree. Developers whose normal workflow is one-worktree-per-branch — a common setup — are locked out of Construction with no supported path forward.

## What's there today

`assertNotSiblingWorktree` (`core/tools/aidlc-worktree.ts:99-132`) resolves `git rev-parse --show-toplevel` and compares it to the parent of `--git-common-dir`. If they differ (i.e. cwd is inside a linked worktree), it errors before any git op:

> `aidlc-worktree must run from the main repo checkout, not from a sibling worktree at <path>. Bolt worktrees are siblings of the main checkout, not nested.`

The guard runs on `create`, `merge`, and `discard`. The swarm (`aidlc-swarm.ts`) forks a worktree per Unit through this same primitive, so the refusal blocks autonomous Construction too.

## Is the gap real?

Yes. There is **no escape hatch** — grepping the source turns up no `AIDLC_*` override and no `--allow-worktree`-style flag. The `--repo` re-anchoring handles the multi-*repo* case, not the multi-*worktree* case, so it doesn't help. The check is unconditional: a worktree-native developer simply cannot run Construction without first `cd`-ing to the canonical main checkout, and the error doesn't tell them where that is.

The rationale is sound (avoid the dev-worktree-vs-bolt-worktree clash), but the tool assumes the human always drives from the main checkout, which isn't a safe assumption.

### Secondary: comment/placement inconsistency

The guard message says Bolt worktrees are "siblings of the main checkout, not nested," but `worktreePath` (`core/tools/aidlc-lib.ts:1545`) nests them *inside* the repo at `<projectDir>/.aidlc/worktrees/bolt-<slug>`. Worth reconciling the prose while we're here.

## Proposed mitigations

Recommended: ship the actionable error now, land the auto-anchor as the real fix.

1. **Auto-anchor (ideal).** Resolve the main checkout from `--git-common-dir` and run the git ops there regardless of cwd, instead of refusing. Most robust; removes the constraint entirely.
2. **Opt-in override (pragmatic near-term).** `AIDLC_ALLOW_WORKTREE=1` or `--from-worktree` so a worktree-native dev can accept the risk explicitly.
3. **Actionable error (minimum).** Print the resolved main-checkout path in the error: "run from `<path>` or set `<override>`."
4. **Document the constraint.** Surface "run Construction from the main checkout" in the User Guide / troubleshooting — currently unmentioned.

## Acceptance

- A developer working from a linked git worktree can run `aidlc-worktree create/merge/discard` (or Construction end-to-end) without manually relocating to the main checkout, OR gets an error that names the exact directory to run from.
- The guard comment and `worktreePath` placement description agree.

_Files: `core/tools/aidlc-worktree.ts`, `core/tools/aidlc-lib.ts`, `core/tools/aidlc-swarm.ts`._

jstrunk:
# Summary

Closes #567.

`aidlc-worktree` refused to run from any linked git worktree, so a developer whose project root is a
worktree could not run Construction at all — no override, no flag. One worktree and one intent is
enough to hit it; nothing concurrent is involved.

I posted the analysis on #567 before opening this. The short version: the issue's mitigation 1
(auto-anchor the git ops to the main checkout) is expensive, and its mitigation 3 (an error naming
the main checkout) is actively unsafe, because complying with the guard silently changes the base
ref — moving to the main checkout means every per-unit worktree forks from `main` instead of the
branch holding the approved work. This takes a third route that is cheaper than 1 and safe.

## Changes

The guard's own error names the real invariant: *"Bolt worktrees are siblings of the main checkout,
not nested"*. So the rule is **nesting**, not "is a linked worktree".

`assertNotSiblingWorktree` now refuses only a checkout nested INSIDE the main checkout's working
tree — the `.aidlc/worktrees/<bolt>` (or legacy `.claude/worktrees/<dev>`) shape, where creating a
Bolt worktree would place one worktree inside another's tracked tree. A worktree that lives outside
the main checkout is allowed.

Nothing about that layout breaks the invariant:

* git registers such a worktree under the same common dir, so `gitCommonDirHash` and the
  creating-repo binding from #868 are unchanged;
* its Bolt worktrees hang off its own root, not inside anyone else's tree;
* `--base` and the merge target remain bound to the invoking checkout (`repoCwd` is the invoking
  `projectDir`), which is the property that keeps a unit forking from the branch holding approved
  work. There is a test for this rather than a comment.

**Deliberately not in this PR:**

* **No placement change.** `worktreePath` and its ~50 consumers across `aidlc-worktree`,
  `aidlc-state`, `aidlc-swarm`, `aidlc-bolt` and `aidlc-runtime` — including the state and audit
  fork/merge paths and `t69`'s pure-composer contract — are untouched.
* **No gitignore change.** `AIDLC_SHELL_PATHS` (`core/tools/aidlc-lib.ts`) already excludes `.aidlc`
  at the shell-carrying root, so nested Bolt worktrees never enter the source fingerprint. A
  `.aidlc/worktrees/` ignore rule would only tidy `git status`.
* **No opt-in flag.** With the guard narrowed to the case its message describes, `AIDLC_ALLOW_WORKTREE`
  / `--from-worktree` is unnecessary. If you would rather gate the allowance behind a flag
  (mitigation 2 on the issue), that is this patch plus a flag and I am happy to add it.

**One disclosure:** this PR makes Construction-from-a-worktree reachable, which in turn makes the
`bolt-<slug>` ref-namespace collision reachable. Bolt branches live in the shared common dir and the
slug is `boltSlugForUnit(unit)` with no intent component, while `assertUniqueSwarmBoltSlugs` only
checks uniqueness within one batch — so two concurrent intents whose units share a name collide at
`git worktree add -b`. It fails loudly ("branch already exists") rather than corrupting anything, and
it does not affect the single-intent case this PR fixes, but you may want branch namespacing before
or after this lands. I have not filed it separately for want of a reproduction.

## User experience

**Before.** With the project root in a git worktree, `aidlc-worktree create` / `merge` / `discard`
refuse before doing any work, and the error says to run from the main repo checkout without saying
where that is. Following it makes the tool run and builds every unit from whatever the main checkout
has checked out — normally `main`, not the branch with the approved work. So: a hard stop, or a
silent build against the wrong code. The only reliable workaround is to stop using worktrees.

**After.** Construction runs from the worktree, and its units fork from that worktree's HEAD.
Running from inside a Bolt worktree is still refused, with the same message. Running from the main
checkout is unchanged.

## Checklist

* [x] Changes are documented

No fingerprint, epoch or receipt identity gains an input. `gitCommonDirHash` is computed from
`--git-common-dir`, which resolves to the same path from a linked worktree as from the main
checkout, so the creating-repo binding is byte-identical.

## Test plan

`tests/e2e/t06.test.ts` is the file that pins this guard, so it carries both sides:

* Its existing case is **unchanged and still passes**. Its fixture builds the worktree at
  `<fixture>/.claude/worktrees/dev-slug` — nested inside the fixture's working tree — which is
  exactly what remains refused, including both message phrases it asserts.
* New case: `create` from a worktree OUTSIDE the main checkout succeeds, and the Bolt's HEAD equals
  the invoking worktree's HEAD and differs from the main checkout's. The fixture puts the feature
  branch one commit ahead of `main` so the two are distinguishable, which is what makes the base-ref
  assertion meaningful.
* Mutation-checked: restoring the blanket `cwdTop !== mainCheckout` refusal fails the new case while
  the nested case still passes.

Verified locally against `main` at `9288f80`:

* `bun scripts/package.ts --check` — deterministic across two builds for all seven harnesses.
* `bun run typecheck` and `bun run lint` clean.
* `tests/e2e/t06.test.ts` — 2 pass.
* `t02`, `t03`, `t04`, `t05`, `t49`, `t78`, `t162`, `t164` (every other suite whose fixtures note a
  dependency on this guard) — 49 pass / 0 fail.
* `t166`, `t325` — 54 pass / 4 fail, where the same 4 `t166` cases fail on a pristine checkout of
  `9288f80` (confirmed by re-running with these changes stashed).

Not run here: the SDK-driven suites need live model credentials and stop with `Could not load
credentials from any providers` in my environment.

## Acknowledgment


wowzoo (approved):
Approving. The reinterpretation is right — the guard's own sentence already said the rule was nesting — and each of the three safety properties you name holds where I checked it.

Head `68b27fb46`, recorded base `9288f80f4`. The default branch has moved to `d378d0df` since, so everything below was measured on `d378d0df` merged into this head, which merged with no conflicts.

## Why the relaxation is safe, checked rather than taken

- **The binding really is unchanged.** `gitCommonDirHash` (`core/tools/aidlc-worktree.ts:299-300`) hashes whatever path it is handed, so the claim rests on its call sites — and all four hand it a common-dir value: `creatingGitCommonDir` at `:550`, the recorded `worktreeMeta.gitCommonDir` at `:1315`, and `gitCommonDirRealpath(repoCwd)` / `gitCommonDirRealpath(wtPath)` at `:1337-1338` (resolved at `:1308-1309`). Git reports one common dir for a main checkout and all of its linked worktrees, so the hash and the creating-repo binding are identical either way.
- **Nesting under the invoking root is already the normal case.** `worktreePath` puts Bolt worktrees at `<projectDir>/.aidlc/worktrees/bolt-<slug>` (`core/tools/aidlc-lib.ts:9614-9616`), so a worktree inside the shell-carrying root is what the main checkout does too. And that placement is deliberately outside the fingerprinted surface: `AIDLC_SHELL_PATHS` excludes `aidlc` and `.aidlc` at the shell-carrying root (`core/tools/aidlc-lib.ts:14045-14054`). An outside worktree becomes that root, so it inherits the same exclusion.
- **There is no default-to-`main` hazard inside this tool.** `--base` is required (`core/tools/aidlc-worktree.ts:385`) and `rev-parse --verify` runs in `repoCwd` (`:438`, `:446`), so nothing here can silently fork from whatever the main checkout has checked out.

The guard is still applied at all three verbs with the target anchor — `:435` create, `:2227` merge, `:2869` discard — and both sides of the comparison stay canonicalised (`:258`, `:266`).

## Control

Same mode, single-file runs, packager before each arm:

| arm | `tests/e2e/t06.test.ts` |
|---|---|
| `core/tools/aidlc-worktree.ts` reverted to `9288f80f4`, test kept | **1 pass / 1 fail** — only the new case 4 |
| restored | **2 pass / 0 fail**, 12 expect() |

`bun run check` (packager, `--check`, three tsc projects, biome) is clean on the merge result. Measured with bun 1.4.0 on macOS.

The pre-existing refusal case keeps working without being weakened: it asserts `must run from the main repo checkout` and `siblings of the main checkout, not nested` (`t06:138-139`), and both substrings survive in the new message.

## Two P3s

**1. `isNestedInside` reads a child whose name starts with `..` as not nested.** `rel.startsWith("..")` is the escape, so a path *inside* the main checkout slips through when its first segment begins with `..`. Measured:

| parent | child | `relative()` | result |
|---|---|---|---|
| `/x/repo` | `/x/repo/.aidlc/worktrees/bolt-a` | `.aidlc/worktrees/bolt-a` | nested ✅ |
| `/x/repo` | `/x/repo-feature` | `../repo-feature` | not nested ✅ |
| `/x/repo` | `/x/repo/a..b` | `a..b` | nested ✅ |
| `/x/repo` | `/x/repo/..foo` | `..foo` | **not nested** ❌ |

The invariant the predicate needs is that only a leading `..` **path segment** means "outside", not a leading `..` substring. Exotic, and it fails open on a safety guard rather than closed, which is why it is worth a line.

**2. The new case's title claims a property its assertions cannot isolate.** `4: … forks from ITS head` reads as though the invoking worktree determines the fork point, but `--base` is required and branch refs are shared across a repository's worktrees, so `--base feature-x` resolves to the same commit from the main checkout too — the assertion would hold there as well. What the case does pin is the part that is actually new and does matter: creation succeeds from an outside worktree, and the Bolt lands under the invoking worktree at `wtPath(outside, "demo")`. Worth renaming so a later reader does not rely on a stronger guarantee than the test provides.

Neither blocks. Thank you for citing the refreshed line numbers on #567 and for naming the two mitigations you rejected and why — that is what made this quick to check rather than quick to accept.

apackeer (approved):
Thanks, Jeff — the reinterpretation (nesting, not "is a linked worktree") is the right fix, and the write-up on #567 explaining why mitigation 3 silently changes the base ref made this easy to check rather than take on faith.

I reviewed the current head `5d34070d0` against `main` `0e54914f3`. Alignment was settled by @wowzoo's review; the head hasn't expanded beyond that scope.

To get this over the line I rebased the branch onto current `main` (clean; #1149 had touched the same file) and pushed one follow-up commit on top of yours addressing the two P3s from the review:

- `isNestedInside` now escapes only on a leading `..` *segment* (`rel === ".." || rel.startsWith(`..${sep}`)`), so a checkout named e.g. `<main>/..dev` is still refused. Reproduced before the fix, and t06 gained a case that builds a real `..dev` worktree and asserts the refusal — reverting the predicate fails only that case.
- Case 4 is retitled to the property it pins: creation succeeds from an outside worktree and the Bolt lands under the invoking worktree, with `--base` resolving as given. Assertions unchanged.

Validation on the rebased head: packager determinism across all seven harnesses, `t06`/`t78`/`t162`/`t164`/`t49-bolt-sensor-failures` 45/45, typecheck clean on the touched files; CI is green on `5d34070d0`. Approving and merging.

PR #1208, opened 2026-09-16
test: skip POSIX gate-sensor fixture on Windows

apackeer:
# Summary

Follow-up to #1166. Its compiled gate-sensor regression test creates a `#!/bin/sh` executable that native Windows cannot launch, causing the test to fail with `dispatcher spawn failed (ENOENT)`.

## Changes

Add `test.skipIf(process.platform === "win32")` to the POSIX executable fixture and explain its shell dependency in a comment. This follows the existing guard on equivalent fixtures in `tests/unit/t230-dispatcher-routes.test.ts`.

## User experience

Native Windows runs the other gate-sensor cases and explicitly skips this shell fixture. Linux continues to execute the compiled-dispatch regression test.

## Checklist

* [ ] Changes are documented

## Test plan

On the current `main` base with this patch, using Bun 1.3.14:

* `bash tests/run-tests.sh --debug -P 8 --integration --filter '^t311-gate-sensor-enforcement\.'` — Linux: 15 pass, 0 fail; native Windows: 14 pass, 1 skip, 0 fail. Output and per-file runner artifacts were captured for both runs.
* `bun scripts/package.ts --check` — deterministic across all seven harnesses.
* Biome check of the changed file and `git diff --check` — clean.

Before the fix, the new case failed twice on native Windows; a diagnostic repeat exposed `dispatcher spawn failed (ENOENT)` for both artifacts.

## Acknowledgment


wowzoo (approved):
Approving. The guard is right, it matches what the file and the suite already do, and it changes nothing off Windows.

Base `d378d0df29acf64c8853758706ad3bb45f1eff3a`, head `9f2294ec62bf83b3a6d06482bed05db7b207757b`.

## What I verified

The premise holds in the shipped code, not just in the report: the string you saw, `dispatcher spawn failed (ENOENT)`, is produced at `core/tools/aidlc-state.ts:3071` from the `result.error` branch at `:3062` — the blocking arm of the gate-sensor dispatch. A fixture that cannot be launched lands exactly there.

The guard matches practice. Before this change `test.skipIf(process.platform === "win32")` was already in eight test files, including three times in `tests/unit/t230-dispatcher-routes.test.ts` (`:1195`, `:2719`, `:2751`), and `t311` already branches on platform for its symlink fixture at `tests/integration/t311-gate-sensor-enforcement.test.ts:607`. No workflow runs the test suite on Windows — the only Windows jobs are the `windows-lifecycle` install-and-uninstall checks in `.github/workflows/release.yml:274` and `.github/workflows/preview-release.yml:312` — so this is a native developer run, which is exactly where you hit it.

Off Windows it is a no-op, which is what the change should be. Your head sits directly on `d378d0df`, so this is the merge result, not an approximation of it:

| tree | result |
|---|---|
| base `d378d0df` | 15 pass / 0 fail, 151 expect() |
| this head | 15 pass / 0 fail, 151 expect(), 0 skipped |

Running the guarded case by name on macOS gives 1 pass, 14 filtered out — so the predicate fires only on `win32` rather than skipping everywhere. `bun run check` (packager, `--check`, three tsc projects, biome) is clean on the head.

**Evidence scope:** I cannot run native Windows. That half rests on your report, on the platform behaviour, and on the code path above producing the exact message you quoted — not on a run of mine.

## One P3, not introduced here, no change requested

The sibling fixture in the same file has the identical shape and no guard. `dispatcherStub` writes a `#!/usr/bin/env bun` shebang script and `chmodSync(path, 0o755)` (`t311:313-327`), and `:523` hands it to the gate as `AIDLC_COMPILED_EXECUTABLE` for all three of `exit-2`, `malformed` and `timeout`.

That case asserts fail-closed — `status` 1, `Blocking gate sensor evaluation did not pass`, marker `- [-] probe` (`:526-530`). Those are precisely the observations a launch failure produces, through the same `:3062` → `:3071` path, and no assertion in it separates the two. So under the premise this PR rests on — that shebang fixture cannot be launched there — that test stays green while all three modes collapse into one ENOENT and the distinctions it exists to pin go unexercised.

That is the only one left. Sweeping the file for `chmodSync`, `0o755`, shebang lines and `symlinkSync` finds exactly three POSIX dependencies: the `dispatcherStub` shebang and chmod at `:320-325`, the `#!/bin/sh` fixture at `:541-542` that this PR guards, and `symlinkSync` at `:619`, which the pre-existing branch at `:607` already covers. So two of the three are guarded after this change, and the third is the one above.

Nothing to do here — this PR does not make it worse, and the case it guards is the one whose assertions actually detect the launch failure, because it asserts `status` 0 and the argv shape. Worth stating so the next reader does not have to re-derive why one of the two needed the guard.

PR #1230, opened 2026-09-18
fix: recompute the stage diary path when runtime-graph.json is absent

rafaelgsr:
## Symptom

On the first approval gate of a fresh workflow, the mandatory §13 learnings ritual cannot run:

```
$ aidlc engine learnings surface --slug intent-capture
aidlc-learnings: stage "intent-capture" not found in runtime-graph.json
$ echo $?
1
```

Stage protocol §2 Part 0 requires the ritual to complete before `orchestrate report --result awaiting-approval`, so the stage stalls at its gate with no non-manual way forward.

## Reproduction

From an empty directory, with a native install:

```bash
mkdir /tmp/aidlc-repro && cd /tmp/aidlc-repro && git init -q
aidlc init --yes && aidlc config --harness claude --yes
aidlc engine intent create --label repro --scope feature
aidlc engine orchestrate next          # first stage starts
ls aidlc/spaces/*/intents/*/runtime-graph.json   # absent
aidlc engine learnings surface --slug intent-capture   # exit 1, message above
```

The same sequence reproduces from a source checkout against `dist/claude/.claude/tools/aidlc-learnings.ts`.

## Root cause

`runtime-graph.json` is a gitignored, machine-local, re-derivable view. Its only writer is `aidlc-runtime.ts compile`, and its only dispatcher is the `rebuild-stage-graph` PostToolUse hook, which fires only on the transition-class commands `classifyRuntimeCompileCommand` recognises: `state`, `jump`, `bolt`, `unit`, `utility`, `recompose`, `scope change`, `config set`, the read-only routes, and `orchestrate report`.

`init`, `intent create` and `orchestrate next` are not in that set. So a fresh workflow has no graph at all until the `orchestrate report --result awaiting-approval` at its first gate — and §13 runs *before* that report. The first gated stage of every workflow therefore found the file absent, and `readRuntimeStageRow` turned that into a hard `fail(..., 1)`.

The same hole opens mid-workflow on a fresh clone, after a `git clean`, or after a single dropped hook compile — the file is not tracked, so it is legitimately absent in all of those states.

`surface` reads exactly one field from the graph, `memory_path`, which `compile` itself derives:

```ts
memory_path: relativeMemoryPath(phaseInfo.phase, slug, recordPrefix)
```

Nothing in that derivation needs the audit log, so the absent-file case is recoverable without guessing. Every other consumer of the graph already degrades — `runtime summary`, `doctor-bundle`, and `aidlc-orchestrate.ts`'s recompute-and-warn for a missing or stale `bolt_dag`. `learnings surface` was the only hard failure.

## What the fix does

`resolveMemoryPath` prefers a recorded row and otherwise recomputes the same string from the stage graph, then warns on stderr naming the rebuild command:

```
aidlc-learnings: no compiled memory_path for "intent-capture"; derived
aidlc/spaces/default/intents/… from the stage graph. runtime-graph.json is
machine-local and is compiled at the next transition — `aidlc engine runtime
compile` rebuilds it now.
```

## What it does not cover

- It does not change when the hook compiles. The graph still first appears at the first `orchestrate report`; this change makes the one command that needed it earlier able to proceed, and makes the gap visible rather than fatal.
- A **malformed** `runtime-graph.json` still fails with `runtime-graph.json is malformed`. Only an absent file or a missing row is treated as recoverable — absence is ordinary machine-local state, corruption is not.
- A recorded `memory_path` still wins outright, so the legacy flat `aidlc-docs/<phase>/<slug>/memory.md` layout is untouched.
- `learnings persist` is unchanged; it never read the graph.
- With a compiled row present the command behaves exactly as before and prints nothing extra on stderr.

## Test evidence

`tests/unit/t199-learnings-memory-path.test.ts` gains three cases, alongside the four it already had:

- `surface recomputes the diary path when runtime-graph.json was never compiled` — deliberately skips `compile`, asserts exit 0, the right phase, one candidate, the stderr note, and that `surface` stayed read-only.
- `the recomputed diary path is byte-identical to the compiled memory_path` — asserts the recomputed string, then runs `compile` and asserts the row's `memory_path` equals it, then re-runs `surface` and asserts stderr is empty.
- `a malformed runtime-graph.json still fails instead of falling back`.

```
$ bun test tests/unit/t199-learnings-memory-path.test.ts
 8 pass, 0 fail

# with the fix reverted (tests unchanged):
 6 pass, 2 fail
  → aidlc-learnings: stage "user-stories" not found in runtime-graph.json
```

Full local run: `bun tests/run-tests.ts` (smoke + unit + integration), `bun run typecheck`, `bun run lint`, and `bun scripts/package.ts --check` (deterministic across all seven harnesses).

---


apackeer (changes requested):
Thanks for tracking this down — the first-gate stall was real, the diagnosis (§13 runs before the first transition-class command, so the hook has never compiled the graph) is right, and recomputing the one derived field instead of making a read-only tool a writer or widening the hook is the correct shape. The two new absent-graph tests genuinely fail against base and pass on head, and I reproduced the fresh-workflow path end to end (`intent create` → `orchestrate next` → `learnings surface`): graph absent, exit 0, and the derived path is byte-identical to the row a subsequent `runtime compile` wrote.

I reviewed the current head 98750394e against 87cc7f050 (main has since moved one unrelated kiro-ide commit, no file overlap). Internal maintainer change; alignment is the existing §13 contract.

One blocking item and three optional ones:

1. **Please drop the release bump.** `core/tools/aidlc-version.ts`, the README badge, and the `## [2.9.1] - 2026-09-18` CHANGELOG entry are reserved for the release-preparation PR per AGENTS.md's Release Metadata Policy (`t68` passes only because the three surfaces agree, so CI does not catch this). The entry text is good release-note material — please carry it into the release-prep PR instead.
2. *(optional)* `resolveMemoryPath` also recomputes when a row exists but has no string `memory_path` (`core/tools/aidlc-learnings.ts:316-317`, via the `?.memory_path` collapse). Base exited 1 there (`has no memory_path in runtime-graph.json`), and the PR text and `docs/reference/13-runtime-graph.md` say corruption stays fatal. Either recompute only when the row is `null`, or relax the doc to match.
3. *(optional)* The stderr note fires on every normal first gate, so it cannot by itself distinguish the dropped-hook case the CHANGELOG/doc text says it makes visible. Fine to leave as a plain hint; just worth softening that claim in the doc.
4. *(optional)* `13-runtime-graph.md` "not before the first report" is slightly strong — `aidlc status`, `config set`, the copy-channel `bun .claude/tools/aidlc-utility.ts intent-create`, and Kiro IDE's filter bypass can all compile earlier. "May still be absent at the first gate" is accurate.

Also: the **Require Contributor Statement** check needs the license sentence in the PR body.

Validation on head: `t199` 8/8 (6/2 with `aidlc-learnings.ts` reverted to base — the two absent-graph cases fail as expected, the malformed case passes on both), learnings cluster `t97 t99 t112 t158 t174 t306 t68` 80/0, layout matrix (session binding vs shared cursor, non-default space, plugin stage, scope-excluded stage) all resolve correctly, no stranded consumers of the old error strings.

**REQUEST_CHANGES** on item 1 only — happy to approve once it is out.

apackeer (approved):
Thanks again, Rafael — with your fix the first-gate stall is gone and the ritual runs on a fresh install with a byte-identical derived path.

I reviewed the current head 428852d77 against 87cc7f050 (trial merge with main at 3bd1642cd is clean). Internal maintainer change; alignment is the existing §13 contract. To get this over the line I pushed two small commits on top of yours under maintainer-edit:

- `29351f625` drops the `2.9.1` version bump, README badge, and CHANGELOG heading per AGENTS.md's Release Metadata Policy — the entry text is worth reusing in the release-prep PR.
- `428852d77` narrows the fallback to a `null` row: a row that exists without a string `memory_path` fails with the base message (`has no memory_path in runtime-graph.json`), matching what the PR text promises about corruption, with a t199 case that fails without the guard; `13-runtime-graph.md` now says the file *may* be absent at the first gate rather than *cannot exist before*.

Validation on 428852d77: `bun scripts/package.ts --check` deterministic across all 7 harnesses, typecheck and lint clean, t199 9/9, learnings cluster (t68 t97 t112 t158 t174 t306) 76/0, and CI is fully green. I also added the contributor statement to the description.

**APPROVE.**

Issue #1231, opened 2026-09-18
[Docs]: Statusline $cost is a misleading list-price estimate under Bedrock — document the caveat and disable-before-customer-use


SaRedfiche:
## What

The Claude Code statusline renders a running cost segment, e.g.:

```
BR:opus-5[1m] ctx:38% ↑3k ↓66.9k $23.66
```

The `$<usd>` figure is misleading in the harness's shipped, customer-facing configuration, and there is no guidance advising users to turn it off before demos or customer sessions. Two concrete problems:

1. **It's a public list-price estimate, not the real cost.** `costSegment()` → `computeCost()` prices tokens from `DEFAULT_RATES` in `core/tools/aidlc-usage.ts`, which are documented as *"PUBLIC Anthropic list prices."* But the Claude harness ships with `CLAUDE_CODE_USE_BEDROCK=1` (`harness/claude/settings.json` → `env`), so the model is served through Bedrock. The displayed `$` is therefore **not** the Bedrock on-demand rate, **not** a negotiated/Private-Offer/EDP rate, and **not** any subscription price — yet it's presented next to the live Bedrock model id (`BR:opus-5`), which reads as an authoritative bill.

2. **It's dominated by cache-read tokens, so it looks inflated and climbs fast.** Cache reads are priced (`cacheRead` in every rate row) and summed into the USD total. Over a long AI-DLC workflow, cache-read volume dominates, so both the token counts and the `$` climb quickly and read as an inflated spend.

The net effect in a customer-facing setting is a number that looks like a real, large, fast-growing bill but isn't — leaving a more negative than positive impression.

## Why

The number is a *local estimate for the user's own awareness*, not a billing figure. In a customer demo it invites exactly the wrong conversation ("why is this costing $23 already?") about a figure that doesn't correspond to what anyone will actually be charged on Bedrock.

The disable mechanism already exists and works — `AIDLC_DISABLE_USAGE_TRACKING=1` silences the whole seam (documented in `docs/guide/15-troubleshooting.md`, `docs/reference/06-hooks-and-tools.md`, etc.), and it can be set in the gitignored `.claude/settings.local.json` `env` block with no effect on the workflow (it only removes the cost/token segment). The gap is **not** the mechanism — it's that the docs never:

- state the `$` is a list-price estimate that does **not** reflect Bedrock/subscription/negotiated pricing (the default config runs on Bedrock);
- note that cache-read tokens dominate the figure, so it inflates and climbs fast;
[...]

PR #1232, opened 2026-09-18
docs: flag statusline $cost as a misleading list-price estimate under Bedrock

SaRedfiche:
## Why

The Claude Code statusline renders a running cost segment (e.g. `BR:opus-5[1m] ctx:38% ↑3k ↓66.9k $23.66`). The `$<usd>` is a rough **public list-price** estimate, but the harness ships with `CLAUDE_CODE_USE_BEDROCK=1`, so it does not reflect the Bedrock on-demand rate, a negotiated/Private-Offer/EDP rate, or a subscription plan — and it's dominated by cache-read tokens, so it climbs fast and reads as inflated. In a customer-facing session that leaves a more negative than positive impression about a number that isn't a real bill.

The disable mechanism already exists (`AIDLC_DISABLE_USAGE_TRACKING=1`) and is documented; the gap is that nothing tells users the figure is a list-price estimate under Bedrock, that cache-read tokens dominate it, or that they should turn it off before customer-facing use.

Fixes the documentation gap raised in #1231.

## What changed

Documentation + example only — **no behavior change**:

- `docs/guide/15-troubleshooting.md` — added the accuracy caveat (list price, not Bedrock/negotiated/subscription; cache-read-dominated) and an explicit "disable before customer-facing use" recommendation with the `.claude/settings.local.json` `env` snippet, to the existing "Statusline shows a cost segment you don't want" section.
- `docs/guide/13-customization.md` — added the same caveat to the statusline section.
- `docs/guide/02-your-first-workflow.md` — added a short caveat callout to the statusline prose and to the statusline-legend table row.
- `harness/claude/settings.local.json.example` — added `AIDLC_DISABLE_USAGE_TRACKING` as a commented example in the `env` block, so the workaround is discoverable at the point of use.

## How it was verified

- `bun scripts/package.ts && bun scripts/package.ts --check` — deterministic across two independent builds for every harness.
- `bun test` on the affected suites, all green: `t40-settings-hook-config` (incl. "settings.local.json.example is valid JSON"), `t01-file-structure`, `t298-settings-hierarchy`, `t-tui-user-settings-isolation`, `t269-statusline-cost`, `t271-usage-tracking-kill-switch`.
- Edited hand-authored source only (`docs/`, `harness/`); `dist/`/`dist-release/` untouched (gitignored).
- The only failing tests in `t244-install-management` are pre-existing network/timeout cases (update-discovery: socket/refresh timing); they fail identically on a pristine `main` checkout and are unrelated to this change (it touches nothing that suite reads).


apackeer (approved):
Thank you for this, @SaRedfiche — and for #1231, which laid out the code paths so clearly that I could confirm the premise in minutes. You're right that the User Guide never explained what the `$` figure is, and with Bedrock on by default that is easy to misread. Surfacing it where users actually look is a genuine improvement.

I reviewed the current head 98fb721e8 against `main`. The Developer Reference already describes the rate table as defaults rather than a billing claim and documents the override layers, so bringing that into the guide fits the existing design well.

While reading it against the pricing model, I found a few spots worth tightening, and since they were small I pushed two commits on top of yours rather than bouncing it back. In summary:

- Bedrock does bill prompt-cache reads, and its on-demand pricing varies by tier, inference profile, and region, so I softened "does not reflect the Bedrock rate" and "inflated" to "may not match your invoice" and "real usage, not inflation" — the same warning, without a claim we can't back for every account.
- `AIDLC_MODEL_RATES` is documented in the Developer Reference (§ Rate table and overrides) rather than Customization, so the link now points there.
- Added the rate override next to the kill switch, so readers on negotiated pricing have a way to correct the number as well as hide it (with a note that it applies to newly recorded usage).
- Reworded "customer-facing" to "while presenting, screen-sharing, or recording" so the advice reads the same for every kind of user.
- Trimmed the first-workflow callout, which repeated the kill-switch sentence directly above it.

Happy to adjust any of the wording if it drifts from what you intended. Validation on 98fb721e8: `bun scripts/package.ts --check` is deterministic across all seven harnesses, and `t40`, `t01`, `t298`, `t269`, `t271` pass; CI, Markdownlint, Security Scanners, and Deploy Documentation are all green on this head.

Issue #1218, opened 2026-09-17
[Bug]: .DS_Store system file changes causing drift in macOS 17 with aidlc 2.9.0


nishkarshjain:
### Description

Response from Claude CLI on macSO
"The real cause is background os activity (Spotlight/mdworker and/or Finder are actively running and touching files, e.g. DS_Store) faster than the ~1-2 minute window a review dispatch needs, and this engine's review-source binding treats any workspace file change as invalidating, with no config knob exposed."

<img width="1280" height="245" alt="Image" src="https://github.com/user-attachments/assets/87ae2857-24e1-455a-aefa-5393e25cba7b" />

### Steps to Reproduce

Happens in final phases and mostly seems to be during review.

### Expected Behavior

The system files are not in control with us, so they should be ignored by aidlc

### Actual Behavior

Due to continous changes in .DS_Store in macOS 27 indexing system process mdworker, the aidlc work on every intent is getting blocked, showing it as drift and asking everytime diffrently to fix it.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.9.0

### AI-DLC Phase

Operation

### Harness

Claude Code

### AI Model

Claude Sonnet Medium

### Environment

- OS: macOS 27
- Claude Code v2.1.274

kawaji:
Another instance of this, with a trigger that is inside the framework rather than in the OS:
**the verification the reviewer is instructed to run is what invalidates its own review
receipt.**

Observed on aidlc 2.9.0, Kiro CLI 2.22.0, macOS 26.6.2, Construction / code-generation
(`workspace_requires: true`), per-unit adversarial review:

1. `aidlc engine log review --stage code-generation --unit <unit> --reviewer … --iteration 1`
   wrote `reviewed-source-<id>.tsv` and bound the review to it.
2. The reviewer ran. Per the reviewer protocol it must ground findings in machine-checkable
   evidence, so its brief had it run the project's own checks — `bash scripts/check.sh`
   (ruff + pytest) and a `pytest --cov` run. It wrote only its review file.
3. Recording the verdict was refused:
   `Refusing REVIEW_COMPLETED for "code-generation": workspace source changed after
   REVIEW_REQUESTED iteration 1. Restore the requested source state and re-dispatch the
   reviewer.`
4. I hash-compared every path in `reviewed-source-<id>.tsv`: **all 17 claimed paths, plus
   `source-manifest.json`, still matched their recorded sha256.** `git status` showed the
[...]

Issue #1034, opened 2026-09-07
[Bug]: stage completion is unsatisfiable when a unit's verification runs pytest — the freshness check counts `__pycache__` as source while the manifest validator forbids claiming it


mkhan-lb:
**Version:** 2.7.0 (harness: Claude Code)

## Summary

Two checks in the Code Generation completion path disagree about what counts as
application source, and the disagreement is unsatisfiable for any unit whose
declared verification runs a Python test suite.

- **The source-manifest validator rejects gitignored paths.** A directory claim
  that sweeps in an ignored file is refused.
- **The stage-completion freshness check counts gitignored paths as changed
  source** and demands they be claimed or reverted.

`pytest` necessarily produces `__pycache__/*.pyc`. They are gitignored, so they
cannot be claimed; they are regenerated by any subsequent Python run, so reverting
them does not terminate. There is no third option, and completion fails closed.

## Reproduction

1. Run a `feature`-scope workflow to Code Generation with at least one unit whose
   `verification.ciCommand` invokes `pytest`.
2. Run that verification (i.e. do the thing the stage asks for).
3. Write `source-manifest.json` claiming the service directory:
   `{ "path": "services/<svc>/" }`
4. Request a review → **refused**:

```
Cannot record REVIEW_REQUESTED for "code-generation": unit "<svc>" has no valid
source manifest at .../source-manifest.json (writes[0].path: "services/<svc>/"
contains ignored application source "services/<svc>/.coverage")
```

5. Replace the directory claim with explicit per-file paths (37 in our case).
   The review then records.
6. Report stage completion → **refused**:

```
Refusing to complete "code-generation": 81 application-source path(s) changed
during this stage run that no reviewed unit's source manifest claims
(ci/scripts/__pycache__/<module>.cpython-312.pyc,
 tooling/<dir>/__pycache__/<module>.cpython-312.pyc, …)
Add each path to the owning unit's source-manifest.json … or revert the change.
Unclaimed source changes fail closed (RFC #662).
```

7. Take the "revert" branch — delete the caches. Report again → **refused**,
   now because deleting them is itself a post-review source change:

```
Cannot complete "code-generation" because the project source changed after
aidlc-architecture-reviewer-agent reviewed it.
```

8. Re-review to attest to the deletion, then report again → **refused**, with the
   count now 83: any tool invocation in between regenerated them.

## Why it cannot be resolved by the user

The two checks are individually reasonable and jointly contradictory:

[...]

Issue #878, opened 2026-08-23
[RFC]: enforce stale completed-stage results through backward revalidation


djoo-lgcns:
### Summary

Follow-up to #716. That PR establishes deterministic completed-stage validity detection, schema-2 completion receipts, machine-readable advisories for known drift, and fail-open migration. This RFC proposes the smallest enforcement layer on top: block normal workflow routing only when inspection is fully available and reports known stale completed results, while keeping explicit status and backward-jump recovery reachable. Recovery re-executes the earliest affected stage and the completed downstream suffix so normal completion writes fresh receipts. Waiver semantics are deliberately not included in the first implementation unless this discussion concludes they are required.

### Motivation

A completed checkbox records execution history, not whether its result still matches the artifact basis captured at completion. #716 closes the observation gap but intentionally keeps routing advisory.

The review split detection from enforcement because the detector is mature while policy is not. A blanket barrier makes a typo and a scope rewrite equivalent, and a waiver would create new audit-authority and downstream-validity semantics. The reviewer therefore asked for the hard barrier and its recovery contract to have their own design discussion, and later invited this follow-up to be driven separately.

This RFC seeks agreement on that policy before code begins, without reopening the resolver, receipt, propagation, migration, or artifact-vocabulary work already reviewed in #716.

#### Inherited PR1 contract

PR2 treats the merged #716 behavior as fixed substrate:

- inspectStageValidity remains read-only and deterministic.
- Schema-2 STAGE_COMPLETED receipts remain the only tracked basis.
- stage_validity remains the machine-readable diagnosis.
- Receipt-less and legacy completions remain untracked and fail open.
- Receipt-capture and inspection failures remain warnings and fail open.
- Per-render statusline inspection remains out of scope.
- The detector continues to cover AI-DLC-owned artifact validity only, not source trees, Git, CI, deployments, or external systems.
- The maintainer follow-up described on #716 remains intact: untracked-only histories surface through /aidlc --status rather than warning on every next.

#### Proposed enforcement matrix

| Inspection result / command | Proposed behavior |
| --- | --- |
| Clean | Keep the normal directive kind and route normally. |
| Fully inspected, one or more stale / needs-revalidation issues | Return a hard barrier instead of the normal happy-path directive. Carry the same machine-readable stage_validity, including direct roots, downstream stages, and earliest affected stage. |
| Untracked-only history | Route normally; report details through /aidlc --status. |
[...]

PR #1235, opened 2026-09-18
fix: exclude incidental OS/editor artifacts from the source-floor walk (#1099)

### LINKED ISSUE #1099: [Bug]: plan-approval source floor is a gitignore-unaware filesystem walk, so incidental files (.DS_Store) cause spurious drift and forced re-approval
### Description

The code-generation **source floor** used by plan approval is computed by `workspaceSourceFingerprint` → `workspaceSourceState` → `filesystemSourceIdentity` (`core/tools/aidlc-lib.ts:16345`, `:16269`, `:15442`) — a **pure filesystem walk**, not git-aware. In `full` mode it records **every** regular file via `recordFile` (`:16104-16135`) with no gitignore/tracked-only/extension filter (the `textLike` filter at `:15106` applies only to symlinked-external `sourceOnly` subtrees, not the root walk). The hard-exclude denylist (`SOURCE_FINGERPRINT_HARD_EXCLUDED_NAMES`, `:13610`) lists only **directory** names (`.git`, `node_modules`, `.venv`, …).

Consequently a regular **file** like macOS's `.DS_Store` — or any untracked scratch file a developer leaves in the tree — is hashed into the floor. Incidental filesystem activity (Finder writing `.DS_Store`, an editor swap file, a local note) therefore changes the source floor even though no source changed, which trips the plan-approval drift check (`aidlc-testing-posture.ts:1605-1609`) and forces a spurious re-fingerprint + re-presentation of an already-approved plan.

> **Scope / severity — deliberately honest.** #1000 (Change Control) already fixed the *dangerous* half: approval binds to a re-recordable `[Planned Source]` tag, so drift is now recoverable via re-fingerprint/re-approve (or Relaxed mode / the human override) rather than irrecoverably orphaning the plan. This issue is the **residual friction**: the floor still reacts to non-source, untracked files, so users hit avoidable "source drifted, re-approve" cycles from filesystem noise. Medium severity, not a blocker.

### Steps to Reproduce

1. Reach a code-generation plan-approval for a unit; note the source is not-drifted.
2. Without touching any tracked source, let the OS/editor create or update an untracked file in the tree (e.g. `touch .DS_Store` at a walked directory, or save an editor swap file).
3. Re-run the drift check / attempt to proceed.
4. Observe a source-drift result (re-fingerprint / re-approve required) though no tracked source changed.

### Expected Behavior

The source floor reflects only **tracked source** (or an explicit allowlist of source extensions). Incidental, untracked, gitignored, or OS-metadata files (`.DS_Store`, swap files, local scratch) do not change the floor and do not force re-approval.

### Actual Behavior

Any regular file present at walk time is hashed into the floor; `.DS_Store` and untracked files included. The only exclusions are directory names. So filesystem noise drifts the floor and forces spurious re-approval.

### AI-DLC Version
v2 (current on main)

### Release / Commit
2.8.1

### Phase
Construction (code-generation plan approval)

### Harness
Claude Code — source-confirmed; harness-neutral (`core/`). `.DS_Store` is macOS-universal but the defect (untracked/gitignore-unaware floor) is general.

### Additional Context

Same root as the internal "non-deterministic source floor" observation. Proposed fix: make the source fingerprint **gitignore/tracked-aware** (walk only tracked files, or intersect with `git ls-files`), or apply a source-extension allowlist at the root walk — not just a directory denylist. This is orthogonal to and complementary with the #1000 recovery path: #1000 made drift *recoverable*; this stops drift from firing on non-source noise in the first place.

### Acknowledgment

By submitting this issue, I confirm I can share this contribution under the project license.

tejasavora:
## What

Fixes the plan-approval source-floor false positive in #1099: an incidental OS/editor artifact appearing in the workspace during a stage run (`.DS_Store`, `Thumbs.db`, `desktop.ini`, or an editor `*.sw[a-p]`/`*~` swap/backup file) counted as unclaimed application-source drift and failed the stage closed under RFC #662 — forcing a manual `find … -delete` and re-approval to proceed.

`filesystemSourceIdentity` now skips a small, fixed, vendor-neutral set of never-source OS/editor artifacts at any depth, guarded by `registeredPathRelevant` so an explicitly registered path is never dropped.

## Design note — why NOT gitignore-awareness

#1099's title frames the root as a "gitignore-unaware filesystem walk," and a gitignore-aware exclusion was the first thing I built and tested. **It is the wrong fix for this codebase:** the walk *deliberately binds ignored application source* — `t314` ("ignored application source outside generated-output boundaries remains bound") asserts that changing a git-ignored `.ts` still moves the fingerprint. Honoring `.gitignore` for exclusion breaks that invariant. So the boundary model stays (hard-excluded dirs + conditional `build/dist/…` dirs + the source registry); this change only drops files that carry no source and are never the project's own content. The added test documents this: a git-ignored source file stays bound.

## Scope (intentionally minimal)

Only these artifacts are excluded: `.DS_Store`, `Thumbs.db`, `desktop.ini`, and `*.sw[a-p]`/`*~`. No tool- or plugin-specific names. `.gitignore` is not consulted. Registered paths win.

## Verification

- New `tests/unit/t345-source-floor-artifact-exclusion.test.ts` (3 cases): artifacts don't move the fingerprint; a real source file still does (guard still fires); a git-ignored *source* file stays bound.
- `bun run typecheck` clean; `bun scripts/package.ts --check` deterministic across all harnesses; `t68` version/CHANGELOG/README sync green.
- Ran the source-floor test cluster (t305/t311/t314/t321/t327/t328/t337) and confirmed **no net-new failures** vs `origin/main` (the git-fixture tests that fail in my sandbox fail identically without this change).
- Version bumped 2.9.0 → 2.9.1 with CHANGELOG + README badge.

Closes #1099.

---


apackeer (changes requested):
Thank you for this, @tejasavora — both for the careful issue report and for the design note. You were right to build the gitignore-aware version first and then reject it: `t314` pins "ignored application source stays bound" for a reason, and a boundary whose semantics depend on `.git/info/exclude` and `core.excludesFile` (state outside the fingerprint) can't serve as an integrity floor. That reasoning is sound and I want to keep it.

I reviewed the current head `9cc20adc9f74aa21a8e23f5a4a6065ec7b8b88ab` against `87cc7f050dabbfd24ec1e2cf71d0c8d4dc1f3705` (`main`; one unrelated commit, #1204, has landed since). Two independent adversarial reviews plus my own probes ran in a clean worktree at that head. There is no prior maintainer review on #1099 or this PR, so I'm treating this as the alignment pass as well as the code review.

## Where this head stands

For a freshly recorded baseline, the stage-completion gate is genuinely fixed: with a unit reviewed and `src/.DS_Store` created afterwards, `aidlc-state.ts gate-start code-generation` exits 0. `t345` case 1 fails without the lib change and passes with it. Typecheck, `bun scripts/package.ts --check`, and the source-floor cluster (t305/t311/t314/t321/t327/t328/t337/t345/t68) are green at the head.

However, the exclusion as implemented opens the floor in ways the PR's own contract ("never-source artifacts only; a registered path is never dropped") rules out, and it reaches only one of the two representations of the boundary. Details below, ordered by severity.

## Findings

**1. P1 — Name-only match runs before the type check, so whole directories and real assets leave the floor.**
`core/tools/aidlc-lib.ts:16767-16775` tests `entry.name` before `lstatSync` (`:16826`) and the directory/file branches (`:16969`, `:17022`). A directory named `legacy~/` or `src/Thumbs.db/` drops its entire subtree; `/\.sw[a-p]$/` also matches `.swf`, `.swc`, `.swg`, `.swm`. Reproduced: editing `legacy~/app.ts` and `src/asset.swf` leaves the fingerprint unchanged and the listing at `[src/app.ts]`. Driving the real gate: an unreviewed edit under `legacy~/` → `gate-start code-generation` exits 0 and the stage enters `awaiting-approval`. That is a silent bypass of the RFC #662 fail-closed contract.

**2. P1 — A symlink alias defeats explicit registration (regression vs. base).**
With `aa -> zz` and `.aidlc-source-paths.json` = `{"version":1,"paths":["zz/.DS_Store"]}` (or `["zz"]`), the sorted walk visits `aa` first (`:16716-16718`); the registry-relative path is `aa/…`, so the guard at `:16768` does not see the registration and skips the artifact; the `visited` memo (`:16653-16664`, keyed on realpath + mode) then suppresses the later traversal through `zz`. Editing the explicitly registered file leaves the fingerprint unchanged. At base the same probe reports a change. Reproduced for internal and external targets and for exact and directory registration.

**3. P2 — The walk and the Source Commit now disagree.**
`shapeSourceSnapshotIndex` (`aidlc-lib.ts:15155-15163`) restores only shell, hard-excluded, and conditional pathspecs to HEAD; there is no analogue for the new artifact set. Swarm does `git read-tree HEAD` + `git add -A` (`aidlc-swarm.ts:577-580`, `:1374-1389`), so non-gitignored `.DS_Store` / `foo~` / `.x.swp` stay in the tree. Reproduced with a real `swarm prepare → review → finalize`: with an exact `src/app.ts` claim, `finalize` exits 2 with the same "wrote application-source paths outside unit's source manifest (src/.DS_Store, src/foo~, src/x.swp)" refusal the PR sets out to remove; with a `src/` prefix claim it exits 0 and the immutable Source Commit contains the three post-review files whose bytes the fingerprint no longer binds. Any alignment here also has to be type-aware: a deleted, unregistered `.DS_Store` *symlink* must still surface as a source deletion, so a static glob reset is not sufficient.

**4. P2 — Pre-upgrade evidence trips the fixed gate once, with a dead-end remedy.**
A baseline recorded by the old walk lists `src/.DS_Store`; the new walk omits it, so `aidlc-state.ts:3817-3828` marks it changed and `:4075` refuses as unclaimed. The message's remedy ("revert the change") cannot work: recreating the file never re-enters the listing. Reproduced end to end; the recovery review that clears it also sets `recoverySpent:true`, i.e. the upgrade consumes the one bounded recovery allowance. The upgrade note doesn't mention this.

**5. P2 — CI will be red: `tests/.coverage-registry.json` is stale.**
`t345`'s `covers:` header adds one `coveredBy` row; `gen-coverage-registry.test.ts` "committed coverage registry is fresh" fails at the head and passes at base (`bun tests/gen-coverage-registry.ts --check` → exit 1 vs 0; the diff is exactly that entry). `function:isIncidentalArtifactName` joins no enumerated unit (not exported).

**6. P2 — Release metadata belongs to the release-preparation PR.**
`core/tools/aidlc-version.ts:4`, `README.md:8`, and the new `## [2.9.1]` entry in `CHANGELOG.md` need to be reverted per the Release Metadata Policy in `AGENTS.md` (fix PRs don't bump the version, badge, or add a release entry; see #1181 for how releases land). The behavior note is good material for the release-prep entry.

**7. P3 — `registeredPathRelevant` is self/ancestor/descendant**, so every artifact beneath a registered directory stays bound (`{"paths":["src"]}` → `src/.DS_Store` still moves the fingerprint). Defensible, but "excluded at any depth" and "an explicitly registered path is never dropped" both describe it imprecisely.

**8. P3 — Test coverage.** `t345` exercises only `workspaceSourceFingerprint`. Removing the registry guard entirely still passes all three cases; making `git` exit 127 also passes (case 3 ignores git statuses, and since the walk never consults `.gitignore`, case 3 proves the same thing as case 2). Vim writes `.main.ts.swp`; the fixture writes `main.ts.swp`.

## Direction

Findings 1–3 are the reason I'm requesting changes rather than fixes: they show that excluding by basename can't be made safe by tightening the list, and the class you described in your 2026-09-18 comment (`.abapctl/.session.json` — any tool that writes into the watched tree) can't be enumerated in code anyway. What I'd like to steer toward instead:

- **Declared, not inferred.** Exclusions live in the existing `.aidlc-source-paths.json` as exact relative paths / directory prefixes — no globs or basename patterns — with the registry itself, `.aidlc/`, source manifests, and registered paths non-excludable. A small baked-in default (`.DS_Store`, `Thumbs.db`, `desktop.ini`; regular files only, exact case) can ride on the same mechanism and be rendered in the same list.
- **Pinned to the evidence.** The resolved boundary (framework boundary version + hash of the exclusion set) is recorded alongside the fingerprint in the stage baseline, unit snapshots, and the plan-approval binding. Comparisons happen only under an equal boundary; a mismatch is an explicit "source boundary changed" refusal, never silent reinterpretation of old listings. That also turns finding 4 into a clear re-baseline instead of a dead end.
- **Authorized by the protected human receipt.** The exclusion set takes effect for source checks only once it is bound to the human plan-approval receipt — not to `STAGE_STARTED`, which `advance` records automatically and which unit-major Construction doesn't even re-record.
- **One decision function** shared by the walk and `shapeSourceSnapshotIndex`, with the walk exporting its excluded regular / symlink / deleted-at-HEAD sets so the Source Commit tree always equals what the fingerprint bound.

This is bigger than a fix PR, so I'd suggest an RFC. If you'd like to write it, I'll review it promptly; if you'd rather I open it, say so and I'll credit #1099 and this PR's design note as the basis. Either way, thank you again — the gitignore analysis in particular saved the next person a detour.

CI for this head is `action_required` (fork approval pending); no test run exists for it yet. Submitting as **Request changes**.

tejasavora:
Thank you — this is the review I hoped for, and the findings are right.

Findings 1–3 are the ones that matter: basename matching before the type check drops whole directories and over-matches real assets (`.swf`/`.swc`), the symlink-alias-before-realpath case defeats registration outright, and the walk and `shapeSourceSnapshotIndex` diverging means the fix never reaches the Source Commit at all. Those aren't list-tightening bugs — they show basename exclusion is the wrong shape, exactly as you say. 4 (old-baseline dead-end), 5 (stale coverage registry), and 6 (release metadata — my mistake; fix PRs don't bump) are all fair, and 7/8 are correct about the imprecise contract wording and shallow tests.

Your four properties are the design: **declared** (exact paths/prefixes in `.aidlc-source-paths.json` + a small regular-files-only default on the same mechanism), **evidence-pinned** (boundary version + exclusion-set hash recorded with the fingerprint; a mismatch is an explicit "boundary changed" refusal — which turns finding 4 into a clean re-baseline instead of a dead end), **receipt-authorized** (bound to the human plan-approval receipt, not `STAGE_STARTED`), and **one decision function** shared by the walk and the Source Commit, exporting the excluded regular/symlink/deleted-at-HEAD sets so the two representations can't drift. That also handles the class I flagged on 2026-09-18 — `.abapctl/…`, any tool writing into the tree — the right way: declared by the project, not enumerated in framework code.

Scoping it as a core change rather than a fix, the RFC should also close the rest of this cluster, not just #1099: **#1218** is an independent report of the same drift on the **review-source binding** path (which is why property 4 has to be one resolver across the walk, the Source Commit, *and* review binding — not just the floor), and **#1034** is the same class turned unsatisfiable (`__pycache__` counted by the freshness check but un-claimable by the manifest validator). I'll credit #869 as prior art that the walk/commit asymmetry is real, and tie the evidence-pinned boundary to #878's revalidation model.

One genuine fork I'd want your read on before I write the resolver: your "regular files only" default is right for the incidental-artifact set, but the ubiquitous cases in #1034 are gitignored *directories* (`__pycache__`). I'm inclined to keep those on the existing `HARD_EXCLUDED_DIRS`/`CONDITIONAL_DIRS` lists (type-aware, audited) and reserve the declared registry + default for regular-file artifacts and project-specific tool dirs (`.abapctl/`) — rather than baking cache dirs into the default. Sound right, or would you rather a single declared surface carry both?

I'll write the RFC tagged to #1099/#1218/#1034 and this PR, and mark #1235 superseded rather than patch it — findings 1–3 are about the approach, not the details. Thanks again for the depth.


RFC filed: #1273.

tejasavora:
Marking this as **superseded by #1273** and converting to draft. The review findings 1–3 are about the approach (basename exclusion), not the details, so this isn't patchable in place — #1273 carries the declared + evidence-pinned + receipt-authorized + single-resolver design that replaces it and also closes #1218 and resolves #1034. Leaving the branch/diff up as the design-note reference #1273 builds on.

PR #1150, opened 2026-09-12
feat: review Construction at verified Unit and batch checkpoints

apackeer:
# Summary

Construction approvals now follow completed Units and swarm batches. New solo Unit workflows that include a source-producing stage use unit-major guided execution and verified checkpoints; existing workflows and design-only scopes retain their previous flow.

## Changes

- Build and verify the first integrated Unit before later Units when the walking skeleton is enabled, under either selected iteration order.
- Review each completed Unit in unit-major execution; subsequently settle its stage transitions without duplicate approval.
- Separate serial/swarm execution from guided/autonomous completion approval, with typed runtime setters and fresh-human-request protection during Construction.
- Approve the exact live batch's coding plans together while retaining individual protected Plan Approval receipts. Remaining prepared workers keep those approvals and their worktrees when a batch finishes in parts, including after checkpoint revision preparation.
- Review completed swarm batches before dispatching the next batch. Request Changes resumes a matching surviving worktree or creates a fresh child from the already-landed parent source, with fresh approved plans and a validated parent-to-worktree approval handoff.
- Preflight the entire protected batch before creating worktrees: approved parent source must be committed and reproducible. This also applies to legacy autonomous preparation. The tool does not commit parent source.
- Preserve prior revision records and recover interrupted preparation for the same rejection. Surviving dirty source must match the approved baseline; clean source may only fast-forward to committed approved source.
- Validate transported Unit source manifests against the immutable reviewed source before projecting them into a multi-repository workspace. Temporary blob streams stay isolated between sessions and are cleaned on success or failure.
- Restore an explicitly discarded worker from its recorded committed approved source, keeping current approval and other workers intact. Native/unified swarm commands check each named Unit's approval, and native checks refuse unapproved workers before running commands.
- Carry checkpoint metadata through rule-delivery continuation and preserve earlier approvals when autonomy is revoked or unchanged evidence is refreshed.

Identity changes detect changes to the actual reviewed Unit artifacts, claimed code, active attempt, verification command, or named plan/batch membership. Worktree delegation additionally checks the execution target, immutable worktree provenance, current protected parent approval, and reviewed starting source. Routine directive re-issuance and receipt timestamps do not identify different approved work.

Built on #1143 and #1140. Release metadata is unchanged.

## Upgrade note

Finish active swarms before installing this follow-up. Workers prepared by an older runtime lack the new delegated approval receipt and require preparation under the updated runtime before their checks can continue. Preserve any unmerged work before restarting those workers.

## User experience

[...]

PR #1236, opened 2026-09-18
fix: only advise a missing reviewer dispatch record when one was owed

wowzoo:
### Problem

`stage-protocol-reviewer.md` settles when a §12a step-1 dispatch record is owed: "Single-stage
reviews (no `directive.unit`) write no record."

The no-record branch of `core/hooks/aidlc-reviewer-scope.ts` can write a `reviewer-scope` drop
saying the conductor skipped that write. Its conditions are the agent name and a `construction/`
substring in the tool input; it asks nothing about whether a record was owed. So on a scope that
skips `units-generation`, where every Construction stage review is a single-stage review, the
absence the advisory reports is the protocol being followed — and the drop counters it feeds are the
ones `--doctor` surfaces.

I observed this in two separate runs of the same intent, one under a composed scope that skipped
`units-generation` and one that executed it. The run that skipped it produced six identical
advisories between `2026-09-17T20:48:14Z` and `2026-09-18T20:29:46Z` with zero
`REVIEWER_SCOPE_BLOCKED` events, recurring throughout the observed phase. The run that executed it
produced no advisories and three block events.

This is about the diagnostic signal only. A missing record already fails open by design — the record
is the only source of `unit` and `exempt` — and the conductor in the run above did not deviate from
the protocol. I am also not claiming that a scope which skips `units-generation` is wrong; zero-Unit
execution is supported.

### Change

Gate the advisory on the active-directive marker, which the hook can already reach. It carries
`unit` for a per-unit stage and `units` for `invoke-swarm`, and reading it revalidates the state
digest, so a marker written against a different state does not answer the question.

The state content already read for claimed-checkout enforcement is retained, avoiding a second
state-file read during digest validation. The active-directive file is read in the no-record branch
after the reviewer-agent and `construction/` checks and before the rate-limit check.

The predicate uses `unit` and `units` rather than requiring `kind` or version 2, preserving
version-1 markers that carry `unit` but no `kind`.

The record-present and claimed-checkout enforcement paths are unchanged: they retain the same block
decisions and `REVIEWER_SCOPE_BLOCKED` events. Calls without a dispatch record still return 0; this
change only narrows when that branch appends its advisory drop.

### Validation

- `bun run typecheck` passes; `biome check --error-on-warnings` is clean on both changed files;
  `bun scripts/package.ts --check` reports deterministic output across all seven harnesses.
- `tests/unit/t221-reviewer-scope-hook.test.ts`: 124 pass / 0 fail.
- Three affected cases in that suite: two new, and one existing expectation changed. The existing
  "no record -> exit 0 even for a reviewer sibling sweep" case asserted that its sibling-sweep
  fixture produces the advisory drop; with no active directive there is no authority to say step 1
  was owed, so it now asserts the drop is absent.
- Negative control — with the hook reverted to this base and the tests kept, 122 pass / 2 fail: the
  sibling-sweep case and the new single-stage case. The third case, a per-unit marker with no
  record, passes at the parent by design; it pins that the advisory survives where it is correct.
- Eight suites covering this hook, the harness seam's hook-sibling list, documentation parity, the
  coverage registry, the Kiro adapter, the adjacent reviewer hooks and the dispatcher routes: 413
  pass / 0 fail on a base worktree and 415 pass / 0 fail with the change, the two extra passes being
  the new cases and the failure difference empty.

### Non-goals

This PR does not add deterministic read-scope enforcement for single-stage reviews, and it does not
change composer or validator behavior.


apackeer (approved):
Thanks for this, Keehyun — the diagnosis is exactly right, and grounding the advisory in the active-directive marker (with digest revalidation) is the correct authority to use. The write-up made the review straightforward.

I reviewed the head 94562e3b9 against merge-base 87cc7f050; it merges cleanly onto today's `main` (4bfc514fe), and the commits that landed since do not touch the reviewer protocol or this hook (#1150's checkpoint directives skip the reviewer, so no newly-owed case is affected).

One gap turned up in the predicate, which I fixed in a focused follow-up commit (06aeb1973) rather than sending it back:

**Inherited `units` re-created the false advisory after a swarm.** `writeActiveDirectiveMarker` carries `units` forward onto every later v2 marker regardless of kind (`core/tools/aidlc-lib.ts` `requestedUnits = marker.units ?? base.units`), and the marker is only cleared by a `--single` report. So after any `invoke-swarm` in an intent, a later legitimate no-unit review (e.g. `next --stage nfr-requirements --single`, which §12a says writes no record) had a `kind: run-stage` marker with `units: [u1, u2]` and `perUnitReviewOwed` returned true. Reproduced against the real engine (`next` → `invoke-swarm` → `next --single` → reviewer Read): exit 0, advisory drop written. The fix gates the `units` branch on `active.kind === "invoke-swarm"`; `unit` stays field-keyed so version-1 markers still count. The follow-up also adds three version-2 cases through the production writer (live swarm → advisory; swarm then no-unit run-stage → silent, which fails on the original hook; per-unit run-stage → advisory) and updates the fail-open paragraph in `docs/reference/06-hooks-and-tools.md` to the gated trigger.

Verified at 06aeb1973: t221 127 pass / 0 fail (the new swarm→no-unit case is the only failure with the original hook), typecheck clean, biome clean on both changed files, t132 hook-doc parity passes, `bun scripts/package.ts --check` deterministic across all seven harnesses. Also confirmed the per-unit marker is live and digest-valid at review dispatch (drove the engine to a per-unit reviewer-bearing directive; the advisory fires there as intended), the enforcement path is unchanged, and exit stays 0 on every advisory branch.

Non-blocking note for later: §12a conditions the record on `directive.unit`, while the swarm module says "follow §12a" with `directive.units`; the predicate now encodes "a live swarm owes a record", which is worth a one-line clarification in §12a at some point.

CI is green at this head. Approving.

PR #1237, opened 2026-09-19
docs: replace semicolons that break two sequenceDiagram blocks

### LINKED ISSUE #1219: [Docs]: Broken mermaid diagrams in the reference documentation
### Version

v2 (current on main)

Typo or formatting issue

### Location

https://github.com/awslabs/aidlc-workflows/blob/main/docs/reference/03-orchestrator.md

### Description

The full stage lifecycle diagram is broken and renders an error

Image

### Suggested Fix

Replace usage of `;` in the mermaid diagram with other characters like `,` or `-`

justinhsu1477:
Fixes #1219.

## The break

In a mermaid `sequenceDiagram`, `;` is a statement separator. An unquoted semicolon inside a `Note over` body or a message label ends the statement, and what follows parses as an invalid one, so the whole block renders as a parse error rather than a diagram. Two blocks in `docs/reference/03-orchestrator.md` hit this:

| line | block |
|---|---|
| 354 | **Full Stage Lifecycle** - the one in the issue screenshot |
| 552 | **Construction Execution** |

Replaced the two semicolons with a spaced hyphen, which is the dash style the same file already uses in prose (7 existing occurrences).

## Scope

Three other mermaid lines under `docs/` also contain semicolons, in `docs/guide/04-phases-and-stages.md` and `docs/reference/diagrams.md`. All three sit inside **quoted** labels (`|"..."|` and `["..."]`), where the parser does not treat `;` as a separator.

I could not exercise those blocks to confirm either way: they are flowcharts, and mermaid's flowchart path needs a DOM, so it fails in Node with `DOMPurify.addHook is not a function` — including for a control flowchart containing no semicolon at all. Since the failure is environmental rather than about content, editing those lines would be an unverified change, so this PR leaves them alone.

## Verification

- `bun tests/run-tests.ts --smoke --unit` — **RESULT: PASS**, 314 files, 8910 assertions, 0 failures
- `bun scripts/package.ts --check` — deterministic across all 7 harnesses (claude, codex, copilot, cursor, kiro, kiro-ide, opencode)
- Parsed every `sequenceDiagram` block under `docs/` with the mermaid parser: **11 blocks, 0 failures** after the change. Before it, both edited lines reproduce `Parse error on line 3`; a comma and a spaced hyphen both parse clean.
- Platform: macOS

I did not run the integration or e2e levels: they open live-model gates, and this change is two lines of markdown under `docs/` — no `core/`, no `harness/`, no `dist/`, lockfile untouched.

## Possibly worth a separate issue

`docs/reference/09-testing.md` describes Layer 1 as "every change, **no LLM**, seconds" with "**Run:** `bun tests/run-tests.ts` (default, no flags needed)", but the default profile opens live-model gates — which is why `--no-llm` exists. A first-time contributor following that line waits on a model call while expecting seconds. Happy to file it if that reading is right.

## Contributor Statement


apackeer:
Thanks @justinhsu1477 — this is a clean, well-scoped fix, and the write-up (grammar explanation, the sweep of the three quoted flowchart labels, and the honest note about not being able to verify them) made review easy.

I reviewed the current head 515a855cfb39a7cca90c983a4ad9ceb10bbd8a62 against `main` (0e54914f3e71d2899e3e8c64341b1632bdaa6a09). Alignment basis: #1219 reports these diagrams broken on GitHub and suggests exactly this replacement; the change is docs-only and restores existing rendering.

Verification on my side: rendered the base blob on GitHub and reproduced both parse errors (lines 11 and 27) at the two edited statements; parsed all 45 mermaid fences under `docs/` and `README.md` with mermaid 10.9.8, 11.x and 12.0.0 (jsdom for flowcharts) — base fails exactly these two blocks, head passes all 45, including the 11 sequenceDiagrams. I was also able to confirm what you couldn't: the three quoted flowchart labels containing `;` parse fine, so leaving them alone was right. Nothing else in the repository mirrors this diagram text, and there is no version/changelog impact for a docs PR.

Two housekeeping items before this can merge, neither about the code:

1. Please add the contributor statement to the PR description — the `Require Contributor Statement` check needs this exact sentence: *By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/aidlc-workflows/blob/main/LICENSE).* Editing the description re-runs the check.
2. The remaining workflows were waiting on maintainer approval to run for a first-time contributor; I've approved them.

On the `09-testing.md` "Layer 1 / no LLM / default flags" point you raised: yes, that reading looks right — please do file it, it's a real onboarding papercut.

I'll submit the approval once the statement is in and checks are green.

apackeer (approved):
Approving head 515a855cfb39a7cca90c983a4ad9ceb10bbd8a62. Verification details are in my comment above; contributor statement is in, and CI, Markdownlint, Security Scanners, Deploy Documentation and Pull Request Validation are all green on this head. Thanks again for the fix and the thorough write-up.

justinhsu1477:
@apackeer 
Thanks again for the review and the approval! 🙌

I’ve been digging into ways to optimize AIDLC for smaller models lately, so I’m hoping to contribute more as I learn my way around the project. Looking forward to more contributions and discussions

PR #1239, opened 2026-09-19
fix(sensor): route a sensor's path argument from its declared input_schema

### LINKED ISSUE #1221: [Bug]: Sensor dispatcher's hardcoded isCodeSensor allowlist ignores manifest input_schema,    breaking plugin-contributed sensors
### Description

A sensor manifest can declare `input_schema` (documented in `aidlc-sensor-schema.ts` as an "optional invocation contract"), and the docs state that forks/plugins may add custom `aidlc-<id>.md` sensor manifests.

In practice this extension point doesn't work for argument routing: `parseSensorManifest` never parses `input_schema` out of the YAML frontmatter at all, and the dispatcher (`aidlc-sensor.ts`, `handleFire()`) decides whether to pass `--file-path` or `--output-path` using a hardcoded allowlist:

```ts
const isCodeSensor = id === "linter" || id === "type-check";
```

Any plugin-contributed sensor that needs `--file-path` (a "code sensor" per the existing convention) is invisible to this list, so it receives `--output-path` instead.

If the plugin's script only recognizes `--stage`/`--file-path` (matching the shipped code-sensor convention), it exits immediately with an `"unknown flag"` error — reported as `SENSOR_PASSED` with `Note: script-error: exit-<n>`, not `SENSOR_FAILED`, so it fails completely silently.

### Steps to Reproduce

 1. Install/author a plugin sensor manifest with matches: `**/*.{ts,tsx}` and a script that accepts
     only --stage/--file-path (following the linter/type-check convention).
  2. Declare that sensor in a stage's sensors: frontmatter list.
  3. Run the stage so it writes a matching file.
  4. Check the audit log.

### Expected Behavior

The sensor should receive the argument shape its script expects (either viaa declarative signal from its own manifest, or some other plugin-reachable mechanism) and actually execute its check.

### Actual Behavior

`SENSOR_FIRED` → `SENSOR_PASSED` with `Note: script-error: exit-<n>` and a duration in the low tens of milliseconds — far too fast to have run the real check.

Six consecutive fires of a plugin sensor (`chunk-validate`, matches `**/*.{ts,tsx}`) during a code-generation stage all failed identically (13–17ms, `script-error: exit-1`, across `src/time.ts`, `src/time.test.ts`, `src/greeting.ts`, `src/greeting.test.ts`, `src/app.ts`, `src/app.test.ts`).

Invoking the same script manually with the correct flags succeeds normally:

```json
{"pass":true,"exitCode":0,"cached":false,"stage":"code-generation"}
```

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.9.0

### AI-DLC Phase

Construction

### Harness

Claude Code

### AI Model

Claude Sonnet 5

### Environment

- macOS (Darwin 23.6.0)
- aidlc 2.9.0 (runtime 2.9.0)

### Additional Context

**Root cause:**
1. `parseSensorManifest` in `aidlc-sensor-schema.ts` never reads `input_schema` from the manifest frontmatter despite it being a documented supported field.
2. `isCodeSensor` in `aidlc-sensor.ts` is a closed, hardcoded list of the two framework-shipped sensor IDs, with no path for a plugin to join it without editing core.

**Suggested fix:** make argument routing manifest-driven instead of ID-based — parse `input_schema` and pass `--file-path` when it declares a `file_path` key, falling back to `--output-path` otherwise.

This is advisory-severity, so it fails silently rather than blocking — easy to miss without reading the audit log closely.

vnlebaoduy:
# Summary

A sensor manifest may declare `input_schema`, documented in `aidlc-sensor-schema.ts` as the "optional invocation contract". Nothing read it: `parseSensorManifest` never extracted the block, and the dispatcher chose a sensor's path argument from a hardcoded pair instead —

```ts
const isCodeSensor = id === "linter" || id === "type-check";
```

so a plugin-contributed sensor that follows the shipped code-sensor convention (`--stage` / `--file-path`) received `--output-path`, exited on the unknown flag, and was recorded as `SENSOR_FIRED` → `SENSOR_PASSED` with `script-error: exit-<n>` in tens of milliseconds. The extension point was reachable for discovery but not for invocation, and the failure presented as a pass.

Closes #1221.

## Changes

* `core/tools/aidlc-sensor-schema.ts` — `parseSensorManifest` now populates the declared `input_schema`. A small `mappingKeys` helper reads the top-level keys of a nested frontmatter mapping in declaration order, skips deeper nesting (a list of objects under one of those keys belongs to that key, as in `linter`'s `output_schema.violations`), and ends the block at the next unindented line. Only the key set is meaningful; the values are type hints nothing consumes today.
* `core/tools/aidlc-sensor.ts` — new exported `sensorTakesFilePath(manifest, id)` decides the path flag from the declared contract: a sensor that declares `file_path` is invoked with `--file-path`, a sensor that declares any other contract with `--output-path`. A manifest that declares no contract falls back to the shipped id pair, so nothing that predates `input_schema` changes shape. The dispatcher's `isCodeSensor` now calls it.

**This is behaviour-preserving for everything shipped.** The six framework manifests already declare exactly the split the hardcoded list encoded:

| sensor | declared `input_schema` keys |
|---|---|
| `linter` | `file_path` |
| `type-check` | `file_path` |
| `required-sections` | `output_path`, `stage_slug` |
| `traceability` | `output_path`, `stage_slug` |
| `claim-sources` | `output_path`, `stage_slug`, `deliverables` |
| `upstream-coverage` | `output_path`, `stage_slug`, `consumes`, `deliverables` |

So honouring the manifest reproduces today's routing for every shipped sensor and makes the documented contract actually load-bearing for a plugin. I deliberately did not invent a new declarative signal: the issue leaves that open, and `input_schema` is the one the schema already documents and the shipped manifests already populate correctly.

* `tests/unit/t86-sensor-manifest-schema.test.ts` — the parser surfaces a non-empty `input_schema` for all six shipped manifests; exactly `linter` and `type-check` declare `file_path`; every other manifest declares `output_path` and not `file_path`; and `linter`'s nested `output_schema.violations` list never leaks into the input contract.
* `tests/unit/t251-sensor-stdout-noise.test.ts` — `sensorTakesFilePath` routing: a declared `file_path` wins whoever ships the sensor (including an unknown plugin id); a declared contract without `file_path` routes the output path even for the id `linter`, because the manifest is the authority; an absent or empty contract falls back to the shipped pair.

## User experience

Before: a plugin sensor could be discovered by `loadSensors` and fire, but was invoked with the wrong flag, so its script exited immediately and the audit recorded a pass. Nothing in the log said the check had not run — only the implausible duration hinted at it.

After: a plugin declares `input_schema: file_path: string` in its manifest and is invoked the way the shipped code sensors are. Shipped sensors are invoked exactly as before.

## Checklist

* [ ] Changes are documented

## Test plan

On `891669aa` (main):

```
bun scripts/package.ts
bun test tests/unit/t86-sensor-manifest-schema.test.ts tests/unit/t251-sensor-stdout-noise.test.ts   # 25 pass
bun tests/run-tests.ts --unit --filter "<the unit files mentioning sensors>"   # running at submission; result posted as a comment
bun tests/gen-coverage-registry.ts --check    # fresh, guards green, ratchet held
bun scripts/package.ts --check                # deterministic, 7 harnesses
```

Both new cases fail without the source change (`git stash push -- core/ && bun scripts/package.ts && bun test ...` reports 2 failures: the parser surfaces no `input_schema`, and the routing helper does not exist / does not honour a declared contract).

I also drove the shipped projection to confirm the routing directly:

```
plugin declares file_path  : true
plugin declares output_path: false
plugin declares nothing    : false
shipped linter (no decl)   : true
```

## Acknowledgment


apackeer (approved):
Thanks for this. The diagnosis is right, and routing on the declared `input_schema` keeps every shipped sensor on the flag it gets today while giving plugin sensors a way to ask for `--file-path`.

Approving. The maintainers will now prepare this for merge.

PR #1240, opened 2026-09-19
fix(doctor): read hook heartbeats left at the pre-engine-dir path

### LINKED ISSUE #1214: [Bug]: `Hooks have never executed` fails falsely when the engine is newer than the project runtime stamp
### Description

`aidlc doctor` reports a hard failure saying the hooks have never executed, while the hooks are demonstrably firing — and while another row in the very same report lists six hooks with their last-fired timestamps.

The project's record still uses the pre-2.9.0 layout (`<record>/.aidlc-engine/` does not exist). 2.9.0 moved record-level engine files there, with legacy-path fallbacks for existing records. This check appears not to use that fallback, so on a record written by an older release it finds nothing and concludes the hooks never ran.

### Steps to Reproduce

1. Start a workflow on an older release so the record uses the pre-2.9.0 layout.
2. Run `aidlc update` to 2.9.0. (`aidlc config` cannot be run to align the project, because the workflow is still active — see the companion issue.)
3. Run `aidlc doctor` and `aidlc doctor --verbose` from the project root.
4. Fully restart the harness, as the fix text instructs, and run `aidlc doctor` again.

### Expected Behavior

Either the check honours the legacy-path fallback the release notes describe, or it degrades to a warning naming the version skew as the likely cause rather than asserting the hooks never executed.

### Actual Behavior

fail Hooks have never executed although this workflow has progressed 13 stages
fix: verify this harness's hook registration or trust configuration, then fully restart the harness before resuming the workflow

Restarting the harness does not clear it. The condition is the version skew, not hook registration.

The same `aidlc doctor --verbose` run contains, on the same project:

ok Hooks last fired: continue-workflow 2026-09-15T12:19:58Z, log-subagent 2026-09-14T07:19:23Z, plan-approval-guard 2026-09-16T09:54:02Z, run-sensors 2026-09-16T09:54:05Z, session-start 2026-09-15T09:28:05Z, write-audit-log 2026-09-16T09:54:04Z

So one row lists six hooks with their last-fired timestamps while another asserts the hooks have never executed.

### AI-DLC Version

v2 (current on main)

### Release / Commit

2.9.0

### AI-DLC Phase

Construction

### Harness

Kiro IDE

### AI Model

Claude Opus 5

### Environment

- OS: Windows 11
- Harness: Kiro IDE 1.1.14
- AI-DLC: engine 2.9.0, project runtime stamp 2.8.1, project pin 2.8.2
- <record>/.aidlc-engine/ does not exist; the record still uses the pre-2.9.0 layout

## Further evidence that the hooks did execute

Heartbeats under `<record>/.aidlc-hooks-health/`:

```
plan-approval-guard.last   2026-09-16 18:54:02
write-audit-log.last       2026-09-16 18:54:04
run-sensors.last           2026-09-16 18:54:05
```

`hook-debug.log` contains `emitted eventType="ARTIFACT_UPDATED"` lines, and the audit shards hold 73 `ARTIFACT_UPDATED` rows, rising to 77 after four further writes in the same session.

## Impact

This is a `fail`, not a warning, and its remediation points at hook registration and trust configuration. We hit it while diagnosing a genuine hook problem, and a false confirmation that the hooks had never run sent the investigation in the wrong direction for a considerable time.

The skew itself cannot be resolved while a workflow is active, because `aidlc config` refuses to run (companion issue).

## Related

Same class as other doctor checks whose condition does not match what they assert: #1168 (`Runtime hook PATH` probes a PATH source that is not what hooks experience) and #1167 (a deliberate policy refusal surfaces as a hook malfunction). In each case the remediation text sends the user toward an unrelated cause.

vnlebaoduy:
# Summary

2.9.0 moved record-level engine files under `<record>/.aidlc-engine/`. The relocated stores that shipped with it read the legacy name until the new one exists — `sensorsReadDir`, the summary-authorization registry, and the source-review dir all go through `engineReadDirFor`. Hook heartbeats moved in the same release (`<record>/.aidlc-hooks-health/` → `<record>/.aidlc-engine/hooks-health/`) but got no read fallback.

So on a project whose workflow was in flight across the upgrade, the record still holds its heartbeats under the legacy name until the next hook fires, `hookLiveness` sees an absent directory, and doctor cannot distinguish that from hooks that never ran:

```
fail  Hooks have never executed although this workflow has progressed 13 stages
fix:  verify this harness's hook registration or trust configuration, then fully restart the harness before resuming the workflow
```

The stated fix cannot clear it, because hook registration was never the condition.

Closes #1214.

## Changes

* `core/tools/aidlc-lib.ts` — new `hooksHealthReadDir()` mirrors `sensorsReadDir()`: it resolves `<record>/.aidlc-engine/hooks-health/`, falling back to `<record>/.aidlc-hooks-health/` only while the new directory does not exist (`engineReadDirFor` already encodes that rule, including its refusal to revive legacy data behind a malformed or dangling new entry). `hookLiveness` reads through it. Writers still use `hooksHealthDir`, so every hook keeps writing to the new location and the fallback stops applying after the next fire.
* `core/tools/aidlc-utility.ts` — doctor's `.drops` scan is a read of the same directory, so it follows the same fallback; a legacy record's degraded-hook drops were invisible for the same reason.
* `tests/unit/t319-doctor-hooks-blocked.test.ts` — with workflow progress recorded and a heartbeat at the legacy path, doctor no longer prints `Hooks have never executed` and reports `Hooks last fired: write-audit-log …`; once a heartbeat exists at the new path, the new location wins (the legacy path is a fallback, not a merge).
* `tests/.coverage-registry.json` — regenerated for the new exported unit (`bun tests/gen-coverage-registry.ts`); `--check` reports fresh, guards green, ratchet held.

## User experience

Before: any project upgraded to 2.9.0 with a workflow still in flight got a hard doctor failure asserting the hooks had never executed, with a remediation (restart the harness) that could not clear it, until the next hook fired and populated the new directory.

After: the heartbeats already on disk are read, so doctor reports when the hooks last fired. A project that genuinely has no heartbeats anywhere still fails exactly as before.

## Checklist

* [ ] Changes are documented

## Test plan

On `891669aa` (main):

```
bun scripts/package.ts
bun test tests/unit/t319-doctor-hooks-blocked.test.ts       # 12 pass
bun tests/gen-coverage-registry.ts                          # regenerate for the new unit
bun tests/gen-coverage-registry.ts --check                  # fresh, guards green, ratchet held
bun tests/run-tests.ts --unit --filter "gen-coverage-registry|t319-doctor-hooks-blocked|t07-hook-audit-logger|t10-hook-session-start|t30-hook-session-end"   # 5 / 5 PASS
bun scripts/package.ts --check                              # deterministic, 7 harnesses
```

The new case fails without the source change: `git stash push -- core/ && bun scripts/package.ts && bun test tests/unit/t319-doctor-hooks-blocked.test.ts -t "before the engine-dir move"` reports 1 failure.

Driving `hookLiveness` against a real record created by the shipped CLI, with heartbeats placed by layout:

```
                 before                                  after
legacy    dirExists=false entries=[]              dirExists=true  entries=["run-sensors 2026-09-15T12:19:58Z"]
current   dirExists=true  entries=["session-start …"]  unchanged
both      dirExists=true  entries=["session-start …"]  unchanged (new location wins)
```

## Acknowledgment


leandrodamascena (approved):
Thanks for the focused fix for issue #1214. I reviewed the current head 47011a7666fa496be2ccea7262222124d3d83af0 against base eaf8e991deb5ce45967e20f8198d25e64d25e218.

Project alignment is established by the bug report and the existing 2.9.0 migration contract: relocated record stores retain read-only legacy fallbacks while writers use the new engine directory.

I found and fixed two gaps during the community takeover:

- The diagnostic export path still read only `.aidlc-engine/hooks-health`, so exported evidence omitted legacy heartbeats and degraded-drop records. Commit 47011a766 routes that reader through `hooksHealthReadDir` and adds legacy/current precedence coverage.
- The original precedence test wrote the same hook name and timestamp to both locations, so it did not prove that the current directory wins. Commit 47011a766 uses distinct heartbeats and asserts that legacy data is excluded once the current directory exists.

I also merged the current base in c92c3596e and regenerated the coverage registry.

Validation completed locally:

- 43 focused doctor and diagnostic-export tests passed.
- 28 Plan Approval runtime-authority tests passed.
- 36 existing doctor tests passed.
- 1,731 smoke tests passed.
- Typecheck, lint, coverage-registry freshness, and deterministic packaging passed.

All required GitHub checks are green on this head, including all unit shards, contract checks, smoke, production guards, security scanners, documentation, native terminal units, and live OS isolation on macOS, Linux, and Windows.

No blocking findings remain in the reviewed diff. APPROVE.

PR #1250, opened 2026-09-19
fix(hooks): route each engine child through the dispatcher, not a bare bun

### LINKED ISSUE #454: Kiro adapter spawns bare `bun` — ENOENT when the hook PATH lacks ~/.bun/bin

### LINKED ISSUE #1249: [Bug]: three core hooks spawn a bare `bun` child, so on a native install the Stop hook crashes and stops enforcing, and the graph rebuild and sensor dispatch never run
### Description

Three core hooks spawn their engine child by the bare name `"bun"`:

- `core/hooks/aidlc-continue-workflow.ts:1130-1136` — `Bun.spawnSync({ cmd: ["bun", enginePath, "next", …] })`
- `core/hooks/aidlc-rebuild-stage-graph.ts:242-244` — `spawnSync("bun", ["run", runtimeTs, "compile"], …)`
- `core/hooks/aidlc-run-sensors.ts:227-229` — `spawnSync("bun", [sensorTs, "fire", …], …)`

The documented contract says a native install does not need Bun. `docs/guide/15-troubleshooting.md:79-82`: *"The source/development `dist/` projection runs its 17 TypeScript hooks through `bun`. Native installs and versioned release runtimes route those same hooks through `aidlc`."* `docs/guide/18-install-and-lifecycle.md:419-420` says the runtime probe resolves *"`bun` for copy projections or `aidlc` for native projections"*, and `docs/guide/01-getting-started.md` states the installer needs neither Bun nor Node.js.

That contract holds for the hook's **entry point** — the native `settings.json` wires `aidlc engine hook continue-workflow`, so the hook runs in-process inside the compiled binary — but not for the hook's **body**, which then execs a `bun` the install never provided. The native projection still ships the `.ts` files (70 under `tools/`, 18 under `hooks/`), so the `existsSync(enginePath)` precondition passes and the spawn is really attempted.

The doctor cannot warn about it: `runtimeRequirements()` in `core/tools/aidlc-config-diagnostics.ts` detects a Bun requirement by matching `/\bbun\s+[^\n]*(?:aidlc|\.ts)/` over hook command files, which never matches a `"bun",` argument literal. On a native-shaped project it reports `bun: required=false status=not-required` while the hooks require Bun at runtime.

### Steps to Reproduce

Copy the native projection into a fresh project, seed a record with a workflow in flight, and feed the Stop hook its payload twice — once with Bun's directory on PATH, once without:

```
WITH bun           exit=0    enforced=true
WITHOUT bun        exit=1    enforced=false   stderr: … (stack trace at aidlc-continue-workflow.ts:1130)
```

("enforced" = the hook emitted `{"decision":"block", …}`.) Each run uses a fresh project directory, because the hook's own recursion guard releases the stop after two consecutive no-progress blocks and would otherwise mask the second result.

### Expected Behavior

On a native install the hooks spawn their engine child through `aidlc`, exactly as the docs describe. A missing interpreter, if it ever happens, degrades the way each hook already documents: the Stop hook's `null` means fail open, the other two record a drop with a readable reason.

### Actual Behavior

- **Stop hook**: `Bun.spawnSync` throws `ENOENT` before the documented "null → allow the stop" branch at `:1467`, so the hook exits 1 with a stack trace and emits **no decision at all**. Forwarding-loop enforcement is silently dead, and the user sees a `bun` error at the end of every turn.
- **rebuild-stage-graph**: Node's `spawnSync` returns `status: null` with an ENOENT error; the hook only tests `status !== 0`, so it records the uninformative drop `exit undefined:` and `runtime-graph.json` is never recompiled after `GATE_APPROVED` / `STAGE_STARTED`.
- **run-sensors**: same literal, so no sensor fires; recorded as a drop.

### Version

2.9.0 (`main` @ 63b65ded)

### Area

`core/hooks/aidlc-continue-workflow.ts`, `core/hooks/aidlc-rebuild-stage-graph.ts`, `core/hooks/aidlc-run-sensors.ts`

### Additional Context

The repo already fixed this class once: closed issue #454 ("Kiro adapter spawns bare `bun` — ENOENT") was resolved with `executable ? [executable, "engine", …] : [process.execPath, <script>, …]`, which is what `harness/kiro/hooks/aidlc-kiro-adapter.ts:319-323` and `aidlc-codex-adapter.ts:301,321` do today, and what `aidlc-orchestrate.ts:7565` does through the shared `aidlcEngineCommand()` helper in `core/tools/aidlc-runtime-paths.ts`. The three core hooks were simply not migrated. Using that helper also fixes the source-mode case where Bun is installed but `~/.bun/bin` is absent from the hook's non-interactive PATH, since its source branch names `process.execPath` absolutely.

No documentation change is needed: the docs already describe the intended behaviour; the code is what disagrees with them.

PR follows.

vnlebaoduy:
# Summary

Three core hooks spawn their engine child by the bare name `"bun"`:

- `core/hooks/aidlc-continue-workflow.ts:1130` — `Bun.spawnSync({ cmd: ["bun", enginePath, "next", …] })`
- `core/hooks/aidlc-rebuild-stage-graph.ts:242` — `spawnSync("bun", ["run", runtimeTs, "compile"], …)`
- `core/hooks/aidlc-run-sensors.ts:227` — `spawnSync("bun", [sensorTs, "fire", …], …)`

A native install ships no Bun, and the docs already say so: `docs/guide/15-troubleshooting.md:79-82` — *"Native installs and versioned release runtimes route those same hooks through `aidlc`"* — and `docs/guide/18-install-and-lifecycle.md:419-420`, where the runtime probe resolves *"`bun` for copy projections or `aidlc` for native projections"*. That contract holds for the hook's **entry point** (native `settings.json` wires `aidlc engine hook continue-workflow`, so the hook runs in-process inside the compiled binary) but not for its **body**, which then execs a `bun` the install never provided. The native projection still ships the `.ts` files, so the `existsSync` precondition passes and the spawn is really attempted.

The doctor cannot warn: `runtimeRequirements()` detects a Bun requirement by matching `/\bbun\s+[^\n]*(?:aidlc|\.ts)/` over hook command files, which never matches a `"bun",` argument literal, so a native-shaped project reports `bun: required=false status=not-required`.

Closes #1249.

## Changes

* `core/tools/aidlc-runtime-paths.ts` — `aidlcEngineCommand`'s route union gains `"runtime" | "sensor"`. Both dispatcher routes already exist (`aidlc engine runtime compile`, `aidlc engine sensor fire`), and the helper's source branch resolves `aidlc-<route>.ts`, which matches both tool filenames.
* `core/hooks/aidlc-continue-workflow.ts`, `aidlc-rebuild-stage-graph.ts`, `aidlc-run-sensors.ts` — each engine child now goes through `aidlcEngineCommand(...)`: `[executable, "engine", <route>, …]` when there is a compiled executable, `[process.execPath, <sourceTool>, …]` otherwise. The source branch also fixes the Bun-is-installed-but-not-on-the-hook-PATH case, since `process.execPath` is absolute.
* `tests/unit/t-guard-operation-contract.test.ts` — a pin that strips comments and then refuses any `"bun"` literal in code (prose may name the interpreter; these hooks explain why they must not spawn it), plus route coverage for `runtime` / `sensor` in both compiled and source shapes.
* `docs/reference/06-hooks-and-tools.md` — the sensor-dispatch and graph-rebuild steps now name the channel the way the Stop hook's step already did ("through the selected channel"), instead of describing only the Bun form.

## User experience

Before, on a native install with no Bun on the hook PATH: the Stop hook threw `ENOENT` before its documented "null → allow the stop" branch, exiting 1 with a stack trace and emitting **no decision**, so forwarding-loop enforcement was silently dead and the user saw a `bun` error at the end of every turn; the graph rebuild recorded the drop `exit undefined:` and never recompiled `runtime-graph.json`; no sensor fired.

After, the same install routes each child through `aidlc`, exactly as the documentation describes. A copy-channel project is unchanged apart from Bun being named by absolute path instead of by `$PATH` lookup.

## Checklist

* [x] Changes are documented

## Test plan

On `main` @ 63b65ded:

```
bun scripts/package.ts
bun scripts/package.ts --check      # deterministic across 7 harnesses
bun tests/gen-coverage-registry.ts --check   # fresh, guards green, ratchet held
bun tests/run-tests.ts --unit --filter "<the 255 unit files mentioning hooks, spawn, or the engine command>"
                                    # Failed files: 0 — RESULT: PASS
```

The 255-file batch was run with nothing else on the machine. Earlier runs of the same batch went red on *different* files each time (`t321/t322/t328`, then `t220/t221/t222`) while a second runner was active; the unit tier runs serially within one invocation, so those were cross-invocation contention, not this change. The clean run is the one quoted above.

**Reproduction, native, no Bun on PATH.** Fresh project per run, because the Stop hook's own recursion guard releases the stop after two consecutive no-progress blocks and would otherwise mask the second result. "enforced" = the hook emitted `{"decision":"block", …}`:

| | `main` | this branch |
|---|---|---|
| PATH with Bun | exit 0, enforced | exit 0, enforced |
| PATH without Bun | **exit 1, not enforced** (stack trace at `aidlc-continue-workflow.ts:1130`) | **exit 0, enforced** |

**Child argv, traced.** Compiled: `[aidlc, engine, orchestrate, next, --project-dir, <proj>]`, `[aidlc, engine, runtime, compile]`, `[aidlc, engine, sensor, fire, <id>, --stage, <slug>, --output-path, <path>]`, each exiting 0 with the expected effect (`runtime-graph.json` recompiled, a fresh `SENSOR_FIRED` row, zero drops). Source: identical argv to before except `cmd[0]` becomes Bun's absolute path; `bun run <tool>.ts compile` and `bun <tool>.ts compile` produce the same exit code, stdout, and `runtime-graph.json` hash.

**Stop-hook probe marker.** `AIDLC_STOP_HOOK_PROBE=1` still reaches the child, and `.aidlc-engine/engine-touch` is unchanged across the hook's run, so the conversational carve-out the comment calls load-bearing is intact.

**Fail-without-fix.** `git stash push -- core/ && bun test tests/unit/t-guard-operation-contract.test.ts` reports 11 pass / 1 fail; the pin names each pre-fix hook that holds the literal (all three).

## Notes for reviewers

Two things I deliberately did not change:

1. `isCompiledExecutable()` decides "compiled" from `basename(process.execPath)` not starting with `bun`. A Bun binary shipped under another name would mis-classify — but that exposure already exists for every current `aidlcEngineCommand` caller and is out of scope here.
2. `resolveHookPath` prefers the project-installed `<project>/<harness>/hooks/*.ts` over the binary's packaged runtime, so native users pick this up when their project is re-projected by `aidlc update` / `aidlc config`, not from the binary alone.

This is the same shape that closed #454 for the Kiro adapter (`harness/kiro/hooks/aidlc-kiro-adapter.ts:319-323`) and that `aidlc-orchestrate.ts:7565` already uses; the three core hooks were simply not migrated when the native channel landed.

## Acknowledgment


apackeer (approved):
**Approve.** Verified the fix end to end on compiled binaries, not just under bun.

**What I checked**

- `bun scripts/package.ts`, `--check`, `bun tests/gen-coverage-registry.ts --check`, `bun run typecheck`, `bun run lint`: all green at 6a834d4.
- Smoke + unit + integration slice over the 16 test files that reference the three hooks or `aidlcEngineCommand`: 442 assertions, 0 failed.
- Fail-without-fix: with `main`'s three hooks in place the new pin fails (11 pass / 1 fail); with this branch, 12 pass.
- `git merge-tree` onto current `main` (839fa1f) is clean.
- `engine runtime compile` and `engine sensor fire` are existing noun-passthrough routes, and the inner sensor dispatcher was already compiled-aware, so this closes the last bare-interpreter layer.

**Compiled-binary reproduction.** Built `main` and this branch with `bun build --compile` of the `dist-release/claude` dispatcher, seeded a fresh poc project with the real `intent-create` + `next`, and ran each hook as the native `settings.json` wires it (`aidlc engine hook <name>`) with `PATH=/usr/bin:/bin:/usr/sbin:/sbin`:

| hook | `main` binary, no bun | this branch, no bun |
|---|---|---|
| continue-workflow | exit 1, `aidlc: Executable not found in $PATH: "bun"`, no decision | exit 0, `{"decision":"block", ...}` |
| run-sensors | drop `required-sections: Executable not found in $PATH: "bun"`, no SENSOR rows | `SENSOR_FIRED` + outcome row, no drop |
| rebuild-stage-graph | drop `exit undefined:`, no `runtime-graph.json` | `runtime-graph.json` written, no drop |

With bun on PATH both binaries block on the Stop hook as before.

I also confirmed note 2 in the description: this branch's binary against a project whose installed `.claude/hooks` are pre-fix still fails the same way, because `resolveHookPath` prefers the project-installed hook. Native users pick this up on re-projection (`aidlc config` / `aidlc update`), not from a new binary alone. Worth a line in the release entry.

**Non-blocking**

1. The pin stops at the first violating hook (`expect` throws inside the loop), so it names one hook, not all three as the description says. Collect then assert if you want the fuller message.
2. The pin only matches the double-quoted `"bun"` literal. Fine under biome's default quote style; `/["'\`]bun["'\`]/` would also catch single-quoted or template forms.
3. Separate follow-up, not for this PR: on a native install the Stop hook's block reason still tells the agent to run `bun .claude/tools/aidlc-orchestrate.ts next` / `park`. Those template strings in `aidlc-continue-workflow.ts` (around lines 1251-1282) are not reached by the packager rewrite and PR #1060 does not touch this file. I saw this live in the branch binary's output. Happy to file it.

CI note: all fork workflow runs for this head are waiting on maintainer approval, which is why the merge state shows UNSTABLE.

PR #1260, opened 2026-09-20
fix: keep read-only `next` forms conversational on Copilot and normalize launcher options in the Kiro guard

### LINKED ISSUE #1258: [Bug]: Copilot Stop hook blocks after a read-only `/aidlc` query (`--status`, `intent list`, `team-board`) because the adapter claims the `next` as coordination
## Summary

On the Copilot harness, every read-only `/aidlc` query that goes through `next` ends with a spurious forwarding-loop block. The conductor runs `aidlc engine orchestrate next --status` (or `next intent list`, `next --config models`, `next team-board`), the engine answers with a terminal `print` before touching workflow state, the conductor runs the named utility and prints its output, and the Stop hook then answers:

```
AI-DLC coordination evidence is missing or stale. Run one fresh `bun .aidlc/tools/aidlc-orchestrate.ts next`; do not reuse an earlier continuation token.
```

A fresh bare `next` is exactly what a read-only query must not trigger (it advances the active stage). Reproduced on `main` `dc34c1559` with a full `record-human-turn → PreToolUse → next → PostToolUse → utility → PostToolUse → Stop` lifecycle for `--status` and `intent list`; the `team-board` case is the same.

## Root cause

`harness/copilot/hooks/aidlc-copilot-adapter.ts` `orchestrationCommand()` recognises **every** `next` as a coordination command (`:493-494`), so `claimCopilotCommand` (`core/tools/aidlc-lib.ts:7503-7633`) advances `engine_sequence` and sets `needs_rehydrate: true` at PreToolUse. At PostToolUse `settleCopilotCommand` (`:7692-7707`, `:7764-7765`) treats a `print` as non-retainable → `delivery: superseded`, `needs_rehydrate: true`. At Stop, `consumeCopilotConversation` (`:7859-7880`) requires `human_sequence > engine_sequence`, which the claim defeated, and `copilotStopEvidence` (`:7828-7830`) reports `recovery` → block.

The other harness families already agree that these forms are conversational: the engine skips `touchEngineMarker` for read-only flags, `--config`, workspace verbs and `team-board` (`core/tools/aidlc-orchestrate.ts` routeNext), and the transcript classifier `isTerminalUtilityNext` exempts them on Claude/Codex. Copilot is the only harness whose evidence model counts them as engagement.

## Expected

A `next` whose argv the engine answers terminally without workflow inspection — a `READ_ONLY_FLAGS` token, the `--config` alias, a workspace navigation noun (not `intent create`, which mutates), a leading `team-board`, sole `help` — is not claimed as coordination on Copilot: no `--aidlc-attempt-id` rewrite, no `engine_sequence` advance, no settlement. The turn stays conversational and Stop allows it, exactly as `next --status` does on Claude/Codex/Kiro/opencode. Bare `next`, `next --resume`, `next --stage …`, `continue`, `report` and `park` keep the current claim/settle contract.

## Acceptance

- Copilot lifecycle test (t249 shape): after `record-human-turn`, `next --status`, `next intent list` and `next team-board` (accepted or refused argv) each run without a rewrite, the marker's `engine_sequence` is unchanged, and `continue-workflow` returns `""`; a following bare `next` in the same session still claims and Stop still blocks as today.
- One exported predicate in `core/tools/aidlc-lib.ts` defines "read-only `next` argv" for the Copilot adapter and the transcript classifier so the two cannot drift.

Refs: #1109, #1112 (where this was found), #657.

### LINKED ISSUE #1259: [Bug]: Kiro roll-forward guard misreads `next --project-dir <dir> <verb>` as a bare advancing next
## Summary

`harness/kiro/hooks/aidlc-kiro-adapter.ts` `guard-tool-call` derives `nextArgs` from the raw shell text after `next` (`:485-488`) and applies its leading-token exemptions (`compose`, the orchestrator verbs, `classifyTerminalCommand`) to `nextArgs[0]` without stripping the launcher options the orchestrator itself accepts anywhere in argv (`core/tools/aidlc-orchestrate.ts main()`: `--project-dir <dir>` and `--aidlc-attempt-id <id>` are removed before the subcommand is read).

So with a fresh same-turn read-only latch:

- `aidlc engine orchestrate next --project-dir /x compose …`, `… next --project-dir /x intent list`, `… next --project-dir /x team-board` reach the guard with `nextArgs[0] === "--project-dir"`, are classified as bare-advancing and **blocked** (exit 2), while the same commands without the launcher option are allowed.
- `… next --project-dir /x --status` is allowed only because the read-only scan looks at every token.

The same raw argv feeds the first-next fidelity comparison against `aidlc/.aidlc-forwarding-latch` (`:528-551`), so a conductor that prepends `--project-dir .` to the exact user argv is told it "dropped or changed the user's arguments".

## Expected

The guard normalises `nextArgs` the way the engine does — drop `--project-dir <dir>` / `--aidlc-attempt-id <id>` pairs before a `--` literal — before the bare-advancing classification and the forwarding-latch comparison. Everything else (exemptions, exit-2 contract, fail-open) unchanged.

## Acceptance

t180 rows: with a fresh latch, `next --project-dir <dir> compose x`, `next --project-dir <dir> team-board`, `next --project-dir <dir> intent list` exit 0; `next --project-dir <dir>` alone (bare) still exits 2; a seeded forwarding latch is consumed by `next --project-dir <dir> <exact user argv>`.

Refs: #1112 (where this was found).

apackeer:
# Summary

Two inherited defects surfaced while reviewing #1112, fixed together because both are the same class: a harness seam classifying a `next` argv differently from the engine. Closes #1258. Closes #1259.

## Changes

* **`isReadOnlyNextArgv` (aidlc-lib)** — one exported rule for a `next` argv the engine answers with a terminal directive *before* workflow inspection and *without* touching the engine marker: sole `help`/`-h`, a `READ_ONLY_FLAGS` token in flag position (a read-only spelling in the value slot of `--scope`/`--stage`/`--phase`/`--depth`/`--test-strategy`/`--report`/`--claim`/`--release`/`--team`/`--rhythm` is that flag's value, as `parseNextFlags` consumes it; a `--` literal ends the scan), `--config` at any position (the config print or the usage refusal — `parseNextFlags` returns on either; the refusal is now classified `config` so the engine marker is skipped too), a workspace navigation noun (`intent create` excluded — it mutates), a leading `team-board` (accepted or refused). `isTerminalUtilityNext`, the Claude/Codex transcript classifier, delegates to it, and `isEngineEngagementSegment` treats a parsed non-read-only `next` as engagement before the raw read-only regex can exempt it (`next --report --status` composes; `next -- --status` is intent text).
* **Copilot adapter (#1258)** — `orchestrationCommand()` returns `unrelated` for a read-only `next`, so it is not claimed as coordination: no `--aidlc-attempt-id` rewrite, no `engine_sequence` advance, no settlement. Bare `next`, `--resume`, `--stage`, `continue`, `report` and `park` keep the current claim/settle contract.
* **`stripOrchestratorLauncherOptions` (aidlc-lib) + Kiro guard (#1259)** — mirrors the orchestrator's `--project-dir <dir>` / `--aidlc-attempt-id <id>` extraction (a value must follow; literal `--` respected); `guard-tool-call` applies it only for the bare-advancing classification (the `compose` / orchestrator-verb / `classifyTerminalCommand` exemptions). The first-next forwarding-latch comparison stays byte-exact as on `main` — a launcher option the user did not type is an alteration — and this guard does not police project boundaries (its regex never sees the dispatcher-global `--project-dir` position, so a check there would be a false boundary).
* Tests: t249 `24c` runs the full Copilot lifecycle (`record-human-turn → guard-tool-call → next → post-tool → continue-workflow`) for `next --status`, `next intent list`, `next team-board`, `next team-board --status`, `next --config bogus`, `next --config trust extra` in both command forms, asserting no rewrite, unchanged `engine_sequence`, and an allowed Stop; then `next --report --status` is still claimed (attempt id injected, `engine_sequence` advanced, Stop blocks) and a bare-`next` control still blocks; t180 `11` covers the stripped launcher options; t226 pins the transcript classifier on `help`, `-h`, `--version`, `--doctor --export`, `--status --stage …` and every `--config` form as non-engaging, and `--report --status`, `--scope --status`, `-- --status`, `help me build auth`, `plugin list`, `intent create` as engaging (two earlier pins that read a refused mixed `--config` form as engagement are replaced); t114 asserts a refused `--config` leaves the engine marker untouched over an active workflow; t178 pins both predicates; t250's lib stub gains the new adapter import. No version, badge, or CHANGELOG change.

## User experience

Before, on Copilot with an active workflow:

```
/aidlc --status          (or intent list, team-board, --config models)
→ <status output>
→ Stop hook: "AI-DLC coordination evidence is missing or stale. Run one fresh `… aidlc-orchestrate.ts next`"
```

and the conductor's obedient bare `next` advances the stage the user only asked about. After: the status prints and the turn ends, as it already does on Claude, Codex, Kiro and opencode.

Before, on Kiro with a same-turn read-only latch: `… next --project-dir <dir> team-board` (or `compose …`, `intent list`) is blocked as a bare advancing `next`. After: launcher options are ignored for that classification the way the engine ignores them; a bare `next --project-dir <dir>` is still blocked, and the first-next fidelity check is unchanged.

## Checklist

* [ ] Changes are documented

## Test plan

* `bun run check` (package, `--check` determinism, typecheck, lint) and `bun tests/gen-coverage-registry.ts --check` are green.
* `bun test tests/unit/t178-classify-terminal-command.test.ts tests/unit/t226-detector-corpus.test.ts tests/unit/t180-kiro-rollforward-seam.test.ts tests/unit/t114-orchestrate-next.test.ts tests/unit/t249-copilot-adapter.test.ts tests/unit/t250-copilot-adapter-security.test.ts tests/integration/t121-stop-hook-enforce.test.ts tests/integration/t-guard-native-remedies.test.ts` — all pass.
* Regression proof: with only the adapter edits reverted, t249 `24c` fails (the guard rewrites `next --status` with an attempt id) and t180 `11`/`12` fail (exit 2 instead of 0).
* Manual: a packaged Copilot tree in a scratch project, full hook lifecycle for `next --status` and `next intent list` → `continue-workflow` prints nothing (allowed) where `main` `dc34c1559` printed `decision: block`.

## Acknowledgment


apackeer:
Re the AI review on 29fd72554:

- **`plugin` / `knowledge` nouns with a trailing read-only spelling** — correct, the predicate fell through to the flag scan where the engine gives those nouns precedence and does not exclude them from the engine marker. Next push mirrors that precedence (`next plugin sync --status` is engagement on every path) with t178/t226/t249 rows.
- **`intent archive` / `unarchive`** — not changing. Workspace lifecycle commands are already non-engaging on every harness by the repository's existing contract: `routeNext` skips the engine marker for every `workspaceCommand` ("the workspace verbs are excluded: they carry no workflow intent"), and `isTerminalUtilityNext` has exempted every workspace kind except `create-intent` since it was written. This PR makes Copilot agree with that contract; reversing it for archive/unarchive would be a new decision about the forwarding-loop scope, not a defect of this change. The engine's `print` for those verbs names the utility to run, exactly as it does for `space-create` and `intent <switch>` today.

PR #1262, opened 2026-09-20
fix: make the guards a fence for the agents and a gate the human holds the key to

apackeer:
## Why

Two things were wrong at once, and they had the same root.

The rules a stage needs were delivered in chunks behind a 610-character signed token that the conductor had to copy by hand from the end of a 17 KB line. Models corrupted it, replayed it, or read a stale one, and every failure printed an error and stopped. Two audit logs from real runs show the same three sentences over and over: "I corrupted the token by transcription", "The token is stale", "A fresh next confirms".

The guards were refusing the human. They exist to control the agents in the workflow, but nothing told them who was acting, so a person who had just typed "yes, write the code" met a fence built to stop an agent from writing code nobody asked for. Change Control was supposed to relax that, and it only ever reached five sites.

This PR fixes both at the source: the rules travel differently, and the guards now know the difference between the workflow acting on its own and a person deciding.

## What changes for someone using AIDLC

**Rules and continuation**

1. Rules arrive in one message. A stage directive carries its own rules whenever they fit under the transport cap. Measured across every stage of every scope: 189 of 193 stage runs deliver in one message, and the largest is 27,494 bytes against a 28 KiB floor. Nothing to copy.
2. When a bundle does not fit, each part carries an 8-character receipt and the exact command to fetch the next one. A receipt that is mistyped, stale, replayed, or unknown is answered with the step the workflow is actually on, not with an error directive. Four Classic stages take that path today (functional-design, nfr-requirements, nfr-design, infrastructure-design), each as a single part. Two errors remain by design: a legacy Kiro Plan Approval owned by another IDE window says so, and a lock still held after a bounded retry says "run the same command again". When two calls race for the same receipt, the loser is answered with the winner's committed successor on every harness, Kiro IDE and Copilot session-owned markers included, and never restarts delivery at part one.

**One knob: Guard Policy (was Change Control)**

3. `strict | relaxed | off` instead of strict/relaxed. The new name reaches every surface a model reads: `/aidlc --guard-policy`, config key `guard-policy`, scope key `guard_policy`, state line `Guard Policy`, memory heading, audit event `GUARD_POLICY_SET`, and the prose in all seven orchestrator skills, all eleven scope descriptions, the composer, the code-generation stage and the protocols. Every retired spelling is still read for one release; none is written.
4. On a changed approved input (source moved after plan approval, a reviewed document edited, an output saved without the current summary confirmation): strict asks again, naming what changed; relaxed and off record one `CHANGE_ACCEPTED` row, tell you in one line, and continue.
5. On the fences: strict leaves all five up; relaxed lowers plan-approval and review-freeze; off also lowers state-transition and reviewer-scope. Human presence is never lowered by the policy word.

**Five fences, switchable per piece of work**

6. `/aidlc config set guard.<fence> off|on` (or `--guard.<fence>`) for plan-approval, review-freeze, state-transition, reviewer-scope. Off writes `Guards Off: ... (set by you)` and one `GUARD_DISABLED` row; on writes `GUARD_RESTORED` and can raise a fence the policy word lowered (`Guards On`). It comes back on for the next piece of work. Human presence has no per-work switch; the attempt refuses and says so.
7. A guard that stands aside says so in one line and writes one `GUARD_STOOD_ASIDE` row. It never asks "are you sure": the switch is already off. Human presence is the exception: it has no per-work switch and no stand-aside notice. Its one off-switch is the machine-wide `AIDLC_SKIP_HUMAN_PRESENCE_GUARD=1`, set outside the session, and the session-start hook only records that launch environment.
8. What no policy value turns off: the runtime-integrity check. A tool call that invokes a hook (by path, by `aidlc engine hook`, or as an argv array inside a script), sets a session or bypass variable, or edits the session and Plan Approval runtime records is refused even under `off`, and so is a script the agent writes or runs that concretely imports or executes a hook module. Prose, comments, and inert strings that merely name one are untouched. This is defense in depth: hooks and tool calls run as the same user, so the harness's permission model and your review of what the agent runs remain the outer boundary.

**Relaxing from chat**

9. You type it and it applies as you say it. `/aidlc --guard-policy relaxed|off [...]`, `/aidlc config set guard.<fence> off [--intent x --space y]`, or the confirmation words `guard policy relaxed|off`: the prompt-submit hook applies the switch to the session's piece of work when the prompt arrives, writes the state line and audit row, and reports `AIDLC Guard Policy: ...` as hook context where the harness injects it. Works on Windows and on every harness that forwards the prompt. Codex says `$aidlc`. When the harness supplies a session id and the driver is attended, a typed switch is applied or the hook says why not: under `AIDLC_UNATTENDED=1` it reports that human authority is withheld and names the remedy. A payload with no valid session id records presence only and applies no switch. Proven on a live Claude session under production guards by `tests/integration/t-guard-live-chat-lowering.sdk.test.ts`.
10. A plain-words request ("stop asking me to re-approve when files change") is not a switch: the conductor names the exact command for you to type and ends the turn. A plain-words request for strict runs at once.
11. CLI setters never lower on their own, and their refusal names the exact thing to type. A scope change raises the policy to the new scope's default automatically but never lowers it; to lower and change scope, type the lower value first, then change scope. At creation, naming the scope's own default beside `--scope` is recorded as the scope's value, not as a chat lowering. Memory-held `Mode: strict` refuses every lowering and names the file to edit; `AIDLC_UNATTENDED=1` withholds lowering entirely.

**Refusals and recovery**

12. Under strict, source that moved after the plan was approved used to stop code generation with prose alone. The plan-approval hook now ends that refusal with a typed question the conductor renders, with four answers in recommendation order: approve the plan again, show what changed, stop here, or turn the plan-approval check off for this piece of work.
13. Guard-recovery asks are typed. A `lower-fence` choice is human-input guidance that tells you the command to type and executes nothing. Every fence refusal names its way out, so the way past a guard is printed beside the thing that stopped you rather than left in a reference page.

**Status, config, scopes**

14. `/aidlc --status` shows `Guard Policy: <value>` with its source (`from scope x`, `set by you`, `from org.md`) and a `Fences:` line with all five and where each setting came from.
15. `/aidlc config get|list` covers eleven settings (depth, test-strategy, review, guard-policy, sensors, learnings, summary-confirmation, and the four `guard.<fence>` keys) with effective values and sources; they combine in one atomic command.
16. Scope defaults: strict on enterprise, security-patch, and infra; relaxed on the other eight, Classic and Express included, so plan-approval and review-freeze stand aside there by default. Express declares sensors, learnings, and summary confirmation all off. Every scope declares its Guard Policy and its three ceremonies explicitly.

**Composer gate**

17. The proposal carries a Guard Policy row with a one-line rationale that you can flip before approving. Flipping a matched stock proposal turns it into a custom scope declaring that value, and the intent is created from it; nothing to type afterwards. In-flight, the row is read-only and names the routes that change the value.

**Kiro IDE builds that deliver no prompt text**

18. One plain-language note per session: guard settings cannot be lowered for the active piece of work because this build does not deliver the submitted message; to use a lower setting, update Kiro IDE or start a new piece of work from a scope whose default already uses it; selecting strict and turning a fence on still work; a sole retired `Change Control: relaxed|off` line is renamed to `Guard Policy` without changing its value.

## The chain of authority

Every action a guard sees is covered by one of two authorities, or neither.

The human grant is a message recorded after the engine last issued a directive. It lasts until the engine issues the next one, and it covers everything done to carry it out, including by an agent dispatched for it. The engine instruction is the directive currently in force, and it covers the work that directive asks for. Neither is what the fences are for, and neither decides anything a guard does: the grant is written on every stand-aside row (Authority, Grant, Actor) so a reader can see who was at the keyboard, and it is read nowhere else. The keys are the policy word and the per-work switch.

The question is never "who is acting". A developer agent acts on the conductor word, and the conductor on the human. Authority flows down the delegation chain, so a dispatched agent inherits the authority that sent it (stamped on the in-flight ledger at dispatch) and can never mint a grant for itself. Whether the work happens inline or in a subagent is knowledge, not authority.

One function answers who is acting (authorityFor) and one decides what a fence or a changed-input check does (decideGuard). The four fence hooks and the plan-approval drift check route through them. The older human-presence readers in the tools still consult the ledger directly; unifying them is listed under Not in this PR.

| | human grant | engine instruction | neither |
|---|---|---|---|
| Changed input, policy relaxed or off | stands aside, logged | stands aside, logged | stands aside, logged |
| Changed input, policy strict | asks once, naming what changed | asks once, naming what changed | asks once, naming what changed |
| Fence up | holds | holds | holds |
| Fence lowered | stands aside, logged | stands aside, logged | stands aside, logged |

Human presence is the key holder, not a fence. No policy value and no per-work switch lowers it, because "a real person is behind this approval" is the one thing an agent must never decide for itself. It has exactly one off-switch, the machine-wide `AIDLC_SKIP_HUMAN_PRESENCE_GUARD=1`, set outside the session as a deliberate human act.

A general human message is not a key for any fence, and that is deliberate. A live Kiro IDE fixture showed why: an answer to an unrelated question would have admitted a source write before the plan was approved. The key is the exact switch, typed on purpose and recorded: the prompt-submit hook reads the command text the person submitted and applies only the recognized lowering forms, so a question that mentions a switch, a quoted or negated mention, or an unrelated reply applies nothing. The hook is activated only through the dispatcher's hook route (a per-process token; the hook file is inert when executed directly and exports no callable). It does not authenticate who launched the dispatcher: hooks and tool calls run as the same user, and no harness gives a hook an identity a same-user process cannot copy. A process-ancestry check was tried and refused the legitimate path, because Claude runs hook commands and Bash tool calls through the same `sh -c` chain. So the boundary is stated rather than pretended: the runtime-integrity check refuses every recognized tool-call route to the hook and its records, the audit row names the claimed source so a forgery is visible afterwards, and the harness's permission model plus the person's review of what the agent runs are the outer boundary. A held fence presents its recovery question carrying that switch as one of the choices, so turning it takes one move.

## Migration

Every retired name is read for one release and never written. The scope key change_control, the memory heading, the flag --change-control, and the config key change-control all still work; using one prints a single line naming the new name. A state file carrying only a retired `Change Control: relaxed|off` line is renamed to `Guard Policy` on the next prompt, value and source label preserved, without a policy audit row; the state digest treats the two names as one field, so the rename invalidates no issued directive and no bound approval. A file carrying both lines with different policy words resolves to strict and carries a notice until you choose. Old CHANGE_CONTROL_SET audit rows are still read as policy history.

## Audit

GUARD_POLICY_SET replaces CHANGE_CONTROL_SET as the written setting row. New: GUARD_STOOD_ASIDE (which fence, which authority, how the grant was proven) and GUARD_RESTORED. GUARD_DISABLED now also covers a fence switched off for one piece of work. The taxonomy is 105 events after the rebase (this branch and the base each added three). No row was removed and none is forgeable from the audit CLI.

## Not in this PR

**The desktop journey now passes, and fixing it found something worth knowing.**
That gate had not been running at all: the harness looked for an executable name
the current application no longer uses, and a missing binary counts as a skip, so
the file reported success while running nothing live. Pointed at the right binary
it ran and failed, on this branch and equally on the base. The cause was not a
guard: the test was telling the assistant to record an approval for work that
never happened, and the assistant refused, quoting the framework's own rules back
at it. It was right to refuse, and a test must not need it to misbehave to prove
the engine stops misbehaviour. So the assistant is now asked only for what a
person legitimately does at an open gate, the fabricated approval is driven by the
test itself with the check switched on, and every original assertion is unchanged.
Both journeys now run green. Two commits carry this, both test-only and separable
from the rest.

One product question came out of it and is fixed in this PR rather than filed:
asking the engine what to do next while a gate is open and awaiting approval used
to answer with an instruction to run that stage again, sending an assistant doing
the right thing to redo finished, reviewed work that the review freeze would then
refuse. `next` at an open gate now re-presents the gate as a gate-only directive
(`gate_only: true`) that keeps the Review brief and dispatches no stage body or
reviewer; see "Since the first version".

**One protocol test needed fixing, and what it exposed was in the tests.** Rules
now arrive with the stage, so an assistant acts on the whole rule set in its
first turn instead of a later one. That is the point of the change, and it made
two long-standing test flaws visible. The fixture seeded a state file directly
and never wrote the project-description sidecar that intent creation always
writes, so the record was unlike any real one and the assistant probed a file
that was missing. And the assertion behind it treated any failed tool call as a
protocol fault, including a path the assistant merely guessed at, which a live
assistant will always do sometimes. Both are fixed in a separate test-only
commit: the fixture is realistic now, and a call the harness never saw start is
reported as a host rejection instead of failing the journey. That test went from
four failures out of four to three passes out of three. The neighbouring
multi-repo journey fails on the base too and is untouched by this.

Three readers of "a human acted" still coexist: the ledger reader the approve and answer paths use, the last-answer reader, and the turn markers authorityFor reads. They agree in practice and none of them decides a fence any more, but one definition is the design's intent, and folding them reaches into two harness adapters, so it reviews better alone.

Two refusals keep their prose. The recording tool's decision and answer paths and the begin command still refuse strict drift in words alone, because their shared error path wraps an embedded question in one JSON string that the ask parser rejects; fixing that path is its own change. The revision backstop, which fires when evidence moves between the preflight and the approval, is a race that should be re-answered rather than switched, and is left as found. The required-artifacts guard stays as it is on purpose: folding it under a policy word would let a stage complete with no output.

The scope descriptions still explain only the changed-input half of Guard Policy in prose; the fence half lives in the frontmatter, the status line and the reference docs.

The refusal-wording sweep. Every lockout named in the guard inventory can now be
lowered for one piece of work, and the sites that a person actually hits carry
the switch in band. What is left is consistency of voice across the remaining
refusal sites (unit verbs, answer and override refusals, unattended paths, unit
merge): they still speak in engine vocabulary and none of them offers the way
out. That is a mechanical pass with per-family templates written down, and it
reviews better on its own than buried under this one.

Three transport findings, all recorded with file and line. A race loser on a
stateless route, a Copilot-owned marker, or Kiro IDE republishes part one
instead of reading the winner successor, which costs rework rather than
correctness. A replayed continue under a tracked Copilot attempt still returns
an error directive. Each needs either the publication preconditions or the
Copilot claim protocol exercised with fixtures, and guessing at them at the end
of this change would risk more than it buys.

On Kiro IDE the stand-aside line is not delivered, because that host forwards
no hook output. The audit row is the record there. This is a pre-existing
limitation of every hook message on that harness, not something introduced here.

## Verification

On the current head (`bca8b3207`, main merged through `52ac7eda6`):

| Gate | Stamp | Result |
|---|---|---|
| Live Claude session under production guards (typed switch applies; plain words do not; fence switch and status) | 2026-09-23, `t-guard-live-chat-lowering.sdk` | PASS, 51 s here and 57 s in the implementing run |
| Smoke and unit, 364 files | 2026-09-23T10-01-50Z-p3846434 | 0 failed |
| Integration, 131 files | 2026-09-23T11-14-21Z-p58921 | 2 red: t19 (Claude SDK live preflight, not logged in on this machine) and t-e2e-native-cancellation (upstream runner test, load-sensitive, passes alone) |
| Production-guards tier, 2 files (36 cases) | 2026-09-23 | 0 failed |
| Build determinism across seven harnesses, typecheck, lint, coverage registry | 2026-09-23 | all exit 0 |
| CI on the head (eight unit shards, smoke, production guards, native terminal, live OS isolation) | run 35846133159 | green |

Not yet run on this head: the live Kiro CLI protocol and terminal tiers and the live Codex tier. They were green early in the branch and the adapters changed since; they are the remaining confidence check before merge.

The earlier runs below are kept as the record of the first version and its rebase.
Every tier and slice was run on this branch with the required flags, and each
verdict comes from the run stamp rather than the stream.

| Gate | Stamp | Result |
|---|---|---|
| Smoke and unit, 311 files | 2026-09-17T06-39-02Z-p75430 | 4 red, all resolved or environmental |
| Integration, 119 files | 2026-09-17T07-37-06Z-p42280 | 7 red, all triaged and resolved |
| Integration, converted suites | 2026-09-17T07-53-03Z-p78819 | 5 files, 0 failed |
| Deterministic end to end, 33 files | 2026-09-17T10-47-50Z-p91013 | 1 red, fixed and rechecked green |
| Live Claude terminal, 20 files | 2026-09-17T07-59-44Z-p5390 | 0 failed |
| Live Kiro terminal, 3 files | 2026-09-17T12-07-51Z-p44023 | 0 failed |
| Live Kiro protocol, 10 files | 2026-09-17T11-29-20Z-p23481 | 2 red, both shown pre-existing or flaky |
| Live Codex execution, 5 files | 2026-09-17T12-36-24Z-p70642 | 1 red, green on a raised time budget |
| Live Kiro IDE, 3 files | 2026-09-17T13-17-51Z-p96677 | 2 of 3 pass; the third cannot start |
| Full tier after the second commit, 431 files | 2026-09-18T06-12-51Z-p26906 | 9 red: registry regenerated, install test fails on the base, two parity reds and one compile red were edits made during the run, five re-run green below |
| Smoke and unit on the final source, 311 files | 2026-09-18T07-46-10Z-p54596 | 4 red: registry regenerated against a fresh build, install test as above, one base flake re-run green below, one new test fixed and re-run green below |
| Standalone re-runs, 36 files | 2026-09-18T08-40-58Z-p12718 | 0 failed: registry and its readers, plugin compose, the four integration flakes, the packaging trio, and the guard suites |
| Full tier on the rebased tree, 431 files | 2026-09-18T09-05-28Z-p29394 | 5 red, all attributed; the four non-environmental ones pass alone |
| Those four alone, 36 files | 2026-09-18T10-32-51Z-p49041 | 0 failed |
| Deterministic end to end on the rebased tree, 33 files | 2026-09-18T10-37-36Z-p57414 | 0 failed |
| Live Claude terminal on the rebased tree, 20 files | 2026-09-18T11-11-33Z-p83167 | 19 of 20; the one red is a coin-flip assertion, two passes and two failures over four runs |
| Live Kiro protocol, the test that was failing | 2026-09-19 three consecutive runs | 0 failed, after the two fixes below |
| Full tier on the final tree, 431 files | 2026-09-19T05-29-35Z-p15451 | 8 red, every one settled: one fails identically on the base, the other seven pass alone |
| Those seven alone, 19 files | 2026-09-19T06-59-45Z-p34762 | 0 failed |
| Live Kiro terminal on the rebased tree, 3 files | 2026-09-18T15-03-03Z-p77724 | 0 failed |
| Live Codex execution, 5 files | 2026-09-18T20-23-12Z-p94068 | 0 failed, on a raised per-beat budget |
| Live Kiro desktop checkpoint, 3 files | 2026-09-19T04-49-27Z-p90728 | 0 failed; both live journeys run for the first time, 9 of 9 in the checkpoint file. The two neighbouring files are native-Windows-only |

Build determinism passes across all seven harnesses, typecheck is clean, and the
linter reports only its three pre-existing infos. The coverage registry was
regenerated and no class regressed.

Three reds needed attribution rather than a fix, and each was settled with
evidence rather than assumption.

The install-management timing tests fail on this machine because they wait on a
local fixture server. This branch changes no file in the doctor or update path,
and neither of those files mentions anything it adds.

The Kiro protocol journey fails on the base commit too, with the same shape: the
live agent emits an occasional malformed tool call, and any test asserting there
were none is sensitive to it. The sibling utilities test failed twice, passed on
the base commit, then passed here on a third run. No guard stood aside and none
refused in any of those traces.

The Codex compose test was killed at exactly its ten minute per-turn budget and
passes when that budget is raised. This change makes rule delivery faster rather
than slower, so a transport cause is not plausible.

Every live gate is now met. The Kiro IDE checkpoint journey was the last one
outstanding: an earlier run could not start because the desktop app was showing
its sign-in page, and the run recorded above is that gate signed in and driven
through, 9 of 9 in the checkpoint file.

## Since the first version

The branch was rebased onto main and later merged with it repeatedly; the audit count is 105 on all three pinned surfaces. Fifteen adversarial review iterations followed, and three design questions were settled in writing on this PR:

- **A person relaxes the guards from chat.** `4e7737c48` removed lowering from chat; `492e5822d` reverted that, keeping its automatic retired-field rename and the state-digest canonicalization that protects issued directives across the rename.
- **`relaxed` lowers plan-approval and review-freeze.** `da25afefe` had kept those two up under relaxed; `4f76646eb` restores the shipped mapping (strict: all five up; relaxed: plan-approval and review-freeze stand aside for undirected work and log; off: also state-transition and reviewer-scope; human presence never).
- **The hook does not authenticate its caller.** `c6aea00e6` gated the human-turn hook on process ancestry, which refused the real Claude path; `bca8b3207` drops the gate, keeps dispatcher-only activation and the runtime-integrity refusals, and states the same-user boundary in the reference.

Also since: `next` at an open gate re-presents the gate instead of rerunning the stage or its reviewer, keeping the Review brief; a typed `config set|get|list` is routed to the config command and is not workflow engagement for the Stop hook or the Copilot claim gate; the always-loaded onboarding carries the guards rule so a plain-words request is answered in one turn; a Guard Policy flip at the compose gate on a matched proposal becomes a custom scope; continuation race losers receive the winner's committed successor on every harness; the runtime-integrity check refuses concrete imports and executions of hook modules only, not prose.

The review's remaining findings restate decisions recorded here (F1 the same-user boundary, F5 and F16 what relaxed means, F10 scope-derived lowering, F17 terminal prints being non-engaging) and are for the ledger, not for more code. One narrow follow-up remains open: a gate resumed after Review Override is changed to none should source its Review brief from the recorded review (F18).

## Acknowledgment


apackeer:
Heads-up before it lands: the next push reverts `4e7737c48` (fix: address guard authorization findings) as a revert commit, on the author's direction that a person must be able to relax the guards from chat.

What comes back: the typed switch (`/aidlc --guard-policy relaxed|off`, `/aidlc config set guard.<fence> off`, plain-words `guard policy relaxed`) applied by the prompt-submit hook at prompt time, the `lower-fence` recovery remedy, the Kiro IDE adapter's typed-switch path, and the "type the switch" wording in the docs and skills. CLI setters still never lower on their own; the conductor still runs the strict setter directly.

What is kept from `4e7737c48`: the automatic rename of a sole retired `Change Control: relaxed|off` line (`normalizeRetiredGuardPolicyField`, prompt-submit hook and Kiro IDE adapter) together with the state-digest canonicalization that treats `Change Control` and `Guard Policy` as one field and its t330 test, so the rename never invalidates an issued directive or a bound approval.

Also kept: everything since (`0452ec467`..`d02023617`): a compose-gate Guard Policy flip on a matched proposal becomes a custom scope, the runtime-integrity check refuses concrete imports and executions of hook modules only, naming a scope's own default beside `--scope` takes that scope's policy back, and the stage and reviewer contracts say where the plan-approval stop and the review freeze live when a fence stands aside.

The trust boundary stays stated in the reference: hooks and tool calls run as the same user, so the harness's permission model and the person's review of what the agent runs are the outer boundary; the person's typed switch is the authority. Review findings F1 and F10 restate that boundary and are for the ledger, not for more code.

leandrodamascena (approved):

apackeer:
Evidence on `c6aea00e6` (fix: authenticate human-turn hooks and repair CI), for whoever picks this up next:

- CI on this head is red in all four unit shards: t147 (unattended prompt does not mint HUMAN_TURN), t218 8d12 (record-human-turn applies a visible typed fence switch at prompt time, expected `AIDLC Guard Policy:`), t278 (three wave cases), t333 (4 and 9: `off (set by you)`, `relaxed (set by you)`, the hook lowering a fence), the committed coverage registry (stale for t341), and the native Kiro IDE recovery journey.
- The customer path is broken, not only the fixtures: `tests/integration/t-guard-live-chat-lowering.sdk.test.ts` drives a real Claude session against the installed `dist/claude/.claude` under `--production-guards`. On `2482cb4ad` it passed in 51 to 56 s across five runs (state line `- **Guard Policy**: relaxed (set by you)` plus one `GUARD_POLICY_SET` after the person types `/aidlc --guard-policy relaxed`). On `c6aea00e6` the same drive leaves the state at `strict (from scope enterprise)` and writes no row. Run it with the Bedrock env exported and `AIDLC_CLAUDE_SDK_LIVE=1 bash tests/run-tests.sh --debug --integration --production-guards --filter "^t-guard-live-chat-lowering"`.
- Likely cause, from reading `authenticatedHumanTurnSession`: Claude Code runs each hook command through a shell, so the chain is claude -> sh -> `aidlc.ts engine hook` (launcher) -> hook child. `launcher.ppid` is the transient shell, not the host recorded at SessionStart, so `readSessionPidEntry` finds no entry and the hook returns 0 without minting or applying the switch. The SessionStart hook's own parent is a different transient shell, so the recorded PID cannot match either.

Since the author's direction is that a person must be able to relax the guards from chat, that live test is the acceptance check for any authentication design here: it has to stay green on the real hook registration, not only when the hook is spawned by a test.

apackeer:
Decision, agreed between the author and Leandro: **`relaxed` lowers the plan-approval fence again.** The next push restores the fence set the branch shipped and documents everywhere: `strict` leaves all five fences up; `relaxed` lowers plan-approval and review-freeze for work nobody directed, recording one `GUARD_STOOD_ASIDE` row each time while the conductor still asks the approval question; `off` also lowers state-transition and reviewer-scope; human presence is never lowered by the policy word. This reverses the fence mapping in `da25afefe` (which kept those two up under relaxed and lowered reviewer-scope instead). If review-freeze should stay up under relaxed after all, say so; the plan-approval half is the agreed part.

Two smaller things come with it: the gate-only resume keeps a read-only review-brief surface so a resumed gate still shows the existing findings without dispatching another review, and the typed `config set|get|list` route is classified as terminal in the shared `isReadOnlyNextArgv` classifier so Copilot's next-claim and the Stop hook agree with the engine.

Verification of the customer path stays blocked until `c6aea00e6`'s ancestry check accepts the shell Claude Code puts between the host and the dispatcher (see the previous comment); the deterministic suites cover the fence mapping in the meantime, and the live journey is the acceptance check once the hook mints again.

apackeer:
Decision (author): **option A, drop the process-ancestry gate on the human-turn hook.**

What it changes, stated plainly: `authenticatedHumanTurnSession` guards everything the hook mints from a prompt, not only the guard switch: the `HUMAN_TURN` row, protected-question responses, Plan Approval responses, and override requests. Removing it means a shell-capable agent that reaches the hook through a route the runtime-integrity check does not recognize (a module name assembled at run time) could fabricate any of those, and the audit row would carry the claimed `Source: you`; the row makes a forgery visible afterwards, it does not detect it. That is the same-user boundary this branch has stated since the ninth review: hooks and tool calls run as the same user, so the harness's permission model and the person's review of what the agent runs are the outer boundary. No harness gives a hook an identity a same-user process cannot copy, and the ancestry check proved it by refusing the legitimate path: Claude runs hook commands and Bash tool calls through the same `sh -c` chain.

What stays as defense in depth: the dispatcher-only activation with the per-process token (the hook file is inert when executed directly, and `run` is not exported), and the runtime-integrity refusal of every recognized route to the hook and its records from a tool call (paths, `aidlc engine hook`, session and bypass environment assignments, inline and wrapper scripts, aliases, functions, written content that imports or executes the hook module).

The next push removes the gate and the synthetic-host test scaffolding it required, adds a hook-side notice for any future case where a typed switch is not applied so it can never fail silently again, states the boundary in the reference where the gate was described, and proves the outcome three ways: the live journey on a real Claude session under production guards (the typed switch applies; plain words do not), the forged-route cases (a tool call invoking the dispatcher hook route or writing a wrapper that imports the hook is refused; the hook file executed directly mints nothing), and the production-guards tier.

apackeer:
/aida accept F1 Accept the residual risk of a shell-capable agent forging human-authority records through an unrecognized same-user route, as explicitly decided by the author in https://github.com/awslabs/aidlc-workflows/pull/1262#issuecomment-5792195050. The process-ancestry gate rejected legitimate Claude hook execution; the accepted boundary is the harness permission model and human review, with dispatcher-only activation and runtime-integrity checks providing defense in depth.

apackeer:
/aida accept F16 Intentional product behavior confirmed by the author: a user who lowers plan-approval through relaxed/off or the per-work fence switch is choosing continued execution after approved plan, test-instruction, or Testing Contract content changes, without mandatory reapproval. We accept the risk that execution can therefore use content not covered by the earlier approval. The original human approval must remain an accurate historical record; continuation is attributed to the lowered fence and audited as such. Instructions and execution paths are being aligned with this choice instead of adding a non-lowerable reapproval stop.

apackeer:
/aida accept F16 Author-confirmed UX decision: relaxed/off or the per-work plan-approval switch permits continuation after plan, test-instruction, or Testing Contract edits without mandatory reapproval. We accept execution of content not covered by the earlier approval. Preserve that approval as history and audit continuation under the lowered fence. Align instructions and execution paths with this choice.

apackeer:
/aida accept F24 Accept the residual risk that a shell-capable agent running as the same user can replace installed enforcement files through link operations. Runtime integrity is defense in depth, not a filesystem sandbox; the accepted outer boundary remains the harness permission model and human review, consistent with F1.

PR #1264, opened 2026-09-20
fix(doctor): name the runtime compiler in the runtime-graph remedies

### LINKED ISSUE #1164: [Bug]: runtime-graph.json not built for very simple initial stage
### Description

First, apologies if this report is incomplete. Just trying out this project and learning.

I am trying to create a new plugin to replace the core "aidlc" plugin. The docs, and existing tests imply this is supported.

To get my feet wet, creating simple pass through stages that simply create an arfitact recording the name of the stage. The stage routing is based on a scope specific to the plugin. This is just to get a feel for the setup.

I am observing that I cannot exit my first stage because of missing runtime-graph.json

### Steps to Reproduce

1. Create a new simple plugin
- Stages - stage1, stage2 (create <stage>.md artifacts that have a single`# <stage>` line)
- Scope - simple-scope - basically stage1 -> stage2

2. Create a project where only the simple plugin is installed. 

2. Run with /aidlc --scope simple-scope "Create hello world" - the task is not important

3. Observe proper intent creation, stage1 entry, and stage 1 artifact creation

4. Observe attempt to go through the learning step which fails with missing runtime-graph.json

5. Attempted recovery with `/aidlc --resume` eventually fails with the same problem. Some of the output indicates that it tries to run `aidlc engine graph compile` which appears to be the wrong command anyway

### Expected Behavior

User is able to complete stage 1 and eventually execute the full scope

### Actual Behavior

Unable to exit stage1 because runtime-graph.json is misisng

### AI-DLC Version

v2 (current on main)

### Release / Commit

_No response_

### AI-DLC Phase

Initialization

### Harness

Codex CLI

### AI Model

gpt-5.6-sol

### Environment

- OS: MacOS
- Harness: codex-cli 0.154.0

### Additional Context

I had Codex debug the session and it came up with

---
The hook saw the activity, but its filters prevented the runtime compiler from running:
1. aidlc engine intent create recorded STAGE_STARTED, but the hook’s command allowlist intentionally excludes intent commands. It returned before inspecting that event. See [aidlc-lib.ts (line 1663)](<redacted>/.codex/tools/aidlc-lib.ts:1663).
2. Writing the Question artifact did not help because the runtime rebuild hook is attached only to Bash commands, not file writes. See [hooks.json (line 107)](<redacted>/.codex/hooks.json:107).
3. The later reuse command produced QUESTION_ANSWERED and ARTIFACT_REUSED, but the hook only accepts these audit events:
   GATE_APPROVED, STAGE_STARTED, STAGE_AWAITING_APPROVAL, AUDIT_MERGED, UNIT_MERGED, or WORKFLOW_COMPLETED.
   See [aidlc-rebuild-stage-graph.ts (line 188)](/<redacted>/.codex/hooks/aidlc-rebuild-stage-graph.ts:188).
4. The next qualifying event would be STAGE_AWAITING_APPROVAL, but AI-DLC requires the learnings step before recording that event—and the learnings step refuses to run without runtime-graph.json.
So this is a first-stage bootstrapping deadlock: the initial STAGE_STARTED event is skipped, while the next accepted event occurs only after a step that already requires the missing runtime graph.""
---

On the surface this makes sense, but I'm too new to the project to know if it accurate. 
Looking at what it described in (1) the hook does indeed explicitly skip `intent create` which seems like a reasonable place to get this setup, but I lack the context for the decision why this is the case.

(2) Suggests that if my stage had used Bash in some way the graph would have been written. Have not tried it yet but is is not a reasonable requirement that this should be the case.

I can recover the intent with `aidlc engine runtime compile` but this is not ideal

vnlebaoduy:
# Summary

Both runtime-graph findings in the doctor tell the user to re-run the **stage-graph** compiler to repair the **runtime** graph:

- `core/tools/aidlc-doctor-bundle.ts:717` — `runtime-graph-stale`: ``The compiled runtime graph is out of date. Re-run `${aidlcToolInvocation("graph")} compile` ``
- `core/tools/aidlc-doctor-bundle.ts:737` — `runtime-graph-missing`: ``No compiled runtime graph. Re-run `${aidlcToolInvocation("graph")} compile`. ``

`aidlc-graph.ts` compiles `stage-graph.json` + `scope-grid.json`. It contains no reference to `runtime-graph` at all (`git grep -c runtime-graph -- core/tools/aidlc-graph.ts` → 0). `runtime-graph.json` has one writer, `aidlc-runtime.ts compile` (`core/tools/aidlc-runtime.ts:6` — *"Walk audit + memory, write runtime-graph.json"*), and that is what the rebuild hook dispatches (`core/hooks/aidlc-rebuild-stage-graph.ts:1`).

So the remedy runs a real command that cannot change the condition that printed it, on both channels — `bun <harness>/tools/aidlc-graph.ts compile` on a Bun projection, `aidlc engine graph compile` on a native install.

`runtime-graph-missing` fires exactly when a workflow exists but no graph has been compiled yet, which is where a user at the first stage lands. That is step 5 of #1164 — *"it tries to run `aidlc engine graph compile` which appears to be the wrong command anyway"*.

The prose already had this right: `docs/reference/13-runtime-graph.md:174,370` and `docs/guide/12-cli-commands.md:1868` all name `aidlc engine runtime compile`. Only the doctor disagreed, so no documentation change is needed.

Closes #1164. The deadlock that opens that report (`learnings surface` hard-failing on the absent graph) was fixed by #1230 / `4bfc514f`, which did not reference the issue; this is the remaining reported defect.

## Changes

* `core/tools/aidlc-doctor-bundle.ts` — both remedies route through `aidlcToolInvocation("runtime")`. `runtime` is an existing dispatcher route (`core/tools/aidlc.ts:136`), so the native spelling resolves to `aidlc engine runtime compile`.
* `tests/unit/t243-doctor-bundle.test.ts` — test 23b pins both findings: each remedy contains `` `${aidlcToolInvocation("runtime")} compile` `` and does not contain `aidlcToolInvocation("graph")`. Pure in-process assertion over `runDiagnosis`, no fixture or subprocess.

Deliberately unchanged: the four other `aidlcToolInvocation("graph")` call sites (`core/tools/aidlc-utility.ts:4852,4853,4942`, `core/hooks/aidlc-session-start.ts:391`) are all about uncompiled **stage** entries and are correct as they stand.

## User experience

Before: `aidlc doctor` on a project whose runtime graph is missing or stale prints a remedy naming a compiler that does not produce the file. Running it changes nothing, the warning persists, and the accompanying "the rebuild-stage-graph hook may not be firing" hint sends the reader toward hook diagnosis instead.

After: the remedy names `runtime compile`, the same command `learnings surface` already prints on its own recompute warning.

## Checklist

* [x] Changes are documented

## Test plan

Rebased onto `main` @ 2167de80:

```
bun scripts/package.ts
bun scripts/package.ts --check                 # deterministic across 7 harnesses
bun tests/gen-coverage-registry.ts --check     # fresh, guards green, ratchet held
bun tests/run-tests.ts --smoke --unit          # 321 files, 0 failed, 9184 assertions
```

The unit tier was run with nothing else on the machine; it runs serially within one invocation, so a second concurrent runner makes unrelated files go red.

**Fail-without-fix.** With the test in place and only the two production lines reverted (`git stash push -- core/tools/aidlc-doctor-bundle.ts && bun scripts/package.ts && bun test tests/unit/t243-doctor-bundle.test.ts`): **30 pass / 1 fail**, the failure being 23b. Restored: **31 pass / 0 fail**.

**Evidence the named tool cannot do the job.** `aidlc-graph.ts` has zero `runtime-graph` references; the only `writeFileSync` for `runtime-graph.json` in `core/` is in `aidlc-runtime.ts`.

## Acknowledgment


vnlebaoduy:
Folded into #1272 — same file (`core/tools/aidlc-doctor-bundle.ts`), same subsystem, and reviewing them separately meant opening that file twice for one contributor. The commit is unchanged, it is now the second commit on that branch, and #1272's description carries the full rationale and the re-run gates.

Closing this one so the queue holds a single review unit instead of two. #1164 is referenced from #1272.

PR #1268, opened 2026-09-20
feat(config): share a harness-neutral AGENTS.md block across harnesses

apackeer:
# Summary

Follow-up to #1199 (multi-harness coexistence, refs #1196). The root `AGENTS.md` managed block becomes **harness-neutral and byte-identical** for Kiro CLI, Kiro IDE, Codex, Cursor, and OpenCode, so any two of them with distinct engine directories can coexist in one project. Each harness's specific onboarding moves to a native always-on surface inside its own directory. Copilot keeps its full root `AGENTS.md` (its method `@`-imports live there and are rewritten at runtime by `aidlc-includes.ts`) and stays exclusive.

## Changes

- **Templates.** `core/templates/onboarding.md` is now the marker-free neutral root (what AI-DLC does, the `aidlc/` workspace, method/knowledge layers, conventions, resumption, git, and a list of where each harness's onboarding lives). New `core/templates/onboarding-harness.md` carries every per-harness slot plus a leading `frontmatter` slot. The duplicated DocumentKB paragraph is collapsed to one (with `summarize`).
- **Manifests / packager.** `OnboardingSpec.harnessDst` renders the native file: `.kiro/steering/aidlc-onboarding.md` (Kiro CLI via `agents/aidlc.json` `resources`; Kiro IDE via `inclusion: always`), `.cursor/rules/aidlc-onboarding.mdc` (`alwaysApply: true`), `.aidlc/onboarding.md` (listed in `opencode.json` `instructions`), `.codex/onboarding.md` (pointed at from `.codex/config.toml`). Claude and Copilot keep one concatenated file (native part first). Codex's root is rendered by the packager like every other harness; `emit.ts` only rewrites skills paths. The packager asserts byte identity for every `shared: "identical"` root path in both channels, and the emitted descriptor gains `onboarding` (the native file diagnostics track).
- **`aidlc config`.** `RootIntegration.shared` adds `"identical"`; the five harnesses declare it for `AGENTS.md`. The descriptor guard allows identical↔identical; adding a harness next to an install from a release before shared onboarding is refused with `… whose install predates shared onboarding; run aidlc config --harness <name> first` (also under `--force`); exclusive blocks keep `cannot coexist in one project`. Diagnostics track the native onboarding file from the descriptor instead of a Claude-only special case.
- **Includes.** Space repointing (`aidlc-includes.ts`) and the Cursor installer skip `<space>`-style documentation placeholders, which are never space names.
- **Docs.** Install guide (root-integration table, coexistence paragraph), troubleshooting rows, all six harness guides, porting guide (`harnessDst`, `shared: "identical"`, descriptor `onboarding`), repo `AGENTS.md`.

## User experience

| You run | Result |
|---|---|
| `aidlc config --harness codex` on a kiro project | `.codex/` scaffolded; one neutral `AGENTS.md` block; both baselines record the same block hash; later refreshes `preserve` |
| `aidlc config --harness cursor` / `opencode` next to kiro, kiro-ide, or codex | same, when engine directories differ |
| `aidlc config --harness copilot` on a kiro project | `harness copilot shares AGENTS.md with installed kiro; they cannot coexist in one project` |
| adding next to a kiro tree installed by an earlier release | `… whose install predates shared onboarding; run aidlc config --harness kiro first` |
| refreshing an existing single-harness install | root block updates to the neutral text through the normal owner path; the native onboarding file is added as an ordinary managed file |

Still refused: kiro↔kiro-ide (`.kiro`), opencode↔copilot (`.aidlc`), copilot with any `AGENTS.md` harness.

## Checklist

* [x] Changes are documented
* [x] If this change adds an input to any fingerprint, epoch, or receipt identity, the description names the human-visible change it detects — not applicable (no fingerprint/epoch/receipt inputs change)

## Test plan

- `bun scripts/package.ts --check` deterministic across all seven harnesses (includes the new byte-identity assertion); `bun run typecheck`, `bun run lint` clean.
- `bash tests/run-tests.sh --smoke` PASS; unit tier: shard 1/4 PASS in full; the remaining shards' failures under concurrent execution in one worktree were `dist/` regeneration races — the 14 affected files rerun serially through the runner all pass (t244's two timing cases pass in isolation).
- Targeted: t243 (95 → new coexistence/convergence/refusal cases), t294, t151, t239, t275, t248, t174, t274, t304, t230, t156, t157, t266, t-active-space-includes, t146, t55, t40.
- Manual (`bun core/tools/aidlc-init.ts` against `dist-release/*`): kiro→codex add is `AGENTS.md: preserve` with identical baselines and byte-identical shipped roots; kiro+copilot refused; stripped-`shared` sibling refused under `--force` with no `.codex` created.
- Not verified here (no Kiro/Cursor/opencode/Codex binaries on this host): that each CLI actually loads the new native onboarding file. Wiring follows each harness's existing always-on mechanism (agent `resources`, steering `inclusion: always`, `.mdc` `alwaysApply`, `opencode.json` `instructions`); a live check per harness is the reviewer ask.

## Acknowledgment


apackeer:
Declining P3-3 as proposed: requiring a byte-identical block whenever a sibling lacks a descriptor "regardless of baseline" would deadlock two genuine legacy trees that both ship AGENTS.md (each refresh changes the block and names the other — the fixture that passes today); the double-fault it targets (stamp present, descriptor and baseline both lost) is a damaged ≥ 2.8.0 install, distinguishable from a legacy tree by the presence of aidlc-stamp.json, so the right fix is stamp-aware and is left for a follow-up.

Issue #1190, opened 2026-09-15
State-tracking not persisted under unit-major iteration (per-unit progress + workflow-stage transitions)


josephtsimon:
## Summary

Under unit-major Construction iteration, `aidlc-state.md` is not updated as work progresses. Two distinct facets observed in a single end-to-end run.

## Facet 1: Per-unit Construction tracking stays at `[TBD]`

- Units 4-8 completed on-disk (all artifacts written, adversarially reviewed, builds green) between Sep 8-9, 2026
- `aidlc-state.md` last written 2026-09-04 during unit 3 completion
- Per-unit Construction tracking template stayed at `Per unit: [TBD]`
- When review-pipeline code-generation returned READY, the engine looped back to re-run review-pipeline functional-design based on the stale state file
- Engine correctly refused to blindly re-run (audit-first design behavior working as intended), but the near-miss was a destructive re-run of completed reviewed work

## Facet 2: Workflow-level stage in-progress transition also fails

- After routing to build-and-test, `report --result completed` refused with `Stage is still pending`
- The orchestrator (`aidlc-orchestrate.ts`) keys on the checkbox glyph in `aidlc-state.md`
- State-file prose said "In Progress" but the checkbox was still `[ ]` (never flipped to `[-]`)
- The prose-updating path in the earlier state correction had missed flipping the workflow-level checkbox
- Only recovery: manual checkbox glyph edit — which the framework otherwise correctly discourages (`aidlc-state.ts` gate-start / approve are blocked from direct use and redirect to `report`)

## Impact

- Near-miss destructive re-run of completed work
- Tool refuses to advance under normal API when checkbox is stale
- Recovery requires operator to read orchestrator source, identify the specific checkbox glyph, and hand-edit — inconsistent with the framework's audit-first design

## Root cause

Under unit-major iteration, both per-unit stage progress AND workflow-level in-progress transitions are not written back to `aidlc-state.md`. The engine's normal `next → advance` flow that writes `[-]` does not fire on these paths.

## Suggested direction

Fix state-file write-back for unit-major iteration paths in `aidlc-orchestrate.ts` so:
- Per-unit Construction progress increments as units complete
- Workflow-level stage in-progress transitions (`[ ]` → `[-]`) persist to state
- No manual checkbox reconciliation required for the framework to advance

## Evidence

[...]

PR #1272, opened 2026-09-20
fix(doctor): compare the audit against the per-stage checkboxes

vnlebaoduy:
# Summary

`aidlc-state.md` is read by three surfaces: the orchestrator routes on its Stage Progress checkbox glyphs, `/aidlc --status` renders them on all seven harnesses, and the Claude Code statusline renders them too. The audit ledger is the authority those checkboxes are supposed to reflect.

Doctor compares the two in exactly one place. `state-audit-drift` (`core/tools/aidlc-doctor-bundle.ts:676-697`) tests one pair: a recorded `WORKFLOW_COMPLETED` against the state file's `Status=`. Nothing compares the per-stage checkboxes, so when a state write is lost the ledger and every status surface disagree in silence — doctor stays green while `--status` and the statusline under-report the run, and `report --result` refuses the stage as *"still pending"* with no diagnostic naming the cause.

That is #1190 Facet 2: after routing to build-and-test, the prose said In Progress but the checkbox was never flipped, `report` refused, and the only recovery was hand-editing the glyph — which the framework otherwise blocks (`aidlc-state.ts` gate-start / approve redirect to `report`).

The machinery to detect it already exists and is exported: `reconstructTimeline(audit, stateContent)` reads both sides, and its own comment notes that `stateContent` supplies "the checkbox for a stage whose STAGE_COMPLETED never landed". Only the comparison was missing.

Refs #1190. This closes the observability half only — see **Not in scope** below.

## Changes

`core/tools/aidlc-doctor-bundle.ts`

* `stage-state-audit-drift` (warning) — stages the current attempt recorded as started or completed whose Stage Progress checkbox is still `pending`.
* `current-stage-not-started` (warning) — `Current Stage` naming a stage whose checkbox is `pending`. This needs no audit at all, and it is exactly the state `report --result` keys on when it refuses. A hand-corrected state file reaches that shape with no `STAGE_STARTED` of its own, so the ledger comparison alone would stay silent on it.
* Three helpers: `extractCurrentStage`, `checkboxStateBySlug`, `ledgerStageActivity`.

`docs/guide/15-troubleshooting.md` — the doctor check list said "state/audit consistency"; it now says which two comparisons that covers.

`tests/unit/t243-doctor-bundle.test.ts` — seven cases, pure in-process `runDiagnosis`, no fixture or subprocess.

## The three guards, and why each is load-bearing

I had a naive version of this rule reviewed adversarially before writing it, with the reviewer asked to find false positives rather than to validate. It found three, each reproduced by driving the real tools:

| Legitimate action | Why the naive rule fired | Guard |
|---|---|---|
| Backward `aidlc-jump` | `aidlc-jump.ts:374-392` resets every downstream stage to `pending` on purpose, while their earlier `STAGE_STARTED`/`STAGE_COMPLETED` rows stay in the buffer; `reconstructTimeline` floors only at `WORKFLOW_STARTED` (`:372-383`) | floor at the latest `WORKFLOW_STARTED` **or** `STAGE_JUMPED` |
| `next --stage <s> --single` | rows carry `Workflow: single-stage:<slug>` (`aidlc-orchestrate.ts:7442-7449`) and the design deliberately leaves the main checkbox untouched (`:7380-7400`); every other consumer already filters them (`aidlc-runtime.ts:169`, `aidlc-state.ts:303-310`, `aidlc-lib.ts:25598-25610`) | drop `single-stage:` rows |
| State with repeated per-unit blocks | a last-wins slug map reads a completed stage as pending, while `setCheckbox` flips the FIRST match (`aidlc-lib.ts:23366-23378`) | resolve the slug map first-wins |

Each guard is pinned by its own test, and each is proven load-bearing by ablation: removing one turns exactly its own case red and nothing else (see Test plan).

Condition (b) — started-but-unchecked — is kept deliberately. The #1190 shape is a `STAGE_STARTED` with no `STAGE_COMPLETED`, so a completed-only rule would miss the motivating bug entirely.

Severity is `warning`, not `error`: the third false-positive class can still arise from hand-written or legacy state files, and doctor is exactly the tool run on those. Warnings stay visible and exit 0.

## Not in scope

**The writer.** #1190 Facet 1 reports per-unit Construction tracking stuck at `[TBD]`. Reading `main`, that field has a writer for its placeholder and none for its value: `aidlc-utility.ts:6713` writes the literal `Per unit: [TBD]` at state creation, `:8263` copies the existing line verbatim on recompose, and those are the only two references in `core/` or `harness/`. Nothing ever fills it in, on any path or scope — the populated shapes in `tests/fixtures/state-construction.md` (`Per unit: widget-cart`) depict a state the engine cannot produce. A doctor rule is the wrong instrument here: flagging `[TBD]` would fire on every project in Construction, since nothing can clear it. Connecting the field to the advance path is a behaviour change I did not want to pick for you.

**A latent scope-change bug the new rule surfaces.** `handleScopeChange` rebuilds Stage Progress mapping only completed / in-progress / skipped (`aidlc-utility.ts:8266-8270`), so an open gate's `[?]` or `[R]` collapses to `[ ]` with no open-gate guard. `stage-state-audit-drift` fires afterwards — correctly, since `report` then refuses. Reported here rather than fixed, because the fix belongs in the scope-change path.

**Two shipped fixtures depict an impossible state.** `state-construction-bolt1.md` and `state-initialization-done.md` carry `Current Stage` pointing at a `[ ]` stage; `current-stage-not-started` fires on both when fed to `runDiagnosis`. The engine always flips that stage to `[-]` in the same write that sets Current Stage (`aidlc-utility.ts:6726`, `aidlc-state.ts:4629`+`:4635`, `:1561-1562`, `:1888-1893`), and the replay guard at `:4580-4600` treats only in-progress / awaiting-approval / revising as started. No doctor test consumes either fixture, so nothing goes red. Left alone to keep this diff scoped.

## Checklist

* [x] Changes are documented

## Test plan

```
bun scripts/package.ts
bun scripts/package.ts --check                 # deterministic across 7 harnesses
bun tests/gen-coverage-registry.ts --check     # fresh, guards green, ratchet held
bun tests/run-tests.ts --smoke --unit --integration
```

Run on an idle machine (the unit tier runs serially within one invocation, so a second concurrent runner reds unrelated files): **444 files, 2 failed**. Both fail on unmodified `origin/main` too, verified today by running each alone on this branch and on a clean `main` worktree:

| File | this branch, alone | clean `main`, alone |
|---|---|---|
| `t19` | FAIL | **FAIL** — live-SDK preflight, no credentials on this machine |
| `t78-bolt-worktree-lifecycle` | FAIL | **FAIL** — `restore materializes a large index exceeding one MiB` times out at 60000 ms here |

`t243` is green in the full run and alone.

**Fail-without-fix.** Tests in place, only `core/tools/aidlc-doctor-bundle.ts` reverted: t243 **34 pass / 3 fail** (26, 27, 31). Restored: **37 pass / 0 fail**.

**Each guard is load-bearing.** Reverting one guard at a time, everything else intact:

| Guard removed | Red |
|---|---|
| floor at `STAGE_JUMPED` | 28 — backward jump |
| `single-stage:` filter | 29 — isolated run |
| first-wins slug map | 30 — repeated per-unit blocks |

No other test moved in any of the three ablations.

**Empirical false-positive sweep.** Every `tests/fixtures/state-*.md` fed through `runDiagnosis`: only the two fixtures named above fire, for the reason given there.

## Acknowledgment


vnlebaoduy:
## Update — #1264 folded in

I had opened this and #1264 as separate PRs. Both change only `core/tools/aidlc-doctor-bundle.ts` and both are about the same thing — what `runDiagnosis` reports — so reviewing them apart meant opening the same file twice for one contributor. #1264's commit is now the second commit here and I have closed it.

**What came from #1264:** both runtime-graph findings told the user to re-run `graph compile`. `aidlc-graph.ts` compiles `stage-graph.json` and `scope-grid.json` and contains no reference to `runtime-graph` at all; the file has one writer, `aidlc-runtime.ts compile`, which is what the rebuild hook dispatches. The remedy named a real command that could not change the condition that printed it, and `runtime-graph-missing` fires exactly where a first-stage user lands. That is step 5 of #1164. Its test is `23b`.

Also fixed while merging: my drift tests were numbered from 26, but `t243` on `main` already has a test 26. They are now 27-33, so no number is used twice in the file.

Gates re-run on the combined branch at `main` @ 839fa1f3: t243 **38 pass / 0 fail**, lint, typecheck, coverage registry and `package.ts --check` all green.

Closes #1164 as well as the Refs #1190 above.

PR #1282, opened 2026-09-21
fix(ci): harden issue review and move PR judge to Sol

leandrodamascena:
## Problem

The post-push AIDA reviews found trust, completeness, lifecycle, and size-limit gaps in the PR and Issue reviewers. A review could execute proposed reviewer controls with privileged credentials, mistake changed-file text for trusted evidence framing, omit older Issue comments from prompt-injection inspection, authorize non-maintainer deletion triggers, or fail permanently on large mandatory evidence such as `CHANGELOG.md`.

## Result

- Moves Issue authorization into a preflight job so rejected events never enter the per-Issue review concurrency group.
- Uses one effective Issue trigger rule: maintainer-owned issues accept maintainer activity, while external issues require the `ai-review` opt-in label; opted-in conversations accept later human comments, including comments that imitate the AIDA marker.
- Includes every human Issue comment in both model context and publication identity, excludes bot comments, and fails before model execution when the complete human conversation exceeds the explicit 1,000,000-byte bound.
- Requires deletion triggers to satisfy the same authorization rule as other activity. A non-maintainer deleting a comment from an unlabeled maintainer-owned Issue cannot start an expensive review.
- Persists the authorization basis, serializes per-Issue publication, and rechecks Issue metadata, authorization, and complete conversation identity before model execution and before and after publication, with rollback.
- Fetches up to 200 actual Issues through `search/issues` with `is:issue`, avoiding PR-heavy catalog pages.
- Schema-binds the Sol bug classifier and keeps deterministic validation of its selected base-tree tests.
- Moves both PR and Issue final judges to Sol. Fable remains responsible for User Experience and Direction.
- Runs both Sol judges without command, network, browser, app, multi-agent, or file tools.
- Replaces delimiter-concatenated judge input with one JSON evidence envelope. Every record carries explicit provenance, revision/path, encoding, source byte length, and SHA-256; arbitrary file content is JSON-escaped and cannot create a sibling trusted record.
- Keeps the aggregate evidence limit at 2,000,000 bytes. Context text is split into records, binary content uses authenticated metadata, and oversized changed or cited files use bounded line excerpts. Missing line locations produce metadata only and cannot support a finding.
- Runs both PR and Issue evidence preflights before model execution. Large release changes and large diffs degrade to bounded changed-line evidence instead of failing solely because a required file exceeds 400,000 bytes.
- Runs the privileged PR entrypoint through `pull_request_target`, rejects forks, checks out only the default branch, and always sources prompts, validators, and runtime setup from that trusted revision. It never checks out executable PR code.
- Handles the one-time builder bootstrap before the file exists on the default branch by running only that proposed builder once as a disposable user with an empty environment, read-only workspace, isolated network namespace, and timeout. Its validated JSON is reused after credentials are assumed; proposed code does not execute with provider or publication credentials.
- Removes the 20 MiB file-size limit from dependency installation while retaining it for package and test execution, and publishes a bounded HTML-escaped diagnostic for setup failures and timeouts.

## Validation

- `bun test tests/unit/t301-ai-issue-review.test.ts tests/unit/t300-ai-pr-review.test.ts` — 46 tests, 607 assertions.
- `bunx tsc --noEmit`
- `bunx biome check .github/scripts/build-ai-review-evidence.ts .github/scripts/ai-issue-review.ts tests/unit/t300-ai-pr-review.test.ts tests/unit/t301-ai-issue-review.test.ts`
- `markdownlint-cli2 CONTRIBUTING.md .github/prompts/ai-pr-review-common.md .github/prompts/ai-pr-review-judge.md .github/prompts/ai-issue-review-common.md .github/prompts/ai-issue-review-judge.md`
- Parsed both changed workflows as YAML and ran `git diff --check`.
- Verified forged evidence delimiters remain data inside one changed-file record and cannot create trusted records.
- Verified more than 20 MiB of changed snapshots produces bounded metadata and changed-line excerpts instead of an evidence failure.
- Built the real immutable PR #1282 judge envelope at `0b4994470`: 687,986 bytes of the 2,000,000-byte limit, including bounded evidence from the 632,407-byte `CHANGELOG.md`.
- Live structured-output transport with Codex `0.151.0`, `openai.gpt-5.6-sol`, Bedrock, profile `aidlc-v2-github-ai-review`, and both production judge schemas.
- Live no-tool Sol smoke using the same profile: only `agent_message` events, no tool events, `toolAccess: false`, and no credential patterns in the captured transport output.

Bootstrap note: this push intentionally stops the legacy `pull_request` self-review path. The new `pull_request_target` entrypoint can begin running only after its trusted workflow definition exists on the default branch; the current head is therefore validated by deterministic CI and the prior AIDA findings rather than by executing proposed privileged workflow code.

Follow-up to #1280.


leandrodamascena:
Closing this PR because the implementation accumulated too many coupled changes and is no longer a safe or clear unit to review.

The objective, confirmed behavior, trust boundaries, model responsibilities, and acceptance criteria have been preserved in #1292. We will restart the implementation from that issue in smaller reviewable changes. This PR remains available as investigation history and should not be merged.

PR #1295, opened 2026-09-21
fix(claim-sources): keep a definition-shaped line inside a paragraph visible

apackeer:
Follow-up to #1156 (found while validating the confirmation-side block splitter).

## Problem

`referenceAnalysis` in `core/tools/aidlc-sensor-claim-sources.ts` tried to parse a link reference definition at every line. CommonMark §4.7 does not allow a definition to interrupt an open paragraph, so a definition-shaped line written directly under prose or under a list item's text (no blank line between) is that paragraph's lazy continuation and renders as visible text:

```markdown
- The reviewer may be needed for imports. [assumption]
[note]: https://example.test
```

The sensor treated line 2 as a definition. Two observable consequences:

- The visible text was blanked out of the claim block, so the deliverable's assumption was compared without content the reader sees (and a confirmation entry with the same wrapped text could not match it).
- A `[Q1]`-style tag elsewhere in the document was demoted from a visible tag to a resolved link when a lazy-continuation line happened to look like `[Q1]: /url`, so the claim was reported as having no source tag.

## Fix

`referenceAnalysis` now tracks whether a blank-terminated block is open while scanning the container lines and only attempts `referenceDefinitionAt` when no block is open. Blank lines, ATX headings, thematic breaks, setext underlines, and a new container (next list item, deeper block quote) close the block; paragraph text, HTML block content, and table rows keep it open; indented code neither opens a block nor hosts a definition; consecutive definitions still chain. The context predicate from `canContinueReference` is extracted as `continuesContext` and shared. `referenceDefinitionAt`, `documentContainerLines`, and `claimBlocks` are unchanged.

`core/sensors/aidlc-claim-sources.md` gains one sentence stating the rule.

## Tests

`tests/unit/t247-claim-sources-sensor.test.ts`:

- two negative cases (list item and top-level prose followed directly by a definition-shaped line) that fail on the current recognizer with `claim block has no source tag` and pass with the fix;
- one positive guard: consecutive definitions after a heading are all definitions;
- the "Markdown reference metadata" hidden-tag fixture gains the blank line that makes its definition a definition under CommonMark (without it, the title line is visible prose and its tags count).

Validation on this head: t247 96/96, t92 claim-sources cases 2/2, `bun run lint`, `bun run typecheck`, `bun scripts/package.ts --check` deterministic across all seven harnesses.

No version, badge, or CHANGELOG change per the release metadata policy.


apackeer:
Maintainer decision on the AI review of f1109a26 (P1, two triggers):

**Ordered list after an open paragraph (`2. [note]: /url`)** — accepted and fixed. `referenceAnalysis` now applies CommonMark §5.3 before resetting block state: a list may interrupt a paragraph only when it is a bullet list or starts at 1 (compared numerically, so `01.` counts as 1); a new block quote always interrupts; a rejected marker is treated as paragraph continuation and does not persist its list context, so `2. …\n3. [note]: /url` stays one paragraph. Regressions are isolated under `## Review` so the observable is the document-wide label alone.

**Definition-shaped line directly after a pipe table row** — not accepted as a false pass. Under GFM §6.9 a table is broken only by a blank line or the start of another block-level structure; a link reference definition is not a block start (it is parsed out of paragraph content), so a pipe-less `[Q1]: /url` line directly under the last row is a one-cell table row (the spec's own example turns a bare `bar` line into a row; markdown-it's table rule likewise has no definition terminator). The sensor therefore correctly leaves `[Q1]` in the rows as a literal tag. In the sensor's block model that trailing line is reported as an untagged claim block, which is a false failure the manifest permits, not a false pass. Pinned by the Review-isolated table regression in t247.

PR #1297, opened 2026-09-21
refactor(kiro): serve Kiro IDE and Kiro CLI from one distribution

wowzoo:
# Summary

Serve Kiro IDE and Kiro CLI from one distribution instead of two.

Both surfaces already run the same agent runtime and read the same `.kiro/` tree, so the repository shipped two rows — `kiro` and `kiro-ide` — that declared the same engine directory and, since #1268, the same native onboarding path. This retires the `kiro-ide` row and serves both surfaces from `kiro`, while keeping every install made by the retired row readable and upgradable.

**This reopens #1157 under a new number, and only because of where the branch lived.** #1157's head is on a fork, and the AI reviewer is same-repository only (`AI review is disabled for forks` in `ai-pr-review.yml`'s context step — `workflow_dispatch` does not get past it either), so that PR received zero AI reviews in six days. The branch content is identical to #1157's current head; nothing was dropped to escape a review. The two 2026-09-15 `CHANGES_REQUESTED` reviews are accounted for below, finding by finding.

## Changes

`harness/kiro-ide/` is removed — manifest, onboarding fills, hook adapter, agent surfaces — and `harness/kiro/` serves both. The unified row declares `harnessDir: ".kiro"`, `AGENTS.md` with `shared: "identical"`, and `onboarding.harnessDst: "steering/aidlc-onboarding.md"`, so its root block is byte-identical to codex, cursor and opencode while its own setup renders into the native steering file. `.vscode/settings.json` stays as a native root integration for the IDE surface.

A retired `kiro-ide` id stays readable wherever it was persisted. `RETIRED_DISTRIBUTION_SUCCESSOR` in `core/tools/aidlc-runtime-paths.ts` maps it one-directionally to `kiro`, and `currentDistribution()` / `distributionUpgradesTo()` are applied at the readers that resolve a current runtime identity — project stamps, machine `default-harness`, settings keys, config diagnostics, doctor — while the readers that check a historical projection's internal consistency keep the raw value on purpose. The unified manifest also carries the retired row's shipped root-file hashes, so an unmarked `AGENTS.md` or `.gitignore` written by that row is still recognised as framework-owned instead of taking the ambiguity branch.

Absorbing #1268 is most of the diff's recent movement. That change made the root onboarding block harness-neutral and moved every per-harness section into a native file, so this branch takes upstream's neutral `core/templates/onboarding.md` wholesale, folds its own harness prose into the new `onboarding-harness.md` slot set, adds the `frontmatter` slot (`inclusion: always`, which the IDE needs and the CLI already loads through the conductor's steering resource glob), and drops the two slots the new template no longer declares.

## User experience

Before: a user chose between two distributions that wrote the same directory, and the two could never coexist in one project. An install stamped `kiro-ide` had to be reconfigured with an explicit `--harness kiro`.

After: there is one Kiro row. `aidlc config --harness kiro` works on a fresh project and on a project stamped by the retired row, and the root `AGENTS.md` block is shared with the other harnesses that declare it, so a Kiro install can sit beside one of them whenever their engine directories differ. Harness-specific setup moved to `.kiro/steering/aidlc-onboarding.md`, which the IDE loads through its `inclusion: always` frontmatter and the CLI through the `aidlc` agent's steering resources.

## Checklist

* [x] Changes are documented

## Test plan

Local verification on this head:

- `bun scripts/package.ts` — exit 0. The native onboarding file is emitted at `dist/kiro/.kiro/steering/aidlc-onboarding.md`, and `dist/{codex,cursor,kiro,opencode}/AGENTS.md` are byte-identical, so `assertIdenticalRootIntegrations` passes for the unified row.
- `bun run typecheck` — 0 errors across all three projects.
- `bun tests/gen-coverage-registry.ts --check` — `OK (fresh, guards green, ratchet held)`.
- Targeted: `t151` (7), `t148` (16), `t156` (10), `t174` (3), `t243` signature group (5), `t281` (10) — all pass.

Reviewer-facing checks worth running, because each pins a roster this change narrows:

- `tests/unit/t151-onboarding-skeleton.test.ts` — the identical-sharing roster is four here, not five.
- `tests/smoke/t148-kiro-file-structure.test.ts` — `.kiro/steering/` ships exactly the active-memory pointer and the native onboarding file.
- `tests/unit/t243-install-mechanism.test.ts` — the legacy-signature inventory and the retired-stamp migration cases.
- `tests/unit/t281-kiro-mcp-registry.test.ts` — the MCP server-count claim, now read from the native onboarding file.

CI on the fork's last run of this content found one real defect and it is fixed in this branch: `t281` asserted the MCP claim in the root `AGENTS.md`, which #1268 emptied of per-harness prose. One unrelated failure on that run was a 5,000 ms timeout overrun in `t339`, a test this branch does not touch.

## Prior review carried forward

Both 2026-09-15 reviews on #1157 raised three findings each. All six are addressed on this head:

- **P1 — preview publication still invokes the retired distribution** (`preview-release.yml:440`, `:574`). `kiro-ide` no longer appears anywhere in that workflow.
- **P1 — previously valid model settings prevent migration** (`aidlc-settings.ts:199-212`). `normalizeHarnessModelMap` now resolves a retired key through `currentDistribution`, defines the conflict behaviour explicitly (the current key wins; the retired one is residue), and applies the retired value after the loop so the outcome does not depend on key order.
- **P2 — persisted machine defaults are not migrated** (`aidlc-init.ts:4170`). The machine `default-harness` comparison goes through `distributionUpgradesTo`, so a machine retaining `kiro-ide` configures a fresh project without an explicit `--harness`.
- **P2 — read-only configuration and the interactive setup walk reject retired project stamps.** Both normalize now: `aidlc-config-diagnostics.ts` maps the stamp before the model-access decision, and the setup walk resolves a stamped distribution through `currentDistribution`.
- **P3 — plugin documentation still calls `.kiro` ambiguous** (`docs/harness-engineering/10-authoring-a-plugin.md:577`). The remaining mention is a shell comment naming both surfaces (`# open in Kiro IDE or kiro-cli chat`), which is accurate for one row.

If any of these readings is wrong, I would rather hear it here than have it inferred from the diff.

## Acknowledgment


wowzoo:
Thanks for the review. F21 and F19 are fixed below, each with a test that fails without the change. For F2 and F1 I'm asking for a maintainer decision rather than a change, and the reasons are below. My earlier reply to the d0899aba review was never posted, so the measurements it would have carried are included here.

## Fixed

- **F21 (P1): a delegate could rewrite Kiro instruction resources** (`abd9c58b`). For `.kiro`, the runtime-integrity hook now protects `agents/`, `settings/`, `steering/`, `skills/`, `knowledge/` and `aidlc-common/` as whole trees, alongside the hooks, the shipped tools and the session records it already covered. I protected whole trees rather than the shipped file names, because Kiro loads every steering file in the directory: a new file injects instructions as well as an edited one does. No stage or protocol writes these trees through a tool call; the active-memory steering file is written by the engine process. The paths stages do write stay open: `scopes/`, `sensors/` and `tools/data/scope-grid.json`. t242 adds the new paths for the write tools and for shell mutations, and adds the open paths as expected passes. The new refusals fail against the previous classifier. The Kiro guide, the onboarding text and the hooks reference name the protected trees, and the guide says the user edits their own steering, skills and agents there by hand.
- **F19 (P2): the boundary still refused commands outside AI-DLC's grants** (`480d57fa`). The previous fix used a path-prefix reading, which was wider than the grants. The packager now bakes the emitted grant strings into the adapter, and the boundary matches the command against those grants directly, with `*` matching any text. A command that no grant matches goes to Kiro's prompt. That covers `bun .kiro/tools/aidlc-helper.js`, `bun run .kiro/tools/aidlc-helper.js --watch` and `bun .claude/tools/aidlc-x.ts`, with or without a redirection. A command that a grant does match but that names no shipped tool is still refused. For example, `bun .kiro/tools/aidlc-helper.js --out x.ts y` is refused because `bun .kiro/tools/aidlc*.ts *` reaches the `.ts` in an argument. The new permitted cases in t218 fail at the parent.

## F2 (P0) and F1 (P1): please decide these as accepted risk for the source-copy channel

Both findings assume that the project's `.kiro/tools/*.ts` files have been replaced with hostile code. F2 reaches that code through a pre-approved invocation. F1 uses it to rewrite delegation state.

On the source-copy channel, those tool files sit in the same project `.kiro/` directory as the three things that decide what they may do:

- the permission rules that pre-approve them (`.kiro/agents/*.md`);
- the hook that would check them (`.kiro/hooks/`);
- the settings that load both.

Whoever can plant `.kiro/tools/aidlc.ts` can equally rewrite the grant or the hook. A digest check would run in project-resident code, against a list stored in the project, so the party it is meant to stop could edit it. I don't think an in-repository integrity binding can close this, and adding one would advertise a protection that isn't there. The guide already says so: on a source copy, pre-approving a tool means pre-approving that file as the project currently holds it.

What this PR does control is whether a session can turn into that party. I measured this on Kiro IDE, with the conductor delegating to the developer persona under Autopilot:

- **Persona rules do not reach a delegate.** Kiro IDE does not apply a persona's own `permissions` when the persona runs as a delegate. With the persona's `fs_write` deny set to `**`, the delegated developer still wrote `src/` and `.kiro/tools/`. A persona `rm -r*` deny did not stop `rm -r probe-dir` either; Kiro asked for approval instead. The same rules do apply when the persona is selected as the agent. So the persona rules are not the boundary for delegated work.
- **The runtime-integrity hook does reach a delegate.** It runs on every agent's tool calls, delegates included. In the same run it refused the delegate's writes to `.kiro/tools/aidlc-probe.ts`, `.kiro/agents/aidlc.md` and `.kiro/settings/permissions.yaml`, and none of those files changed on disk.

The hook refuses file-tool writes, and recognised shell mutations such as redirection, `cp`, `mv`, `tee` or `sed -i`, when they target the shipped tools, the hooks, the session records or any of the `.kiro` trees listed under F21. That set includes the grants themselves. Pre-approved commands must also be one simple command that names a shipped tool directly. So the tampered-tools premise requires someone outside the session to have written the repository. That is the trust decision a user makes when opening it in Kiro, and AI-DLC's grants cannot reverse it.

On the relay half of F2, terminal output reaches the model only as a file path, never inline. The model reads it the same way it reads any other file in the repository.

A host-owned dispatcher is the direction that removes the tool code from the project, and the native channel's installed `aidlc` binary already is one. Moving delegation identity into host-owned state belongs with it. I'd take both as a follow-up rather than widen this PR. The residual limits are already pinned as tests (`5d17`–`5d19`), so they can't be mistaken for closed.

## Verification

For each commit I ran locally: `scripts/package.ts`, `typecheck` (0 errors), `lint` (0 warnings), and the affected tests. Results on `480d57fa`: t218 133/0, t243-install-mechanism 120/0, t147 87/0, t242-state-transition-guard 44/0, t243-doctor-bundle 31/0, t242-plugin-state 27/0, t252 9/0. Each new assertion was shown to fail against the code it replaces.

wowzoo:
Thanks for the review. The CI failure on `480d57fa` (t174) came from my own guide wording and is fixed in `9d4ec8b7`. Three of F21's four parts are fixed in `28103713`, with tests that fail without the change. For the root `AGENTS.md` part I'm proposing to leave the file writable, for the reasons below. F2 and F1 are unchanged.

## F21 (P1): fixed

The runtime-integrity hook is what bounds a delegate on Kiro, so each of these is a hook change (`28103713`):

- **Active-space memory.** On a Kiro install, the hook now refuses tool-call writes, moves and removals under `aidlc/spaces/<space>/memory/`, and removal of any directory that holds it. The conductor loads that memory as a resource, and the shipped steering includes it in every session, so it is standing instruction to every agent. You asked for mediated exceptions for legitimate updates. Those updates never come through a tool call: the learnings ritual (`aidlc-learnings.ts`) and the practices promotion (`practices-promote`) write memory from the engine process, which this hook does not see. The method routes memory updates through those tools and tells the orchestrator not to edit memory directly (`practices-discovery.md` Step 7, and the language-persistence rule in `core/memory/org.md`). So no exception had to be carved out, and no workflow step that writes memory today is refused.
- **`.vscode/settings.json`.** Protected on a Kiro install, because the IDE reads `kiroAgent.trustedCommands` from it. This matches how Claude's `settings.json` is already treated.
- **Links.** The shared shell target extractor now handles `ln` and `link`. A link counts as a write at its new name. A hard link also counts as a write at its source, because a later write through the new name changes the source's content and no realpath can reveal that. `cp -l` is handled the same way. A symbolic link's source is not treated as a target, because a write through the link already resolves to the real path, which the hook classifies. Review-freeze and the plan-approval guard use the same extractor, so they now treat links as writes too.

t242 adds three tests. They cover direct writes through all four write tools, shell renames and removals, symbolic and hard links into protected trees, hard links out of them, and a write through an existing symbolic-link alias. They also check that the unprotected cases still pass: a project without Kiro, intent records, other `.vscode` files, and ordinary links. All three tests fail against the previous hook and extractor. The Kiro guide, the onboarding text and the hooks reference now list the new locations. The guide also says the user edits memory and `.vscode/settings.json` directly, and that memory updates from a workflow still land.

## F21 (P1), root `AGENTS.md`: proposing to leave it writable

Protecting the root `AGENTS.md` would not close this route. It would, however, refuse a common request on every Kiro project.

- **Kiro loads the file by itself, not because AI-DLC asks it to.** Kiro's steering documentation says a workspace-root `AGENTS.md` is picked up automatically and always included. It also says `AGENTS.md` files are discovered in subdirectories throughout the workspace, and each one is loaded as steering. In the Kiro IDE delegate runs I measured for my previous reply, the delegated developer's panel listed `AGENTS.md` under Included Steering, even though that persona's `resources` do not name it. So removing `file://AGENTS.md` from the conductor's resources would not stop Kiro from loading it.
- **Protecting only the root file leaves the rest open.** A delegate could create `src/AGENTS.md` and have it loaded the same way. Closing the route completely would mean refusing writes to every file named `AGENTS.md` anywhere in the workspace.
- **That would refuse ordinary work.** "Update AGENTS.md" is a routine request, and a Code Generation stage building a project may create one. The hook's protected set is AI-DLC's own grants, instructions and records. `AGENTS.md` is the user's project instruction file, and Kiro loads it whether or not AI-DLC is installed.

A delegate's edit to `AGENTS.md` is an ordinary workspace change, and it appears in the diff the user reviews. The guide gives the same advice for other delegated writes. If you'd still prefer to refuse agent writes to every `AGENTS.md`, I can add that. I'd rather have that decided explicitly than slip it in as a side effect of F21.

## F2 (P0) and F1 (P1)

My reply to the previous review ([comment](https://github.com/awslabs/aidlc-workflows/pull/1297#issuecomment-5819778391)) still describes my position: on the source-copy channel, anyone who can replace `.kiro/tools/*.ts` can also rewrite the grants and hooks that would check it. For that reason I'm asking for a maintainer decision on these two findings rather than an in-repository integrity binding. The only change since then is that the set the hook refuses is now larger, as described above.

## Verification

For each commit I ran locally, in this order: `scripts/package.ts`, `typecheck` (0 errors), `lint` (0 warnings), then the affected tests. Results on `28103713`: t242-state-transition-guard 47/0, t218 133/0, t334 98/0, t265 65/0, t250 40/0, t264 29/0, t-guard-plan-continuation-swarm 21/0, t241 17/0, t174 3/0. t276 is 39/1. The single failure is in test 23, and it fails the same way locally on main at `20008a5d`.

PR #1314, opened 2026-09-22
fix: always sort audit rows before deriving the stage run floor

### LINKED ISSUE #1313: [Bug]: unit lifecycle deadlocks once an intent has more than one audit shard: the run-floor reader skips the cross-shard sort the writer applies
### Description

Once an intent's `audit/` directory holds more than one shard, the per-unit lifecycle reader and writer can disagree about the current stage attempt, and every per-unit Construction stage deadlocks.

`aidlc-state.ts unit start` succeeds, then `unit complete` for the same unit is refused with:

```
it is not the active unit (no unit is active — start it first)
```

`aidlc-state.md` still shows `Active Unit` and `Unit State: in-progress`, and `/aidlc --doctor` reports every check passing, so the message sends the operator looking at the wrong thing.

### Root cause

`latestMainWorkflowStageRunFloorForProject` (`core/tools/aidlc-lib.ts`) passes `auditRows !== undefined` into `latestMainWorkflowStageRunFloorFromRows` as a `preSorted` flag. It assumes that a caller who supplied rows also sorted them. That is not true for `readAuditShardEvents`, whose documented contract is to return rows shard-major (filename order, append order within a shard) and leave the `(timestamp, shardIndex, pos)` tie-break to the caller.

The two sides of the unit lifecycle therefore compute the floor differently:

* Writer: `aidlc-state.ts unit start` calls the function without rows, so the rows are read and sorted, and the receipt is stamped with the correct floor.
* Reader: `currentUnitLifecycleRows` reads the shards itself and forwards the raw rows, so the sort is skipped and the floor is taken from whichever boundary event happens to sit last in the last-read shard.

With one shard the two agree by luck, because filename order is chronological order. With two or more shards, the chronologically latest boundary is often not in the last-read shard. On the workspace where this was found the reader landed on a `WORKFLOW_STARTED` from eleven days earlier. No `UNIT_STARTED` receipt carries that floor, so `activeUnitCheckpoint` returns `null`, `unit complete` refuses, and `unit start` never objects to a duplicate either.

A second shard appears whenever the same intent is driven from a second clone or machine (the clone id file is gitignored, so a fresh clone starts a new shard). That is exactly the case the shard design exists for, and it is not exercised by a single-developer run, which is why this survived.

The swarm unit-convergence flooring routes through the same function with caller-supplied rows and has the same defect.

### Steps to Reproduce

1. Seed an intent whose own shard holds a `STAGE_JUMPED` boundary at T2, and a second shard, later in filename order, holding a `WORKFLOW_STARTED` at T1 < T2.
2. Compute the floor both ways:

   ```ts
   latestMainWorkflowStageRunFloorForProject(proj, "functional-design", true);
   latestMainWorkflowStageRunFloorForProject(proj, "functional-design", true, undefined, readAuditShardEvents(proj));
   ```

3. Run `aidlc-state.ts unit start --stage functional-design --unit unit-a`, then `unit complete` for the same unit.

### Expected Behavior

Both calls return `STAGE_JUMPED:<T2>#1`. `unit complete` commits.

### Actual Behavior

The first call returns `STAGE_JUMPED:<T2>#1`, the second returns `WORKFLOW_STARTED:<T1>#1`. `unit complete` is refused with "no unit is active".

### AI-DLC Version

v2 (current on main)

### Release / Commit

Observed on 2.7.1. Still present on `main` at `30313035`.

### AI-DLC Phase

Construction (per-unit stages, unit lifecycle receipts)

### Harness

Claude Code, but harness-neutral: the defect is in `core/tools/aidlc-lib.ts`.

### Additional Context

Suggested fix: drop the `preSorted` parameter and always sort inside `latestMainWorkflowStageRunFloorFromRows`. The `.filter()` ahead of the sort already returns a fresh array, so no caller's rows are mutated, and sorting already-ordered input is a no-op, so no caller that was correct changes result.

Two smaller follow-ups worth considering separately: have `unit complete` name the floor mismatch instead of reporting "no unit is active", and have it reconcile against the `Active Unit` fields in `aidlc-state.md`, which were correct the whole time.

### Acknowledgment

By submitting this issue, I confirm I can share this contribution under the project license.

mahezsh:
# Summary

Always sort audit rows before deriving the main-workflow stage run floor, so a caller that passes in rows from `readAuditShardEvents` gets the same floor as a caller that lets the function read them. Fixes #1313.

## Changes

* `latestMainWorkflowStageRunFloorFromRows` drops its `preSorted` parameter and sorts unconditionally. `latestMainWorkflowStageRunFloorForProject` no longer derives `preSorted` from `auditRows !== undefined`, which assumed that supplied rows were sorted rows. `readAuditShardEvents` returns rows shard-major by contract, so that assumption failed for `currentUnitLifecycleRows` (unit lifecycle reader) and for the swarm unit-convergence flooring, both of which forward raw rows.
* The `.filter()` ahead of the sort already returns a fresh array, so no caller's rows are mutated, and sorting already-ordered input is a no-op, so no caller that was correct changes result.
* Regression test in t260: two shards where the chronologically latest boundary (`STAGE_JUMPED`) sits in the shard read first and an older `WORKFLOW_STARTED` sits in the shard read last. Asserts the writer-path and reader-path floors agree, then drives the real `unit start` and `unit complete` through `aidlc-state.ts` and asserts completion commits. Fails on `main` (reader returns `WORKFLOW_STARTED:...#1`), passes with the fix.
* No version, badge, or CHANGELOG change, per the Release Metadata Policy.

## User experience

Before, on any intent driven from two clones (two files under `<record>/audit/`), every per-unit Construction stage deadlocks:

```
aidlc-state.ts unit start --stage functional-design --unit unit-a     → ok
aidlc-state.ts unit complete --stage functional-design --unit unit-a  → refused:
  it is not the active unit (no unit is active — start it first)
```

while `aidlc-state.md` shows `Active Unit: unit-a` and `Unit State: in-progress`, and `--doctor` passes.

After, `unit complete` commits and the workflow advances. Existing audit history needs no repair: the receipts `unit start` already wrote carry the correct floor and become visible as soon as the reader agrees with the writer.

## Checklist

## Test plan

* `bun test tests/unit/t260-unit-lifecycle-receipts.test.ts` with and without the `aidlc-lib.ts` change: the new test fails before, passes after.
* `bun test` on t201, t260, t272, t305, t340, t341, t342, t343, t344 (every suite that covers the run-floor functions) is green.
* `bun run check` (package `--check`, typecheck, lint) and `bun tests/gen-coverage-registry.ts --check` are green. The one lint info in t260 reproduces on untouched `main`.

## Acknowledgment


apackeer (approved):
Thanks @mahezsh, this is a tight fix with a regression test that reproduces the deadlock, and the write-up in #1313 made it quick to verify. Approving. The maintainers will now prepare this for merge.

PR #1305, opened 2026-09-22
feat(aida): findings ledger with /aida maintainer commands

leandrodamascena:
## Problem

AIDA re-derives every finding from scratch on each head and has no verified channel for a maintainer's decision. In practice a finding a maintainer rejected came back reworded on every push, and a maintainer comment saying "this is the accepted contract, do not report it again" was never seen: issue comments were not part of the review context, and even the decision language already in the prompts could only act on data that never arrived. #1303 tried to fix this with a convergence contract written as prompt prose and could not converge — the reviewer re-litigated the prose on every head.

This PR gives AIDA durable memory and a verified decision channel **without changing any decision rule**. It is part 1 of the follow-up to #1301.

## What changes

**Findings ledger** (`.github/scripts/ai-pr-ledger.ts`, new). One bot-owned comment per PR, created on the first review and edited in place after each one. Every finding gets a stable id (`F1`, `F2`, …) anchored to **content hashes of the lines it cites**, not to titles or line numbers, so a follow-up review recognizes the same finding across commits. The comment carries a human-readable table plus the JSON, and a digest of that JSON: a hand edit is detected, its decisions are not honored, and they are reset with an audit event until a verified command restates them.

**Maintainer commands** (`.github/workflows/ai-pr-ledger.yml`, new). `/aida accept F# <reason>`, `/aida reject F# <reason>`, `/aida reopen F#`, `/aida status`, `/aida full` as PR comments.
- Authority is **repository write permission**, checked through `GET /repos/{repo}/collaborators/{user}/permission` on every command (`write`, `maintain`, or `admin`). `author_association` is deliberately not used.
- The workflow is isolated from the review workflow on purpose: `issue_comment` trigger, `contents: read` + `pull-requests: write` only, no model credentials, no cloud role, default-branch checkout only, `created` events only, bot authors ignored. (t300 pins the review workflow against `issue_comment`; that pin still holds.)
- Reactions report the outcome: 👍 applied, 👎 no write access (nothing else happens), 😕 usage error with a short reply.
- **P0/P1 findings can be accepted but not rejected**: a blocking finding a maintainer takes on stays visible under *Accepted risks* with their name and reason; it cannot be silenced.

**Review workflow integration** (`ai-pr-review.yml`, `ai-pr-review.ts`).
- The ledger is fetched **after** the context digest, so a `/aida` decision never forces a new full review of an unchanged head; `/aida status` re-renders instead.
[...]

PR #1316, opened 2026-09-22
feat(aida): judge dispositions bind restatements by id; security lenses emit structured evidence

leandrodamascena:
Fourth PR of the AIDA review-stability series (#1305 ledger, #1308 identity and refresh, #1312 incremental scope). This one targets the growth pattern visible on #1312: the judge restated open findings under new ids while the old ids stayed retained, and the exemption of security findings from the scope depended on a regex over lens prose.

## Dispositions (the structural fix for duplicates)

The judge's output gains a required top-level `ledger` array with one entry per **open** ledger finding:

- `{"id": "F3", "disposition": "still-open", "findingIndex": 0}` — the defect still holds and `findings[0]` is its restatement. The validator binds the id to that finding, so the judge no longer has to carry `ledgerId` per finding, and the restatement may use new wording and new lines after a partial fix.
- `{"id": "F3", "disposition": "still-open", "findingIndex": null}` — still holds, not restated: kept verdict-bearing.
- `{"id": "F5", "disposition": "resolved", "findingIndex": null}` — the head corrected it: closed at this head with an auditable event, even when the exact cited lines are unchanged (a fix can live elsewhere).

An explicit id binds an **open** entry outright (the worst case of a wrong id is a defect that stays visible under an older id). Accepted and rejected entries remain unreachable from model output. An open entry the judge leaves undisposed falls back to today's presence-based retention and the review says so ("N open entries left undisposed by the judge"). Contradictions (an id disposed twice, resolved-and-restated, one finding bound to two ids) are validation errors.

## Structured security evidence (kills the F2 → F9 → F11 lineage of #1312)

The security and prompt-attack lenses now answer under `ai-pr-review-lens-schema.json` — `marker`, `status` (`complete` | `blocked`), and `candidates` with the same evidence shapes as the final review — into `.ai-review-lenses/prompt-injection.json` and `security.json`. The workflow checks the marker with `jq`. The validator reads the cited lines and files from that JSON; a finding that cites any of them is never deferred by the incremental scope, whatever category the judge assigns. The regex over markdown is gone.

## Convergence property

The identity guarantee this PR implements is deterministic and tested (`t347`): a defect an open ledger entry already names keeps that id across heads, however its wording or cited lines change, and a disposition of `resolved` closes a P0/P1 only with change evidence. The intended effect over a PR's life is that fix pushes do not grow the set of open findings with duplicates of entries already named.

## Verification

- `tests/unit/t347-aida-dispositions.test.ts` (new): binding and every contradiction; reconcile with resolved / still-open / undisposed; end-to-end restatement under new wording keeping its id, a declared fix resolving, and the rendered summary; schema and prompt pins.
- `t346` updated for the JSON lens outputs and workflow wiring; `t345` and `t300` pins updated (`ledger` in the schema's required list, `.json` lens files).
- Lint, typecheck (after `bun scripts/package.ts`), coverage registry, YAML.


leandrodamascena:
/aida accept F2 F3 F6 F8 F9 merged; F2, F3, F6 and F9 were fixed in c9dd61cc0 (merged as 3c54ec1ac) before the round-4 review could dispose of them, and F8 was addressed by rewording the PR description

PR #1326, opened 2026-09-22
feat(inception): author cross-references as clickable anchor links

### LINKED ISSUE #1323: [Feature]: Make inter-document references (ADR-1, FR-3, …) clickable links
### Description

AIDLC artifacts reference other artifacts by stable ID — `ADR-1`, `FR-3`, `US-1`, `NFR-2`, `ENT-001`, `unit-4`, etc. ADRs are already authored as relative markdown links in `core/knowledge/aidlc-architect-agent/adr-template.md` (e.g. `[ADR-015](./ADR-015-migrate-to-dynamodb.md)`), so they render clickable in any markdown viewer. Every other reference type, however, is written as a **bare token**, so a reader can't jump from `FR-3` to the requirement it names.

This proposes an **authoring convention** (no engine code): when a stage writes a cross-reference to another artifact, it emits a relative markdown link to the referenced ID's anchor — `[FR-3](./requirements.md#fr-3)` — and IDs are authored as anchorable headings so the link resolves. Because relative markdown links render across every AIDLC runtime's built-in preview (Kiro, Cursor, Claude Code, Codex, Copilot, opencode), this needs no custom viewer.

Scope of this change (the surgical, "Option A" path):
- Author each stable ID as an anchorable heading in its defining artifact (anchor lives in the shared artifact, e.g. `requirements.md#fr-3`) rather than restructuring to one-file-per-item.
- Update the relevant stage/knowledge files to instruct agents to emit references as relative links to those anchors.
- Add a carve-out in `core/memory/org.md` so the "preserved literal token / cross-reference" rule does not conflict with "write the reference as a link" (the ID text stays literal; it gains a link target).
- Note: `stage-protocol.md`'s rule to expand IDs in questions posed to the *user* (`FR3` → "the requirement that … (FR3)") is about human-facing questions, not artifact-to-artifact references, and is unchanged.
- Add a lightweight gate/sensor check that flags a cross-reference written as a bare token where a link target exists, keeping authoring honest without a rewrite engine.

### Use Case

A reader (or reviewing agent) tracing a decision currently has to grep for `FR-3` across artifacts to find what it means. Clickable references let them jump straight to the definition — improving navigability of generated specs, requirements, domain designs and reviews, especially in larger intents where the ID→text mapping is not memorable. It also strengthens traceability: a review lane can follow a reference to its source instead of assuming the mapping.

### Version

v2 (current on main)

### Area

Inception

### Additional Context

Prior art in-repo: `core/knowledge/aidlc-architect-agent/adr-template.md` already uses the exact link form this proposal generalizes. First implementation target is the highest-traffic reference producers — requirements-analysis and domain-design — before extending the convention method-wide.

SaRedfiche:
# Summary

AIDLC artifacts reference each other by stable ID (ADR-1, FR-3, US-1, NFR-2, ENT-001, unit-4). ADRs already render as clickable relative-markdown links; every other ID family was a bare token, so a reader could not click from `FR-3` to the requirement it names. This adds an authoring convention (no engine change) so those references become clickable links, backed by an advisory sensor.

## Changes

- **Stages** (`requirements-analysis`, `domain-design`): author each stable ID as a heading carrying a portable HTML anchor `<a id="fr1-2"></a>`, and write a cross-reference to it as a relative markdown link (`[FR1.2](#fr1-2)` / `[FR1.2](./requirements.md#fr1-2)`). Portable `<a id>` is used rather than a Kramdown `{#..}` heading attribute because `{#..}` does not render on GitHub.
- **`core/memory/org.md`**: states the convention as the single source of truth — anchor derivation (ID lower-cased, non-alphanumeric runs to single hyphens) and two exceptions that stay bare (an ID at its own definition point; a token naming no defined artifact yet). The preserved-literal-token rule is unchanged: the ID text stays literal; a cross-reference may wrap it in a link. The lower-cased anchor is invisible to the claim-sources / traceability / required-sections sensors (verified against their source).
- **New advisory `xref-links` sensor** (`core/tools/aidlc-sensor-xref-links.ts` + `core/sensors/aidlc-xref-links.md`): deterministic, same-file only. Flags a bare stable-ID token when the same file already defines an anchor it could link to, skipping fenced code blocks, inline code, inline/reference-style links, HTML tags, and the two convention exceptions. Wired into the dispatcher, routes, dynamic-import registry, and native sensor-list build. It deliberately does not flag missing-anchor / wrong-id / cross-file references (those need a record-wide index) — the manifest says so.

## User experience

**Before:** a reader tracing `FR-3` or `ADR-1` across generated specs, requirements, and decision logs had to grep for the ID to find its definition — bare tokens with no navigation.

**After:** references author as clickable links that jump to the defining anchor in any renderer that shows the `<a id>` target (GitHub included), and an advisory sensor nudges authors when a same-file reference was left bare next to an anchor it could point at. Advisory only — it never blocks a gate.

## Checklist

* [x] Changes are documented

## Test plan

- `bun run check` (package projection, typecheck, biome lint) — green.
- `bun test tests/unit/t329-sensor-xref-links.test.ts` — 25 cases: anchor derivation, both anchor forms (`<a id>` and `{#..}`), same-file flag, the two exceptions, inline-code / fenced-code / inline+reference-style link masking, prose `<`/`>` operators around an ID, nested differing-length fences, unterminated fence (no EOF blindness), hyphenated + single-segment IDs, per-line dedup, anchor-defined-after-use, CRLF, empty file, the three exit-1 read-error paths, and the dispatcher `matches` glob accept/reject plus the FAILED-verdict + detail-file mapping.
- `bun test tests/unit/t230-dispatcher-routes.test.ts` — sensor enumeration updated for the new sensor.

## Acknowledgment

Closes #1323


SaRedfiche:
**Intent:** Make AIDLC inter-document references (ADR-1, FR-3, US-1, …) clickable relative-markdown links via an authoring convention plus an advisory same-file `xref-links` sensor — no engine change.
**Not a goal:** Cross-file/missing-anchor/wrong-id link verification (needs a record-wide index — deferred); restructuring IDs to one-file-per-item.

PR #1131, opened 2026-09-11
fix: state the minted record-dir shape in the onboarding template (#1009)

vnlebaoduy:
# Summary

`core/templates/onboarding.md:47` told every session at startup that the active intent's record dir is `aidlc/spaces/<active-space>/intents/<slug>-<id8>/`. The engine mints `<YYMMDD>-<label>` (`aidlc-utility.ts` intent-create; `aidlc-lib.ts` keeps `<slug>-<id8>` only as `recordDirMatches()`'s back-compat fallback), and every other instruction surface already states the minted shape. The template was the one outlier, and the t174 docs legacy-ref gate could not catch it because its own fixture comment described `<slug>-<id8>` as the current shape.

Closes #1009.

## Changes

* `core/templates/onboarding.md` — the record-dir line now reads `<YYMMDD>-<label>`; renders into all seven `CLAUDE.md` / `AGENTS.md` onboarding files.
* `tests/unit/t174-docs-legacy-refs-gate.test.ts` — fifth retired-token lane: any `<slug>-<id8>` occurrence in `docs/**/*.md`, `core/templates/*.md`, or a harness orchestrator `SKILL.md` must be pinned in the allowlist. Nothing is pinned, so the lane has zero survivors.
* `tests/fixtures/docs-legacy-refs.json` — `_comment` states the correct shape and documents lane (5).

No version bump, badge, or CHANGELOG entry, per the Release Metadata Policy in `AGENTS.md`.

## User experience

Before: a fresh session reads `<slug>-<id8>` from `CLAUDE.md` / `AGENTS.md`, then `<YYMMDD>-<label>` from `stage-definition.md` — two instruction files disagreeing about the same path.

After: every onboarding file states the shape the engine actually writes.

## Checklist

* [x] I have reviewed the [contributing guidelines](https://github.com/awslabs/aidlc-workflows/blob/main/CONTRIBUTING.md)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented
* [ ] If this change adds an input to any fingerprint, epoch, or receipt identity, the description names the human-visible change it detects

## Test plan

On `355903d6` (2.8.2):

```
bun scripts/package.ts
bun test tests/unit/t174-docs-legacy-refs-gate.test.ts   # 3 pass
bun test tests/unit/t68-version-changelog-sync.test.ts   # 7 pass
bun run lint && bun run typecheck
bun scripts/package.ts --check                           # deterministic, 7 harnesses
```

[...]

PR #1368, opened 2026-09-23
fix(onboarding): render user-typed skill names with the harness skill prefix

### LINKED ISSUE #1341: [Bug]: `/aidlc-session-cost` is not a valid command
### Description

The `/aidlc-session-cost` command does not work — it is reported as not a valid command.

**Ask:** Fix the cost-tracking command or document the correct one.

### Steps to Reproduce

Observed during AI-DLC v2 testing / customer workshops. See Description for the specific scenario. (Consolidated from internal V2 feedback tracking.)

### Expected Behavior

The workflow completes the described step without the failure/friction noted above.

### Actual Behavior

See Description — the behavior above was observed instead.

### AI-DLC Version

v2 (current on main)

### Release / Commit

_No response_

### AI-DLC Phase

Not phase-specific

### Harness

Other

### AI Model

_No response_

### Environment

_No response_

### Additional Context

_No response_

vnlebaoduy:
Closes #1341.

## The bug

`{{INVOKE}}` is a shell command: `bun <harness-dir>/tools/aidlc.ts` on a bun install, `aidlc` on a native one. `core/templates/onboarding-harness.md` glued skill-name suffixes onto it, so the onboarding a bun install ships reads:

```
typed as `bun .claude/tools/aidlc.ts-session-cost`, `bun .claude/tools/aidlc.ts-replay`, ...
```

That is the reported "not a valid command". It affected every user-invocable skill named in harness onboarding — the three session skills, the document skill, the per-stage runners, and `-init`.

## The token for this already existed

`scripts/onboarding.ts` declares `{{SKILL_INVOKE}}` and `OnboardingFills.invoke` ("The invoke command this harness uses, e.g. `/aidlc` or `$aidlc`"), and `renderOnboarding` substitutes it and guards against a leftover marker. Every harness already declares its value. No template used the token.

This moves the eight command names onto `{{SKILL_INVOKE}}` and leaves the five genuine shell invocations (`{{INVOKE}} engine runtime summary --json`, `{{INVOKE}} --doctor`, `{{INVOKE}} --stage <slug> --single`, `{{INVOKE}} engine gen runners`, `{{INVOKE}} knowledge <verb>`) where they belong.

## Before / after

Rendered through `renderOnboarding` with each harness's own fills:

| harness | before | after |
|---|---|---|
| claude, copilot, cursor, kiro, kiro-ide, opencode | `bun <dir>/tools/aidlc.ts-session-cost` | `/aidlc-session-cost` |
| codex | `bun <dir>/tools/aidlc.ts-session-cost` | `$aidlc-session-cost` |

Codex is why this uses the token rather than a literal `/aidlc-…`: it declares `$aidlc`, and hardcoding the slash form would have been wrong there. Any harness that needs a different prefix changes one line in its own `onboarding.fills.ts`.

## Shipped-bytes consequence

Copilot's manifest sets `onboarding: { dst: "AGENTS.md", projectRoot: true }` with no `harnessDst`, so the harness skeleton is embedded in its root `AGENTS.md` and those bytes change. Its previously shipped hash moves into `harness/copilot/manifest.ts` so a copy-channel install still carrying it stays adoptable on refresh, and t243's inventory pins the new current hash — the same handling #1131 needed.

The other five AGENTS.md harnesses project the skeleton to a separate file (`steering/aidlc-onboarding.md`, `onboarding.md`, `rules/aidlc-onboarding.mdc`), so their root bytes are unchanged and they need no entry. I checked this against the rebuilt projections rather than assuming the #1131 shape: the packager appended no new hash for those five.

## Tests

`t151` gains a pin with two halves. The class guard is `expect(HARNESS).not.toMatch(/\{\{INVOKE\}\}-/)` — `{{INVOKE}}` followed directly by a suffix is always this mistake. The behavioural half renders the skeleton under both a `/aidlc` and a `$aidlc` harness and asserts each skill name carries that prefix and that no `aidlc.ts-` survives.

Ablation: reverting the template to `{{INVOKE}}-` reddens it (7 pass / 1 fail), and t243's legacy-signature inventory independently catches the copilot byte change.

Gates: `bun run check` green. `t151` 8/8, `t243` 116/116. Full unit tier 341 files, 3 failures — `t-tui-runtime`, `t-tui-native-private-root`, `t340-plan-approval-batch` — each verified failing identically on clean `main` @ a007c849, same tests and lines.

## Acknowledgment


apackeer (approved):
Thanks for this. The diagnosis is right, and the fix renders the correct skill prefix on all seven harnesses in both the copy and native channels. The copilot legacy-hash handling matches what #1262 did.

I pushed one follow-up commit (aba29e78) for two more examples in the same lines that rendered commands the CLI rejects:

- `{{INVOKE}} --stage <slug> --single` rendered `bun .claude/tools/aidlc.ts --stage ...`, which fails with `unknown command '--stage'`. It's the skill flag, so it's now `{{SKILL_INVOKE}} --stage <slug> --single`, matching `docs/guide/12-cli-commands.md`.
- `{{INVOKE}} knowledge <verb>` rendered `bun .claude/tools/aidlc.ts knowledge ...`, which fails with `unknown command 'knowledge'`. Knowledge is an engine noun, so it's now `{{INVOKE}} engine knowledge <verb>`.

t151 now pins both for every harness and channel, and copilot's new AGENTS.md hash replaces the unshipped one in t243 (no extra legacy entry needed).

Approving. We'll take it from here and merge.

Issue #1378, opened 2026-09-23
Plan Approval: fingerprint blocks on a complete Testing Contract (D)


SaRedfiche:
Part of #1374.

## Problem

The fingerprint binds the code-generation plan, which embeds a `## Testing Contract` JSON block produced by `aidlc-testing-posture.ts render`. If `render` is still running or wrote a partial contract (a backgrounded `testing-posture render &` fed a PARTIAL Testing Contract on the stalled customer run), `parseTestingContract` returns null, `contractValid` is false, and the approval is refused later with:

> Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.

The fingerprint command will happily fingerprint a plan whose Testing Contract is incomplete, so the failure surfaces only at the gate.

## Verbatim error string(s) addressed

- `Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.`
- `Plan Approval requires the current Testing Contract` (`planApprovalQuestionEvidence`, `core/tools/aidlc-testing-posture.ts`)

## Proposed behavior

Before fingerprinting, require a COMPLETE Testing Contract. If `testing-posture render` is still running or the embedded contract is incomplete/partial, refuse with a distinct message:

> Testing Contract still rendering — wait for render to complete (or re-run it synchronously) before fingerprinting.

Options: make `render` synchronous by contract, or gate the fingerprint on a render completion marker so a partial `## Testing Contract` can never be fingerprinted.

## Acceptance criteria

- `fingerprint` refuses with the "Testing Contract still rendering / incomplete" message when the embedded contract is partial or absent, instead of fingerprinting it.
- A complete, current Testing Contract fingerprints exactly as today.
- The guard invariant is unchanged; this only moves the incomplete-contract detection from the gate to fingerprint time.

Issue #1377, opened 2026-09-23
Plan Approval: auto-resolve --session from active SessionStart context (C)


SaRedfiche:
Part of #1374.

## Problem

A `--checkpoint plan-approval` `decision`/`answer` requires `--session <id>` from the invoking SessionStart context (`core/tools/aidlc-log.ts`). When it is omitted — a fresh session, or a conductor that did not thread the id — the command refuses:

> Plan Approval requires --session <id> from the invoking SessionStart context.

The refusal does not tell the caller WHICH id to pass, so the conductor must go find it. Meanwhile `resolveSessionIdFromAncestry(projectDir)` already exists and is used elsewhere in the same file (the `withdrawProtectedQuestions` fallback), so the id is usually resolvable without asking.

## Verbatim error string(s) addressed

- `Plan Approval requires --session <id> from the invoking SessionStart context.` (emitted in `handleDecision`, `handleAnswer`, and `handlePlanApprovalBatch`, `core/tools/aidlc-log.ts`)

## Proposed behavior

When `--session` is omitted at a plan-approval `decision`/`answer`, auto-resolve it from the active SessionStart context via `resolveSessionIdFromAncestry(projectDir)` and use that id. If it cannot be resolved, fail with the exact command to run **including the resolved-or-required id**, e.g.:

> Plan Approval requires --session <id>. Could not auto-resolve it from the SessionStart context; pass `--session <the SessionStart id>`.

## Acceptance criteria

- With `--session` omitted and an ancestry-resolvable session, the plan-approval decision/answer succeeds using the resolved id (the receipt binds the same id it always would).
- With `--session` omitted and no resolvable session, the command fails with a message naming the exact `--session` argument to add.
- An explicitly passed `--session` still wins and is unchanged.
- The receipt's session binding is identical to the pre-change behavior when the id is the same — no new approval path, no guard loosening.

Issue #1376, opened 2026-09-23
Plan Approval: preflight batching all preconditions in one message (B)


SaRedfiche:
Part of #1374.

## Problem

The plan-approval gate reports precondition failures **one at a time**. A conductor fixes the missing `--session`, re-runs, then discovers the fingerprint is stale, re-runs, then discovers the recorded choice is wrong. Each round-trip is a separate turn; on the stalled customer run these compounded into ~90 minutes. The three preconditions are checked in different places (`core/tools/aidlc-log.ts` for session and choice; `core/tools/aidlc-testing-posture.ts` for the fingerprint).

## Verbatim error string(s) addressed

- `Plan Approval requires --session <id> from the invoking SessionStart context.`
- `Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.`
- `Refusing to record Plan Approval: received reply <reply>. Valid choices are "Approve Plan" or "Request Changes".`

## Proposed behavior

Add an approval **preflight** that validates all preconditions together and reports EVERY failing one in a single message:

- session present (or auto-resolvable — see issue C),
- fingerprint current (recorded == recomputed over current plan/instructions/contract),
- choice matches the offered gate options.

Instead of stopping at the first failure, collect and report the full set so the conductor fixes them in one pass.

## Acceptance criteria

- With multiple preconditions failing, the preflight lists all of them in one message, not just the first.
- Each reported item names the concrete remedy: pass/auto-resolve session, re-fingerprint, use an exact offered choice.
- The guard invariant is unchanged — preflight is advisory batching of the SAME checks the gate already enforces; it never approves or loosens anything.

Issue #1375, opened 2026-09-23
Plan Approval: signal fingerprint staleness at edit time (A)


SaRedfiche:
Part of #1374.

## Problem

Plan Approval binds a fingerprint over the code-generation plan, unit-test instructions, and the Testing Contract. When any of those files is edited after the fingerprint is recorded, the approval is **silently invalidated** — nothing observes the edit. The conductor only discovers the invalidation later, at the decision/answer/dispatch gate, where `core/tools/aidlc-testing-posture.ts` recomputes the fingerprint and refuses with the verbatim string:

> Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.

This silent-until-the-gate behavior was a major contributor to a ~90-minute stall.

## Verbatim error string(s) addressed

- `Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.` (emitted in `planApprovalQuestionEvidence` and mirrored in `evaluateCodeGenerationApproval`, `core/tools/aidlc-testing-posture.ts`)

## Proposed behavior

When a plan / unit-test-instructions / Testing-Contract file is edited while a fingerprint already exists for that stage+unit, emit an **observable transition** rather than silently invalidating:

> approval fingerprint invalidated — re-fingerprint before approving

Turn the silent invalidation into a visible state change the conductor can act on immediately (re-fingerprint), before ever presenting the plan.

## Acceptance criteria

- A plan/instructions/Testing-Contract edit made after a fingerprint exists produces the observable `approval fingerprint invalidated — re-fingerprint before approving` signal.
- The guard invariant is unchanged: the gate still refuses an unapproved/stale-fingerprint dispatch. This only makes invalidation observable earlier; it does not loosen the guard.
- No new fingerprint bypass: the recorded-vs-current comparison is untouched.

PR #1379, opened 2026-09-23
feat(plan-approval): surface fingerprint staleness at edit time and auto-resolve --session

SaRedfiche:
# Summary

Improves the code-generation **plan-approval** failure ergonomics without weakening the approval guard. Implements improvements **A** and **C** from #1374: an edit-time signal that a bound file will invalidate the approval, and auto-resolution of `--session` from the active SessionStart context. Both changes are advisory or plumbing — the guard still refuses any unapproved or stale-fingerprint dispatch exactly as before.

Part of #1374. Improvements B (#1376) and D (#1378) remain issues.

## Changes

| Improvement | What it does |
|---|---|
| **A** — edit-time signal (#1375) | When a write touches an approval-bound file (`code-generation-plan.md`, `unit-test-instructions.md`) while a live fingerprint exists, the guard prints a prospective notice that editing the file will invalidate the approval and it should be re-fingerprinted before approving. The notice never changes the allow/block decision, and the file read is gated behind the cheap bound-file check so non-bound writes skip it on the PreToolUse hot path. |
| **C** — auto-resolve `--session` (#1377) | A plan-approval `decision`/`answer` that omits `--session` reuses the existing `resolveSessionIdFromAncestry()` to fill the id from the live SessionStart context. An explicit `--session` still wins. When it cannot resolve, the message names the exact argument to add. |

The break-glass override path is pinned by a session-consistency round-trip test (the request is consumed and the receipt bound under one id) plus a cross-session refusal test, so a future split between the authorization-check id and the receipt-bind id fails the suite.

## User experience

Before: a conductor could omit `--session` and hit `Plan Approval requires --session <id> from the invoking SessionStart context.`, and editing a plan/instructions/Testing-Contract file after fingerprinting silently invalidated the approval — the conductor only discovered it later at the gate, via `Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.` This guess-and-retry loop was the root of a ~90-minute customer stall.

After: the session is auto-filled when the harness omits it, and editing a bound file surfaces the impending invalidation immediately so the conductor re-fingerprints before presenting the plan, instead of discovering the staleness only at the gate.

## Checklist

* [ ] Changes are documented

This change adds no input to any fingerprint, epoch, or receipt identity. The fingerprint comparison and the receipt's session binding are unchanged; A only observes bound-file edits earlier, and C fills an omitted id with the same value the caller would have passed.

## Test plan

`bun test tests/unit/t328-plan-approval-runtime-authority.test.ts` — 38 pass / 0 fail. New tests: the edit-time signal fires on a bound-file edit and stays silent with no live fingerprint or a non-bound write; the guard still allows a bound-file edit (exit 0) when the fingerprint machinery throws; the break-glass override round-trip consumes the request and binds the receipt under one session id; a request typed under one session is refused when answered under a different `--session`. `bun run typecheck` clean; `bun scripts/package.ts --check` deterministic (no dist drift).

**Why no screenshot:** the diff is TypeScript engine/harness logic and tests with no rendered UI surface.

**Intent:** make the plan-approval gate surface its state before a refusal — signal bound-file edits at edit time and auto-fill the session — without weakening the approval guard.
**Not a goal:** changing what the guard allows or blocks, the fingerprint comparison, or the receipt's identity.


SaRedfiche:
Implements **A** (#1375) and **C** (#1377) from the tracker #1374. **B** (#1376) and **D** (#1378) remain as issues.

Verbatim gate strings this addresses:
- `Plan Approval requires --session <id> from the invoking SessionStart context.` — C auto-resolves the id from the SessionStart context; an explicit `--session` still wins.
- `Plan Approval fingerprint does not match the active intent, target, stage attempt, plan, instructions, and Testing Contract. Re-run the fingerprint command, re-present the plan, and approve again.` — A surfaces the impending invalidation at edit time so it is observable before the gate, instead of only at the refusal.

The approval guard invariant is unchanged: A is an advisory notice that never alters the allow/block decision, and C fills an omitted `--session` with the same id the caller would have passed. The break-glass override path is pinned by a session-consistency round-trip test plus a cross-session refusal test.

